PR #173, opened 2021-07-13
feat: Add OpenModal helper for testing modals.

stephenh:
An unintended artifact of moving the `ModalBody` / `ModalFooter` to using portals it that if tests render a `SomeModalComponent` directly, the content goes in the portals ... but the divs aren't mounted in the DOM yet, b/c that only happens when `openModal` is called.

So, this is just a wrapper to easily call `openModal` for tests ... would eventually be nice to not use this, and somehow have the divs already mounted.

Fwiw another approach would be to make another wrapper, like:

```
const r = await render(
  <ModalComponent id=foo />,
  withOpenModal,
  withApollo(...)
);
```

Where `withOpenModal`'s `wrap` function would basically have a `useEffect` that calls `openModal`.

This seems a little more straight forward, but not sure, the `withOpenModal` is a little less code...

### DIFF
--- a/src/components/BeamContext.tsx
+++ b/src/components/BeamContext.tsx
@@ -8,7 +8,7 @@ import { EmptyRef } from "src/utils/index";
 /** The internal state of our Beam context; see useModal and useSuperDrawer for the public APIs. */
 export interface BeamContextState {
   // SuperDrawer contentStack
-  contentStack: MutableRefObject<ContentStack[]>;
+  contentStack: MutableRefObject<readonly ContentStack[]>;
   modalState: MutableRefObject<ModalProps | undefined>;
   canCloseModalChecks: MutableRefObject<Array<() => boolean>>;
   /** The div for ModalBody to portal into; note this can't be a ref b/c Modal hasn't set the ref at the time ModalBody renders. */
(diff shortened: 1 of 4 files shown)


src/components/BeamContext.tsx
```diff
@@ -8,7 +8,7 @@ import { EmptyRef } from "src/utils/index";
 /** The internal state of our Beam context; see useModal and useSuperDrawer for the public APIs. */
 export interface BeamContextState {
   // SuperDrawer contentStack
-  contentStack: MutableRefObject<ContentStack[]>;
+  contentStack: MutableRefObject<readonly ContentStack[]>;
```
KoltonG, line 11: **Question:** This didn't seem to help with the `.push` - does it help with other things?

src/components/Modal/OpenModal.tsx
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from "react";
+import { useModal } from "src/components/Modal/useModal";
+
+/**
+ * A component for testing open modals in unit tests.
+ *
+ * Current, calling `render(<ModalComponent />)` in a test currently doesn't work, because
```
KoltonG, line 7:
```suggestion
 * Calling `render(<ModalComponent />)` in a test currently doesn't work, because
```

KoltonG (approved):
🍣 interesting!

src/components/BeamContext.tsx
stephenh: It did cause compile errors for me when I directly to put the `push` calls back, :shrug:

PR #174, opened 2021-07-13
feat: Add ModalSize.xl.

stephenh:
Pending confirmation that this is how we want to solve the procurement modal being larger than `lg`.


KoltonG (approved):
🍣

PR #175, opened 2021-07-13
fix: Fix card style header being too tall.

stephenh:


src/components/GridTable.tsx
```diff
@@ -137,15 +137,17 @@ export const condensedStyle: GridStyle = {
...
-  firstNonHeaderRowCss: Css.$,
+  firstNonHeaderRowCss: Css.mt2.$,
   cellCss: Css.p2.my1.bt.bb.bGray400.$,
   firstCellCss: Css.bl.add({ borderTopLeftRadius: "4px", borderBottomLeftRadius: "4px" }).$,
   lastCellCss: Css.br.add({ borderTopRightRadius: "4px", borderBottomRightRadius: "4px" }).$,
-  rootCss: Css.dig.$,
   // Undo the card look & feel for the header
   headerCellCss: {
     ...defaultStyle.headerCellCss,
-    ...Css.add({ border: "none", borderRadius: "unset" }).p1.m0.mb1.xsEm.gray700.$,
```
stephenh, line 148: Really odd, but the `mb1`, which is margin-bottom, was causing the header cells to all be "too tall" by ~8px.

We can work around it by leaving `mb1` off of here, and just compensate by having the `firstNonHeaderRowCss` have a `mt2` (so the same total of 16px between).

src/components/GridTable.tsx
```diff
@@ -137,15 +137,17 @@ export const condensedStyle: GridStyle = {
 /** Renders each row as a card. */
 export const cardStyle: GridStyle = {
   ...defaultStyle,
-  firstNonHeaderRowCss: Css.$,
+  firstNonHeaderRowCss: Css.mt2.$,
   cellCss: Css.p2.my1.bt.bb.bGray400.$,
   firstCellCss: Css.bl.add({ borderTopLeftRadius: "4px", borderBottomLeftRadius: "4px" }).$,
   lastCellCss: Css.br.add({ borderTopRightRadius: "4px", borderBottomRightRadius: "4px" }).$,
-  rootCss: Css.dig.$,
```
stephenh, line 144: I'd wanted to try `inline-grid` but I think it's easier to get more deterministic widths if this is always `grid` and then callers just put it into a div of width 550 or what not.

KoltonG (approved):
🍣 what an interesting margin issue!

PR #176, opened 2021-07-13
fix: Make SuperDrawer Content Stable when ConfirmationModal Appears

KoltonG:
When using the canClose checks with the SuperDrawer, the drawer content would unmount when any checks failed. This causes implementation problems when using forms inside drawer content which would cause the form to lose it's state.


src/components/SuperDrawer/SuperDrawer.tsx
```diff
@@ -102,9 +103,9 @@ export function SuperDrawer(): ReactPortal | null {
                 {/* Left */}
                 <div css={Css.df.itemsCenter.$}>
                   <div css={Css.xl2Em.gray900.mr2.$} {...testId.title}>
-                    {title}
+                    {modalState.current?.title || title}
```
stephenh, line 106: Neat.

stephenh (approved):

PR #179, opened 2021-07-16
fix: Fix date field width, match others.

stephenh:
Making them behave width-wise like our other form fields:

image

(We didn't have this width convention figured out yet with date field was written.)


KoltonG (approved):
Thanks

PR #182, opened 2021-07-21
feat: Add Alert based on TW example.

stephenh:


src/components/Alert.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <div css={Css.df.flexColumn.childGap3.$}>
+      <Alert onClose={action("close")}>
+        <span css={Css.b.$}>Holy Smokes!</span>
+        &nbsp;
+        <span>Something seriously bad happened.</span>
+      </Alert>
+      {/* An example with an error message that wraps. */}
+      <Alert onClose={action("close")}>
+        <div css={Css.b.$}>Holy Smokes!</div>
+        <div>Something seriously bad happened.</div>
```
ksk5280, line 21: **thought:** I assume it's ok to pass through just text without any html/extra styles? Can you add a story to show this as an example?

ksk5280 (approved):
👍

src/components/Alert.stories.tsx
stephenh: Good idea! Adding...

src/components/Alert.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export interface AlertProps {
+  children: ReactNode;
+  onClose: () => void;
+}
+
+export function Alert(props: AlertProps) {
+  const { children, onClose } = props;
+  return (
+    <div css={Css.df.itemsStart.justifyBetween.bgRed100.bRed400.red700.p2.br4.$} role="alert">
```
KoltonG, line 13: **Suggestion:** Heads up that there is no border added to this component. The `bRed400` adds the colour but I think you need to add the border itself 🙈

src/components/Alert.tsx
```diff
@@ -0,0 +1,18 @@
+import { ReactNode } from "react";
+import { IconButton } from "src/components/IconButton";
+import { Css, Palette } from "src/Css";
+
+export interface AlertProps {
+  children: ReactNode;
+  onClose: () => void;
+}
+
+export function Alert(props: AlertProps) {
```
KoltonG, line 10: **Question:** Is there going to be a Context and Portal approach made for this alert component to make it easier to use similar to the Modal and SuperDrawer?

src/components/Alert.tsx
stephenh: @KoltonG good catch! I remember typing `ba` and it looked like that didn't work, but I think my webstorm autocomplete was just being funny. Will add!

src/components/Alert.tsx
stephenh: @KoltonG good question; I didn't really have anything in mind other than the look & feel. You're thinking of basically a toast-ish sort of thing where alerts are auto-floated on top of the current context?

I like that; as-is pages would just have to layout the alert inline where ever they wanted.

PR #183, opened 2021-07-21
fix: Update TextAreaField padding to match TextField.

stephenh:
When looking at a screen that has both regular `TextField`s as well as `TextAreaField`s it looked weird that the `TextAreaField`'s content was indented more than the regular text fields.


chr1sjf0x (approved):
🍣 🚀

PR #185, opened 2021-07-21
chore: Inline ComboBox into SelectFieldBase.

stephenh:
Trying to simplify things on the way to fixing the "Empty means 'All'" behavior that is inherited from Filters that we don't want for regular MultiSelects.

### DIFF
--- a/src/inputs/internal/SelectFieldBase.tsx
+++ b/src/inputs/internal/SelectFieldBase.tsx
@@ -1,4 +1,4 @@
-import { CollectionChildren, Selection } from "@react-types/shared";
+import { Selection } from "@react-types/shared";
 import { Key, ReactNode, useEffect, useRef, useState } from "react";
 import { useButton, useComboBox, useFilter, useFocusRing, useOverlayPosition } from "react-aria";
 import { Item, useComboBoxState, useMultipleSelectionState } from "react-stately";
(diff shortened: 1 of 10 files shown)


src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -69,133 +137,28 @@ export function SelectFieldBase<O, V extends Value>(props: SelectFieldBaseProps<
...
-    <ComboBox<O, Key>
-      {...beamSelectFieldBaseProps}
-      multiselect={multiselect}
-      filteredOptions={fieldState.filteredOptions}
-      inputValue={fieldState.inputValue}
-      selectedKeys={fieldState.selectedKeys}
-      selectedOptions={fieldState.selectedOptions}
-      getOptionLabel={getOptionLabel}
-      getOptionValue={(o) => valueToKey(getOptionValue(o))}
-      onSelectionChange={(keys) => {
```
stephenh, line 87: I'm not sure how to really read this diff, but basically all of the props + functions here that were based to `ComboBox` are now just declared / mapped inline because `SelectFieldBase` _is_ `ComboBox`.

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -31,27 +39,87 @@ export interface SelectFieldBaseProps<O, V extends Value> extends BeamSelectFiel
...
+    getOptionLabel,
+    getOptionValue,
+    getOptionMenuLabel = getOptionLabel,
+    sizeToContent = false,
+    values,
+    ...otherProps
   } = props;
 
-  const { contains } = useFilter({ sensitivity: "base" });
+  function onSelectionChange(keys: Selection): void {
```
stephenh, line 63: See red code below, but this is all "the same code", but just as a `function` declared in `SelectFieldBase` and used here directly, instead of passed as a prop to the now-inlined `ComboBox`.

src/inputs/internal/SelectFieldInput.tsx
```diff
@@ -73,7 +66,6 @@ export function SelectFieldInput<O, V extends Value>(props: SelectFieldInputProp
   const readOnlyStyles = isReadOnly ? Css.bn.pl0.pt0.add("backgroundColor", "unset").$ : {};
   const tid = useTestIds(inputProps); // data-testid comes in through here
   const isMultiSelect = state.selectionManager.selectionMode === "multiple";
-  const [isSizeCurrentlyBasedOnContent, setIsSizeBasedOnContent] = useState(true);
```
stephenh, line 76: This boolean `useState` was just being set true/false in focus/blur, and we already have an `isFocused` hanging around, so used that instead.

chr1sjf0x (approved):
🍣

PR #186, opened 2021-07-22
feat: Add nothingSelectedText to customize All behavior.

stephenh:
The "All" behavior was hard-coded for filters and not something we want for like editing in a form.

### DIFF
--- a/src/components/Filters/utils.tsx
+++ b/src/components/Filters/utils.tsx
@@ -82,6 +82,7 @@ export function getFilterComponents<F>(props: GetFilterComponentsOpts<F>) {
           inlineLabel
           sizeToContent={!inModal}
           onSelect={(values) => updateFilter(filter, key, values)}
+          nothingSelectedText="All"
         />,
         inModal,
         filterDef.label,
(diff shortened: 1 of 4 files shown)


src/components/Filters/utils.tsx
```diff
@@ -82,6 +82,7 @@ export function getFilterComponents<F>(props: GetFilterComponentsOpts<F>) {
           inlineLabel
           sizeToContent={!inModal}
           onSelect={(values) => updateFilter(filter, key, values)}
+          nothingSelectedText="All"
```
chr1sjf0x, line 85: thought (non-blocking): I wonder if something like `placeholderText` would be better (that's what Retool calls this anyway) 🤷  Granted the current name is descriptive 😆

chr1sjf0x (approved):
🍣

src/components/Filters/utils.tsx
stephenh: Oh good point ... you're right that's the canonical term for text fields. ...and this is eventually ending up on a text field. I didn't make that connection since this is the select/multi-select ... Might keep for now but open to a breaking change later if that ends up being a better idea.

PR #188, opened 2021-07-22
feat: Create Ability To Export Mocks For Beam Inputs [ch5604]

DeanGilewicz:
# ch5604
- export mocks for the following inputs:
   - DateField
   - SelectField
   - MultiSelectField
   - RichTextField
- Move `RichTextField` from `components` into `inputs`
- write unit tests for the mocks


KoltonG:
You can probably remove the package.lock files created since we are using `yarn` here.

src/mocks/components/MockRichTextField.tsx
```diff
@@ -0,0 +1,25 @@
+import { camelCase } from "change-case";
+import { useState } from "react";
+import { RichTextFieldProps } from "src/components/RichTextField";
+import { useTestIds } from "src/utils";
+
+export function RichTextField(props: RichTextFieldProps) {
```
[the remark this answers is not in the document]
DeanGilewicz: so been working to try and find a cute way to abstract the Jest mocks into `Beam` itself but have not been successful.

The mocks are being exported alongside the components in Beam but this means you still need to manually define each mock in your application's `setupTests.js` such as:

```
jest.mock("@homebound/beam/dist/inputs/DateField", () => {
  return require("@homebound/beam/dist/inputs/DateField.mock");
});

jest.mock("@homebound/beam/dist/inputs/SelectField", () => {
  return require("@homebound/beam/dist/inputs/SelectField.mock");
});

jest.mock("@homebound/beam/dist/inputs/MultiSelectField", () => {
  return require("@homebound/beam/dist/inputs/MultiSelectField.mock");
});
```

As a next iteration, it would be great if we could do these mocks automatically. The idea is to export a function from Beam containing all jest mocks and then in application's `setupTests.js` we could just import and execute this function.

src/inputs/index.ts
```diff
@@ -1,11 +1,14 @@
...
 export * from "./CheckboxGroup";
 export * from "./DateField";
+export * as MockDateField from "./DateField.mock";
 export * from "./MultiSelectField";
+export * as MockMultiSelectField from "./MultiSelectField.mock";
 export * from "./NumberField";
 export { RadioGroupField } from "./RadioGroupField";
 export type { RadioFieldOption, RadioGroupFieldProps } from "./RadioGroupField";
 export * from "./SelectField";
+export * as MockSelectField from "./SelectField.mock";
```
stephenh, line 11: Do we need these exports? Like if downstream projects are doing things like:

```
jest.mock("@homebound/beam/dist/inputs/SelectField", () => {
  return require("src/mocks/MockSelectField");
});
```

And mocking at the module level...

Granted, doesn't hurt either, just curious.

DeanGilewicz: so my thought is that once this is available we would remove all of the application's mocks and instead reference these exported mocks from beam like:
```
jest.mock("@homebound/beam/dist/inputs/SelectField", () => {
  return require("@homebound/beam/dist/inputs/SelectField.mock");
});
```
keeping all mocks centralized. This should also make it easier for us to look into providing a great abstraction (not having to define each and every mock)

DeanGilewicz: just realized I rambled on and didn't answer your question. No don't need to export them so I'll remove

stephenh: +1 to your snippet, I guess my point was that we don't need `MockSelectField` exported from an `index.ts` file then?

Some of the issues @bdow and I have been running into today has been from Beam's `index.ts` files exporting both "for production components" + "for testing utils", and so we're nudging the `index.ts` style exports to be only what's really really production code.

stephenh (approved):

src/inputs/index.ts
DeanGilewicz: funnily enough I was just looking into your comment again after following the PR and I agree to removing the mocks from being explicitly exported. I'll update

PR #189, opened 2021-07-22
feat: Add BoundSwitchField, FormLines.

stephenh:
For Trade Partner Contact screens in internal-internal.

FormLines is kinda WIP, but figure it might as well be WIP here instead of copy/pasting it across internal-frontend and procurement-frontend.

### DIFF
--- a/src/forms/FormLines.stories.tsx
+++ b/src/forms/FormLines.stories.tsx
@@ -1,5 +1,5 @@
 import { Meta } from "@storybook/react";
-import { FormLines } from "src/forms/FormLines";
+import { FieldGroup, FormLines } from "src/forms/FormLines";
 import { TextField } from "src/inputs/TextField";
 
 export default {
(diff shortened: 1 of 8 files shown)


src/inputs/Switch.tsx
```diff
@@ -13,6 +14,8 @@ export interface SwitchProps {
   disabled?: boolean;
   /** Input label */
   label: string;
+  /** Where to put the label. */
+  labelStyle?: "form" | "inline";
```
stephenh, line 18: The current switch had labels off to the right, but when in a form with other fields, I think it's more consistent to put the label on top.

src/inputs/Switch.tsx
```diff
@@ -42,27 +46,30 @@ export function Switch(props: SwitchProps) {
     <label
       {...hoverProps}
       css={{
-        ...switchLabelDefaultStyles,
```
stephenh, line 45: I inlined a bunch of these styles that were constants b/c the idea of tachyons/utility css is to have the css defined inline / as close to the markup as possible.

src/forms/FormLines.tsx
```diff
@@ -0,0 +1,44 @@
...
+        "& > *": Css.mb3.$,
+      }}
+    >
+      {children}
+    </div>
+  );
+}
+
+/** Groups multiple fields side-by-side. */
+export function FieldGroup(props: {
```
KoltonG, line 31: **Question:** Is this like FormColumn?

src/forms/FormLines.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <TextField label={"First"} value={"first"} onChange={() => {}} />
+      <TextField label={"Last"} value={"last"} onChange={() => {}} />
+    </FormLines>
+  );
+}
+
+export function SideBySide() {
+  return (
+    <FormLines>
+      <FieldGroup>
```
KoltonG, line 31: **Thoughts**: Any thought on making it `FieldLines` of `formGroup` to keep the names consistent?

KoltonG (approved):
🍣

src/forms/FormLines.tsx
stephenh: Well, it's more like "a multi-field line", i.e.:

```
| first name          | // 1st line
| last name          | // 2nd line
| middle | phone  | // a FieldGroup one the 3rd line
```

PR #191, opened 2021-07-24
feat: Form lines, widths, required labels

stephenh:
Building out a form for Trade Partner Contacts and infra-izing various things about forms.

The biggest change is that most `...Field`s learned a new `required` prop, that if set to `true` or `false`, they'll add `Required` / `Optional` to their label (the text "Required" or "Optional" is currently controlled by a new `FormContext`).

Also the `Bound...Field`s learned to pass `field.required` in as the new `required` prop, so for the most part everything should wire together automatically/for free, based on `rules: [required]` in the form config.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,6 +1,6 @@
 module.exports = {
   stories: ["../src/**/*.stories.tsx"],
-  addons: ["@storybook/addon-links", "@storybook/addon-essentials", "storybook-addon-performance/register"],
+  addons: ["@storybook/addon-links", "@storybook/addon-essentials"],
   // https://storybook.js.org/docs/react/configure/typescript#mainjs-configuration
   typescript: { check: false },
   webpackFinal: async (config) => {
--- a/src/components/Label.tsx
+++ b/src/components/Label.tsx
@@ -6,15 +6,18 @@ interface LabelProps {
   // We don't usually have `fooProps`-style props, but this is for/from react-aria
   labelProps?: LabelHTMLAttributes<HTMLLabelElement>;
   label: string;
+  suffix?: string;
   // If set, it is recommended to wrap in an element with `position: relative;` set, as the label will have an absolute position.
   hidden?: boolean;
 }
 
 /** An internal helper component for rendering form labels. */
-export function Label({ labelProps, label, hidden, ...others }: LabelProps) {
+export function Label(props: LabelProps) {
+  const { labelProps, label, hidden, suffix, ...others } = props;
   const labelEl = (
     <label {...labelProps} {...others} css={Css.dib.sm.gray700.mbPx(4).$}>
       {label}
+      {suffix && ` ${suffix}`}
     </label>
   );
   return hidden ? <VisuallyHidden>{labelEl}</VisuallyHidden> : labelEl;
--- a/src/forms/BoundDateField.tsx
+++ b/src/forms/BoundDateField.tsx
@@ -1,8 +1,5 @@
 import { FieldState } from "@homebound/form-state";
 import { Observer } from "mobx-react";
-import { useContext } from "react";
-import { FormContext } from "src/forms/FormContext";
-import { getLabelSuffix } from "src/forms/labelUtils";
 import { DateField, DateFieldProps } from "src/inputs";
 import { useTestIds } from "src/utils";
 import { defaultLabel } from "src/utils/defaultLabel";
@@ -16,17 +13,16 @@ export type BoundDateFieldProps = Omit<DateFieldProps, "value" | "onChange" | "o
 /** Wraps `TextField` and binds it to a form field. */
 export function BoundDateField(props: BoundDateFieldProps) {
   const { field, onChange = (value) => field.set(value), label = defaultLabel(field.key), ...others } = props;
-  const settings = useContext(FormContext);
   const testId = useTestIds(props, field.key);
   return (
     <Observer>
       {() => (
         <DateField
           label={label}
-          labelSuffix={getLabelSuffix(settings, field)}
           value={field.value || undefined}
           onChange={onChange}
           errorMsg={field.touched ? field.errors.join(" ") : undefined}
+          required={field.required}
           onBlur={() => field.blur()}
           onFocus={() => field.focus()}
           {...testId}
--- a/src/forms/BoundDateField.tsx
+++ b/src/forms/BoundDateField.tsx
@@ -1,5 +1,8 @@
 import { FieldState } from "@homebound/form-state";
 import { Observer } from "mobx-react";
+import { useContext } from "react";
+import { FormContext } from "src/forms/FormContext";
+import { getLabelSuffix } from "src/forms/labelUtils";
 import { DateField, DateFieldProps } from "src/inputs";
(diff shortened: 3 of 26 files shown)


src/components/Label.tsx
```diff
@@ -6,15 +6,18 @@ interface LabelProps {
   // We don't usually have `fooProps`-style props, but this is for/from react-aria
   labelProps?: LabelHTMLAttributes<HTMLLabelElement>;
   label: string;
+  suffix?: string;
```
stephenh, line 9: For doing `label=First Name`, `suffix=Required` the idea being we don't want "required" to leak into the react-aria label directly. I assume they have like aria-required attributes? Not sure, but just seems good to keep separate.

src/forms/BoundDateField.tsx
```diff
@@ -13,12 +16,14 @@ export type BoundDateFieldProps = Omit<DateFieldProps, "value" | "onChange" | "o
...
 export function BoundDateField(props: BoundDateFieldProps) {
   const { field, onChange = (value) => field.set(value), label = defaultLabel(field.key), ...others } = props;
+  const settings = useContext(FormContext);
   const testId = useTestIds(props, field.key);
   return (
     <Observer>
       {() => (
         <DateField
           label={label}
+          labelSuffix={getLabelSuffix(settings, field)}
```
stephenh, line 26: Teaching the bound fields to auto-add `First Name (Required)` where the "Required" is only added if the `FormLines` parent opts-in to a given label suffix style.

Personally, I'd prefer to have this hard-coded, like we _always_ put `(Required)` and/or `(Optional)` after field lines, but I'm anticipating design wanting more flexibility.

src/forms/FormLines.stories.tsx
```diff
@@ -10,34 +12,114 @@ export default {
...
+      <FieldGroup>
+        <TextField label="Title" value="Engineer" onChange={() => {}} />
+        <span />
+      </FieldGroup>
+      <FieldGroup>
+        <TextField label="First" value="first" onChange={() => {}} />
+        <TextField label="Middle" value="middle" onChange={() => {}} />
+        <TextField label="Last" value="last" onChange={() => {}} />
+      </FieldGroup>
+      <FieldGroup widths={[2, 1, 2]}>
```
stephenh, line 51: `FieldGroup` learned a new `widths` which if numbers are `fr` units, but can also be "30px". They're fed into a `gridTemplateColumn`.

src/forms/labelUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+export function getLabelSuffix(settings: FormSettings, field: FieldState<any, any>): string {
+  switch (settings.labelSuffix) {
+    case "required":
+      return field.required ? "(Required)" : "";
+    case "optional":
+      return field.required ? "" : "(Optional)";
+    case "none":
+      return "";
+    case "both":
+      return field.required ? "(Req)" : "(Opt)";
```
stephenh, line 13: I know "Req" and "Opt" is probably too cute, but I've got a form where "Postal Code (Optional)" would wrap:

image

Granted, I can one-off set the `labelSuffix` directly on this field, but, dunno, _maybe_ this is okay. :-) Especially in "both" mode seeing "Required" and "Optional" spelled out every gets kinda tedious anyway.

I'll also see what design thinks. I believe so far they've preferred spelling out "Required" and "Optional" instead of ambiguous (but short!) symbols like `*`, which length aside I generally agree with.

.storybook/main.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   stories: ["../src/**/*.stories.tsx"],
-  addons: ["@storybook/addon-links", "@storybook/addon-essentials", "storybook-addon-performance/register"],
```
stephenh, line 3: This addon, for some reason, started cause TypeErrors / `state is not defined` errors when trying to load any story, so I had to remove it... :shrug: ?

src/inputs/NumberField.tsx
```diff
@@ -42,6 +45,7 @@ export function NumberField(props: NumberFieldProps) {
   } = props;
 
   const factor = type === "percent" || type === "cents" ? 100 : type === "basisPoints" ? 10_000 : 1;
+  const labelSuffix = getLabelSuffix(required);
```
KoltonG, line 48: **Question:** Wondering why we are not utilizing the `formState` object to determine `required` (using the Bound fields)? I'm concerned that we could have `formState` with required but not set that on the `FormLine` - maybe I'm not following.

KoltonG (commented):
I might not be following on why the label prefix being determined by FormLine vs the Bound fields which have access to formState?

src/inputs/NumberField.tsx
stephenh: It does happen that way:

* A page has `const config = { firstName ... rules: [required] }`
* The page creates a `formState`
* The page does `BoundTextField field=formState.firstName`
* `BoundTextField` sets `TextField.required` based on `firstName.required`

.storybook/main.js
stephenh: Yep, done

src/components/Label.tsx
```diff
@@ -6,15 +6,18 @@ interface LabelProps {
...
 }
 
 /** An internal helper component for rendering form labels. */
-export function Label({ labelProps, label, hidden, ...others }: LabelProps) {
+export function Label(props: LabelProps) {
+  const { labelProps, label, hidden, suffix, ...others } = props;
   const labelEl = (
     <label {...labelProps} {...others} css={Css.dib.sm.gray700.mbPx(4).$}>
       {label}
+      {suffix && ` ${suffix}`}
```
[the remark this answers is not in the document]
stephenh: Right, this is an open question out to UX; going to see what they say, and keep the defaults to be "no suffix" for now.

I'd still like to land this infra though, and can adapt it as UX decides what they'd like.

src/inputs/NumberField.tsx
stephenh: (So, more concretely, this `NumberField` component itself isn't using `formState` b/c our `src/inputs` fields are decoupled from `formState`, and it's the `Bound...Field`s that are the `formState` <-> `src/inputs` glue. So that mapping of "take the form fields `required`-ness, based on the rule", happens in the `BoundNumberField`.)

stephenh:
> I might not be following on why the label prefix being determined
> by FormLine vs the Bound fields which have access to formState?

In this case, `FormLines` is just a set to set the _preferences_ / settings for what the prefixes might be, and the Bound fields still determine "oh this specific field is / is not required".

I could have added another way to set the `FormContext` i.e. by a dedicated `FormContext.Provider` call, but it seemed cute to do that via a prop on the existing `FormLines` component, in terms of where the caller would probably know/what to set those preferences on a per-form basis.

Again with the disclaimer that I'd love for "customizable suffixes" to just go away entirely, and we hard-code "required always means * and optional always means Opt", but I'm anticipating UX asking for per-form behavior.

stephenh:
I'm using some of the form layout for a contact modal that I need to get out for review, so going to merge. All of the suffix/required/optional generation is opt-in / no breaking changes, and pending decisions from design.

PR #194, opened 2021-07-27
fix: Fix RichTextField blur using stale handler.

stephenh:


src/components/RichTextField.tsx
```diff
@@ -55,6 +55,10 @@ export function RichTextField(props: RichTextFieldProps) {
   // Use a ref for onChange b/c so trixChange always has the latest
   const onChangeRef = useRef<RichTextFieldProps["onChange"]>(onChange);
   onChangeRef.current = onChange;
+  const onBlurRef = useRef<RichTextFieldProps["onBlur"]>(onBlur);
+  onBlurRef.current = onBlur;
+  const onFocusRef = useRef<RichTextFieldProps["onFocus"]>(onFocus);
+  onFocusRef.current = onFocus;
```
stephenh, line 61: We only hook the `trix-change` / `trix-blur` / `trix-focus` handlers once in a `useEffect`, because we don't control the lifetime of the trix JS object, which is fine, but means we need to use Refs to ensure we're always calling the latest callback that the user provided.

src/forms/BoundRichTextField.tsx
```diff
@@ -24,7 +24,7 @@ export function BoundRichTextField(props: BoundRichTextFieldProps) {
           // TODO: Potentially support this in the future?
           // errorMsg={field.touched ? field.errors.join(" ") : undefined}
           onBlur={() => field.blur()}
-          onFocus={() => field.blur()}
+          onFocus={() => field.focus()}
```
stephenh, line 27: Also fixing a copy/paste bug while here

DeanGilewicz (approved):
Looks good to me and thanks for providing the context for this change in your comments

PR #198, opened 2021-07-29
feat: Add dedicated cell sortValue.

stephenh:
Previously we had just `value` which was used for both filtering + sorting.

But you may want filtering to use text like "Accepted" and "Rejected" but, for whatever reason, have "Rejected" sort before "Accepted" so have sort values of like 2 and 1 respectively.

### DIFF
--- a/src/components/GridTable.tsx
+++ b/src/components/GridTable.tsx
@@ -732,6 +732,8 @@ export type GridCellContent = {
   alignment?: GridCellAlignment;
   /** Allow value to be a function in case it's a dynamic value i.e. reading from an inline-edited proxy. */
   value?: MaybeFn<number | string | Date>;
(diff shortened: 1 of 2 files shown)


src/components/GridTable.tsx
```diff
@@ -1079,7 +1081,19 @@ function sortBatch<R extends Kinded>(
   });
 }
 
-function maybeValue(value: ReactNode | GridCellContent): any {
+/** Look at a row and get its sort value. */
+function sortValue(value: ReactNode | GridCellContent): any {
+  // Check sortValue and then fallback on value
+  const maybeFn = value && typeof value === "object" && "value" in value ? value.sortValue || value.value : value;
```
KoltonG, line 1087: Neat! Nice re-use of code!

KoltonG (approved):
🍣

PR #201, opened 2021-07-30
chore: Add story for default style.

stephenh:
Just to make it easier to compare/contrast default vs. condensed.


stephenh:
Really simple copy/paste so going to merge...

PR #202, opened 2021-08-04
feat: Add Button.disabled can be a tooltip

stephenh:
This is a) a blatant copy/paste from one of @DeanGilewicz 's WIP PRs + b) a long-standing need to be able to `s/BP Button/Beam Button` in BP, because we have quite a few places that use current BP Button's `disabledReason` prop.

### DIFF
--- a/src/components/Button.stories.tsx
+++ b/src/components/Button.stories.tsx
@@ -4,7 +4,7 @@ import { Button, ButtonProps } from "src";
 import { Css } from "src/Css";
 
 export default {
-  title: "Components/Buttons",
+  title: "Components/Button",
   component: Button,
   args: { onClick: action("onPress") },
   argTypes: {
--- a/src/components/Button.stories.tsx
+++ b/src/components/Button.stories.tsx
@@ -20,7 +20,7 @@ export default {
   },
 } as Meta<ButtonProps>;
 
-export function Buttons(args: ButtonProps) {
+export function ButtonVariations(args: ButtonProps) {
   const buttonRowStyles = Css.df.childGap1.my1.$;
   return (
     <div css={Css.dg.flexColumn.childGap2.$}>
@@ -129,4 +129,8 @@ export function Buttons(args: ButtonProps) {
       </div>
     </div>
   );
+}
+
+export function ButtonWithTooltip() {
+  return <Button disabled={true} disabledReason="You cannot currently perform this operation." label="Upload" />;
 }
--- a/src/components/Button.stories.tsx
+++ b/src/components/Button.stories.tsx
@@ -132,5 +132,18 @@ export function ButtonVariations(args: ButtonProps) {
 }
 
 export function ButtonWithTooltip() {
-  return <Button disabled="You cannot currently perform this operation." label="Upload" />;
+  return (
+    <Button
+      disabled={
+        <div>
+          You <b>cannot</b> currently perform this operation because of:
+          <ul>
+            <li>reason one</li>
+            <li>reason two</li>
+          </ul>
+        </div>
+      }
+      label="Upload"
+    />
+  );
 }
--- a/src/components/Button.stories.tsx
+++ b/src/components/Button.stories.tsx
@@ -132,5 +132,5 @@ export function ButtonVariations(args: ButtonProps) {
 }
 
 export function ButtonWithTooltip() {
-  return <Button disabled={true} disabledReason="You cannot currently perform this operation." label="Upload" />;
+  return <Button disabled="You cannot currently perform this operation." label="Upload" />;
 }
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -2,6 +2,7 @@ import { AriaButtonProps } from "@react-types/button";
 import { ReactNode, RefObject, useMemo, useRef } from "react";
 import { useButton, useFocusRing, useHover } from "react-aria";
 import { Icon, IconProps } from "src";
+import { Tooltip } from "src/components/Tooltip";
 import { Css } from "src/Css";
 import { BeamButtonProps, BeamFocusableProps } from "src/interfaces";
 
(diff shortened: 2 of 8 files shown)


src/interfaces.ts
```diff
@@ -10,6 +10,8 @@ export interface BeamFocusableProps {
 export interface BeamButtonProps {
   /** Whether the interactive element is disabled. */
   disabled?: boolean;
+  /** If set, will show a tooltip about why the button is disabled. */
+  disabledReason?: string;
```
chr1sjf0x, line 14: FWIW: I had been musing with @KoltonG about whether it would be possible/good to have semi-direct integration with potential operations. Like:
```suggestion
  /** If a potential operation is provided, will show a tooltip about why the button is disabled when the operation is not allowed. */ 
  disabled?: boolean | PotentialOperation;
```

I think whether this is a good idea is questionable, since it would directly tie the tooltip content to the potential operation disabled reasons... when in most cases you'd probably want some additional processing of those reasons 🤔 

🤷

chr1sjf0x (approved):
🍣

src/components/Button.stories.tsx
```diff
@@ -130,3 +130,7 @@ export function Buttons(args: ButtonProps) {
     </div>
   );
 }
+
+export function ButtonWithTooltip() {
+  return <Button disabled={true} disabledReason="You cannot currently perform this operation." label="Upload" />;
```
KoltonG, line 135: **Thoughts:** On having the disabled prop take in a `boolean` or a string for the message? Just thinking that disabled and disabledReason would always be in sync.

KoltonG (approved):
Nice! Will be nice to have this built in!

src/interfaces.ts
stephenh: Hrm... yeah, that would be pretty cute, would be a easy if the reasons were just strings:

```
type PotentialOperation2 {
  allowed: Boolean!
  disabledReasons: [Reason!]!
}

type Reason {
  message: String!
  entity: HasBlueprintUrl
}

interface HasBlueprintUrl {
  blueprintUrl: Url!
}
```

But since they're this complex `Reason` type, with pretty bespoke Blueprint-y things in it, yeah I'm not sure (as you already mentioned) we'd want to leak that directly into Beam...

Maybe if it was:

```
disabled: boolean | ReactNode
```

Where if `disabled` is every "not a boolean" / i.e. a `ReactNode` it knows that means "put the react node in the tooltip".

And then we had an adapter like:

```
<Button disabled=renderPotentialOperation(query.canDoFoo) />
```

Where the helper method would do app-specific rendering of the `PotentialOperation2` instead "just some react nodes to put into the tooltip"...

stephenh: @chr1sjf0x / @KoltonG pushed up another version that does:

```
disabled: boolean | ReactNode
```

Wdyt?

src/components/Button.tsx
```diff
@@ -55,6 +52,17 @@ export function Button({
       {endAdornment && <span css={Css.ml1.$}>{endAdornment}</span>}
     </button>
   );
+
+  // If we're disabled b/c of a non-boolean ReactNode, show it in a tooltip
+  if (isDisabled && typeof disabled !== "boolean") {
+    return (
+      <Tooltip title={disabled} delay={100}>
```
KoltonG, line 59: Nice ☝️! Are you facing any _type_ issues with the **Tooltip** `title` prop (I am not sure if that was updated to accept a ReactNode or just a string.

src/components/Tooltip.stories.tsx
```diff
@@ -1,23 +1,31 @@
...
-        </TooltipComponent>
+        </Tooltip>
       ))}
     </div>
   );
 }
+
+export function TooltipDisabled() {
+  return (
+    <Tooltip title="Tooltip Info" disabled={true}>
```
KoltonG, line 27: Interesting! I never knew a **Tooltip** could be disabled!

src/components/Tooltip.tsx
```diff
@@ -10,27 +10,30 @@ import { Css } from "src/Css";
 
 interface TooltipProps {
   /** The content that shows up when hovered */
-  title: string;
+  title: ReactNode;
```
KoltonG, line 13: Thank you! @kbesingeryeh the fix got in!

src/components/Tooltip.tsx
```diff
@@ -10,27 +10,30 @@ import { Css } from "src/Css";
...
+    { delay, isDisabled: disabled },
     state,
     triggerRef,
   );
   const { tooltipProps } = useTooltip(_tooltipProps, state);
 
   return (
     <>
-      {React.cloneElement(children, { ref: triggerRef, ...triggerProps })}
+      <span ref={triggerRef} {...triggerProps}>
```
KoltonG, line 34: **Question:** Any advantage to adding a wrapping `span`?

src/interfaces.ts
```diff
@@ -8,8 +8,12 @@ export interface BeamFocusableProps {
...
 
 export interface BeamButtonProps {
-  /** Whether the interactive element is disabled. */
-  disabled?: boolean;
+  /**
+   * Whether the interactive element is disabled.
+   *
+   * If a ReactNode, it's treated as a "disabled reason" that's shown in a tooltip.
+   */
+  disabled?: boolean | ReactNode;
```
KoltonG, line 16: 💯

src/interfaces.ts
KoltonG: Looks awesome!

src/components/Tooltip.tsx
stephenh: I'm pretty sure `cloneElement` ignores the `ref` prop (both `ref` and `key` are ignored, and the children's original ref/key are still used). I'm not really sure if this fixes anything per se, I'd just noticed that "`cloneElement` doesn't support ref" while doing other things.

KoltonG: It seems that cloneElement supports `ref`s otherwise the Tooltip wouldn't work (maybe its ignored if it is given) 😢 Are you facing a case that you need to pass a ref to the Tooltip component (that might be why, since it's being overwritten)?

Here's an old conversation on the subject https://github.com/facebook/react/issues/8873#issuecomment-275423780

@oagboola and I approached it this way to reduce the wrapping elements needed for the Tooltip and add the Popper ref to the child element (assuming that the child would not have a ref of its own). All good either way!

stephenh: I was just looking at:

https://reactjs.org/docs/react-api.html#cloneelement

"Clone and return a new React element using element as the starting point. The resulting element will have the original element’s props with the new props merged in shallowly. New children will replace existing children. key and ref from the original element will be preserved."

KoltonG: Yeah, it's that preserved behaviour I think is what is special - like if theres are no overrides then it will keep it the original ref. I'll CodeSandbox this for fun!

KoltonG: Here's a CodeSandbox showing that the children's ref is overridden by the clone elements ref! Still a very interesting result! 

https://codesandbox.io/s/reactcloneelement-ref-override-gyyhm

stephenh: Oh I see, it "preserves it" as in "it won't drop the existing one", but you're allowed to override it...playing devils advocate, if the caller did want to use the ref, i.e. the `parentRef` in your example, this approach drops it?

I'll put back though, given it's not broken as I originally though.

PR #204, opened 2021-08-05
fix: Fix filter types with enums.

stephenh:
This one took awhile to figure out, but the key fix was instead of:

```typescript
type X<T> = T extends string ? Array<T> : never;
```

Using:

```typescript
type X<T> = [T] extends [string] ? Array<T> : never;
```

To disable enum filters getting turned into `{ options: Stage.Construction[] } | { options: Stage.PreCon[] }` ... when really we want just `{ options: Stage[] }`.

Huge kudos to SO: https://stackoverflow.com/questions/53996797/typescript-conditional-type-array-of-union-type-distribution


src/components/Filters/types.ts
```diff
@@ -1,22 +1,24 @@
...
 // - V might be `number[]` and could be used for a multiselect that getOptionValue returned numbers
 // - V might be `boolean` and could be used for ...boolFilter...
-export type FilterDef<V> = V extends Array<infer U>
-  ? U extends Key
-    ? { kind: "multi" } & MultiFilterProps<any, U>
-    : never
-  : V extends boolean | undefined
+
+// All of the extra brackets are to avoid `Stage` turning into `Stage.One` | `Stage.Two`:
+// https://stackoverflow.com/questions/53996797/typescript-conditional-type-array-of-union-type-distribution
```
stephenh, line 17: This `FilterDef` change is the core fix; specifically moving from `V extends Key` to `[V] extends [Key]` (actually `[Value]` here as a separate/related clean up).

stephenh:
This is fairly esoteric + need to unblock, so going to merge.

PR #206, opened 2021-08-05
feat: Add ability for ButtonMenu to be disabled

DeanGilewicz:
- https://app.clubhouse.io/homebound-team/story/6162/add-ability-for-buttonmenu-to-be-disabled

### DIFF
--- a/src/components/ButtonMenu.tsx
+++ b/src/components/ButtonMenu.tsx
@@ -21,13 +21,14 @@ interface ButtonMenuProps {
   placement?: "left" | "right";
   // for storybook purposes
   defaultOpen?: boolean;
+  isDisabled?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/ButtonMenu.tsx
```diff
@@ -21,13 +21,14 @@ interface ButtonMenuProps {
   placement?: "left" | "right";
   // for storybook purposes
   defaultOpen?: boolean;
+  isDisabled?: boolean;
```
stephenh, line 24: Minor, but consistency we typically use `disabled`; granted the react-aria API we use internally for some implementation details uses `isDisabled` so you'll see renames from our public API `disabled` --> `isDisabled` where we cross the beam/react-aria API boundaries.

DeanGilewicz: thanks for the insight - I'll update

stephenh (approved):

PR #208, opened 2021-08-06
fix: Break out Superdrawer validation

khalidwilliams:
# [CH5749](https://app.clubhouse.io/homebound-team/story/5749/beam-superdrawer-opens-and-closes-when-previous-next-buttons-are-clicked)

## Description (what was changed/updated/added.. etc…)

This PR fixes an issue with the Superdrawer where the drawer would open and close when the Previous/Next buttons were pressed. The fix was achieved by:

- pulling the Superdrawer validation checks that were previously done within useSuperdrawer's `onClose` method into a new private function, `validateDrawerChange`. This allows the `openInDrawer` method to validate that ta new drawer can be opened, without closing the current drawer. 
- Always having the confirmation modal's `Close` button run `closeModal` 

## Screenshots (if UI was changed)

**Before (buggy behavior)**

https://user-images.githubusercontent.com/36432040/128536372-9373ea48-c16e-4eb6-8992-771110fd59cf.mov

**After**

https://user-images.githubusercontent.com/36432040/128536394-d88e8eb9-8c6c-4e69-a07f-5efc72b5528b.mov

Mental Checklist
- [x] Updated tests for this change.
- [x] Tested changes locally.
~- [ ] Updated Documentation (if necessary).~


ogrodnek:
whoa, sweet @khalidwilliams ! this made my day! :)

src/components/SuperDrawer/ConfirmCloseModal.tsx
```diff
@@ -23,7 +23,10 @@ export function ConfirmCloseModal({ onClose }: ConfirmCloseModalProps) {
       <p css={Css.base.gray700.$}>Any changes you've made so far will be lost.</p>
       <div css={Css.mt4.gap1.$}>
         <Button variant="tertiary" label="Cancel" onClick={closeModal} />
-        <Button label="Close" onClick={onClose} />
+        <Button label="Close" onClick={() => {
+          onClose();
+          closeModal();
```
stephenh, line 28: I wonder if the order of `closeModal` vs. `onClose` matters here? Probably not?

Either way, maybe throw in a comment like:

```
// Order doesn't really matter; close this modal + tell the caller to actually do their close
```

I know this isn't directly part of your PR, but just to clarify some of this code, wdyt about while you're here renaming `ConfirmCloseModalProps.onClose` to `ConfirmCloseModalProps.onConfirm`?

src/components/SuperDrawer/SuperDrawer.stories.tsx
```diff
@@ -182,6 +182,52 @@ const Books: Book[] = [
...
+    header: {},
+    data: { indent: 2, onClick: openRow },
+  };
+
+  return (
+    <div>
+      <h1 css={Css.xl3Em.mb5.$}>Books</h1>
+      <p css={Css.base.mb3.$}>List of books from various authors</p>
+      <GridTable<Row>
+        as="table"
```
stephenh, line 217: Is the `as` meaningful for the test? Maybe drop just for simplicity if not

src/components/SuperDrawer/SuperDrawer.stories.tsx
```diff
@@ -182,6 +182,52 @@ const Books: Book[] = [
...
+        columns={[titleColumn, authorColumn]}
+        rowStyles={rowStyles}
+        rowLookup={rowLookup}
+        rows={[
```
stephenh, line 221: This is just a story, but might be able to be `rows={simpleRows(Books)}`...

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -64,40 +64,48 @@ export function useSuperDrawer(): UseSuperDrawerHook {
   } = useContext(BeamContext);
   const { openModal } = useModal();
 
+  /**
+   * Validates whether a new Superdrawer can be opened / whether the current Superdrawer can be closed.
+   * Breaking the validation into a separate function allows openInDrawer
+   * to validate the drawer's state without closing it
+   * */
+  function validateDrawerChange (onClose: () => void) {
```
stephenh, line 72: Maybe we could make this:

```
function validateDrawerChange(doChange: () => void): void {
  ...
```

Note that besides the `onClose` --> `doChange` rename (open to better names for `doChange`), specifically using a `void` return type...

I know you're just moving this code around, but the previous "sometimes return true, sometimes invoke confirm, sometimes return false" behavior was kind of confusing to me...

I think a potentially better api is to instead of sometimes checking the return value/sometimes not, just assume that `doChange` is goingn to be called...

We might have to change the `useSuperDrawer.closeDrawer` from boolean to void, but I think that'd be a good change to make anyway...

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -64,40 +64,48 @@ export function useSuperDrawer(): UseSuperDrawerHook {
...
+    const contentStackLength = contentStack.current.length;
+    if (contentStackLength === 0) {
+      return true;
```
stephenh, line 75: I.e. here instead of `return true` we'd do `doChange(); return;`

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -64,40 +64,48 @@ export function useSuperDrawer(): UseSuperDrawerHook {
...
+
+    // Attempt to close each drawer details
+    for (let i = contentStackLength - 2; i >= 0; i--) {
+      for (const canCloseDrawerDetail of canCloseDetailsChecks.current[i] ?? []) {
+        if (!canCloseDrawerDetail()) {
+          openModal({
+            title: "Confirm",
+            content: <ConfirmCloseModal onClose={onClose} />,
+          });
+          return false;
```
stephenh, line 86: And then here we'd just `return`

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -64,40 +64,48 @@ export function useSuperDrawer(): UseSuperDrawerHook {
...
+        return false;
+      }
+    }
+  }
+
+
   // Separate close actions to reference then in actions
   const closeActions = useMemo(() => {
     return {
       /** Attempts to close the drawer. If any checks fail, a error modal will appear */
```
stephenh, line 107: While you're here, maybe change `error modal` to `confirmation modal` on the line above?

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -110,6 +118,10 @@ export function useSuperDrawer(): UseSuperDrawerHook {
           // Reset Modal state
           modalState.current = undefined;
         }
+        const maybeValidated = validateDrawerChange(onClose);
+        if (maybeValidated !== undefined) {
+          return maybeValidated
+        }
 
         onClose();
```
stephenh, line 126: I.e. I think this "extra" onClose call could go away because line 121 would be just `validateDrawerChange(onClose)`

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -151,12 +163,17 @@ export function useSuperDrawer(): UseSuperDrawerHook {
         // When opening a new SuperDrawer, check if we can close the previous
         // SuperDrawer content. Bail if it fails
         // TODO: There needs to be try again logic here...
-        if (!closeActions.closeDrawer()) {
-          return false;
+        debugger;
```
stephenh, line 166: Snuck in :-)

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -151,12 +163,17 @@ export function useSuperDrawer(): UseSuperDrawerHook {
...
+        const maybeValidated = validateDrawerChange(openNewModal);
+        if (maybeValidated === false) {
+          return maybeValidated
         }
+        function openNewModal () {
```
stephenh, line 171: Maybe rename to `doOpen()`? Or something without "modal" in it, since that term is also used for `useModal` things.

...or actually maybe even just make it a lambda, i.e. maybe this becomes:

```
openInDrawer(opts) {
  validateDrawerChange(() => {
    contentStack.current = [{ kind: "open", opts }];
  })
};
```

stephenh (approved):
Thanks for tackling this! I really thought fixing the "don't animate on drawer prev/next" was going to require more work; didn't realize it was just because of an "extra" close call.

src/components/SuperDrawer/useSuperDrawer.tsx
khalidwilliams: I like the change! I'll make the update.

KoltonG: +1! Wondering if we can change the name to something that can match validating if we can close the **drawer** and if we can close the **detail drawer**.

I am biased to `canCloseDrawer(() => {...})` and `canCloseDrawerDetail(i, () => {...})`

src/components/SuperDrawer/useSuperDrawer.tsx
KoltonG, line 86: Might be able to move this into it's own function to also be used in the `closeDrawerDetail`

src/components/SuperDrawer/ConfirmCloseModal.tsx
KoltonG: Another thing is that most of our drawer closing logic does include closing the modal. Could we just assume that?

KoltonG (approved):
Looking great! Thank you for the cleanup!

src/components/SuperDrawer/useSuperDrawer.tsx
stephenh: @KoltonG if you're thinking of names for `validateDrawerChange` I was thinking of `maybeChangeDrawer`... or `changeDrawerIfConfirmed`...

khalidwilliams: @KoltonG  are you referring to pulling out the `for` loops that check the drawerDetails, or the whole validation function. 
@stephenh I'm cool with `maybeChangeDrawer`

src/components/SuperDrawer/useSuperDrawer.tsx
khalidwilliams: Good call on the lambda since we don't need that extra `openNewModal` call now.

src/components/SuperDrawer/ConfirmCloseModal.tsx
khalidwilliams: @KoltonG are you saying we shouldn't rename the props?

src/components/SuperDrawer/useSuperDrawer.tsx
khalidwilliams: @KoltonG I realized we get into the same issue of needing to return a `Boolean` from a broken out function, since if we broken this out there would be two function responsible for calling doChange. The child function, called within `maybeChangeDrawer` would need to tell `maybeChangeDrawer` to break as well, otherwise `doChange` could get called twice. 

I might leave out this refactor for the time being at least.

khalidwilliams: Eh, scratch that. Even with the Boolean pattern, this feels clearer
```
    for (let i = contentStackLength - 2; i >= 0; i--) {
      if (!canCloseDrawerDetails(i, doChange)) {
        return;
      }
    }
```

src/components/SuperDrawer/SuperDrawer.stories.tsx
khalidwilliams: We still need the map to give the Books ids, but simpleRows makes it a one-liner!

src/components/SuperDrawer/SuperDrawer.stories.tsx
khalidwilliams: It keeps the styling in line with the other story in this file 🤷🏾‍♂️ I'd argue for keeping it for visual consistency.

stephenh: Cool, sgtm

src/components/SuperDrawer/ConfirmCloseModal.tsx
KoltonG: Sorry! I meant that most, maybe all, of the `onClose` functions (the ones in closeDrawer and closeDrawerDetail) that are passed to the "Confirmation Modal" when a check fails includes logic to remove the modalContent via `modalContent.current = null`. 

If that is the case, just wondering if we still need to be explicit with `closeModal()`? I'm not sure if I missed why we included this.

khalidwilliams: Ahh I see. The doChange function that openInDrawer pases didn't explicitly touch the modalContent. I suppose we could change that, though?

src/components/SuperDrawer/useSuperDrawer.tsx
KoltonG: This was for breaking out the check code for:
- Validating if a drawer detail can be closed
- Validating if a drawer can be closed (which also includes validating that each drawer detail can be closed -> which could use the first validator).

So closeDrawerDetails would be something like:
```
validateDrawerDetailChange(contentStack.current.length - 2, () => {... code to close detail drawer and modal})
```

And the current `validateDrawerChange` would include `validateDrawerDetailsChange` in it's loop to check if each drawer details can be closed!

src/components/SuperDrawer/ConfirmCloseModal.tsx
KoltonG: Got it! Alright if it didn't handle that then let's keep this but make sure we don't have `modalContent.current = null` in the lambda being passed in.

src/components/SuperDrawer/useSuperDrawer.tsx
khalidwilliams: @KoltonG since the `canCloseChecks` are only being iterated over in one place, with the exception of readability, I'm not sure I see a ton of benefit in further breaking things out. Would you want to talk over it at some point today?

PR #216, opened 2021-08-07
feat: Bundle with esbuild for better vite support.

stephenh:
We still produce the current CommonJS output, i.e. for unit tests in node, but also now have the `module: ...` key in package.json that points to our ESM-specific build. 

This mean we don't have to patch Beam anymore in procurement-frontend.


package.json
```diff
@@ -32,6 +34,7 @@
     "format": "prettier --loglevel warn --write \"**/*.{ts,tsx,css,md}\""
   },
   "dependencies": {
+    "@react-aria/utils": "^3.8.0",
```
stephenh, line 37: These two new dependencies were ones that we used, but didn't have in the package.json, so they weren't correctly marked as external.

bdow (approved):

PR #217, opened 2021-08-08
fix: Render TextFields as just text when readOnly.

stephenh:
To support flipping between read/edit on forms that use `BoundTextField` / `BoundTextAreaField`.

### DIFF
--- a/src/inputs/TextFieldBase.tsx
+++ b/src/inputs/TextFieldBase.tsx
@@ -20,7 +20,7 @@ import { useTestIds } from "src/utils/useTestIds";
 interface TextFieldBaseProps
   extends Pick<
       BeamTextFieldProps,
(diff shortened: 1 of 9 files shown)


src/inputs/TextFieldBase.tsx
```diff
@@ -79,26 +80,47 @@ export function TextFieldBase(props: TextFieldBaseProps) {
   return (
     <div css={Css.df.flexColumn.w100.maxw(px(550)).$} {...groupProps}>
       {label && !hideLabel && <Label labelProps={labelProps} label={label} suffix={labelSuffix} {...tid.label} />}
-      <ElementType
```
stephenh, line 82: This diff looks big, but most of this just got indented, recommend the "ignore whitespace" view.

chr1sjf0x (approved):
🍣

PR #219, opened 2021-08-09
fix: Add 'data-readonly' attribute to help with testing.

bdow:
Previously we were able to check for the 'readonly' attribute added to the `<input />` element. Now that the read only version is not an input element, we are adding an explicit data-readonly attribute to help identify that the TextField is in the expected state


khalidwilliams (approved):
Thank you @bdow !

DeanGilewicz (approved):

PR #220, opened 2021-08-10
fix: Add check to see if content stack is empty in useSuperDrawer.closeDrawer

khalidwilliams:
Previously useSuperDrawer's `closeDrawer` method assumed that the content stack had at least one entry. However, a [recent update](https://github.com/homebound-team/beam/releases/tag/v2.41.2) to the control flow of `useSuperDrawer` requires `closeDrawer`'s `onClose` callback to double check that the content stack has content, in case `closeDrawer` is called manually.


bdow (approved):

src/components/SuperDrawer/useSuperDrawer.test.tsx
```diff
@@ -149,6 +149,17 @@ describe("useSuperDrawer", () => {
...
   });
+
+  it("should not throw when attempting to closeDrawer with an empty content stack", () => {
+    // Given a useSuperDrawer and BeamContext hook
+    const hook = renderHook(useSuperDrawer, { wrapper }).result.current;
+    // And a closed SuperDrawer
+    // When we call onClose
+    // Then we do not expect to have an error thrown
+    act(() => {
+      expect(hook.closeDrawer).not.toThrow("Cannot read property 'kind' of undefined");
```
stephenh, line 160: Just curious, but if this threw "throw new Error(bad)" wouldn't that still pass this assertion, b/c the threw something that was not "Cannot read property..."? 

Like maybe just `.not.toThrow()` would be safer?

Seems like I should know off the top of my head, but I'm not sure...

stephenh (approved):

src/components/SuperDrawer/useSuperDrawer.test.tsx
khalidwilliams: That's a good call... the type error is the specific one I was having an issue with, but surfacing any other unexpected ones is safer for sure.

PR #221, opened 2021-08-10
fix: Revert "feat: Bundle with esbuild for better vite support."

stephenh:
Reverts homebound-team/beam#216


bdow (approved):
bummer.

KoltonG (approved):
Hopefully, we can get this working again!

PR #226, opened 2021-08-11
fix: SelectField interaction fixes

bdow:
Fix 1 - Handling focus state is now done without the use of React-Aria. It wasn't needed to use the focus ring, as the focus is displayed regardless of how the user interacts with it. (FocusRing is specifically for displaying keyboard focus). Instead we handle setting a focus value locally via setState
Fix 2 - Reset the SelectField/MultiSelectField input's value onBlur if not properly set.


stephenh (approved):
Nice! Really glad this fix wasn't that bad. Makes sense just moving focus to something we control.

The interactions here are nuanced enough between us / react-aria / react-stately / etc. let's cover with a test but then just ship it!

PR #228, opened 2021-08-12
fix: Fix filtering on GridCellContent.content.

stephenh:


bdow (approved):

src/components/GridTable.tsx
```diff
@@ -1118,7 +1120,14 @@ function sortValue(value: ReactNode | GridCellContent): any {
 
 /** Look at a row and get its filter value. */
 function filterValue(value: ReactNode | GridCellContent): any {
-  const maybeFn = value && typeof value === "object" && "value" in value ? value.value : value;
+  let maybeFn = value;
+  if (value && typeof value === "object") {
+    if ("value" in value) {
+      maybeFn = value.value;
+    } else if ("content" in value) {
+      maybeFn = value.content;
```
stephenh, line 1128: We weren't looking for `{ content: ... }`. I think the assumption was that `{ content: ... }` is usually react components, but I've got a usage that is `{ content: "some string"; indent: 1 }` so looking at content is worthwhile.

PR #232, opened 2021-08-13
feat: Add GridTable helpers

bdow:
Adds column helpers for supplying default alignment for the entire column based on content type.
Adds CollapseToggle component for use with GridTable


src/components/Table/columns.tsx
```diff
@@ -0,0 +1,25 @@
+import { GridColumn, Kinded } from "src/components/Table/GridTable";
+
+/**
+ * Provides default styling for a GridColumn representing a Date
+ * Defaults to a `0fr` grid-template-column width, which results in the column taking up only the width required to fit the content.
```
stephenh, line 5: Interesting; I know we've tried a few different approaches to widths...I think started out with `fr`, then did some `px`, then some of the min-max+fit-content stuff you add; the last few I'd moved back to `fr` b/c fr is "static" insofar as it doesn't change based on the content of the table

...but then `0fr` sounds different, i.e. "based on content" == "column widths might change as the user type-ahead filters"?

I'm good with trying this, just wondering.

stephenh (approved):

src/components/Table/columns.tsx
bdow: Hmm.. just playing around with putting an input field in here and it doesn't look great. Would like to maybe catch up on where we're at with deciding default template column sizes.

bdow: Yea, nevermind. Doing the `0` thing was a bit too cute. It ends up wrapping the table row's content if there is a non-breaking space anywhere in the content. It looked fine for our date & numeric values, but potentially we might have a space in these columns and wouldn't want the content to wrap.

src/components/Table/CollapseToggle.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export interface GridTableCollapseToggleProps {
+  isHeader?: boolean;
+}
+
+export function CollapseToggle({ isHeader }: GridTableCollapseToggleProps) {
+  const { isCollapsed, toggleCollapse } = useContext(GridCollapseContext);
+  const iconKey = isCollapsed ? "chevronRight" : "chevronDown";
+  const headerIconKey = isCollapsed ? "chevronsRight" : "chevronsDown";
+  return <IconButton onClick={toggleCollapse} icon={isHeader ? headerIconKey : iconKey} />;
```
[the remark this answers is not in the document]
bdow: Yea, the problem is the table cell's have this default padding. I'm hoping that padding definitions can be worked out as part of the redesign of tables. I think we're getting by with our overrides for now that I wouldn't want to do more work trying to figure this out correctly, but instead we can bring this up with design as part of the redesign of tables.

PR #234, opened 2021-08-13
feat: Allow using Tabs/TabContent separately.

stephenh:


src/components/Tabs.tsx
```diff
@@ -29,33 +29,40 @@ export interface TabsProps<V extends string, X extends Properties> {
...
 export function TabsWithContent<V extends string, X extends Only<TabsContentXss, X>>(props: TabsProps<V, X>) {
-  const { selected, tabs, contentXss = {}, ...others } = props;
-  const selectedTab = tabs.find((tab) => tab.value === selected) || tabs[0];
-  const tid = useTestIds(others, "tab");
-
-  const onlyOneTabEnabled = tabs.filter((t) => !t.disabled).length === 1;
-
+  const onlyOneTabEnabled = props.tabs.filter((t) => !t.disabled).length === 1;
   return (
     <div>
```
bdow, line 39: **suggestion:** Can we remove this surrounding div and replace with a fragment while we're in here?

src/components/Tabs.tsx
```diff
@@ -29,33 +29,40 @@ export interface TabsProps<V extends string, X extends Properties> {
...
+  const { selected, tabs, contentXss = {} } = props;
+  const selectedTab = tabs.find((tab) => tab.value === selected) || tabs[0];
+  return (
+    <div
+      aria-labelledby={`${selectedTab.value}-tab`}
+      id={`${selectedTab.value}-tabPanel`}
+      role="tabpanel"
+      tabIndex={0}
+      {...tid.panel}
+      css={{ ...Css.mt2.$, ...contentXss }}
```
bdow, line 57: If we allow `contentXss` to supply more than just `Margin`, it might allow us to get the scrolling behavior of nested content within the tab panel as we want.

src/components/Tabs.tsx
stephenh: Yep, done

bdow (approved):

PR #235, opened 2021-08-13
fix: Allow conditional form lines in JSX.

stephenh:


src/forms/FormLines.test.tsx
```diff
@@ -0,0 +1,20 @@
+import { FormLines } from "src/forms/FormLines";
+
+describe("FormLines", () => {
+  it("can be used with boolean expressions", () => {
+    const lines = (
+      <FormLines>
+        {false && <div>maybe line one</div>}
```
stephenh, line 7: The return value of this is like `boolean | element` which broke the previous typing.

And earlier prototype of FormLines really wanted to restrict what its children were, but now we're more lenient and use things like `FieldGroup`.

PR #237, opened 2021-08-17
feat: CH-6240: Update Beam Button to have feature parity with BP buttons.

bdow:
Allow Beam buttons to render as links, both absolute and relative (using react-router).
Allow Beam buttons to render tooltips for non-disabled buttons

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -1,5 +1,4 @@
 import { AriaButtonProps } from "@react-types/button";
-import { PressEvent } from "@react-types/shared";
 import { ReactNode, RefObject, useMemo, useRef } from "react";
 import { useButton, useFocusRing, useHover } from "react-aria";
 import { Link } from "react-router-dom";
@@ -9,19 +8,16 @@ import { Css } from "src/Css";
 import { BeamButtonProps, BeamFocusableProps } from "src/interfaces";
 import { isAbsoluteUrl, noop } from "src/utils";
 
-export interface ButtonProps extends Omit<BeamButtonProps, "onClick">, BeamFocusableProps {
+export interface ButtonProps extends BeamButtonProps, BeamFocusableProps {
   label: string;
   variant?: ButtonVariant;
   size?: ButtonSize;
   icon?: IconProps["icon"];
+  /** Displays contents after the Button's label. Will be ignored for Buttons rendered as a link with an absolute URL */
   endAdornment?: ReactNode;
   /** HTML attributes to apply to the button element when it is being used to trigger a menu. */
   menuTriggerProps?: AriaButtonProps;
   buttonRef?: RefObject<HTMLElement>;
-  /** If function, then it is the handler that is called when the press is released over the target. Otherwise if string, it is the URL path for the link */
-  onClick?: ((e: PressEvent) => void) | string;
-  /** Text to be shown via a tooltip when the user hovers over the button */
-  tooltip?: ReactNode;
 }
 
 export function Button(props: ButtonProps) {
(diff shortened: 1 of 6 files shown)


src/components/Button.tsx
```diff
@@ -1,62 +1,92 @@
...
   variant?: ButtonVariant;
   size?: ButtonSize;
   icon?: IconProps["icon"];
   endAdornment?: ReactNode;
   /** HTML attributes to apply to the button element when it is being used to trigger a menu. */
   menuTriggerProps?: AriaButtonProps;
-  buttonRef?: RefObject<HTMLButtonElement>;
+  buttonRef?: RefObject<HTMLElement>;
+  /** If function, then it is the handler that is called when the press is released over the target. Otherwise if string, it is the URL path for the link */
+  onClick?: ((e: PressEvent) => void) | string;
```
bdow, line 22: Moved the `onClick` definition here to support `string` as well as a callback function. I haven't put it into `BeamButtonProps` interface as that is also used by IconButton. However, if we think we'd want to support URLs/paths on IconButton, then once the implementation of this is flushed out via discussions on this PR, then I'm happy to add support of links to IconButtons as well.

src/components/Tooltip.tsx
```diff
@@ -67,7 +63,12 @@ function Popper({ triggerRef, content, placement = "auto" }: PopperProps) {
   });
 
   return (
-    <div ref={popperRef} style={styles.popper} {...attributes.popper} css={Css.bgGray900.white.px1.py("4px").br4.xs.$}>
+    <div
+      ref={popperRef}
+      style={{ ...styles.popper, maxWidth: "320px" }}
```
bdow, line 68: Adding a max-width here as the tooltip could get pretty long if the `content` didn't explicitly include line breaks. Waiting to confirm with design on what the right size is here.

src/components/Button.tsx
```diff
@@ -1,62 +1,92 @@
...
+        <Link {...buttonAttrs} to={onPress} className={navLink}>
+          {buttonContent}
+        </Link>
+      )
+    ) : (
+      <button {...buttonAttrs}>{buttonContent}</button>
+    );
+
+  // If we're disabled b/c of a non-boolean ReactNode, or the caller specified tooltip text, then show it in a tooltip
+  if ((isDisabled && typeof disabled !== "boolean") || tooltip) {
```
bdow, line 87: There are now two ways of including a Tooltip, which I'm not too excited about. Feels more like bloat of the component. Though, continuing to allow the `disabled` property to be both a boolean and the Tooltip content does save an additional property for buttons utilizing `disabled` as the Tooltip content. Not if that "saving an extra property" is worth the additional logic in the component. Happy to hear other thoughts.

src/components/Button.tsx
```diff
@@ -1,62 +1,92 @@
...
+      ...(isFocusVisible ? focusRingStyles : {}),
+    },
+  };
 
-  // If we're disabled b/c of a non-boolean ReactNode, show it in a tooltip
-  if (isDisabled && typeof disabled !== "boolean") {
+  const button =
+    typeof onPress === "string" ? (
+      isAbsoluteUrl(onPress) ? (
+        <a {...buttonAttrs} href={onPress} className={navLink} target="_blank" rel="noreferrer noopener">
```
bdow, line 74: This is making an assumption that every absolute URL should open in a new tab. Because of that assumption, I think we'd also be safe to always show `<Icon icon="linkExternal" />` within the button. I've added this in on the Storybook examples via `endAdornment`, though probably just have it done by default.

src/components/Button.tsx
stephenh: Yeah, we didn't anticipate "tooltips for non-disabled buttons" when we first added this...I'm good with the current "there are technically two ways" approach.

stephenh (approved):

src/components/Button.tsx
bdow: Added to IconButton as well.

src/components/Tooltip.tsx
bdow: Confirmed with design. Will be going with 320px.

PR #238, opened 2021-08-17
feat: Bump node and rtl-util versions

bdow:


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 node-image: &node-image
-  image: circleci/node:14.8.0
+  image: circleci/node:16.4.0
```
stephenh, line 4: Might need to be `cimg/node:16.4.0`

DeanGilewicz (approved):

.circleci/config.yml
bdow: Cool, thanks!

stephenh (approved):

PR #240, opened 2021-08-17
fix: Remove dependency of rtl-util from build

bdow:
Bumping rtl-utils back to 2.46.0.
Remove exporting of Beam's RTL and Storybook helpers, which have a direct dependency on @homebound/rtl-utils. This was causing rtl-util to be bundled with our production builds in frontend applications.


stephenh (approved):
:crossed_fingers:

PR #244, opened 2021-08-18
feat: Allow for dynamic and static height Modals

bdow:
Also adding type of `IconKey`


src/components/Modal/Modal.tsx
```diff
@@ -35,6 +35,7 @@ export interface ModalProps {
  */
 export function Modal(props: ModalProps) {
   const { title, size = "md", content, forceScrolling } = props;
+  const isFixedHeight = typeof size !== "string";
```
stephenh, line 38: Won't the `size = md` right before this throw off the "default is dynamic"?

stephenh: Oh, this is "the default _width_" is will `md`, but height will default..

bdow: I don't think so? Maybe I'm missing something?
The height will be dynamic if the `size` attribute is defined as a `ModalSize`. If no size is passed in, then it defaults to a ModalSize and that allow for dynamic by default.
If `size` is not a `string`, meaning it has been defined as an `object`, with `width` and `height`, that is when we treat it as `isFixedHeight`.

bdow: Right right right. Okay. Will add in the comment.

stephenh: Yep, I get it now

stephenh (approved):

PR #245, opened 2021-08-19
feat: Allow toggle filter's enabled value to be 'false'

bdow:


src/components/Filters/FilterModal.tsx
```diff
@@ -18,14 +18,22 @@ export function FilterModal<F>(props: FilterModalProps<F>) {
   const [modalFilter, setModalFilter] = useState<F>(filter);
 
   const updateFilter = useCallback((currentFilter: F, key: keyof F, value: any | undefined) => {
```
bdow, line 20: Would like to try and make a reusable `updateFilter` function. There is enough logic in here now that we should share it between the `FilterModal` and `Filter`. 
They are accomplishing the same task, but just using different functions to set the filter: `setModalFilter` vs `onChange`.

bdow:
Another wrinkle similar to this...:
We have another filter called "teamMembersByRole", where the value would look something like:
```
[{ userId: "iu:1", role: ProjectRole.ProjectManager }, ... ]
```
Though, for the MultiSelectField we'd want to just show a list of InternalUsers. So the value of the selected options (["iu:1", "iu:2", ...]` and the filter's value (`[{ userId: "iu:1", role: Foo }, ...]` do not match. 

Gets a little tricky - we could do something like:
```const teamMembersByRole = multiFilter({
      label: "Project Manager",
      options: internalUsers.map((iu) => ({ userId: iu.id, role: ProjectRole.ProjectManager, name: iu.name })) ?? [],
      getOptionValue: (o) => o.userId,
    });
```
But we really need the `getOptionValue` to be an object, as that is the value we want to send over the wire to the filter... thinking, thinking,...

stephenh (approved):

PR #252, opened 2021-08-23
feat: Spike addressing layouts with internal scrolling containers

bdow:


src/components/Filters/Filters.stories.tsx
```diff
@@ -22,6 +22,7 @@ export default {
   component: Filters,
   title: "Components/Filters",
   decorators: [withDimensions(), withRouter(), withBeamDecorator],
+  parameters: { layout: "fullscreen" },
```
bdow, line 25: Prevents the `padding: 1rem;` from being placed on the story's `<body />` tag. I thought about removing it everywhere, but I think it is better to only remove it in stories where the padding causes issues.

src/components/Layout/NestedScrollLayoutContext.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface NestedScrollLayoutContextProps {
+  scrollableEl: HTMLElement;
+}
+
+export const NestedScrollLayoutContext = createContext<NestedScrollLayoutContextProps>({
+  scrollableEl: document.createElement("div"),
+});
+
+// type NestedScrollLayoutXss = Xss<Margin | Padding | "flexGrow" | "flexShrink">;
+type NestedScrollLayoutXss = Xss<any>;
```
bdow, line 13: I decided that any styles should be able to apply to the wrapping element here. For example, the TestSideNav component needs to be able to have some of its content scrollable, while the top part fixed. Though, the whole SideNav component can be toggled in/out of view, so we need to wrap that transition logic around everything within `NestedScrollLayoutContextProvider`'s root element.

src/components/Layout/NestedScrollLayoutContext.tsx
```diff
@@ -0,0 +1,38 @@
+import { createContext, PropsWithChildren, useEffect, useMemo, useRef } from "react";
+import { Css, Only, Xss } from "src/Css";
+
+interface NestedScrollLayoutContextProps {
+  scrollableEl: HTMLElement;
+}
+
+export const NestedScrollLayoutContext = createContext<NestedScrollLayoutContextProps>({
+  scrollableEl: document.createElement("div"),
```
bdow, line 9: Wondering if this really needs to be reset in the `NestedScrollLayoutContextProvider` function?

src/components/Layout/TestLayout.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { PreventScrollContainer } from "src/components/Layout/PreventScrollContainer";
+import { ScrollableContent } from "src/components/Layout/ScrollableContent";
+import { Css, Margin, Only, Padding, Xss } from "src/Css";
+import { zeroTo } from "src/utils/sb";
+
+export function TestLayout({ children }: PropsWithChildren<any>) {
+  return (
+    <PreventScrollContainer>
+      <TestTopNav />
+      <main css={Css.df.mh0.$}>{children}</main>
```
bdow, line 13: Both `df` and `mh0` are required here to ensure the main content shows the scrollbar and doesn't exceed the remaining height of the viewport.

bdow: Since this is necessary to do along side of `<PreventScrollContainer />`, maybe  this `<main />`(or optionally a different element) with these styles should be included via something like `<PreventScrollContainerContent />`? Hmm.. not sure about this. Would be nice if all apps didn't need to know which styles will allow the scrolling within `main`, but can't think of any good naming.

src/components/Layout/PreventScrollContainer.tsx
```diff
@@ -0,0 +1,7 @@
+import { PropsWithChildren } from "react";
+import { Css } from "src/Css";
+
+/** Intended to wrap the whole application to prevent the brower's native scrolling behavior while also taking the full height of the viewport */
+export function PreventScrollContainer({ children }: PropsWithChildren<{}>) {
```
stephenh, line 5: Kinda wonder if this could be inside of `NestedScrollLayoutContextProvider`? Like wouldn't they be used in pairs where a top-level `PreventScrollContainer` would have an immediate child of `NestedScrollLayoutContextProvider`?

Or maybe not, looking at the stories, still reading...

bdow: That is what I was thinking too at first. Though, because I wanted to use `NestedScrollLayoutContextProvider` twice, once in the SideNav, and another within the page's content. Because of this, I figured it'd be better not to wrap the whole application, but rather wrap it further down into the page where we know the scrolling will take place.
With that said, this was just a first go. I'm happy to explore other approaches.

src/components/Layout/Layout.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+function ExamplePageComponent({ contentAboveTable }: { contentAboveTable?: boolean }) {
+  const tabs: Tab[] = [
+    { value: "overview", name: "Overview", render: () => <></> },
+    { value: "lineItems", name: "Line Items", render: () => <></> },
+    { value: "history", name: "History", render: () => <></> },
+  ];
+  return (
+    <NestedScrollLayoutContextProvider>
```
bdow, line 49: I'm thinking this would move up one more level and probably be in like a `PageLayout` or `ProjectLayout` component. That way not _every_ page needs to remember to wrap itself in this... though every page would need to remember to wrap itself in `<ScrollableContent />` where applicable.

src/components/Layout/NestedScrollLayoutContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+});
+
+// Allow any css to be applied to the NestedScrollLayout container.
+type NestedScrollXss = Xss<any>;
+interface NestedScrollLayoutContextProviderProps<X> {
+  xss?: X;
+}
+
+export function NestedScrollProvider<X extends Only<NestedScrollXss, X>>({
```
stephenh, line 18: I think it'd be nice to hide the `Provider` / "this uses `content`" aspect of this name...

Maybe just `ScrollableParent`? To match the `ScorllableContent`?

src/components/Layout/NestedScrollLayoutContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const NestedScrollLayoutContext = createContext<NestedScrollLayoutContextProps>({
+  scrollableEl: document.createElement("div"),
+});
+
+// Allow any css to be applied to the NestedScrollLayout container.
+type NestedScrollXss = Xss<any>;
+interface NestedScrollLayoutContextProviderProps<X> {
+  xss?: X;
```
stephenh, line 15: Since you're not putting any limits on this (which I'm good with, makes sense for this), you could do just `xss?: Properties` w/ the `Properties` from the `Css.ts` file.

And skip the `type ...` on line 13`, the `Only` on line 18, etc...

src/components/Layout/Layout.stories.tsx
```diff
@@ -0,0 +1,202 @@
+import { Meta } from "@storybook/react";
+import { PropsWithChildren, ReactNode, useState } from "react";
+import { IconButton } from "src/components/IconButton";
+import { NestedScrollProvider } from "src/components/Layout/NestedScrollLayoutContext";
+import { PreventBrowserScroll } from "src/components/Layout/PreventBrowserScroll";
+import { ScrollableContent } from "src/components/Layout/ScrollableContent";
```
stephenh, line 6: Maybe import the three of these lines 4-6 from `src/index` to highlight that they're exported (i.e. I don't think they are now :-D ).

stephenh (approved):
Looks great, excited to see how this plays out in BP.

PR #253, opened 2021-08-23
fix: Fix null is valid value, actionColumn defaults to not sorting.

stephenh:


stephenh:
Minor fixes and want to include my BP beam bump, so going to merge.

PR #254, opened 2021-08-24
fix: Add Create Portal to Tooltip component

kbesingeryeh:
Added a fixed strategy so the tooltip component will always sit on top of anything. 

Screen Shot 2021-08-23 at 3 03 37 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,10 +25,10 @@ _To see the latest designs, check out the [Figma](https://www.figma.com/file/aWU
 
 Beam is specifically "Homebound's design system". Given this extremely narrow purpose, we can lean into the simplicity of:
 
-* We don't need to support everything for everyone
-* We can prefer API/UX consistency & simplicity over configuration & complexity
+- We don't need to support everything for everyone
+- We can prefer API/UX consistency & simplicity over configuration & complexity
 
-The most concrete manifestation of this is that we want to *provide as few props as possible*.
+The most concrete manifestation of this is that we want to _provide as few props as possible_.
 
 Fewer props generally means:
 
@@ -41,7 +41,7 @@ All of these points are generally in stark contrast to traditional, "big" UI too
 
 For them, a MUI application in Company A shouldn't have to look & behave exactly like a MUI application in Company B. Which makes sense.
 
-But for Beam at Homebound, we specifically *want* a component that behaves in our App A to look & behave the same as it does in our App B.
+But for Beam at Homebound, we specifically _want_ a component that behaves in our App A to look & behave the same as it does in our App B.
 
 ## Beam and Open Source
 
--- a/src/components/Modal/useModal.tsx
+++ b/src/components/Modal/useModal.tsx
@@ -1,8 +1,8 @@
 import { useContext, useMemo, useRef } from "react";
 import { BeamContext } from "src/components/BeamContext";
 import { Callback, CheckFn } from "src/types";
+import { maybeCall } from "src/utils";
 import { ModalProps } from "./Modal";
-import {maybeCall} from "src/utils";
 
 export interface UseModalHook {
   openModal: (props: ModalProps) => void;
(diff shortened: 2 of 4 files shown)


ksk5280 (commented):
@kbesingeryeh I think you might want to change the commit title to `fix` instead of `style` because I don't think `style` will trigger a release.

stephenh:
@bdow @KoltonG @kbesingeryeh / all I saw this impressively detailed post go by just a few days ago, which is also about popper coincidentally, and that attaching it to a dummy div via a portal would be a good strategy:

https://atfzl.com/don-t-attach-tooltips-to-document-body

If we did that with a `z-index` that puts it above the main content, would that also fix this issue?

Not really sure the pros/cons of each...

src/forms/BoundToggleChipGroupField.tsx
```diff
@@ -4,10 +4,7 @@ import { ToggleChipGroup, ToggleChipGroupProps } from "src/inputs";
 import { useTestIds } from "src/utils";
 import { defaultLabel } from "src/utils/defaultLabel";
 
-export type BoundToggleChipGroupFieldProps = Omit<
-  ToggleChipGroupProps,
-  "values" | "onChange" | "label"
-> & {
+export type BoundToggleChipGroupFieldProps = Omit<ToggleChipGroupProps, "values" | "onChange" | "label"> & {
```
kbesingeryeh, line 7: This is a prettier change when I ran yarn format

src/forms/BoundToggleChipGroupField.tsx
```diff
@@ -20,15 +17,7 @@ export function BoundToggleChipGroupField(props: BoundToggleChipGroupFieldProps)
...
-      {() => (
-        <ToggleChipGroup
-          label={label}
-          values={field.value || []}
-          onChange={onChange}
-          {...testId}
-          {...others}
-        />
-      )}
+      {() => <ToggleChipGroup label={label} values={field.value || []} onChange={onChange} {...testId} {...others} />}
```
kbesingeryeh, line 20: This is a prettier change when I ran yarn format

src/components/Tooltip.tsx
```diff
@@ -55,22 +56,19 @@ function Popper({ triggerRef, content, placement = "auto" }: PopperProps) {
   const [arrowRef, setArrowRef] = useState<HTMLDivElement | null>(null);
 
   const { styles, attributes } = usePopper(triggerRef.current, popperRef.current, {
+    strategy: "fixed",
```
bdow, line 59: Will this cause the tooltip to stay in place when the user scrolls the page, or scrolls the trigger out of the viewport? I know the tooltip will disappear soon after the user is no longer hovering over the trigger, but just curious if the `fixed` strategy will prevent the Tooltip staying next to the trigger?

src/components/Modal/useModal.tsx
```diff
@@ -1,8 +1,8 @@
 import { useContext, useMemo, useRef } from "react";
 import { BeamContext } from "src/components/BeamContext";
 import { Callback, CheckFn } from "src/types";
+import { maybeCall } from "src/utils";
 import { ModalProps } from "./Modal";
-import {maybeCall} from "src/utils";
```
kbesingeryeh, line 5: This is a prettier change when I ran yarn format

README.md
```diff
@@ -41,7 +41,7 @@ All of these points are generally in stark contrast to traditional, "big" UI too
 
 For them, a MUI application in Company A shouldn't have to look & behave exactly like a MUI application in Company B. Which makes sense.
 
-But for Beam at Homebound, we specifically *want* a component that behaves in our App A to look & behave the same as it does in our App B.
+But for Beam at Homebound, we specifically _want_ a component that behaves in our App A to look & behave the same as it does in our App B.
```
kbesingeryeh, line 44: This is a prettier change when I ran yarn format

src/components/Tooltip.tsx
kbesingeryeh: We removed the fixed strategy and it still works

bdow (approved):
Good quick fix! Thanks!

KoltonG (approved):
🍣

PR #260, opened 2021-08-25
feat: Modal Title to be used as child component

bdow:


src/components/Modal/TestModalContent.tsx
```diff
@@ -17,6 +17,7 @@ export function TestModalContent(props: { initNumSentences?: number; showLeftAct
   const [leftActionDisabled, setLeftActionDisabled] = useState(false);
   return (
     <>
```
stephenh, line 19: Just talking out loud, but I was musing the other day that b/c these `ModalXxx` components are directly siblings of each other, and I think we generally expect them to stay that way, if portals are overkill vs. an approach that looked like:

```
<Modal>
  <ModalHeader>
  <ModyBody>
  <ModalFooter>
<Modal>
```

And then inside of `Modal` it would be like:

```
const header = props.children.find(e => e.type === ModalHeader);
const footer = props.children.find(e => e.type === ModalFooter);
const body = props.children.find(e => e.type === ModalBody);
```

...I'm having a hard time finding a SO post with a canonical example of doing that...

I'd still like to use the portal approach for things like "having SuperDrawerFooter defined several layers down in the react tree and then portallng up as a sibling", i.e. for footers that change based on the step/tab the drawer is showing...

bdow: I would be fine with that. How would our API look then?:
```
// CustomModal implementation, would this need to use the <Modal /> component directly?
function CustomModal() {
  return ( <Modal><ModalHeader>Test</ModalHeader> .... </Modal> );
}

// trigger for Modal, no different than today
openModal({ content: <CustomModal /> });

// BeamContext would just return the `modalRef.current`, rather than <Modal {...modalRef.current} />
{modalRef.current && drawerContentStackRef.current.length === 0 && modalRef.current}
```
?
I _think_ that set up would still work with all of our `canClose` checks.

The above should work if we want the implementer to use the <Modal/> component directly. Otherwise, if we want to leave the BeamContext as is with `<Modal {...modalRef.current} />`, then `CustomModal` would wrapped in a fragment, rather than `<Modal />`. Though, I'm not sure how React sees children when wrapped in a fragment, maybe the fragment is still the one and only child if it wraps the entire component?

stephenh: Yeah, that `CustomModal` + `Modal` + `ModalHeader` etc. is what I had in mind...dunno...

And right, I don't think a component can easily "see through" a `Fragment` to get nested children...

Although there is this library: https://www.npmjs.com/package/react-nanny which purports it is possible. I'm a little hesitant though b/c the guy also has a blog post which asserts that `PropTypes` are necessary to do this sort of "finding children by type" due to minification:

https://mparavano.medium.com/find-filter-react-children-by-type-d9799fb78292

...but I was pretty sure there was a way to do the "find child by type" w/o PropTypes...like maybe I'm assuming create-react-app doesn't minify component names or something like that...

PR #269, opened 2021-08-27
feat: Sort nested rows.

stephenh:
We hadn't gotten around to sorting nested rows.

To sort nested rows, we need to know the original non-flatted structure.

I originally thought we could/should decipher this by going through the flattened list and observing the `kind` values. I.e. if we saw `kind=parent, kind=child, kind=child, kind=grandchild, kind=parent` we could probably guess the tree.

That said, it's a lot simpler to just be passed the tree as-is, because currently all of the code that passes nested rows is `flatMap`-ing from a nested structure anyway.

So, we make a large breaking change by removing the `parentIds` key that used to hint at the structure of parent/children, and instead just add a `GridDataRow.children` to make `rows` now a recursive/non-flat structure.

We still flatten before handing off to the css grid/table/virtuoso renders.

Despite being a in-theory large breaking change, I think updating the code in internal-frontend to "stop flat mapping" will be pretty trivial.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -49,9 +49,9 @@ const nestedColumns: GridColumn<NestedRow>[] = [
   },
   {
     header: () => "Name",
-    parent: (row) => <div>{row.name}</div>,
-    child: (row) => <div css={Css.ml2.$}>{row.name}</div>,
-    grandChild: (row) => <div css={Css.ml4.$}>{row.name}</div>,
+    parent: (row) => ({ content: <div>{row.name}</div>, value: row.name }),
+    child: (row) => ({ content: <div css={Css.ml2.$}>{row.name}</div>, value: row.name }),
+    grandChild: (row) => ({ content: <div css={Css.ml4.$}>{row.name}</div>, value: row.name }),
   },
 ];
 
@@ -649,6 +649,32 @@ describe("GridTable", () => {
     );
     expect(cell(r, 0, 0)).toHaveTextContent("Name");
     expect(cell(r, 1, 0)).toHaveTextContent("bar");
+    expect(row(r, 2)).toBeUndefined();
+  });
+
+  it("can filter child rows", async () => {
+    // Given a parent that won't match the filter
+    const rows: GridDataRow<NestedRow>[] = [
+      { kind: "header", id: "header" },
+      {
+        kind: "parent",
+        id: "1",
+        name: "p1",
+        children: [
+          // And one child does match the filter
+          { kind: "child", id: "p1c1", name: "child foo" },
+          // And the other does not
+          { kind: "child", id: "p1c2", name: "child bar" },
+        ],
+      },
+    ];
+    // When we filter by 'foo'
+    const r = await render(<GridTable filter="foo" columns={nestedColumns} rows={rows} />);
+    // Then we show the header
+    expect(cell(r, 0, 1)).toHaveTextContent("Name");
+    // And the child that matched
+    expect(cell(r, 1, 1)).toHaveTextContent("foo");
+    // And that's it
     expect(row(r, 2)).toBeUndefined();
   });
 
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -294,8 +294,7 @@ describe("GridTable", () => {
         />,
       );
       // Then when sorted by the 2nd column
-      const { sortHeader_1 } = r;
-      click(sortHeader_1);
+      click(r.sortHeader_1);
       // Then the `value: undefined` row is first
       expect(cell(r, 1, 0)).toHaveTextContent("a");
       expect(cell(r, 2, 0)).toHaveTextContent("d");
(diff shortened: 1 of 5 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -297,18 +297,23 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
     }
     return rows;
   }, [columns, rows, sorting, sortState]);
   const noData = !rows.some((row) => row.kind !== "header");
 
-  // For each row, keep it's ReactElement (React.memo'd) + its filter values, so we can re-filter w/o re-evaluating this useMemo.
-  const allRows: RowTuple<R>[] = useMemo(() => {
-    return maybeSorted.map((row) => {
+  // Filter + flatten + component-ize the sorted rows.
+  let [headerRows, filteredRows]: [RowTuple<R>[], RowTuple<R>[]] = useMemo(() => {
```
stephenh, line 312: We used to have two steps, `allRows` was the react components and memo'd, and then we filtered against the flattened list.

When the change to nesting, both filtering and collapsing are best/easiest done pre-flattening, so now we have a single step that both flattens + filters + component-izes things. I'm pretty sure this is a net zero effect in terms of re-renders but we should probably have some like unit tests that use leaky state / global vars to count / assert against re-renders while filters/etc. change...

src/components/Table/GridTable.tsx
```diff
@@ -647,11 +673,11 @@ type DiscriminateUnion<T, K extends keyof T, V extends T[K]> = T extends Record<
  * column being sorted, in which case we use the GridCellContent.value.
  */
 export type GridColumn<R extends Kinded, S = {}> = {
-  [P in R["kind"]]:
+  [K in R["kind"]]:
```
stephenh, line 676: Renamed P to K for no real reason...

src/components/Table/GridTable.tsx
```diff
@@ -647,11 +673,11 @@ type DiscriminateUnion<T, K extends keyof T, V extends T[K]> = T extends Record<
...
     | string
-    | (DiscriminateUnion<R, "kind", P> extends { data: infer D }
-        ? (data: D, id: string) => ReactNode | GridCellContent
-        : (row: DiscriminateUnion<R, "kind", P>) => ReactNode | GridCellContent);
+    | (DiscriminateUnion<R, "kind", K> extends { data: infer D }
+        ? (data: D, row: GridRowKind<R, K>) => ReactNode | GridCellContent
+        : (row: GridRowKind<R, K>) => ReactNode | GridCellContent);
```
stephenh, line 680: Instead of just passing the `DiscriminateUnion` (i.e. the specific kind like `{ kind: parent, data: ParentData }` we are also really passing the `id` + `children` keys that are part of `GridDataRow`, so this is just telling the type system that, i.e. so column renderers could access `children` if they wanted.

src/components/Table/GridTable.tsx
```diff
@@ -759,18 +785,18 @@ type MaybeFn<T> = T | (() => T);
 /**
  * The data for any row in the table, marked by `kind` so that each column knows how to render it.
  *
- * This should contain very little presentation data, i.e. the bulk of the keys should be data for
- * the given `kind: ...` of each row (which we pull in via the `DiscriminateUnion` intersection).
+ * Each `kind` should contain very little presentation logic, i.e. mostly just off-the-wire data from
+ * a GraphQL query.
```
stephenh, line 789: Just copy-editing comments

src/components/Table/GridTable.tsx
```diff
@@ -759,18 +785,18 @@ type MaybeFn<T> = T | (() => T);
...
+ * The presentation concerns instead mainly live in each GridColumn definition, which will format/render
+ * each kind's data for that specific row+column (i.e. cell) combination.
  */
 export type GridDataRow<R extends Kinded> = {
   kind: R["kind"];
   /** Combined with the `kind` to determine a table wide React key. */
   id: string;
   /** A list of parent/grand-parent ids for collapsing parent/child rows. */
-  parentIds?: string[];
+  children?: GridDataRow<R>[];
```
stephenh, line 799: This is the big API change

src/components/Table/GridTable.test.tsx
```diff
@@ -382,6 +381,51 @@ describe("GridTable", () => {
...
+      expect(cell(r, 1, 0)).toHaveTextContent("1");
+      expect(cell(r, 2, 0)).toHaveTextContent("1");
+      expect(cell(r, 3, 0)).toHaveTextContent("2");
+      expect(cell(r, 4, 0)).toHaveTextContent("2");
+      expect(cell(r, 5, 0)).toHaveTextContent("1");
+      expect(cell(r, 6, 0)).toHaveTextContent("2");
+      // And when we reverse the sort
+      click(r.sortHeader_0);
+      // Then the data is sorted by 2 (2 1) then 1 (2 1)
+      expect(cell(r, 1, 0)).toHaveTextContent("2");
```
bdow, line 421: nice!

src/components/Table/GridTable.test.tsx
```diff
@@ -608,12 +652,44 @@ describe("GridTable", () => {
     expect(row(r, 2)).toBeUndefined();
   });
 
+  it("can filter child rows", async () => {
```
bdow, line 655: Good test to make sure we still show the parents of rows that match the filter.

bdow (approved):

PR #270, opened 2021-08-29
feat: Support disabled options in Select and MultiSelect fields

bdow:


src/inputs/SelectField.stories.tsx
```diff
@@ -56,6 +56,7 @@ function Template(args: SelectFieldProps<any, any>) {
         <h1 css={Css.lg.$}>{args.compact ? "Compact" : "Regular"}</h1>
         <TestSelectField
           {...args}
+          disabledKeys={["1"]}
```
stephenh, line 59: Maybe `disabledValues` to match `values`?

src/inputs/MultiSelectField.test.tsx
```diff
@@ -104,6 +104,32 @@ describe("MultiSelectFieldTest", () => {
...
+    );
+    // When opening the menu
+    fireEvent.focus(age());
+    const optionTwo = getByRole("option", { name: "Two" });
+    // Then expect the disabled option to have the correct aria attributes
+    expect(optionTwo).toHaveAttribute("aria-disabled", "true");
+    // And when clicking on that option
+    click(optionTwo);
+    // Then the `onSelect` callback is not called
+    expect(onSelect).not.toHaveBeenCalled();
```
stephenh, line 130: What would happen if:

* `values=[a, b]`
* `disabledValues=[b]`
* the user clicks on `c`

Is the new `onSelect` either:

1. `[a, b, c]` (i.e. it adds `c` but ignores that b is both selected but disabled)
2. `[a, c]` (i.e. while adding `c` it also realizes that `b` is disabled and so filters it out)
3. Neither i.e. we blow up with a runtime error if a value is in both `values` and `disabledValues`

...I'm kinda ambivalent, but would like to have a guess of what is correct and cover it with a test.

stephenh (approved):

src/inputs/MultiSelectField.test.tsx
bdow: Very good question, and I don't know the answer. I'll give it a shot, though I'd expect that disabled means you can't change the selection status of the option, vs you cannot select this option. But we'll see...

stephenh: That sounds reasonable to me, you just can't change it...

PR #271, opened 2021-08-29
feat: Refactor filters using oo.

stephenh:
Sometimes FP and ADTs are great, but sometimes the "bring your own
behavior" of OO is great too.

This de-kinds the FilterDefs and instead makes the filters a list
of instances/classes that know how to render themselves.

This removes a lot of the type-inference gunk from FilterDefs where
it was trying to guess what kinds should be supported based on the
type of the filter.

Really we don't care what kind the filter is, as long as it can
handle the given filter value.

It also makes it possible to "bring-your-own-filter", by implementing the `Filter` interface.

The `singleFilter`/etc. factory methods don't have any API changes, so this should be a non-breaking changing.


src/components/Filters/BooleanFilter.tsx
```diff
@@ -0,0 +1,60 @@
...
+  defaultValue?: undefined | boolean;
+}
+
+const defaultBooleanOptions: BooleanOption[] = [
+  [undefined, "Any"],
+  [true, "Yes"],
+  [false, "No"],
+];
+
+export function booleanFilter(opts: BooleanFilterProps): (key: string) => Filter<boolean> {
```
stephenh, line 21: Each filter type is in its own file, and has both the rendering + value updating logic for it, instead of those being spread around in misc places.


Also the `fooFilter` functions technically return a lambda, i..e the `key: string => ...` which is so that the filter classes can take `key` as a constructor, i.e.: `{ marketId: singleSelect(...) }` will be passed `marketId` so that the `SingleFilter` class can use `marketId` primarily to create the default label.

stephenh:
> Sometimes FP and ADTs are great, but sometimes the 
> "bring your own behavior" of OO is great too.

Also full disclaimer that I originally pushed the kind-based approach :-| so mea culpa/etc. :-)

src/hooks/usePersistedFilter.ts
```diff
@@ -25,8 +25,8 @@ export function usePersistedFilter<F>({ storageKey, filterDefs }: UsePersistedFi
     () =>
       Object.fromEntries(
         safeEntries(filterDefs)
-          .filter(([, def]) => def.defaultValue !== undefined)
-          .map(([key, def]) => [key, (def as any).defaultValue]),
+          .filter(([key, def]) => def(key as string).defaultValue !== undefined)
+          .map(([key, def]) => [key, def(key as string).defaultValue]),
```
bdow, line 29: Was this broken?

src/components/Filters/Filters.tsx
```diff
@@ -80,10 +67,14 @@ function Filters<F, G extends Value = string>(props: FilterProps<F, G>) {
   return (
     <div css={Css.df.itemsCenter.childGap1.$} {...testId}>
       {maybeGroupByField}
-      {pageFilters.map((c, idx) => (
-        <div key={idx}>{c}</div>
+
+      {safeEntries(pageFilters).map(([key, f]: [keyof F, Filter<any>]) => (
+        <div key={key as string}>
+          {f.render(filter[key], (value) => onChange(updateFilter(filter, key, value)), testId, false)}
```
bdow, line 73: Aha! I was trying to figure out where/how is the Filter class's `render` method being called with these props? When I first saw the classes, I assumed it was a react Class Component, but now I realize it is just a generic class. That makes more sense.

src/hooks/usePersistedFilter.ts
stephenh: No, but `def.defaultValue` now has to be `def(key).defaultValue` b/c of the "`singleFilter` returns a factory that takes the filter's key" approach.

src/components/Filters/Filters.tsx
stephenh: Oh yeah, that is confusing. :-) At first I had `renderInPage` and `renderInModal` as separate methods, so it was more obviously "not a react component"; but for most filters, those are ~90% the same logic, so I changed it to a single `render` method with a `modal` prop...

Would not be against a "not-`render`" name for that method...ideas?

src/components/Filters/ToggleFilter.tsx
```diff
@@ -0,0 +1,53 @@
...
+        selected={(value || false) === enabledValue}
+        label={this.label}
+        labelStyle={inModal ? "filter" : "inline"}
+        onChange={(value) => {
+          // The switch being "off" might actually mean "true" to the backend.
+          // I.e. "Hide do not use" being "off" --> { includeDoNotUse: true }
+          const backendValue = !value ? !enabledValue : enabledValue;
+          // If the backend value is already the enabled value (i.e. the default), drop it
+          // I.e. `{ includeDoNotUse: false }` --> drop it, otherwise keep it.
+          setValue(backendValue === enabledValue ? undefined : backendValue);
```
bdow, line 35: Makes sense to move this local to the component instead of adding the logic in the `updateFilter`. Thanks for cleaning this all up.

stephenh: Yeah, agreed breaking up `updaterFilter` is nice; the `kind: ...` / ADT idiom really pushed that as the default way to implement things in a way I didn't anticipate.

src/components/Filters/MultiSelectFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+    );
+  }
+
+  get label(): string {
+    return this.props.label || defaultLabel(this.key as string);
+  }
+
+  get defaultValue(): V[] | undefined {
+    return this.props.defaultValue;
+  }
```
bdow, line 69: **thought:** Would it be worth creating a FilterBase class that these can extend for shared methods such as `get label()` and `get defaultValue()`?

src/components/Filters/SingleSelectFilter.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const { label, defaultValue, ...props } = this.props;
+    return (
+      <SelectField<O, V>
+        {...props}
+        compact
+        value={value}
+        label={this.label}
+        inlineLabel
+        sizeToContent={!inModal}
+        onSelect={(value) => setValue(value || undefined)}
```
bdow, line 30: Re: `value || undefined`, thinking about the case where a SelectField might have a boolean value, though I actually don't think that'd be possible, should use BooleanFilter in that case. Hmm... probably fine.

bdow (approved):
Thanks for the clean up and thoughtful refactor. I think this will be much easier for people to wander into and understand

src/components/Filters/Filters.tsx
bdow: `render` makes too much sense to rename. Maybe a comment above the class definition saying "Not a react class component"? I dunno, probably should be easy to realize as it doesn't `extend React.Component`

src/components/Filters/MultiSelectFilter.tsx
stephenh: Yeah, I was worried that was "too OO" :-) but agreed that'd be a nice clean up.

src/components/Filters/SingleSelectFilter.tsx
stephenh: Oh yeah...now that I know `??` works on `null` to, this should be `??`...will work in...

PR #272, opened 2021-08-31
fix: Allow SelectField to hide labels

bdow:


src/forms/FormStateApp.tsx
```diff
@@ -63,6 +63,7 @@ export function FormStateApp() {
   );
 
   const sports = [
+    { id: undefined as any, name: "Undecided" },
```
bdow, line 66: Throwing in some other examples for FormState at the same time.

src/forms/FormStateApp.tsx
```diff
@@ -136,7 +137,14 @@ export function FormStateApp() {
...
           <div>
-            <Button label="Read Only" onClick={() => setReadOnly(!readOnly)} />
+            <div>
+              <Button label="Read Only" onClick={() => setReadOnly(!readOnly)} />
+            </div>
+
+            <div css={Css.mt1.$}>
+              <strong>Form Values:</strong>
+              <pre>{JSON.stringify(formState.value, null, 2)}</pre>
+            </div>
```
bdow, line 147: Printing out the Form State's values as it changes. Found this to be helpful when playing around and understanding what data is stored in formState.

stephenh: Nice addition!

PR #274, opened 2021-09-01
chore: Bump truss.

stephenh:
Feel free to sanity check/push back on whether the abbreviations are too extreme or not.

They are shorter, but dunno, IMO in reading the diff, they do look a lot more "like tachyons", and stand out less, so I'm still +1/think its an improvement...

### DIFF
--- a/src/components/Alert.tsx
+++ b/src/components/Alert.tsx
@@ -10,7 +10,7 @@ export interface AlertProps {
 export function Alert(props: AlertProps) {
   const { children, onClose } = props;
   return (
-    <div css={Css.df.itemsStart.justifyBetween.bgRed100.ba.bRed400.red700.p2.br4.$} role="alert">
+    <div css={Css.df.aifs.jcsb.bgRed100.ba.bRed400.red700.p2.br4.$} role="alert">
       <div>{children}</div>
       <IconButton icon="x" color={Palette.Red700} onClick={onClose} />
     </div>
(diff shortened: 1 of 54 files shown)


src/components/Alert.tsx
```diff
@@ -10,7 +10,7 @@ export interface AlertProps {
 export function Alert(props: AlertProps) {
   const { children, onClose } = props;
   return (
-    <div css={Css.df.itemsStart.justifyBetween.bgRed100.ba.bRed400.red700.p2.br4.$} role="alert">
+    <div css={Css.df.aifs.jcsb.bgRed100.ba.bRed400.red700.p2.br4.$} role="alert">
```
bdow, line 13: I think these abbreviations will help people unfamiliar with the css definitions learn the full terms. For instance, just saying `justifyBetween` isnt' super clear that it becomes `justify-content: space-between`. I think `jcsb` will make people aware of the true output in the actual css rules.

src/components/Modal/Modal.tsx
```diff
@@ -78,12 +78,7 @@ export function Modal(props: ModalProps) {
         <FocusScope contain restoreFocus autoFocus>
           <div
             css={
-              Css.br24.bgWhite.bshModal
-                .maxh("90vh")
-                .df.flexColumn.wPx(width)
-                .mh(px(height))
-                .if(isFixedHeight)
-                .hPx(height).$
+              Css.br24.bgWhite.bshModal.maxh("90vh").df.fdc.wPx(width).mh(px(height)).if(isFixedHeight).hPx(height).$
```
bdow, line 81: **nit:** Could update `mh(px(...))` to `mhPx()` now.

src/components/Alert.tsx
stephenh: Agreed

bdow (approved):

PR #276, opened 2021-09-01
feat: Add 'All' to single select filters.

stephenh:
This matches multi select / boolean filter / etc. behavior, and means users don't have to add their own "All" options that can act slightly weird due to the "empty vs. All vs. actual value" tri-state.

### DIFF
--- a/src/components/Filters/Filters.stories.tsx
+++ b/src/components/Filters/Filters.stories.tsx
@@ -123,39 +123,6 @@ export function GroupBy() {
   );
 }
 
-export function GroupByViewAll() {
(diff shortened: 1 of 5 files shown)


src/components/Filters/Filters.stories.tsx
```diff
@@ -123,39 +123,6 @@ export function GroupBy() {
   );
 }
 
-export function GroupByViewAll() {
```
stephenh, line 126: This was an example story of the previous "you need to add your own all" that we don't need anyway.

bdow (approved):
Looks Good! Nice that it only changes the Filter behavior.

PR #278, opened 2021-09-01
fix: Fix labels showing up in the modal.

stephenh:
Probably going to add a story/test later before merging...


src/components/Filters/BooleanFilter.tsx
```diff
@@ -31,7 +31,8 @@ class BooleanFilter extends BaseFilter<boolean, BooleanFilterProps> implements F
         label={this.label}
         // We use `String(value)` so that `undefined` becomes "undefined"
         value={String(value)}
-        inlineLabel
+        hideLabel={inModal}
+        inlineLabel={!inModal}
```
stephenh, line 35: Makes me wonder about like a `labelStyle = hidden | inline` sort of thing...

PR #280, opened 2021-09-02
fix: Prevent Filter field resizing on blur

bdow:


src/inputs/internal/SelectFieldInput.tsx
```diff
@@ -208,10 +212,22 @@ export function SelectFieldInput<O, V extends Value>(props: SelectFieldInputProp
...
+            // 2. If is MultiSelect and only one option is chosen, then use the length of that option to define the width to avoid size jumping on blur.
+            // 3. Use `nothingSelectedText`
+            // 4. Default to "1"
+            // And do not allow it to grow past a size of 20.
             sizeToContent
-              ? isFocused
-                ? Math.max(String(inputProps.value || "").length, 20)
-                : String(inputProps.value || "").length || 1
+              ? Math.min(
+                  String(
```
bdow, line 223: Wrapping with `String()` specifically as `inputProps.value` can be ` string | number | readonly string[] | undefined`

src/inputs/internal/SelectFieldInput.tsx
```diff
@@ -208,10 +212,22 @@ export function SelectFieldInput<O, V extends Value>(props: SelectFieldInputProp
...
+                    inputProps.value ||
+                      (isMultiSelect && selectedOptions.length === 1 && getOptionLabel(selectedOptions[0])) ||
+                      nothingSelectedText ||
```
stephenh, line 226: I have a suspicion we have this logic of "determine what the text value is" already somewhere else, so would be great to reuse if possible so it stays up to date, but otherwise makes a lot of sense.

(If we can't use the calc elsewhere, i.e. in slack you mentioned "what might _become_ the content of the input", then might an in-source comment about why we can't.)

stephenh (approved):

src/inputs/internal/SelectFieldInput.tsx
bdow: Added a comment as the other logic for determining the input's value is kind of spread out through the SelectFieldBase and intermixed with other state logic for the field.

PR #281, opened 2021-09-06
feat: Nested cards

stephenh:
Screenshot from the story (the story only has a single `name` column):

image

Here with a `gap1` turned on to show the DOM structure:

image

So the DOM is still flat, and we alternate "content rows" and "chrome rows" (that have the open/close rounded borders + spacers).

And note that even though this chrome row is a single div to css grid/react-virtuoso:

image

It is 1 outer div + three inner divs: close the previous row, add a spacer between cards, open the new row.

Also for the content rows:

image
 
The 1st and last cells (in this case it's the same cell b/c there is only 1 column) have N-levels of pad-left/pad-right added around their normal `GridColumn`-generated output, which achieves the nudging-in look.

Other than that, everything is still stock gridtable.

### DIFF
--- a/.storybook/.babelrc
+++ b/.storybook/.babelrc
@@ -4,5 +4,6 @@
     ["@emotion/babel-preset-css-prop"],
     ["@babel/preset-react", { "runtime": "automatic", "importSource": "@emotion/react" }],
     ["@babel/preset-typescript"]
-  ]
+  ],
+  "plugins": ["@babel/proposal-class-properties", "@babel/proposal-object-rest-spread"]
 }
--- a/src/components/Table/GridRowLookup.ts
+++ b/src/components/Table/GridRowLookup.ts
@@ -1,0 +1,84 @@
+import { MutableRefObject } from "react";
+import { VirtuosoHandle } from "react-virtuoso";
+import { DiscriminateUnion, GridColumn, GridDataRow, Kinded, RowTuple } from "src/components/Table/GridTable";
+
+/**
+ * Allows a caller to ask for the currently shown rows, given the current sorting/filtering.
+ *
+ * We will probably end up generalizing this into a GridTableApi that exposes more
+ * actions i.e. scrolling to a row and selection state.
+ */
+export interface GridRowLookup<R extends Kinded> {
+  /** Returns both the immediate next/prev rows, as well as `[kind].next/prev` values, ignoring headers. */
+  lookup(
+    row: GridDataRow<R>,
+    additionalFilter?: (row: GridDataRow<R>) => boolean,
+  ): NextPrev<R> &
+    {
+      [P in R["kind"]]: NextPrev<DiscriminateUnion<R, "kind", P>>;
+    };
+
+  /** Returns the list of currently filtered/sorted rows, without headers. */
+  currentList(): readonly GridDataRow<R>[];
+
+  /** Scroll's to the row with the given kind + id. Requires using `as=virtual`. */
+  scrollTo(kind: R["kind"], id: string): void;
+}
+
+interface NextPrev<R extends Kinded> {
+  next: GridDataRow<R> | undefined;
+  prev: GridDataRow<R> | undefined;
+}
+
+export function createRowLookup<R extends Kinded>(
+  columns: GridColumn<R>[],
+  filteredRows: RowTuple<R>[],
+  virtuosoRef: MutableRefObject<VirtuosoHandle | null>,
+): GridRowLookup<R> {
+  return {
+    scrollTo(kind, id) {
+      if (virtuosoRef.current === null) {
+        // In theory we could support as=div and as=table by finding the DOM
+        // element and calling .scrollIntoView, just not doing that yet.
+        throw new Error("scrollTo is only supported for as=virtual");
+      }
+      const index = filteredRows.findIndex(([r]) => r.kind === kind && r.id === id);
+      virtuosoRef.current.scrollToIndex({ index, behavior: "smooth" });
+    },
+    currentList() {
+      return filteredRows.map((r) => r[0]);
+    },
+    lookup(row, additionalFilter = () => true) {
+      const rows = filteredRows.map((r) => r[0]).filter(additionalFilter);
+      // Ensure we have `result.kind = {}` for each kind
+      const result: any = Object.fromEntries(getKinds(columns).map((kind) => [kind, {}]));
+      // This is an admittedly cute/fancy scan, instead of just `rows.findIndex`, but
+      // we do it this way so that we can do kind-aware prev/next detection.
+      let key: "prev" | "next" = "prev";
+      for (let i = 0; i < rows.length; i++) {
+        const each = rows[i];
+        // Flip from prev to next when we find it
+        if (each.kind === row.kind && each.id === row.id) {
+          key = "next";
+        } else {
+          if (key === "prev") {
+            // prev always overwrites what was there before
+            result[key] = each;
+            result[each.kind][key] = each;
+          } else {
+            // next only writes first seen
+            result[key] ??= each;
+            result[each.kind][key] ??= each;
+          }
+        }
+      }
+      return result;
+    },
+  };
+}
+
+function getKinds<R extends Kinded>(columns: GridColumn<R>[]): R[] {
+  // Use the 1st column to get the runtime list of kinds
+  const nonKindKeys = ["w", "sort", "sortValue", "align"];
+  return Object.keys(columns[0] || {}).filter((key) => !nonKindKeys.includes(key)) as any;
+}
--- a/src/components/Table/GridRowLookup.ts
+++ b/src/components/Table/GridRowLookup.ts
@@ -1,6 +1,7 @@
 import { MutableRefObject } from "react";
 import { VirtuosoHandle } from "react-virtuoso";
 import { DiscriminateUnion, GridColumn, GridDataRow, Kinded, RowTuple } from "src/components/Table/GridTable";
+import { dropChromeRows } from "src/components/Table/nestedCards";
 
 /**
  * Allows a caller to ask for the currently shown rows, given the current sorting/filtering.
@@ -42,14 +43,16 @@ export function createRowLookup<R extends Kinded>(
         // element and calling .scrollIntoView, just not doing that yet.
         throw new Error("scrollTo is only supported for as=virtual");
       }
-      const index = filteredRows.findIndex(([r]) => r.kind === kind && r.id === id);
+      const index = filteredRows.findIndex(([r]) => r && r.kind === kind && r.id === id);
       virtuosoRef.current.scrollToIndex({ index, behavior: "smooth" });
     },
     currentList() {
-      return filteredRows.map((r) => r[0]);
+      return dropChromeRows(filteredRows).map((r) => r[0]);
     },
     lookup(row, additionalFilter = () => true) {
-      const rows = filteredRows.map((r) => r[0]).filter(additionalFilter);
+      const rows = dropChromeRows(filteredRows)
+        .map((r) => r[0])
+        .filter(additionalFilter);
       // Ensure we have `result.kind = {}` for each kind
       const result: any = Object.fromEntries(getKinds(columns).map((kind) => [kind, {}]));
       // This is an admittedly cute/fancy scan, instead of just `rows.findIndex`, but
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -311,9 +311,13 @@ export function NestedCardsFirstCell() {
           <div />
         </div>
         <div css={Css.display("contents").$}>
+          <div data-level="level1" css={Css.pl1.$}>
+            Milestone 1
+          </div>
           <div data-level="level1">Milestone 1</div>
-          <div data-level="level1">Milestone 1</div>
-          <div data-level="level1">Milestone 1</div>
+          <div data-level="level1" css={Css.pr1.$}>
+            Milestone 1
+          </div>
         </div>
 
         {/*Row is not closing, so spacer level1.*/}
@@ -329,11 +333,15 @@ export function NestedCardsFirstCell() {
         </div>
         <div css={Css.display("contents").$}>
           <div data-level="level1" css={Css.pl1.$}>
-            <div data-level="level2">Group 1</div>
+            <div data-level="level2" css={Css.pl1.$}>
+              Group 1
+            </div>
           </div>
           <div data-level="level2">Group 1</div>
           <div data-level="level1" css={Css.pr1.$}>
-            <div data-level="level2">Group 1</div>
+            <div data-level="level2" css={Css.pr1.$}>
+              Group 1
+            </div>
           </div>
         </div>
         <div data-card-close="level2">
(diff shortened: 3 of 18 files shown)


.storybook/.babelrc
```diff
@@ -4,5 +4,6 @@
     ["@emotion/babel-preset-css-prop"],
     ["@babel/preset-react", { "runtime": "automatic", "importSource": "@emotion/react" }],
     ["@babel/preset-typescript"]
-  ]
+  ],
+  "plugins": ["@babel/proposal-class-properties", "@babel/proposal-object-rest-spread"]
```
stephenh, line 8: For some reason this is needed to do `private foo: ...` in classes; really thought the `preset-typescript` would take care of that... in retrospect, maybe our tsconfig output is "too new" and leaving the syntax instead of decompiling it down. But eh, probably fine...

src/components/Table/GridRowLookup.ts
```diff
@@ -0,0 +1,87 @@
+import { MutableRefObject } from "react";
+import { VirtuosoHandle } from "react-virtuoso";
+import { DiscriminateUnion, GridColumn, GridDataRow, Kinded, RowTuple } from "src/components/Table/GridTable";
+import { dropChromeRows } from "src/components/Table/nestedCards";
+
+/**
```
stephenh, line 6: Extracted this file from `GridTable.ts`, nothing new/changed.

src/components/Table/GridTable.stories.tsx
```diff
@@ -199,6 +203,323 @@ export function NestedRows() {
...
+      value: row.name,
+    }),
+    grandChild: (row) => ({
+      content: <div css={Css.xs.$}>{row.name}</div>,
+      value: row.name,
+    }),
+  };
+  const spacing = { brPx: 4, pxPx: 8, spacerPx: 8 };
+  const nestedStyle: GridStyle = {
+    nestedCards: {
```
stephenh, line 224: `GridStyle` learned a new `nestedCards` that:

a) takes the background color and optional border color of each card (based on kind)
b) enables within GridTable creation of the nested card DOM stuff (i.e. no other flag / `GridTable` prop is necessary, we can just look at the style being used and do the right thing)

src/components/Table/GridTable.stories.tsx
```diff
@@ -199,6 +203,323 @@ export function NestedRows() {
...
+      style={nestedStyle}
+      sorting={{ on: "client", initial: [0, "ASC"] }}
+    />
+  );
+}
+
+// Every row by-definition opens or closes a card.
+// Every row has a space between.
+
+export function NestedCardsProofOfConcept() {
```
stephenh, line 244: Will delete this but this is what I was using to prototype the DOM structure by hand before automating.

src/components/Table/GridTable.stories.tsx
```diff
@@ -199,6 +203,323 @@ export function NestedRows() {
...
+        </div>
+        <div data-card-close="level1">
+          <div data-level="level1" />
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export function OneOffInlineTable() {
```
stephenh, line 492: Not related to the nested cards, but thought I'd copy/paste this over from internal-frontend where it's our "you don't really need gridtable to do this tiny/one-off layout" snippet.

src/components/Table/GridTable.tsx
```diff
@@ -15,51 +15,12 @@ import { Link } from "react-router-dom";
...
 import { navLink } from "src/components/CssReset";
 import { Icon } from "src/components/Icon";
+import { createRowLookup, GridRowLookup } from "src/components/Table/GridRowLookup";
+import { maybeAddCardPadding, NestedCards } from "src/components/Table/nestedCards";
 import { Css, Margin, Only, Palette, Properties, Xss } from "src/Css";
 import { useTestIds } from "src/utils/useTestIds";
 import tinycolor from "tinycolor2";
 
-/** A helper for making `Row` type aliases of simple/flat tables that are just header + data. */
-export type SimpleHeaderAndDataOf<T> = { kind: "header" } | ({ kind: "data" } & T);
```
stephenh, line 23: Got extracted to `simpleHelpers`

src/components/Table/GridTable.tsx
```diff
@@ -345,29 +331,44 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
-        headerRows.push([row, makeRowComponent(row)]);
-        return;
-      }
-
       const passesFilter =
         filters.length === 0 ||
         filters.every((filter) =>
           columns.map((c) => applyRowFn(c, row)).some((maybeContent) => matchesFilter(maybeContent, filter)),
         );
+      // Even if we don't pass the filter, one of our children might, so we continue on after this check
```
stephenh, line 344: I'm thinking that a child passing the filter means we should probably include it's parent, especially for nested cards? Or even just tables in general. It's not what we do now, but not sure why/if it's super on purpose.

src/components/Table/GridTable.tsx
```diff
@@ -345,29 +331,44 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
       const isCollapsed = collapsedIds.includes(row.id);
       if (!isCollapsed && row.children) {
-        row.children.forEach(visit);
+        visitRows(row.children);
       }
+
+      nestedCards && nestedCards.endRow();
     }
 
-    maybeSorted.forEach(visit);
```
stephenh, line 370: Instead of just `maybeSorted.forEach(visit)` we call `visitRows` to kick things off b/c that has the logic to add spacers between cards.

src/components/Table/GridTable.tsx
```diff
@@ -397,52 +398,14 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   const { rowLookup } = props;
   if (rowLookup) {
     // Refs are cheap to assign to, so we don't bother doing this in a useEffect
-    rowLookup.current = {
```
stephenh, line 400: Extracted to GridRowLookup

src/components/Table/simpleHelpers.ts
```diff
@@ -0,0 +1,36 @@
+import { GridDataRow } from "src/components/Table/GridTable";
+
+/** A helper for making `Row` type aliases of simple/flat tables that are just header + data. */
+export type SimpleHeaderAndDataOf<T> = { kind: "header" } | ({ kind: "data" } & T);
```
stephenh, line 4: This file is extracted from `GridTable.ts`, no changes, just trying to break the file up a bit. Should do more in a follow up PR.

src/components/Table/nestedCards.tsx
```diff
@@ -0,0 +1,177 @@
...
+ * Takes the current buffer of close row(s), spacers, and open row, and creates a single chrome DOM row.
+ *
+ * This allows a minimal amount of DOM overhead, insofar as to the css-grid or react-virtuoso we only
+ * 1 extra DOM node between each row of content to achieve our nested card look & feel, i.e.:
+ *
+ * - chrome row (open)
+ * - card1 content row
+ * - chrome row (card2 open)
+ * - nested card2 content row
+ * - chrome row (card2 close, card1 cloard)
```
KoltonG, line 154:
```suggestion
 * - chrome row (card2 close, card1 close)
```

src/components/Table/nestedCards.tsx
```diff
@@ -0,0 +1,177 @@
...
+        ))}
+      </div>,
+    ]);
+    // clear the buffer
+    chromeBuffer.splice(0, chromeBuffer.length);
+  }
+}
+
+export function dropChromeRows<R extends Kinded>(filteredRows: RowTuple<R>[]): [GridDataRow<R>, ReactElement][] {
+  return filteredRows.filter(([r]) => !!r) as [GridDataRow<R>, ReactElement][];
```
KoltonG, line 176: **question:** Is this due to all chrome rows being of `kind` `undefined`?

src/components/Table/nestedCards.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+        border-color: rgba(236,235,235,1);
+        border-left-style: solid;
+        border-left-width: 1px;
+        border-right-style: solid;
+        border-right-width: 1px;
+        border-top-style: solid;
+        border-top-width: 1px;
+      }
+
+      <div
```
KoltonG, line 70: **question:** Is this the merging of the open-top or open bottom divs into a single div?

src/components/Table/nestedCards.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+      </div>
+    `);
+  });
+
+  it("can make open card w/two levels", async () => {
+    const r = await render(makeOpenOrCloseCard([parentCardStyle, childCardStyle], "open"));
+    expect(r.firstElement).toMatchInlineSnapshot(`
+      .emotion-0 {
+        background-color: rgba(247,245,245,1);
+        padding-left: 8px;
```
KoltonG, line 50: **question:** I was thinking that this should also have the border-top left and right too for the parents opening row style?

src/components/Table/GridTable.tsx
```diff
@@ -109,6 +70,28 @@ export interface GridStyle {
   firstRowMessageCss?: Properties;
   /** Applied on hover if a row has a rowLink/onClick set. */
   rowHoverColor?: string;
+  /** Styling for nested cards (see `cardStyle` if you only need a flat list of cards). */
+  nestedCards?: Record<string, NestedCardStyle>;
```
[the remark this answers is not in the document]
KoltonG: Just looking more into the code and I think this is is just the `kind`?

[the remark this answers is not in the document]
stephenh: Yep. I had some types in here to enforce/document that, but it was a little fussy adding the generic to GridStyle when it doesn't already have one, so took it out for now.

src/components/Table/nestedCards.tsx
stephenh: Thanks!

src/components/Table/nestedCards.tsx
stephenh: Basically yes; technically there is not _kind_ `undefined` per se, but yes they don't have a `GridDataRow` so the tuples of `filteredRows` used to be:

* `[row1, <RowComponent />`
* `[row2, <RowComponent />`
* `[row3, <RowComponent />`
* `[row4, <RowComponent />`

Are now:

* `[undefined, <chrome row/>`
* `[row1, <RowComponent />`
* `[undefined, <chrome row />`
* `[row2, <RowComponent />`
* ...

So yeah, this strips out the extra rows, for places that want to still see the real/original rows.

src/components/Table/nestedCards.test.tsx
stephenh: This is drawing a 2nd-level row open, so something like:

```
/--------------------------\
| 1st-level card        |
|  /--------------------\   | <--- this row is what this test is drawing
|  |  2nd-level       |  |
```

So the parent's border-top left/right are above us, and for this row we just straight lines for the parent.

src/components/Table/nestedCards.test.tsx
stephenh: No, this is just the overlay container that comes from our `withBeamRTL` being in every test. This test doesn't cover the "glue close-previous-card + spacer + open-next-card rows into a single chrome div".

src/components/Table/GridTable.stories.tsx
bdow: **thought:** Maybe GridTable should provide a default `nestedStyle`, probably would be good to just default to whatever we're going to need first. Then it'd also be a good reference for developers to see how it should/can be set up.

src/components/Table/GridTable.stories.tsx
bdow: Makes me feel like we should provide a SimpleTable component or something... Maybe not? I dunno. Just feels like this could easily get inconsistent if we start having these little one-off tables with localized styles everywhere.

src/components/Table/GridTable.tsx
bdow: I agree. I think the parent should be shown when a child matches the filter regardless of whether it is the nested style. It would give better context to the user. 

FWIW: I think I like the term "matchesFitler" vs "passesFilter". When first reading your comment I thought "...pass the filter..." meaning passing it as a prop or something. Not a big deal, but just thinking.

src/components/Table/GridTable.stories.tsx
stephenh: Yeah. I balance 100% agreement with "...but that is what mocks just do..." i.e. designers have the expectation of being able to bespoke layout a mini-table.

Maybe we could nudge them towards consistency with something like a `SimpleTable` + 2-3 out-of-the-box styles, i.e. like grid tables style `const`s, maybe with links to previous/standard figma mocks for each style (i.e. probably feature-specific figmas b/c we're extracting these as we go), so that when a new mock goes by with a "not grid table" table (as in it's not a sorting/filtering/data table and instead just some mini-info laid out nicely), we could link to "this mock did that, and it looked like x, can we just use that style?".

src/components/Table/GridTable.tsx
stephenh: Yeah, will probably do the "show parent if child matches" is a separate/follow up breaking change.

Also agree on the variable name, will rename.

src/components/Table/nestedCards.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+  bColor: Palette.Gray200,
+  brPx: 6,
+  pxPx: 6,
+  spacerPx: 6,
+};
+
+describe("nestedCards", () => {
+  it("can make open card w/one level", async () => {
+    const r = await render(makeOpenOrCloseCard([parentCardStyle], "open"));
+    expect(r.firstElement).toMatchInlineSnapshot(`
```
bdow, line 25: **thought:** I would push to have these "tests" be storybook snapshots instead of jest. Just seeing the styles and minimal html output isn't immediately clear on what this is accomplishing/validating. Without seeing the changes visually it is kind of a difficult test to glance over and understand.

src/components/Table/GridTable.tsx
```diff
@@ -901,7 +845,12 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
             ? rowClickRenderFn(as)
             : defaultRenderFn(as);
 
-        return renderFn(idx, cellCss, content, row, rowStyle);
+        let rendered = renderFn(idx, cellCss, content, row, rowStyle);
```
bdow, line 848: **thought (non-blocking):** There is so much going on here that it'd be nice if the `idx` value was more contextual, like maybe `colIdx` just so people quickly know this is the index of the column vs row.

bdow: Granted, you can look back at the `columns.map` and see what the idx is referencing.

src/components/Table/nestedCards.tsx
```diff
@@ -0,0 +1,177 @@
...
+export function makeOpenOrCloseCard(openCards: NestedCardStyle[], kind: "open" | "close"): JSX.Element {
+  let div: any = null;
+  const place = kind === "open" ? "Top" : "Bottom";
+  const btOrBb = kind === "open" ? "bt" : "bb";
+  // Create nesting for the all open cards, i.e.:
+  //
+  // | card1 | card2  --------------   card2 | card1 |
+  // | card1 | card2 / ... card3 ... \ card2 | card1 |
+  // | card1 | card2 | ... card3 ... | card2 | card1 |
+  //
```
bdow, line 80: Nice visualizations. Very helpful understanding what this is doing.

bdow (approved):
This all looks good to me. I'm happy with this being our first go-round with Nested Cards. The way it is set up should allow us to try introducing other styling configurations (like testing out pseudo elements possibly). 
The NestedCard class is well structured, seems like a lot of thought went into it, especially around the `chromeBuffer`, kind of takes some guess work out of figuring out the next chrome/spacer row to apply between/around cards.
My only other thoughts were regarding the names of `beginRow` and `endRow`, I think I'd nudge to call those `beginCard` `endCard`. That way it is specific to knowing the is specifically the card's row, not a spacer/chrome row.

PR #282, opened 2021-09-07
fix: pass disabled prop to select element

ksk5280:
Passing the disabled prop into the select element to allow testing that a `SelectField` is disabled.


stephenh (approved):
Thanks!

PR #283, opened 2021-09-09
fix: Better simpleRows type.

stephenh:
Something about TypeScript 4.4 made the type inference for `simpleRows` wonky, and this fixes it... :shrug:


stephenh:
Kind of a silly/annoying type fix, so going to merge to use in my WIP internal-frontend branch.

PR #284, opened 2021-09-09
feat: Add FormHeading proof-of-concept.

stephenh:
Our `FormHeadings` are consistently `baseEm`, which is almost so tiny to just `Css...` it's not worth component-izing. But, I dunno, I like it...

I'm also trying out a `cloneElement`-based approach for doing "only `mt4` the 2nd FormHeading".

Maybe should use `not nth-type-of` selectors or something, but I think that would require a hard-coded CSS class name for the selector to add margin only to `FormHeading`s and not other children.

I've been wanting to at least get this `cloneElement` approach working for awhile now, so admittedly ~80% of my motivation for this PR was using it as a good test case for `cloneElement` hacking. :-)

### DIFF
--- a/src/forms/FormHeading.tsx
+++ b/src/forms/FormHeading.tsx
@@ -10,7 +10,7 @@ export interface FormHeadingProps {
 export function FormHeading(props: FormHeadingProps) {
   const { title, xss, isFirst = false } = props;
   return (
-    <div
+    <h3
       css={{
         ...Css.baseEm.$,
         // Add space before the heading, but only if it's not first.
@@ -19,7 +19,7 @@ export function FormHeading(props: FormHeadingProps) {
       }}
     >
       {title}
-    </div>
+    </h3>
   );
 }
 
--- a/src/forms/FormHeading.tsx
+++ b/src/forms/FormHeading.tsx
@@ -1,0 +1,27 @@
+import { Css, Margin, Xss } from "src/Css";
+
+export interface FormHeadingProps {
+  title: string;
+  xss?: Xss<Margin>;
+  // This is passed automatically by FormLines
+  isFirst?: boolean;
+}
+
+export function FormHeading(props: FormHeadingProps) {
+  const { title, xss, isFirst = false } = props;
+  return (
+    <div
+      css={{
+        ...Css.baseEm.$,
+        // Add space before the heading, but only if it's not first.
+        ...(!isFirst && Css.mt4.$),
+        ...xss,
+      }}
+    >
+      {title}
+    </div>
+  );
+}
+
+// https://github.com/gaearon/react-hot-loader/issues/304#issuecomment-456569720
+FormHeading.isFormHeading = true;
--- a/src/forms/FormLines.tsx
+++ b/src/forms/FormLines.tsx
@@ -1,15 +1,15 @@
-import { ReactNode, useContext } from "react";
+import React, { Children, cloneElement, ReactNode, useContext } from "react";
 import { Css } from "src/Css";
 import { FormContext, LabelSuffixStyle } from "src/forms/FormContext";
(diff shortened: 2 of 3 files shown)


src/forms/FormLines.tsx
```diff
@@ -27,7 +27,8 @@ export interface FormLinesProps {
  */
 export function FormLines(props: FormLinesProps) {
   const settings = useContext(FormContext);
-  const { children, width = "md", labelSuffix = settings.labelSuffix } = props;
+  const { children, width = "full", labelSuffix = settings.labelSuffix } = props;
```
stephenh, line 30: I'm also changing the default form width to "full" so it just uses whatever the parent width is... Lately I've been in a lot of modals, and full is a better default for that...

src/forms/FormLines.tsx
```diff
@@ -37,7 +38,15 @@ export function FormLines(props: FormLinesProps) {
           "& > *": Css.mb2.$,
         }}
       >
-        {children}
+        {Children.map(children, (child) => {
+          if (child && typeof child === "object" && "type" in child && (child.type as any).isFormHeading) {
```
stephenh, line 42: See the comment on `FormHeading.tsx` line 27, but this is the only reliable method I could get working in storybook with fast refresh.

I.e. any approach that used `child.type === FormHeading` or `child.type === createElement(FormHeading).type` would break after hot reloads and throw off the layout b/c the `isFirst` prop would never be set (b/c the hot-patched code would never match on the `child.type` check anymore).

src/forms/FormHeading.tsx
```diff
@@ -0,0 +1,27 @@
...
+  title: string;
+  xss?: Xss<Margin>;
+  // This is passed automatically by FormLines
+  isFirst?: boolean;
+}
+
+export function FormHeading(props: FormHeadingProps) {
+  const { title, xss, isFirst = false } = props;
+  return (
+    <div
```
bdow, line 13: Since this is a "Heading" component I'm tempted to use a `hX` tag, but I also don't know which one should we use... I'd guess h3/h4, but that is kind of an odd assumption.

src/forms/FormLines.tsx
```diff
@@ -37,7 +38,15 @@ export function FormLines(props: FormLinesProps) {
...
+            const clone = cloneElement(child, { isFirst: firstFormHeading });
+            firstFormHeading = false;
+            return clone;
```
bdow, line 45: This is kind of sneaky. Though, this will only work if <FormHeading /> is a direct child of the <FormLines />,.... just playing devils advocate, this would be a problem someone were to use it like this:
```
<FormLines>
  <Tooltip title="Some extra info regarding the first FormHeading">
    <FormHeading />
  </Tooltip>
  <FormLine />
  <FormHeader /> (this one would be tagged as `isFirst`)
  ...
</FormLines>
```

bdow (approved):

src/forms/FormHeading.tsx
stephenh: Ah yeah, good call, I get kind of lazy on that; will make it an `h3`.

src/forms/FormLines.tsx
stephenh: Right, agreed on the must-be-direct-child con...

I was thinking, that con probably means we shouldn't use this for `ModalFooter` things b/c you'd have to have immediate children like:

* Modal
  * ModalHeader
  * ModalFooter

And I think, well especially for superdrawer, we want actions/footers to be nested down in the react tree somewhere like:

* MyDrawerPaage
  * Tab2
    * SuperDrawerActions

...so, yeah, can't use this approach in those sort of places, but will probably try it here and see how it goes.

PR #285, opened 2021-09-10
fix: Allow propagation of onPointerDown event through Popover when open

bdow:
Try it out with the Filters component: 
https://60466f7d43c37c0021788867-jgwefxrlvi.chromatic.com/?path=/story/components-filters--filter


stephenh (approved):
Great find

PR #286, opened 2021-09-10
fix: fix toggle filter

Chieze-Franklin:
I'm working with the assumption that when the switch is on we want the boolean field participating in the filtering:
- Set it to `true` under normal circumstances
- Set it to `false` when `enabledValue === false`

And when the switch is off we don't want that field participating in the filtering, so set it to `undefined`.


src/components/Filters/ToggleFilter.tsx
```diff
@@ -20,16 +20,19 @@ class ToggleFilter extends BaseFilter<boolean, ToggleFilterProps> implements Fil
...
-        onChange={(value) => {
-          // The switch being "off" might actually mean "true" to the backend.
-          // I.e. "Hide do not use" being "off" --> { includeDoNotUse: true }
-          const backendValue = !value ? !enabledValue : enabledValue;
-          // If the backend value is already the enabled value (i.e. the default), drop it
-          // I.e. `{ includeDoNotUse: false }` --> drop it, otherwise keep it.
-          setValue(backendValue === enabledValue ? undefined : backendValue);
+        onChange={(on) => {
+          // Basically, when the switch is off we return undefined
+          // to signify that no filtering should occur on this field
```
stephenh, line 31: This is a great comment; I was struggling to articulate what this behavior should be, but using "on and off" vs. "frontend vs. backend value" is much more clear. Thanks!

stephenh (approved):

KoltonG (commented):
Awesome fix @Chieze-Franklin!

PR #288, opened 2021-09-10
fix: Scroll to focused ListBox option. Fix menu trigger toggle bug

bdow:


src/inputs/internal/ListBox.tsx
```diff
@@ -36,6 +37,15 @@ export function ListBox<O, V extends Key>(props: ListBoxProps<O, V>) {
 
   const [listHeight, setListHeight] = useState(maxListHeight);
   const isMultiSelect = state.selectionManager.selectionMode === "multiple";
+  const virtuosoRef = useRef<VirtuosoHandle>(null);
+  const focusedItem = state.collection.getItem(state.selectionManager.focusedKey);
+
+  // Handle scrolling to the item in focus when navigating options via Keyboard
+  useEffect(() => {
```
stephenh, line 44: Neat

src/components/internal/Popover.tsx
```diff
@@ -11,17 +12,24 @@ interface PopoverProps extends React.PropsWithChildren<any> {
...
-      shouldCloseOnInteractOutside: () => {
+      shouldCloseOnInteractOutside: (e) => {
         // By default when passing `isDismissable: true` then Popover will `stopPropagation` of the PointerDown event, which is used nearly everywhere in React-Aria (like for clicking buttons)
         // We do not want that propagation to be stopped, but we still want the overlay to be dismissable.
         // When providing `isDimissable: true`, then you can also provide this callback function, `shouldCloseOnInteractOutside`
         // By returning `false` in this function it will no longer call `stopPropagation`, but it also will not call `onHide` for us, so we need to call `onClose` ourselves.
+
+        // If we clicked the trigger element (or within it), then that will call a `state.toggle` for us.
+        // Return early if that happens, otherwise we'd call `onClose`, then the trigger would toggle it back open.
+        if (triggerRef.current?.contains(e)) {
```
stephenh, line 29: Getting to the point where would love to have some tests for these things; seems likely to regress as we poke/fight various react-aria things. Don't have to block the PR on it if this is fixing a prod bug.

stephenh (approved):

src/components/internal/Popover.tsx
stephenh: Granted, maybe this is past jsdom-testable-behavior? Not sure.

bdow: Maybe, but regardless I agree. We need to start having more testing around our Beam components.

PR #290, opened 2021-09-10
fix: Fix TributeJS error. No longer allowSpaces for mentions due to bug

bdow:


src/inputs/RichTextField.tsx
```diff
@@ -169,10 +169,14 @@ function attachTributeJs(mergeTags: string[], editorElement: HTMLElement) {
   const tribute = new Tribute({
     trigger: "@",
     lookup: "value",
-    allowSpaces: true,
+    /** Not using allowSpaces due to a bug: {@link https://github.com/zurb/tribute/issues/606} */
+    allowSpaces: false,
```
stephenh, line 173: Does this mean we can't auto-complete names like `First Last`? Like does it assume we've stripped spaces from names and/or use usernames like `@shaberman`?

stephenh (approved):

src/inputs/RichTextField.tsx
bdow: We can still mention names with a First and Last, but you can't type out the first name, then a space, and then the last name. The Mentions menu will close once you type a space. Right now however, even after you select a mention you get stuck in the mention look up. Maybe this is what we want for now? I kind of thought this was a terrible experience as it is right now.

stephenh: Cool, I'm fine with that, maybe in-source comment that we expect "First Last" to not be auto-completeable, but it's net-net still the best UX?

stephenh: (Granted, maybe `allowSpaces` basically says the same thing, but dunno, just letting the maintainer we really really meant to do this. :-)  )

PR #291, opened 2021-09-12
feat: Allow non-card kinds in nested cards.

stephenh:
I.e. the "add task" use case.

image

Note that "Add" is getting pulled to the top, even though it's the "last child" when passed in, by the client-side sorting (because "Add" comes from "child g1c1").

So we'll either need to implement the "don't sort levels one or two" logic, or IMO what I'd prefer is something more like tagging rows as special i.e. `{ kind: "add", pin: "first" | "last" }` where the client-side sorting would recognize rows with the `pin` key set and automatically shove them first/last as appropriate, either before/after the otherwise normally sorted children. 

---

Update: I was looking for a distraction from the chores I'm supposed to be doing, so went ahead and added `GridDataRow.pin` so now Add will stay last.

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -166,7 +166,7 @@ const rows: GridDataRow<NestedRow>[] = [
       // Put this "grandchild" in the 2nd level to show heterogeneous levels
       { kind: "grandChild", id: "p1g1", name: "grandchild p1g1" },
       // Put this "kind" into the 2nd level to show it doesn't have to be a card
-      { kind: "add", id: "add" },
+      { kind: "add", id: "add", pin: "last" },
     ],
   },
   // a parent with just a child
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -143,7 +143,8 @@ type HeaderRow = { kind: "header" };
 type ParentRow = { kind: "parent"; id: string; name: string };
 type ChildRow = { kind: "child"; id: string; name: string };
 type GrandChildRow = { kind: "grandChild"; id: string; name: string };
-type NestedRow = HeaderRow | ParentRow | ChildRow | GrandChildRow;
+type AddRow = { kind: "add" };
+type NestedRow = HeaderRow | ParentRow | ChildRow | GrandChildRow | AddRow;
 
 const rows: GridDataRow<NestedRow>[] = [
   { kind: "header", id: "header" },
@@ -164,6 +165,8 @@ const rows: GridDataRow<NestedRow>[] = [
       },
       // Put this "grandchild" in the 2nd level to show heterogeneous levels
       { kind: "grandChild", id: "p1g1", name: "grandchild p1g1" },
+      // Put this "kind" into the 2nd level to show it doesn't have to be a card
+      { kind: "add", id: "add" },
     ],
   },
   // a parent with just a child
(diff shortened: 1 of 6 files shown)


src/components/Table/GridTable.stories.tsx
```diff
@@ -218,13 +223,18 @@ export function NestedCards() {
...
     }),
+    add: () => "Add",
   };
   const spacing = { brPx: 4, pxPx: 8, spacerPx: 8 };
   const nestedStyle: GridStyle = {
     nestedCards: {
-      parent: { bgColor: Palette.Gray500, ...spacing },
-      child: { bgColor: Palette.Gray200, bColor: Palette.Gray600, ...spacing },
-      grandChild: { bgColor: Palette.Green200, bColor: Palette.Green400, ...spacing },
+      topLevelSpacerPx: 8,
```
stephenh, line 231: Besides allowing a `kind: add` that doesn't have an entry in the `nestedCards` config (i.e. to tell us its not a card), I also changed how the spacer determines its height.

It used to be:

* If we just visited a row, the spacer was that row's `card.spacerPx`

But that was kinda weird, i.e. if I was in a parent card, and then went "child kind 1, child kind 2, child kind 1", the spacer would depend on which card and not "I'm in the parent card, all spacers between my children is X".

But the reason I'd done that is that we also need a "top-level" spacer, i.e. when there are no cards open (we have just closed the top-level card, and are going to show the nest top-level card), there is no "current open card" and we can't do `openCard.spacerPx`.

Hence adding just a new `topLevelSpacerPx`. It's the spacing between the top-most cards.

src/components/Table/GridTable.tsx
```diff
@@ -61,8 +61,19 @@ export interface GridStyle {
   firstRowMessageCss?: Properties;
   /** Applied on hover if a row has a rowLink/onClick set. */
   rowHoverColor?: string;
-  /** Styling for nested cards (see `cardStyle` if you only need a flat list of cards). */
-  nestedCards?: Record<string, NestedCardStyle>;
```
stephenh, line 65: Made a new `NestedCardsStyle` to have both the `topLevelSpacerPx` + the previous/existing per-kind map to `NestedCardStyle`.

src/components/Table/GridTable.tsx
```diff
@@ -333,32 +344,34 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
           columns.map((c) => applyRowFn(c, row)).some((maybeContent) => matchesFilter(maybeContent, filter)),
         );
       // Even if we don't pass the filter, one of our children might, so we continue on after this check
+      let isCard = false;
       if (matches) {
-        nestedCards && nestedCards.beginRow(row);
+        isCard = nestedCards && nestedCards.maybeOpenCard(row);
```
stephenh, line 349: If this row doesn't have a card style, then we return false, which tells us later that we don't need to close the card.

src/components/Table/GridTable.stories.tsx
```diff
@@ -164,6 +165,8 @@ const rows: GridDataRow<NestedRow>[] = [
       },
       // Put this "grandchild" in the 2nd level to show heterogeneous levels
       { kind: "grandChild", id: "p1g1", name: "grandchild p1g1" },
+      // Put this "kind" into the 2nd level to show it doesn't have to be a card
+      { kind: "add", id: "add", pin: "last" },
```
stephenh, line 169: I went back/forth between putting the `pin` key directly into the `GridDataRow`, or putting in the `GridRowStyle` which is where most of our layout-ish settings (like indent, row links, etc.) goes...

I dunno, maybe it makes more sense to add in the `GridRowStyle`...

bdow (approved):

PR #292, opened 2021-09-13
feat: Generalize toggleFilter for non-boolean on/off values.

stephenh:
We don't need it immediately, but just following up the trend from @Chieze-Franklin 's PR, it seems like `toggleFilter` would really use any values for "on" / "off" instead of just booleans.

### DIFF
--- a/src/components/Filters/ToggleFilter.test.tsx
+++ b/src/components/Filters/ToggleFilter.test.tsx
@@ -1,0 +1,83 @@
+import { useState } from "react";
+import { toggleFilter } from "src/components/Filters";
+import { ToggleFilterProps } from "src/components/Filters/ToggleFilter";
+import { click, render } from "src/utils/rtl";
+import { useTestIds } from "src/utils/useTestIds";
+
+describe("ToggleFilter", () => {
+  it("uses true/undefined by default", async () => {
+    // Given a default boolean filter
+    const r = await render(<TestFilter toggleFilter={{}} />);
+    // It's initially empty
+    expect(r.filter_filter()).not.toBeChecked();
+    expect(r.value()).toHaveTextContent("undefined");
+    // And when they click it, we turn true
+    click(r.filter_filter());
+    expect(r.filter_filter()).toBeChecked();
+    expect(r.value()).toHaveTextContent("true");
+    // And when they click it again, we go back to undefined
+    click(r.filter_filter());
+    expect(r.filter_filter()).not.toBeChecked();
+    expect(r.value()).toHaveTextContent("undefined");
+  });
+
+  it("can use false/undefined when onValue is set", async () => {
+    // Given a default boolean filter
+    const r = await render(<TestFilter toggleFilter={{ onValue: false }} />);
+    // It's initially empty
+    expect(r.filter_filter()).not.toBeChecked();
+    expect(r.value()).toHaveTextContent("undefined");
+    // And when they click it, we turn true
+    click(r.filter_filter());
+    expect(r.filter_filter()).toBeChecked();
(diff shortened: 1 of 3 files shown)


src/components/Filters/ToggleFilter.tsx
```diff
@@ -3,36 +3,45 @@ import { Filter } from "src/components/Filters/types";
 import { Switch } from "src/inputs/Switch";
 import { TestIds } from "src/utils/useTestIds";
 
-export type ToggleFilterProps = { label?: string; enabledValue?: boolean; defaultValue?: boolean };
+export type ToggleFilterProps<V> = {
+  label?: string;
+  onValue?: V | undefined;
+  offValue?: V | undefined;
```
stephenh, line 9: This is a breaking change, but doubling down on the "on/off" language instead of enabled/disabled b/c for whatever reason it seems to make things more clear.

src/components/Filters/ToggleFilter.tsx
```diff
@@ -3,36 +3,45 @@ import { Filter } from "src/components/Filters/types";
...
+ * Provides a two-state "on/off" filter.
+ *
+ * By default the on/off values are `on === true` and `off === undefined`.
+ *
+ * You can flip the on/off values by passing `onValue: false`, in which case
+ * `on === false` and off === undefined`.
+ *
+ * Or you can set on/off directly, by passing both `onValue` and `offValue`.
+ *
+ * @param props
```
bdow, line 23: Please remove this, probably added by WebStorm.

src/components/Filters/ToggleFilter.tsx
```diff
@@ -3,36 +3,45 @@ import { Filter } from "src/components/Filters/types";
...
-        selected={value === undefined ? false : enabledValue === false ? !value : value}
+        selected={value === undefined ? false : value === onValue}
         label={this.label}
         labelStyle={inModal ? "filter" : "inline"}
         onChange={(on) => {
-          // Basically, when the switch is off we return undefined
-          // to signify that no filtering should occur on this field
-          // And when the switch is on we return:
-          // - true (normal behaviour)
-          // - false (if enabledValue is false)
```
stephenh, line 34: Was debating whether to keep or remove these comments, but I think the new impls for both `selected` and `onChange` are sufficiently clear now that we probably don't need them...

src/components/Filters/ToggleFilter.tsx
stephenh: Done!

src/components/Filters/ToggleFilter.tsx
bdow: 100%. The naming convention here is super clear now.

src/components/Filters/ToggleFilter.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    click(r.filter_filter());
+    expect(r.filter_filter()).not.toBeChecked();
+    expect(r.value()).toHaveTextContent("false");
+  });
+
+  it("can use foo/bar when onValue/offValue are both set", async () => {
+    // Given a default boolean filter
+    const r = await render(<TestFilter toggleFilter={{ onValue: "foo", offValue: "bar" }} />);
+    // It's initially empty
+    expect(r.filter_filter()).not.toBeChecked();
```
bdow, line 60: **question/suggestion:** Can we have another test to validate that this Toggle _can_ be initially checked?

stephenh: Yep, added.

bdow (approved):

PR #293, opened 2021-09-13
feat: Support vertical layout for Filters

bdow:


src/components/Filters/types.ts
```diff
@@ -30,5 +30,11 @@ export interface Filter<V> {
   defaultValue: V | undefined;
 
   /** Renders the filter into either the page or the modal. */
-  render(value: V | undefined, setValue: (value: V | undefined) => void, tid: TestIds, inModal: boolean): JSX.Element;
+  render(
+    value: V | undefined,
+    setValue: (value: V | undefined) => void,
+    tid: TestIds,
+    inModal: boolean,
+    vertical: boolean,
```
bdow, line 38: Wondering if we should maybe do a "placement"(?) property and combine these instead?

src/components/Filters/Filters.tsx
```diff
@@ -34,13 +36,14 @@ function Filters<F, G extends Value = string>(props: FilterProps<F, G>) {
...
     // If we have more than 4 filters,
-    if (impls.length > numberOfPageFilters + 1) {
+    if (!vertical && impls.length > numberOfPageFilters + 1) {
       // Then return the first three to show on the page, and the remainder for the modal.
       return [
         Object.fromEntries(impls.slice(0, numberOfPageFilters)) as FilterImpls<F>,
         Object.fromEntries(impls.slice(numberOfPageFilters)) as FilterImpls<F>,
       ];
     }
+    debugger;
```
stephenh, line 46: Snuck in :-)

src/components/Filters/types.ts
stephenh: Ah yeah, I like that...although it assumes the vertical place will _never_ put things in the modal...well, or actually that would be nice b/c then it would be inModal=true vertical=false...  So yeah, makes sense.

Want to do that in this PR or just as a sometime follow up?

stephenh (approved):
Nice!

src/components/Filters/types.ts
bdow: Since there is not rush for this to go in, I think I'd prefer to do it as part of this PR.

bdow: Thought about it, and played around with alternative approaches, but I didn't think it was much cleaner than this. So will leave for now. Happy to revisit at another time.

PR #294, opened 2021-09-14
fix: SingleSelectFilter and MultiSelectFilter `defaultValue` prop not working

KoltonG:
When using the `defaultValue` prop for the Single and Multi Select Filter, no values would appear since the default value was not being given to the underlying `value` prop.

### DIFF
--- a/src/components/Filters/MultiFilter.tsx
+++ b/src/components/Filters/MultiFilter.tsx
@@ -47,7 +47,7 @@ class MultiFilter<O, V extends Value> extends BaseFilter<V[], MultiFilterProps<O
         {...props}
         compact={!vertical}
         label={this.label}
-        values={value || []}
+        values={value || defaultValue || []}
         hideLabel={inModal}
         inlineLabel={!inModal && !vertical}
         sizeToContent={!inModal && !vertical}
--- a/src/components/Filters/MultiFilter.test.tsx
+++ b/src/components/Filters/MultiFilter.test.tsx
@@ -1,0 +1,30 @@
+import { useState } from "react";
+import { FilterDefs, Filters } from "src/components/Filters";
+import { ProjectFilter, stageFilter, stageFilterDefaultValue } from "src/components/Filters/testDomain";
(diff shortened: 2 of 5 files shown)


src/components/Filters/MultiFilter.tsx
```diff
@@ -47,7 +47,7 @@ class MultiFilter<O, V extends Value> extends BaseFilter<V[], MultiFilterProps<O
         {...props}
         compact={!vertical}
         label={this.label}
-        values={value || []}
+        values={value || defaultValue || []}
```
stephenh, line 50: Hm, I don't think this is right, b/c the intent of `defaultValue` was to be the initial value for the filter, and not necessarily "what the field must always be".

I.e. a valid flow would be:

* You have a default filter that defaults to `projects=mine`
* You clear the filter to be just `projects=undefined`

This would be put back `mine`, each time you tried to clear the filter.

KoltonG: Good point, in that case, we could use the `tick` trick to keep track of the initial render and subsequent renders. WDYT? Did @bdow and yourself have another implementation in mind? This is needed for the weeklyUpdates work.

stephenh: Naive question, but why do you think this is not working? FWIW usually when I think that `defaultValue` is not working, it's because `usePersistedFilter` is pulling things from session storage.

Specifically, this is the code that handles "just for the very first filter, use default values":

```
  const defaultFilter = useMemo(
    () =>
      Object.fromEntries(
        safeEntries(filterDefs)
          .filter(([key, def]) => def(key as string).defaultValue !== undefined)
          .map(([key, def]) => [key, def(key as string).defaultValue]),
      ),
    [filterDefs],
  );
```

In `usePersistedFilter`. We can pull that out into like a `createDefaultFilter(filterDefs)` if the weekly updates page needs to call it directly i.e. from a custom `useMemo` if it doesn't want to use `usePersistedFilter`.

KoltonG:
Closing since it's working!

PR #295, opened 2021-09-15
feat: Tabs support routes

bdow:
Allows Tabs to accept a 'path' property which will render the Tab component as a link, using React Router to navigate between tabs.
Selected tab will be based on the route instead of passing a "selected" property into the Tabs component.
Change default Tab element to be a Button which will allow for native 'on Enter key' handling in DOM. In order to achieve this we also need to properly set focus on the element as the user navigates between them via Keyboard.

### DIFF
--- a/src/components/Tabs.test.tsx
+++ b/src/components/Tabs.test.tsx
@@ -1,13 +1,13 @@
 import { fireEvent } from "@testing-library/react";
 import { useState } from "react";
 import { Palette } from "src/Css";
-import { click, render } from "src/utils/rtl";
-import { getNextTabValue, Tab, TabsWithContent } from "./Tabs";
+import { click, render, withRouter } from "src/utils/rtl";
+import { getNextTabValue, RouteTab, Tab, TabsWithContent } from "./Tabs";
 import { TabValue, TestTabContent, testTabs } from "./testData";
 
 describe("TabsWithContent", () => {
   it("should display content of selected tab", async () => {
-    const r = await render(<TestTabs />);
+    const r = await render(<TestTabs />, withRouter());
     // tab panel should initially display Tab 1 Content
     expect(r.tab_panel()).toHaveTextContent("Tab 1 Content");
     // when we click on the 4th tab
@@ -18,7 +18,7 @@ describe("TabsWithContent", () => {
 
   it("should update the active tab", async () => {
     // Given we start out on tab1
-    const r = await render(<TestTabs />);
+    const r = await render(<TestTabs />, withRouter());
     expect(r.tabs_tab1()).toHaveStyleRule("color", Palette.LightBlue700);
     // when an external state change moves us to the 2nd tab
     click(r.goToTab2);
@@ -28,7 +28,7 @@ describe("TabsWithContent", () => {
 
   it("should reset the active tab on blur", async () => {
     // Given we start out on tab1
-    const r = await render(<TestTabs />);
+    const r = await render(<TestTabs />, withRouter());
     expect(r.tabs_tab1()).toHaveStyleRule("color", Palette.LightBlue700);
     // And we've moved to the 2nd tab
     fireEvent.keyUp(r.tabs_tab2(), { key: "ArrowRight" });
@@ -41,7 +41,7 @@ describe("TabsWithContent", () => {
 
   it("cannot click on disabled tabs", async () => {
     // Given we start out on tab1
-    const r = await render(<TestTabs />);
+    const r = await render(<TestTabs />, withRouter());
     expect(r.tabs_tab1()).toHaveStyleRule("color", Palette.LightBlue700);
     // And we try to click on the 3rd disabled tab
     click(r.tabs_tab3);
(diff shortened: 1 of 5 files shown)


src/components/Tabs.tsx
```diff
@@ -25,6 +28,15 @@ export interface TabsProps<V extends string, X extends Properties> {
   contentXss?: X;
 }
 
+// Tabs can be rendered as Links (omit "onChange") and we'll use React-Router for matching (omit "selected")/
```
stephenh, line 31: Per omitting `onChange` and `selected`, should we omit `value` from `RouteTab`'s extension of `Tab`?

src/components/Tabs.tsx
```diff
@@ -106,88 +130,86 @@ export function Tabs<V extends string>(props: TabsProps<V, {}>) {
...
 }
 
-interface TabProps<V extends string> extends BeamFocusableProps {
+interface TabImplProps<V extends string> extends BeamFocusableProps {
   /** active indicates the current tab is highlighted */
   active: boolean;
-  disabled: boolean;
-  label: string;
-  icon?: IconKey;
-  value: V;
```
stephenh, line 137: Nice, these just come for free through the `tab`.

stephenh (approved):
Nice!

src/components/Tabs.tsx
bdow: No, we'll still want `value`. And the reason why we'd want to use both `path` and `value` is so the do the matching an pass the query variables to the tab panel component... which I'm realizing now that I am not doing. To do this we'll need to call `selectedTab.render()` as part of a `<Route path={path} render={selectedTab.render} />`. 
If we do not do this then the contents of each tab wouldn't have access to the current route params. They'd need to do their own `useRouteMatch` locally vs a `useParams`. Which wouldn't be terrible, but would be different from what we're doing today. Thoughts?

stephenh: Ah, do you mean like this in the BP one:

```
    <div data-testid="tabs_content" css={xss}>
      {tabs.map((tab) => (
        <Route key={tab.value} path={tab.path} render={tab.render} exact />
      ))}
    </div>
```

I guess it's using `tab.value` for the `key` ... but otherwise it's using the `tab.path: string | string[]` for what I think you're describing? 

I.e. it sounds like we still only need the `path`...

bdow: You're right we don't need `value` in `<Route />` component, but `value` is needed for the `href` of each tab.

stephenh: Oh I gotcha, that makes sense.

Wdyt about omitting value and adding in a dedicated href key?

bdow: Sure, that makes sense. Definitely a better API.

src/components/Tabs.tsx
```diff
@@ -117,31 +123,34 @@ export function Tabs<V extends string>(props: TabsProps<V, {}> | RouteTabsProps<
   }
 
   // bluring out resets active to whatever selected is
-  function onBlur() {
-    setActive(selected);
+  function onBlur(e: FocusEvent) {
+    // only reset active state if we've moved focus out of the tab list
+    if (!(ref.current && ref.current.contains(e.relatedTarget as Node))) {
+      setActive(selected);
+    }
```
bdow, line 130: This was causing some unintended behavior after introducing `document.getElementById(`${nextTabValue}-tab`)?.focus();` in `onKeyUp`. Now `onBlur` is being triggered immediately after the `onKeyUp` call, where previously `onBlur` was only called via clicking/pressing away from the element. Because of this then `onBlur` had an old copy of `selected`, and would revert `active` to its previous value.
Adding the `contains` logic to ensure we only execute onBlur logic if the focus changed to outside of the tab-list.

src/components/Tabs.tsx
```diff
@@ -25,6 +29,17 @@ export interface TabsProps<V extends string, X extends Properties> {
   contentXss?: X;
 }
 
+// Tabs can be rendered as Links (omit "onChange") and we'll use React-Router for matching (omit "selected")/
+export interface RouteTabsProps<V extends string, X> extends Omit<TabsProps<V, X>, "onChange" | "selected" | "tabs"> {
+  tabs: RouteTab<V>[];
+}
+// A Route Tab has a `href` prop rather than a `value` prop
+export interface RouteTab<V extends string> extends Omit<Tab<V>, "value"> {
+  href: V;
```
KoltonG, line 38: Is this the path to be routed?

src/components/Tabs.test.tsx
```diff
@@ -78,17 +79,114 @@ describe("TabsWithContent", () => {
...
   });
+
+  it("renders tabs as links", async () => {
+    const router = withRouter("/tab1");
+    // Given tabs with `path` values
+    const testTabs: RouteTab<string>[] = [
+      { name: "Tab A", path: "/tab1", href: "/tab1", render: () => <TestTabContent content="Tab 1 Content" /> },
+      { name: "Tab B", path: "/tab2", href: "/tab2", render: () => <TestTabContent content="Tab 2 Content" /> },
+    ];
+    const r = await render(<TabsWithContent tabs={testTabs} />, router);
```
KoltonG, line 101: What happens when you give `selected` to this component? Would the path or selected take priority?

KoltonG (commented):
Thank you! Checking out the Chromatic!

PR #297, opened 2021-09-15
fix: Fix FormLines margins not collapsing.

stephenh:


src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -253,7 +253,7 @@ export function SelectFieldBase<O, V extends Value>(props: SelectFieldBaseProps<
   };
 
   return (
-    <div css={Css.dif.fdc.w100.maxw(px(550)).$} ref={comboBoxRef}>
+    <div css={Css.df.fdc.w100.maxw(px(550)).$} ref={comboBoxRef}>
```
stephenh, line 256: So this being `dif` meant it was inline which meant that any margin on it would not collapse on block elements below it:

https://stackoverflow.com/questions/66953818/why-dont-inline-flex-containers-margins-collapse

:-/

bdow (approved):

PR #299, opened 2021-09-16
fix: Do not allow ListBox to show on top of trigger element

bdow:
A bug was originally discovered when triggering the ListBox on top of the SelectField where an option would be immediately selected due to selection happening onPointerUp and the ListBox being triggered onPointerDown. This was originally fixed by setting 'shouldSelectOnPressUp: false' for the Options within the ListBox. Though, this traded one bug for another. Recently discovered was that the subsequent onPointerDown event fired after selecting an Option could have adverse effects, such as trigger a modal to close if the ListBox was rendered on top of the Modal's underlay element.

To address this issue, we went back to the root problem and now prevent the Menu from ever rendering above its trigger element. This is done by utilizing the OverlayPosition props, which include a 'maxHeight' property we use to define the ListBox's height. This ensures it will only take up the space available between the trigger and the edge of the viewport. A height is needed to be set due to the virtualization library used to render the ListBox's options.

### DIFF
--- a/src/inputs/internal/ListBox.tsx
+++ b/src/inputs/internal/ListBox.tsx
@@ -1,4 +1,4 @@
-import { Key, MutableRefObject, useEffect, useRef, useState } from "react";
+import React, { Key, MutableRefObject, useEffect, useRef, useState } from "react";
 import { DismissButton, useListBox } from "react-aria";
 import { SelectState } from "react-stately";
 import { Virtuoso } from "react-virtuoso";
@@ -11,11 +11,11 @@ interface ListBoxProps<O, V extends Key> {
   compact: boolean;
   listBoxRef: MutableRefObject<HTMLDivElement | null>;
   state: SelectState<O>;
-  maxListHeight?: number;
   selectedOptions: O[];
   getOptionLabel: (opt: O) => string;
   getOptionValue: (opt: O) => V;
   contrast?: boolean;
+  positionProps: React.HTMLAttributes<Element>;
 }
 
 /** A ListBox is an internal component used by SelectField and MultiSelectField to display the list of options */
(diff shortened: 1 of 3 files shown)


src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -245,6 +245,7 @@ export function SelectFieldBase<O, V extends Value>(props: SelectFieldBaseProps<
     shouldFlip: true,
     isOpen: state.isOpen,
     onClose: state.close,
+    placement: "bottom",
```
bdow, line 248: `placement: 'bottom'` is the default, but hard coding it now because we make assumptions on this default in the ListBox component.

src/inputs/internal/ListBox.tsx
```diff
@@ -24,17 +24,24 @@ export function ListBox<O, V extends Key>(props: ListBoxProps<O, V>) {
...
   const { listBoxProps } = useListBox({ disallowEmptySelection: true, ...otherProps }, state, listBoxRef);
 
+  const positionMaxHeight = positionProps.style?.maxHeight;
+  // The maxListHeight will be based on the value defined by the positionProps returned from `useOverlayPosition` (which will always be a defined as a `number` based on React-Aria's `calculatePosition`).
+  // If `maxHeight` is set use that, otherwise use `273` as a default (`42px` is the min-height of each option, so this allows
+  // 6.5 options in view at a time (doing `.5` so the user can easily tell if there are more).
+  const maxListHeight = positionMaxHeight && typeof positionMaxHeight === "number" ? positionMaxHeight : 273;
+  // We define some spacing between the ListBox and the trigger element, and depending on where the list box renders (above or below) we need to adjust the spacing.
+  // We can determine if the position was flipped based on what style is defined, `top` (for positioned below the trigger), and `bottom` (for above the trigger).
+  // The above assumption regarding `top` and `bottom` is true as long as we use `bottom` as our `OverlayPosition.placement` (set in SelectFieldBase).
```
stephenh, line 43: So the `placement: bottom` down below doesn't actually mean "always bottom"? I.e. it sounds like you're still detecting that sometimes it could be above.

stephenh (approved):

src/inputs/internal/ListBox.tsx
bdow: Right, we also set `shouldFlip: true`. I'll add that into the comment.

PR #300, opened 2021-09-16
fix: Add OpenModal.keepOpen for doing tape-style measurements.

stephenh:
I went through much gnashing of teeth to very specifically pad form elements in a modal, particularly b/c all of the tape / measure / etc. extensions I found (for both chrome and FF), and also FF's built-in measure feature, all would cause the modal to close.

This adds a `keepOpen` prop to `OpenModal` that skips the `openModal`/`closeModal` hook and just redirectly renders the `Modal` in the DOM.

I'm planning on making this the default for any story that uses `modal: true` in blueprint, b/c I figure testing close is not really that important in stories.

### DIFF
--- a/src/components/Modal/OpenModal.tsx
+++ b/src/components/Modal/OpenModal.tsx
@@ -1,11 +1,20 @@
 import { useEffect } from "react";
-import { ModalProps } from "src/components/Modal/Modal";
+import { Modal, ModalProps } from "src/components/Modal/Modal";
 import { useModal } from "src/components/Modal/useModal";
 
+export interface OpenModalProps {
+  /** The custom modal content to show. */
+  children: JSX.Element;
+  /** The size to use. */
+  size?: ModalProps["size"];
+  /** Whether to force the modal to stay open. This is useful for stories where ruler/tape extensions cause the modal to close. */
+  keepOpen?: boolean;
+}
+
 /**
- * A component for testing open modals in unit tests.
+ * A component for testing open modals in stories and unit tests.
  *
- * Current, calling `render(<ModalComponent />)` in a test currently doesn't work, because
+ * Currently, calling `render(<ModalComponent />)` in a test currently doesn't work, because
  * nothing has called `useModal` to get the header & footer mounted into the DOM.
  *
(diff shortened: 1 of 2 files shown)


src/components/Modal/OpenModal.tsx
```diff
@@ -21,9 +30,17 @@ import { useModal } from "src/components/Modal/useModal";
...
+  const { size, children, keepOpen } = props;
+  useEffect(() => {
+    if (!keepOpen) {
+      openModal({ size, content: children });
+    }
+  }, [keepOpen, openModal, size, children]);
+  if (keepOpen) {
+    return <Modal size={size} content={children} />;
+  } else {
+    return <div>dummy content</div>;
```
bdow, line 44: was this just necessary to appease a consistent return type?

bdow (approved):

src/components/Modal/OpenModal.tsx
```diff
@@ -1,11 +1,20 @@
...
-import { ModalProps } from "src/components/Modal/Modal";
+import { Modal, ModalProps } from "src/components/Modal/Modal";
 import { useModal } from "src/components/Modal/useModal";
 
+export interface OpenModalProps {
+  /** The custom modal content to show. */
+  children: JSX.Element;
+  /** The size to use. */
+  size?: ModalProps["size"];
+  /** Whether to force the modal to stay open. This is useful for stories where ruler/tape extensions cause the modal to close. */
```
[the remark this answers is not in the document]
stephenh: Yeah browser extensions. After trying a bunch of extensions, I looked for a storybook addon, and there is one built-it but it didn't look like it really did tape/between-element measurements, plus I assumed it'd have the same always-closing-the-modal issue as the extensions did.

src/components/Modal/OpenModal.tsx
stephenh: Hm, no? This is just what it was before, i.e. the previous "keepOpen=false" behavior was to return some dummy content that is the pretend/top-level page content that then `openModal` renders on top of.

PR #302, opened 2021-09-16
fix: Move open/close padding to dedicated columns.

stephenh:


src/components/Table/GridTable.tsx
```diff
@@ -347,6 +349,9 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
       let isCard = false;
       if (matches) {
         isCard = nestedCards && nestedCards.maybeOpenCard(row);
+        if (isCard && nestedCards) {
```
KoltonG, line 352: Is `nestedCards` a sort of feature flag?

src/components/Table/GridTable.tsx
```diff
@@ -294,7 +294,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   }, [columns, rows, sorting, sortState]);
 
   // Filter + flatten + component-ize the sorted rows.
-  let [headerRows, filteredRows]: [RowTuple<R>[], RowTuple<R>[]] = useMemo(() => {
+  let [headerRows, filteredRows, maxCardPadding]: [RowTuple<R>[], RowTuple<R>[], number | undefined] = useMemo(() => {
```
KoltonG, line 297: It would be really great to move this into a function! Its a ~100 line useMemo :P

src/components/Table/GridTable.tsx
```diff
@@ -477,6 +494,7 @@ function renderTable<R extends Kinded>(
   filteredRows: RowTuple<R>[],
   firstRowMessage: string | undefined,
   stickyHeader: boolean,
+  maxCardPadding: number | undefined,
```
KoltonG, line 497: Wondering if we can even support nested cards via the <table> approach anymore? Are we using maxCardPadding inside this anywhere?

src/components/Table/GridTable.tsx
```diff
@@ -580,39 +599,48 @@ function renderVirtual<R extends Kinded>(
...
-          }}
-          data-testid={id}
-        >
-          {children}
-        </div>
-      );
-    });
-  },
-);
+const VirtualRoot = memoizeOne<
```
KoltonG, line 602: I'm so confused on what useMemo does after reading the docs for this 🙈

src/components/Table/GridTable.tsx
```diff
@@ -580,39 +599,48 @@ function renderVirtual<R extends Kinded>(
...
-      // Default to auto, but use `c.w` as a fr if numeric or else `c.w` as-if if a string
-      .map((c) => (typeof c.w === "string" ? c.w : c.w !== undefined ? `${c.w}fr` : "auto"))
-      .join(" ")
-  );
+function calcGridColumns(columns: GridColumn<any>[], maxCardPadding: number | undefined): string {
+  let sizes = columns.map((c) => {
+    // Default to auto, but use `c.w` as a fr if numeric or else `c.w` as-if if a string
+    return typeof c.w === "string" ? c.w : c.w !== undefined ? `${c.w}fr` : "auto";
+  });
+  // If we're doing nest cards, we add extra 1st/last cells...
```
KoltonG, line 639:
```suggestion
  // If we're doing nested cards, we add extra 1st/last cells...
```

src/components/Table/GridTable.tsx
```diff
@@ -580,39 +599,48 @@ function renderVirtual<R extends Kinded>(
...
+  if (maxCardPadding) {
+    sizes = [`${maxCardPadding}px`, ...sizes, `${maxCardPadding}px`];
```
KoltonG, line 641: Neat and this is what creates `gtc`?

src/components/Table/nestedCards.tsx
```diff
@@ -107,27 +112,15 @@ export function makeOpenOrCloseCard(openCards: NestedCardStyle[], kind: "open" |
...
-  const isFirst = columnIndex === 0;
-  const isFinal = columnIndex === columns.length - 1;
-  if (!isFirst && !isFinal) {
-    // Even if we don't need the nested color+padding of each open card, at
-    // least add the background color of the closed open card.
-    const card = openCards[openCards.length - 1];
-    return !card ? div : <div css={Css.bgColor(card.bgColor).$}>{div}</div>;
-  }
+export function maybeAddCardPadding(openCards: NestedCardStyle[], column: "first" | "final"): any {
+  let div: any = <div>&nbsp;</div>;
```
KoltonG, line 116: Curious as to why you had to add this content? Was it to give it a height?

src/components/Table/nestedCards.tsx
```diff
@@ -173,7 +166,8 @@ export function maybeCreateChromeRow(
   if (chromeBuffer.length > 0) {
     filteredRows.push([
       undefined,
-      <div css={Css.gc(`span ${columns.length}`).$}>
+      // We add 2 to account for our dedicated open/close columns
+      <div css={Css.gc(`span ${columns.length + 2}`).$}>
```
KoltonG, line 170: Nice

KoltonG (approved):
Amazing! Thank you for the fix!

src/components/Table/nestedCards.tsx
```diff
@@ -53,6 +53,11 @@ export class NestedCards {
     maybeCreateChromeRow(this.columns, this.filteredRows, this.chromeBuffer);
   }
 
+  maxCardPadding(current: number | undefined): number {
+    const padding = this.openCards.map((c) => c.pxPx).reduce((a, b) => a + b, 0);
```
KoltonG, line 57: @stephenh maybe here we have to consider the bColor for the extra one.

src/components/Table/GridTable.tsx
stephenh: Yeah, it turns on all of the "extra DOM elements" work needed to do the card styling.

src/components/Table/GridTable.tsx
stephenh: Yep

src/components/Table/nestedCards.tsx
stephenh: Right, otherwise the inner divs didn't render.

src/components/Table/nestedCards.tsx
```diff
@@ -53,6 +53,11 @@ export class NestedCards {
     maybeCreateChromeRow(this.columns, this.filteredRows, this.chromeBuffer);
   }
 
+  maxCardPadding(current: number | undefined): number {
+    const padding = this.openCards.map((c) => c.pxPx + (!!c.bColor ? 1 : 0)).reduce((a, b) => a + b, 0);
```
KoltonG, line 57: Nice!

KoltonG (approved):

src/components/Table/GridTable.tsx
stephenh: Via what approach? I think Github lost part of it, oh like the `table` approach?

No, nested cards aren't really supported here, but the param is necessary, even if its unused, b/c right now all of the rendertable / rendercss / rendervirtual methods must have the same method signatures.

PR #303, opened 2021-09-16
fix: Support setting padding as part of nested scrolling layout

bdow:
As I started to pull `<ScrollableParent />` and `<ScrollContent />` components into Blueprint, I found it quite cumbersome to define the consistent padding between all the pages. This change allows for setting that consistent padding, but also provides a `<FullBleed />` component that utilizes the set padding to have the component expand the full width of the `<ScrollableParent />`

### DIFF
--- a/src/components/Layout/FullBleed.tsx
+++ b/src/components/Layout/FullBleed.tsx
@@ -1,0 +1,16 @@
+import { cloneElement, ReactElement } from "react";
+import { useScrollableParent } from "src/components/Layout/ScrollableParent";
+import { Css, px } from "src/Css";
+
+/** Provides a way to extend the full width of the ScrollableParent */
+export function FullBleed({ children, omitPadding = false }: { children: ReactElement; omitPadding?: boolean }) {
+  const { paddingX } = useScrollableParent();
+  return !paddingX
+    ? children
+    : cloneElement(children, {
+        style: Css.mlPx(paddingX * -1)
+          .w(`calc(100% + ${px(paddingX * 2)})`)
+          .if(!omitPadding)
+          .pxPx(paddingX).$,
+      });
+}
--- a/src/components/Layout/FullBleed.tsx
+++ b/src/components/Layout/FullBleed.tsx
@@ -1,6 +1,6 @@
 import { cloneElement, ReactElement } from "react";
 import { useScrollableParent } from "src/components/Layout/ScrollableParent";
-import { Css, px } from "src/Css";
+import { Css } from "src/Css";
 
 /** Provides a way to extend the full width of the ScrollableParent */
 export function FullBleed({ children, omitPadding = false }: { children: ReactElement; omitPadding?: boolean }) {
@@ -8,8 +8,7 @@ export function FullBleed({ children, omitPadding = false }: { children: ReactEl
   return !paddingX
     ? children
     : cloneElement(children, {
-        style: Css.mlPx(paddingX * -1)
-          .w(`calc(100% + ${px(paddingX * 2)})`)
+        style: Css.mxPx(paddingX * -1)
           .if(!omitPadding)
           .pxPx(paddingX).$,
       });
--- a/src/components/Layout/FullBleed.tsx
+++ b/src/components/Layout/FullBleed.tsx
@@ -4,12 +4,10 @@ import { Css } from "src/Css";
 
 /** Provides a way to extend the full width of the ScrollableParent */
 export function FullBleed({ children, omitPadding = false }: { children: ReactElement; omitPadding?: boolean }) {
-  const { paddingX } = useScrollableParent();
-  return !paddingX
+  const { pr, pl } = useScrollableParent();
+  return !pr && !pl
     ? children
     : cloneElement(children, {
-        style: Css.mxPx(paddingX * -1)
-          .if(!omitPadding)
-          .pxPx(paddingX).$,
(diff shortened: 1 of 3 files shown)


src/components/Layout/FullBleed.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { Css, px } from "src/Css";
+
+/** Provides a way to extend the full width of the ScrollableParent */
+export function FullBleed({ children, omitPadding = false }: { children: ReactElement; omitPadding?: boolean }) {
+  const { paddingX } = useScrollableParent();
+  return !paddingX
+    ? children
+    : cloneElement(children, {
+        style: Css.mlPx(paddingX * -1)
+          .w(`calc(100% + ${px(paddingX * 2)})`)
```
stephenh, line 12: Just curious, why do we need both negative margin + the width calc?

bdow: Yes, I honestly didn't even think of doing with with a margin-right value as well. I can update to do it that way instead.
My thought process was that we're moving it to the left x number of pixels, so have it be x number of pixels wider. But you're right, we could use negative margins in both directions to stretch it

bdow: Updated.

src/components/Layout/FullBleed.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { useScrollableParent } from "src/components/Layout/ScrollableParent";
+import { Css } from "src/Css";
+
+/** Provides a way to extend the full width of the ScrollableParent */
+export function FullBleed({ children, omitPadding = false }: { children: ReactElement; omitPadding?: boolean }) {
+  const { paddingX } = useScrollableParent();
+  return !paddingX
+    ? children
+    : cloneElement(children, {
+        style: Css.mxPx(paddingX * -1)
```
stephenh, line 11: I'm kinda surprised setting the `style` like this works...but I get it! The `.$` just returns a hash of CSS styles, so :shrug: it doesn't literally have to be output by emotion.

src/components/Layout/Layout.stories.tsx
```diff
@@ -165,20 +156,23 @@ function TestLayout({ children }: PropsWithChildren<{}>) {
...
       {/* Required to use `overflowHidden` as the prevent the `TestLayout`'s scrollbar from kicking in. */}
       <div css={Css.df.overflowHidden.$}>
         <TestSideNav />
-        <ScrollableParent xss={Css.fg1.$}>{children}</ScrollableParent>
+        <ScrollableParent xss={Css.fg1.$} paddingX={increment(3)}>
+          {children}
+        </ScrollableParent>
       </div>
-    </TestLayout>
+    </PreventBrowserScroll>
```
bdow, line 175: One slight bummer about supplying the padding on the ScrollableParent component is that nesting ScrollableParents could lead to doubling up the padding. This would mean that we can't really nest Layouts.

src/components/Layout/ScrollableParent.tsx
```diff
@@ -29,8 +32,9 @@ export function ScrollableParent({ children, xss }: PropsWithChildren<Scrollable
        * Otherwise, the flex-item's min-height/width is based on the content of the flex-item, which maybe overflow the container.
        * See https://stackoverflow.com/questions/42130384/why-should-i-specify-height-0-even-if-i-specified-flex-basis-0-in-css3-flexbox */}
       <div css={{ ...Css.mh0.mw0.df.fdc.$, ...xss }}>
-        {children}
-        <div css={Css.overflowAuto.$} ref={scrollableRef}></div>
+        <div css={Css.pxPx(context.paddingX).$}>{children}</div>
```
stephenh, line 35: Ah, I see, I was wondering about, instead of a new `paddingX` prop, we just tried to pulling the padding from the `xss`, but you're really setting the padding in here, and not on the root element...

Still, kind of wonder if we could do that...

Like the caller does `ScrollableParent xss=Css.px1.$` which gets passed in as having `padding-left` and `padding-right` set, but we a) see that they're set and put them into the context value, b) unset the in the `xss` and c) actually set it on this div right here...

I dunno, probably too fancy, but would be cute...

src/components/Layout/ScrollableParent.tsx
```diff
@@ -3,21 +3,24 @@ import { Css, Properties } from "src/Css";
...
 
 const ScrollableParentContext = createContext<ScrollableParentContextProps>({
   scrollableEl: document.createElement("div"),
+  paddingX: 0,
 });
 
 // Allow any css to be applied to the ScrollableParent container.
 interface ScrollableParentContextProviderProps {
   xss?: Properties;
+  paddingX?: number;
```
stephenh, line 17: See other comment about a probably-too-cute way to just have this come from `xss`, but if still keeping this as a separate prop (which is fine/a lot simpler), wdyt about `px` to double down on our truss/tachyons naming?

stephenh (approved):
Nice! Worked out that we already had the context in place to pass this around, this is makes a ton of sense vs. "just put px1 everywhere". :-)

src/components/Layout/ScrollableParent.tsx
bdow: Yea, I get what you're saying. I wonder if that'd would be a bit much. Because we'd need to check for the existence of both paddingLeft and paddingRight, and then check they are the same value to assume this is what the user really wants to do.

bdow: Granted, I guess we don't need to see that they are the same,.. we can use each individually and have a `pl` value and a `pr` value.

PR #305, opened 2021-09-17
fix: Tooltip positioning issue when used inside of SuperDrawerContent

DeanGilewicz:
- define a z-index for tooltip to ensure it is displayed on top of other overlays


bdow (approved):
Looks good. Make sure you prefix your the PR title with `fix:` and before merging it the Squash & Merge (commit) title also has the same prefix.

PR #307, opened 2021-09-18
fix: Use custom fr calc to fix virtual alignment

stephenh:
Initially I was only going to support `%` and `px` units in the `as=virtual` mode.

But after futzing around with how to handle default/unset columns, with some sort of cute "well, split the rest of the remaining space equally via a `calc` formula", I discovered that, for whatever reason, we if make a `calc` expression that seems like literally what `fr` is supposed to do internally, for some reason the alignment issues between the header and body rows just magically goes away.

Note that, at least for me, with the native `fr` left in, it was still hard to get the two to misalign, but it was possible, and I can reproduce it by putting "the real `fr`" back. But ow with "the fake `fr`s", AFAICT they are always lined up.

I added some dup columns to the Virtualized Filtering story to help test.


src/components/Table/GridTable.test.tsx
```diff
@@ -581,6 +582,40 @@ describe("GridTable", () => {
...
+
+    it("as=virtual rejects relative units", () => {
+      expect(() => calcGridColumns("virtual", [{ w: "10fr" }, { w: 2 }] as any, undefined)).toThrow(
+        "as=virtual only supports px or percentage units",
+      );
+    });
+
+    it("as=virtual with both px and default", () => {
+      expect(calcGridColumns("virtual", [{ w: 200 }, {}, {}] as any, undefined)).toEqual(
+        "200px 33.33333333333333% 33.33333333333333%",
```
stephenh, line 614: @bdow @KoltonG how fancy do you think we should get in this scenario? I.e.

* Leave as-is, it makes sense?
* Change it to `200px 50% 50%`
* Change it to `200px calc((100% - 200px) / 2) calc((100% - 200px) / 2)` (I don't even know if that is possible)

stephenh: Oh hey, the 3rd option is doable, so added that...

stephenh: ...I think I just re-implemented the `fr` unit...

chr1sjf0x (approved):

src/components/Table/GridTable.tsx
```diff
@@ -631,16 +631,88 @@ const VirtualRoot = memoizeOne<
...
+ *
+ * Because of two things:
+ *
+ * a) react-virtuoso puts the header in a different div than the normal rows, and
+ *
+ * b) content-aware sizing just in general look janky/constantly resize while scrolling
+ *
+ * When we're as=virtual, we change our default + enforce only fixed-sized units (% and px)
+ */
+export function calcVirtualGridColumns(columns: GridColumn<any>[], maxCardPadding: number | undefined): string {
```
KoltonG, line 645: I think the challenge came when using `minmax`, `min-content`, `max-content` for the `w` attribute since both tracks couldn't talk to each other. I think here you just allow px, % and fr values.

PR #308, opened 2021-09-18
fix: Remove hardcoded width from NavLink side variant

bdow:
The width isn't really necessary. Having a hard coded with is more trouble than its worth, and the size should be flexible based on content and parent element.


stephenh (approved):
Makes sense

PR #309, opened 2021-09-20
fix: ScrollableParent component to BeamRTL wrapper

bdow:


stephenh (commented):
Little mixed, I think ScrollableParent seems too/very layout specific, vs. the BeamProvider which is more clearly global context stuff.

I.e. in internal-frontend we have two ScrollableParents, one for the side bar, and one for the right hand, which I think is a nudge that the scope of a ScrollableParent isn't global/so we could leave it out.

bdow:
> Little mixed, I think ScrollableParent seems too/very layout specific, vs. the BeamProvider which is more clearly global context stuff.
> 
> I.e. in internal-frontend we have two ScrollableParents, one for the side bar, and one for the right hand, which I think is a nudge that the scope of a ScrollableParent isn't global/so we could leave it out.

I agree. Will close and try the other behavior that we chatted about over Slack.

PR #310, opened 2021-09-20
fix: Provide escape hatch for rendering ScrollableContent children not within ScrollableParent context

bdow:


src/components/Layout/ScrollableContent.tsx
```diff
@@ -11,8 +11,14 @@ export function ScrollableContent({
...
   children: ReactNode;
   virtualized?: boolean;
-}): ReactPortal {
+}): ReactPortal | JSX.Element {
   const { scrollableEl, pl } = useScrollableParent();
+
+  // Escape hatch specifically for tests where a "ScrollableParent" context may not be present.
+  if (!scrollableEl) {
+    return <>{children}</>;
+  }
```
bdow, line 20: Wondering if this should have some sort of warning associated with it to tell the user that scrolling may not work. But also don't want to clutter up the test output.

src/components/Layout/ScrollableContent.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <ScrollableContent>
+          <div data-testid="pageContent">Hello World!</div>
+        </ScrollableContent>
+      </ScrollableParent>,
+    );
+
+    expect(r.pageContent()).toBeTruthy();
+  });
+
+  it("renders ScrollableContent children in inline if not wrapped in ScrollableParent", async () => {
```
chr1sjf0x, line 17: nit: 
```suggestion
  it("renders ScrollableContent children inline if not wrapped in ScrollableParent", async () => {
```

src/components/Layout/ScrollableContent.tsx
chr1sjf0x: Is this `only` expected to occur in tests?

chr1sjf0x (approved):
🍣

src/components/Layout/ScrollableContent.tsx
stephenh: Yeah, right ... I'm pretty sure create-react-app sets `NODE_ENV` differently between tests vs. dev build vs. production build ... but then procurement-frontend uses vite, which doesn't use `process.env.NODE_ENV`, it uses the newer `import.meta.FOO_BAR` or something like that...

I think I'd be fine skipping it for now and coming back to figuring out a warning if it proves troublesome.

stephenh (approved):
Nice!

src/components/Layout/ScrollableContent.tsx
bdow: Yea, though could end up happening in Storybook too. Granted most of the time we use storybook just for visual diffs, but this could mean the output in storybook may differ from the app if the user does not include the ScrollableParent/portal in storybook, but does include it in the App. Why that differs is because it could change layout arrangements.

PR #313, opened 2021-09-21
fix: Allow wip value to be negative in onChange

ksk5280:
This is to fix a bug where the negative value in a `NumberField` input is stripped in the `onChange` function. This causes an issue in a field where I have a rule that is checking the value of the input and returning an error if the value is positive, which it always is when the negative sign is removed.

video of issue:
https://share.getcloudapp.com/d5uRQW7Z

rule that is not working with this bug:
https://github.com/homebound-team/internal-frontend/pull/1327/files#diff-4ee1020222eca7718138803724c9c886778201d0fa508b945411f4c858cd2b35R580-R585


src/inputs/NumberField.tsx
```diff
@@ -120,7 +120,7 @@ export function NumberField(props: NumberFieldProps) {
       // This is called on each DOM change, to push the latest value into the field
       onChange={(rawInputValue) => {
         // If the wip value is invalid, i.e. it's `10b`, don't push that back into the field state
-        const wip = Number((rawInputValue || "").replace(/[^0-9\.]/g, ""));
+        const wip = Number((rawInputValue || "").replace(/[^-0-9\.]/g, ""));
```
KoltonG, line 123: **praise:** Great find!!

KoltonG (commented):
🍣

KoltonG (approved):

src/inputs/NumberField.tsx
bdow, line 123: Thanks for fixing, could you also please add a test?

bdow: Thanks for adding. Very odd that the test passed for negative numbers without the change you made. Makes me want to look into why that'd happen.

src/inputs/index.ts
```diff
@@ -2,7 +2,7 @@ export * from "./Checkbox";
 export * from "./CheckboxGroup";
 export * from "./DateField";
 export * from "./MultiSelectField";
-export * from "./NumberField";
+export type { NumberField, NumberFieldProps } from "./NumberField";
```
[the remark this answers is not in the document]
ksk5280: I got this ts error when I didn't add `type` after the export: `Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.ts(1205)`

I can do this without getting that error:
```
export { NumberField } from "./NumberField";
export type { NumberFieldProps } from "./NumberField";
```

src/inputs/NumberField.tsx
ksk5280: I'm thinking of changing this to just `wip = parseFloat(rawInputValue)` instead of using regex. Any objections to that?

src/inputs/index.ts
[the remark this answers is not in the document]
bdow: Okay, that works for me. Thanks for updating!

src/forms/BoundNumberField.tsx
```diff
@@ -1,6 +1,6 @@
 import { FieldState } from "@homebound/form-state";
 import { Observer } from "mobx-react";
-import { NumberField, NumberFieldProps } from "src/inputs";
+import { NumberField, NumberFieldProps } from "src/inputs/NumberField";
```
chr1sjf0x, line 3: question: What's up with all these import changes?

chr1sjf0x (approved):
Thanks @ksk5280

src/inputs/NumberField.test.tsx
```diff
@@ -91,6 +91,35 @@ describe("NumberFieldTest", () => {
   });
 });
 
+describe("parseRawInput function", () => {
+  it.each([
```
bdow, line 95: Cool test!

PR #314, opened 2021-09-21
[DO NOT MERGE]: SchedulesV2 Exploration

KoltonG:
This PR is to share some exploratory code to determine the following:

- Can GridTable be styled similarly to Schedules  V2?
- Can GridTable support Drag and Drop?
- Can GridTable support GroupBy?
- Can GridTable support Bulk Selections / Actions?
- Can GridTable support Expand / Collapse at multiple levels?
- Can GridTable support sorting as expected for SchedulesV2?


package.json
```diff
@@ -40,8 +40,10 @@
     "date-fns": "^2.21.3",
     "dompurify": "^2.3.0",
     "framer-motion": "^4.1.11",
+    "html2canvas": "^1.3.2",
```
KoltonG, line 43: My bad, will change to devDependencies

src/components/Table/styles.tsx
```diff
@@ -0,0 +1,43 @@
+import { Css, Palette } from "src";
+import { GridStyle } from ".";
+
+/** Our original table look & feel/style. */
+export const defaultStyle: GridStyle = {
```
bdow, line 5: Thanks for breaking this out. Makes sense

src/components/Icon.tsx
```diff
@@ -43,6 +43,16 @@ export function Icon(props: IconProps) {
  */
 export const Icons = {
   // Actions
+  drag: (
```
KoltonG, line 46: Let me know if this is intrusive? I used the SVG for SchedulesV2,

src/components/Table/GridTable.tsx
```diff
@@ -69,8 +69,10 @@ export interface GridStyle {
 }
 
 export interface NestedCardsStyle {
-  /** The space between top-level cards. */
-  topLevelSpacerPx: number;
+  /** Space between each card. */
+  spacerPx: number;
+  /** Width of the first and last "hidden" columns used to align nested columns. */
+  firstLastColumnWidth?: number;
```
KoltonG, line 75: Since calculating this was causing horizontal layout shifts and we seemed to have static numbers, we moved to this.

src/components/Table/GridTable.tsx
```diff
@@ -95,56 +97,13 @@ export interface NestedCardStyle {
...
-  spacerPx: number;
 }
 
 export interface GridTableDefaults {
   style: GridStyle;
   stickyHeader: boolean;
 }
 
-/** Our original table look & feel/style. */
-export const defaultStyle: GridStyle = {
```
KoltonG, line 108: My OCD...

src/components/Table/GridTable.tsx
```diff
@@ -345,21 +302,20 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
-        if (isCard && nestedCards) {
-          maxCardPadding = nestedCards.maxCardPadding(maxCardPadding);
-        }
         filteredRows.push([row, makeRowComponent(row)]);
       }
 
       const isCollapsed = collapsedIds.includes(row.id);
-      if (!isCollapsed && row.children) {
+      if (!isCollapsed && !!row.children?.length) {
+        nestedCards && matches && nestedCards.addSpacer();
```
KoltonG, line 318: This was important to have a spacer between the parent and child.

src/components/Table/GridTable.tsx
```diff
@@ -496,10 +454,10 @@ function renderTable<R extends Kinded>(
   headerRows: RowTuple<R>[],
   filteredRows: RowTuple<R>[],
   firstRowMessage: string | undefined,
-  stickyHeader: boolean,
-  maxCardPadding: number | undefined,
+  _stickyHeader: boolean,
```
KoltonG, line 457: This is for better TS ignores.

src/components/Table/GridTable.stories.tsx
```diff
@@ -237,7 +237,7 @@ export function NestedCardsThreeLevels() {
   const spacing = { brPx: 4, pxPx: 4, spacerPx: 1 };
   const nestedStyle: GridStyle = {
     nestedCards: {
-      topLevelSpacerPx: 8,
+      spacerPx: 8,
```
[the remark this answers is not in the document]
stephenh: Do you mean renamed as in it's _always_ the `spacerPx` now and not just the top-level spacer px? Should we rename `spacerPx` from each card-level settings then?

src/components/Table/GridTable.tsx
stephenh: Just checking but this is adding a spacer between the parent and the 1st child after it, but not any space after a child? I.e. it's

```
* parent 1
* spacer
* child 1
* child 2
* child 3
* parent 2
* ```

And not:

```
* parent 1
* spacer
* child 1
* child 2
* child 3
* spacer
* parent 2
* ```

?

I think that makes sense, but that we don't need space after a child, going into the next parent, but just checking.

src/components/Table/GridTable.tsx
```diff
@@ -370,19 +326,21 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
       const length = rows.length;
       rows.forEach((row, i) => {
         if (row.kind === "header") {
+          nestedCards && nestedCards.maybeOpenCard(row);
```
stephenh, line 329: This is to make the header look like a card? I kinda thought the 1st set of mocks had "doesn't look like a card" headers, but I guess now they do? That's cool, just checking.

src/components/Table/GridTable.tsx
stephenh: Cool

src/components/Table/nestedCards.tsx
```diff
@@ -42,22 +50,14 @@ export class NestedCards {
     this.openCards.pop();
   }
 
-  addSpacerBetweenChildren() {
-    const openCard = this.openCards[this.openCards.length - 1];
-    // If we're between two top-level cards, there is no open card, so fallback on topLevelSpacerPx
-    const height = openCard?.spacerPx || this.style.nestedCards!.topLevelSpacerPx;
```
stephenh, line 48: Cool, yeah let's remove the `NestedCardStyle.spacerPx` then.

src/components/Table/GridTable.stories.tsx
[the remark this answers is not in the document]
KoltonG: Sorry for the confusion! I've renamed the `topLevelSpacerPx` to `spacerPx` and removed the `spacerPx` from the `NestedCardStyle` type. The `NestedCardStyle` will no longer have the `spacerPx` field.

src/components/Table/GridTable.tsx
KoltonG: Yes, correct! The "after child" situation seemed to look good in the POC so I opted to not muck with it.

src/components/Table/GridTable.tsx
KoltonG: It shouldn't, looking at the screenshot, but I think that is because I removed the borderRadius. 

I will see if I can remove this and still get the style that I want.

- [ ] Verify if this change is needed.

src/components/Table/nestedCards.tsx
KoltonG: I believe I did! Will double check.

- [ ] Validate that `NestedCardStyle.spacerPx` is removed.

KoltonG: I've confirmed that it's removed and also removed any lingering usages.

src/components/Table/GridTable.tsx
stephenh: Ah okay, cool, that makes sense.

PR #315, opened 2021-09-21
fix: Support scrolling fallback in ScrollableParent if ScrollableContent not defined

bdow:


src/components/Layout/Layout.stories.tsx
```diff
@@ -223,10 +250,11 @@ function TestProjectLayout({ children }: PropsWithChildren<{}>) {
   return (
     <PreventBrowserScroll>
       <TestTopNav />
-      {/* Required to use `overflowHidden` as the prevent the `TestLayout`'s scrollbar from kicking in. */}
+      {/* Required to use `overflowHidden` to prevent the `PreventBrowserScroll`'s scrollbar from kicking in,
+          which would scroll both the side nav and the main content at the same time. */}
       <div css={Css.df.overflowHidden.$}>
         <TestSideNav />
-        <ScrollableParent xss={Css.fg1.px3.$}>{children}</ScrollableParent>
+        <ScrollableParent xss={Css.px3.$}>{children}</ScrollableParent>
```
bdow, line 257: `fg1` is now set on ScrollableParent by default.

src/components/Layout/ScrollableParent.tsx
```diff
@@ -28,21 +40,31 @@ export function ScrollableParent(props: PropsWithChildren<ScrollableParentContex
...
+    pr: paddingRight ?? 0,
+    setPortalTick: setTick,
+  };
 
   useEffect(() => {
     scrollableRef.current!.appendChild(scrollableEl);
   }, [scrollableEl]);
 
+  // Whenever the `tick` changes, reevaluate the needs for fallback scrollable styles
+  useEffect(() => setHasScrollableContent(scrollableEl.childNodes.length > 0), [tick]);
```
stephenh, line 59: I wonder if we need another `useState` here for this and instead could just do:

```
const hasScrollableContent = scrollableEl.childNodes.length > 0;
```

?

It seems cheap to compute on each render.

src/components/Layout/ScrollableParent.tsx
```diff
@@ -28,21 +40,31 @@ export function ScrollableParent(props: PropsWithChildren<ScrollableParentContex
...
+
   return (
     <ScrollableParentContext.Provider value={context}>
       {/* mh0/mw0 will respect the flexbox boundaries of the "flex-direction" if set on a parent.
        * Otherwise, the flex-item's min-height/width is based on the content of the flex-item, which maybe overflow the container.
        * See https://stackoverflow.com/questions/42130384/why-should-i-specify-height-0-even-if-i-specified-flex-basis-0-in-css3-flexbox */}
-      <Tag css={{ ...Css.mh0.mw0.df.fdc.$, ...otherXss }}>
-        <div css={Css.pl(context.pl).pr(context.pr).$}>{children}</div>
+      <Tag css={{ ...Css.mh0.mw0.fg1.df.fdc.$, ...otherXss }}>
+        <div css={Css.pl(context.pl).pr(context.pr).if(!hasScrollableContent).h100.overflowAuto.$}>{children}</div>
```
stephenh, line 67: Nice!

src/components/Layout/Layout.stories.tsx
```diff
@@ -82,6 +69,47 @@ export function VirtualizedScrolling() {
   );
 }
 
+export function WithoutScrollableParent() {
+  return (
+    <PreventBrowserScroll>
+      <TestTopNav />
+      <TestHeader title="Change Event - Mud Room" />
+      <p css={Css.py1.$}>
+        This is page "forgot" to use the "ScrollableParent" component, though still can scroll thanks to an overflow
```
stephenh, line 78: Crazy, I didn't realize how many permutations of forgetting you were covering. Pretty cool!

stephenh (approved):

src/components/Layout/ScrollableParent.tsx
bdow: Hmm.. Yea, I just kind of went all hook based for this, but it seems even this `useEffect` is not necessary, or the `tick` value. So this would become:
```
const [, setTick] = useState(0);
const hasScrollableContent = scrollableEl.childNodes.length > 0;
```
And that works! Pretty cool... I really figured I'd have to watch that `tick` value change, but I guess just calling a State Hook triggers the re-render.

stephenh: Nice!

PR #316, opened 2021-09-21
feat: WIP Frozen columns

stephenh:


KoltonG:
Nice to see this WIP! Excited to see how it comes out!

src/components/Table/GridTable.tsx
```diff
@@ -957,6 +957,12 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
           // Then apply any header-specific override
           ...(isHeader && style.headerCellCss),
           ...(isHeader && stickyHeader && Css.sticky.top(stickyOffset).z1.$),
+          // Frozen prototype
+          ...(columnIndex === 0 ? Css.sticky.z1.left0.bgWhite.$ : {}),
+          ...(columnIndex === 1 ? Css.sticky.z1.left(`${columns[columnIndex - 1].w}`).bgWhite.$ : {}),
```
KoltonG, line 962: **question:** I see that this is assuming that there is a `w` attribute for each column. I'm wondering if we can have this working without having an explicit `w` associated? Refs maybe?

src/components/Table/GridTable.stories.tsx
```diff
@@ -95,10 +95,14 @@ export function VirtualFiltering() {
...
-      { header: "Value", data: ({ value }) => value },
-      { header: "Action", data: () => <div>Actions</div>, clientSideSort: false },
+      { header: "Value", data: ({ value }) => `${value} `.repeat(10) },
+      { header: "Value", data: ({ value }) => value, w: "200px" },
+      { header: "Value", data: ({ value }) => value, w: "200px" },
+      { header: "Value", data: ({ value }) => value, w: "200px" },
+      { header: "Value", data: ({ value }) => value, w: "200px" },
+      { header: "Value", data: ({ value }) => value, w: "200px" },
+      { header: "Value", data: ({ value }) => value, w: "200px" },
+      { header: "Action", data: () => <div>Actions</div>, clientSideSort: false, w: "100px" },
```
KoltonG, line 105: **question:** Did you have any thoughts on the API?

src/components/Table/GridTable.tsx
stephenh: I think eventually / on a future pre-merge iteration it will use a helper method that is very similar to our new grid template column calc, i.e. enforce using `px` or `%` or `fr`-or-our-homegrown-`fr`-if-virtual for any columns that are before the last fixed column.

src/components/Table/GridTable.stories.tsx
stephenh: No. I still kinda like the `GridColumn.frozen = true`  but would probably do a `GridTable.stickyColumns = { left: N, right: N }` API. Or maybe `GridTable.frozen = { header: true, left: N, right: N }` and remove `stickyHeader`.

KoltonG: Maybe all around the term `sticky` might work better since it's supported in CSS?

stephenh: Yeah, but imo sticky is an implementation detail and in discussing with users/PMs/googling for ideas I think this is most often called frozen header/frozen columns, so more tempted to use that term.

PR #318, opened 2021-09-24
feat: Spike on supporting colspan in GridTable

bdow:


src/components/Table/GridTable.stories.tsx
```diff
@@ -634,3 +635,41 @@ export function WrappedHeaders() {
...
+type TotalsRow = { kind: "total"; totalPriceInCents: number };
+type ColspanRow = HeaderRow | DataRow | TotalsRow;
+export function ColSpan() {
+  const nameCol = column<ColspanRow>({
+    header: "Name",
+    data: ({ name }) => name,
+    total: () => ({ content: "Totals:", colspan: 3, alignment: "right" }),
+  });
+  const detailCol = column<ColspanRow>({ header: "Details", data: ({ role }) => role, total: "" });
+  const dateCol = dateColumn<ColspanRow>({ header: "Date", data: ({ date }) => date, total: "" });
```
bdow, line 649: Kind of a bummer to always need to put a `total` property in here as we're just going to span over it. But not sure how else this could know?

stephenh: ...right, because it depends on the order that the columns are rendered in.

Wonder if we could have like an `export const skippedColumn = ...` sort of thing that by convention we use in these spots to indicate "yes we know it's being skipped"?

Or maybe something like:

```
const detailCol = column<ColspanRow>({
  header: "Details",
  data: () =>,
  ...ignoreColspan("total", "otherKind"),
});
```

And `ignoreColspan` would just create a hash with `total: () => throw new Error("should not render")` or something like that...

src/components/Table/GridTable.tsx
```diff
@@ -959,6 +966,8 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
           ...(isHeader && stickyHeader && Css.sticky.top(stickyOffset).z1.$),
           // If we're within a card, use its background color
           ...(currentCard && Css.bgColor(currentCard.bgColor).$),
+          // Add in colspan css if needed
+          ...(currentColspan > 1 ? Css.gc(`${columnIndex + 1} / span ${currentColspan}`).$ : {}),
```
stephenh, line 970: Just curious, I thought this could be just `span 3`, but we do need the `3 / span 3`?

src/components/Table/GridTable.stories.tsx
```diff
@@ -634,3 +635,47 @@ export function WrappedHeaders() {
...
+        { kind: "data", id: "2", name: "Bar", role: "VP", date: "01/29/86", priceInCents: 1_524_99 },
+        { kind: "data", id: "3", name: "Biz", role: "Engineer", date: "11/08/18", priceInCents: 80_65 },
+        { kind: "data", id: "4", name: "Baz", role: "Contractor", date: "04/21/21", priceInCents: 12_365_00 },
+        { kind: "total", id: "total", totalPriceInCents: 14_083_64 },
+      ]}
+    />
+  );
+}
+
+113_00 + 1_524_99 + 80_65 + 12_365_00;
```
stephenh, line 681: ? :-)

bdow: Whoops, left that in. Wanted to make the story as real as possible so did a little math. :)

src/components/Table/GridTable.tsx
bdow: Yea, we might be able to get away with `grid-column: span 3`, but I guess I was trying to be a bit more explicit with the definition. `grid-column` is shorthand for `grid-column-start` and `grid-column-end`.
`grid-column: 2 / span 3` translates to:
```
grid-column-start: 2;
grid-column-end: span 3;
```
`grid-column: span 3` translates to:
```
grid-column-start: span 3;
grid-column-end: auto;
```
The big difference here really is just the `auto`, which should translate to a default of `span 1`. 
After all that, yes you are right! ha. We could just do `grid-column: span 3`. I just kind of figured I'd be more explicit, but doesn't seem necessary.

bdow: (updated to just use `span {currentColspan}`

src/components/Table/GridTable.stories.tsx
stephenh: Pretty good to leave this as-is for now and figure out something cuter later.

stephenh (approved):

PR #319, opened 2021-09-24
feat: Support 'emptyCell' GridStyle.

bdow:
Allows implementer to define a default empty cell value for the whole table. For instance, in many tables in Blueprint we want to show the "-" character if there is no data for that cell

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -10,6 +10,7 @@ import {
   condensedStyle,
   dateColumn,
   defaultStyle,
+  emptyCell,
   GridColumn,
   GridDataRow,
   GridRowLookup,
@@ -684,9 +685,10 @@ export function CustomEmptyCell() {
   const nameColumn: GridColumn<Row> = { header: "Name", data: ({ name }) => name };
   const valueColumn: GridColumn<Row> = { header: "Value", data: ({ value }) => value };
   const actionColumn: GridColumn<Row> = { header: "Action", data: () => <div>Actions</div> };
+  const fourthColumn: GridColumn<Row> = { header: "Really Empty", data: () => emptyCell };
   return (
     <GridTable<Row>
-      columns={[nameColumn, valueColumn, actionColumn]}
+      columns={[nameColumn, valueColumn, actionColumn, fourthColumn]}
       style={{ ...condensedStyle, emptyCell: <>&mdash;</> }}
       rows={[
         { kind: "header", id: "header" },
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -33,7 +33,7 @@ export default {
   parameters: { layout: "fullscreen", backgrounds: { default: "white" } },
 } as Meta;
 
-type Data = { name: string; value: number };
+type Data = { name: string | undefined; value: number | undefined };
 type Row = SimpleHeaderAndDataOf<Data>;
 
 export function ClientSideSorting() {
@@ -644,6 +644,7 @@ export function ColSpan() {
     header: "ID",
     data: ({ id }) => id,
     // Not putting the colspan here just as a POC that we can colspan somewhere other than from the first column.
+    // We should expect this to show the `emptyCell` fallback
     total: "",
   });
   const nameCol = column<ColspanRow>({
@@ -666,6 +667,7 @@ export function ColSpan() {
   return (
     <GridTable<ColspanRow>
       columns={[idCol, nameCol, detailCol, dateCol, priceCol]}
+      style={{ ...condensedStyle, emptyCell: <>&mdash;</> }}
       rows={[
         simpleHeader,
         { kind: "data", id: "1", name: "Foo", role: "Manager", date: "11/29/85", priceInCents: 113_00 },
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -1014,11 +1016,19 @@ const ObservedGridRow = React.memo((props: GridRowProps<any, any>) => (
...
+  content: ReactNode | GridCellContent,
+  isHeader: boolean,
+  canSortColumn: boolean,
+  style: GridStyle,
+): ReactNode {
   if (typeof content === "string" && isHeader && canSortColumn) {
     return <SortHeader content={content} />;
   } else if (isContentAndSettings(content)) {
     return content.content;
+  } else if (style.emptyCell && !content) {
```
stephenh, line 1029: I wonder if we should support like a "export const noReallyEmpty" value that columns could use for cases that really do need empty? Seems like something that could come up?...

stephenh (approved):
Looks great!

src/components/Table/GridTable.tsx
bdow: Yea, probably a good idea. I was thinking that too, for instance there is sometimes content that we only want to show when the table is in `edit` mode, otherwise hide it in `read`.

bdow: Do you think that'd be a property on the column, or would we want each cell to be able to define that?

stephenh: Yeah, I was thinking it'd be a return value from a cell, i.e. something like:

```
const secondColumn: ... = {
   data: emptyCell,
}
```

Where `emptyCell`-as-a-`const` means literally empty.

src/components/Table/GridTable.stories.tsx
```diff
@@ -666,6 +667,7 @@ export function ColSpan() {
   return (
     <GridTable<ColspanRow>
       columns={[idCol, nameCol, detailCol, dateCol, priceCol]}
+      style={{ ...condensedStyle, emptyCell: <>&mdash;</> }}
```
stephenh, line 670: Wdyt about just putting this directly into our `defaultStyle`? 

Granted, Blueprint could do a `setDefaultStyle` during boot to customize it; I think procurement-frontend did that to get a slightly-different-default-`style`, but I _think_ Blueprint is just using the stock `defaultStyle`...

src/components/Table/GridTable.test.tsx
```diff
@@ -1023,6 +1023,26 @@ describe("GridTable", () => {
...
+    const r = await render(
+      <GridTable<Row>
+        columns={[nameColumn, valueColumn]}
+        style={{ emptyCell: <>empty</> }}
+        rows={[
+          simpleHeader,
+          // And some content is undefined and empty strings
+          { kind: "data", id: "2", name: "", value: 2 },
+          { kind: "data", id: "1", name: "a", value: undefined },
+          { kind: "data", id: "3", name: "c", value: 1 },
```
stephenh, line 1038: Kind of a nit, but these rows are probably un-ordered from a client-side sorting test; maybe order them to just be slightly easier to follow...

stephenh: Also what if `value` was `0`? Would we replace that b/c it's falsely? Should we?

It seems like `null` / `undefined` / even `""` should definitely be replaced with `emptyCell` ... not sure about `0` though...seems like a wrinkle...

src/components/Table/GridTable.stories.tsx
bdow: Outside of Blueprint, it seems a bit heavy handed to assume the default style is to always have some sort of content in an empty cell. I wouldn't think that is super apparent to the user.

src/components/Table/GridTable.tsx
bdow: Yea, that could work. What do you think of that vs something like:
```
const secondColumn = {
  data: "",
  allowEmpty: true,
}
```
Might be easier for implementers to discover the available prop vs finding the exported const helper. Or I guess they could even do:
```
const secondColumn = {
  data: <></>,
  // or
  data: " ",
}
```
That'd give the same result.

src/components/Table/GridTable.stories.tsx
stephenh: Ah yeah, I guess that's fair. Are you thinking of adding like a `setDefaultStyle` call on the blueprint side, or specifically opt-ing in on a table by table basis?

src/components/Table/GridTable.tsx
stephenh: Hm, yeah, the `GridColumn.allowEmpty` is across the whole column though, vs the `data` being per-cell. I guess we could do both...

I kinda like having just one, and since `data: emptyCell` is a superset of `allowEmpty` (i.e. you can do `data: emptyCell` a few times in a column to make the whole column effectively not empty, but you can't use `allowEmpty` to make only some of the cells empty), that's where I'm currently leaning.

But not against both/or start with one and see how it goes.

bdow: Whoops, I definitely wrote that out wrong. I was actually thinking of it more like:
```
const secondColumn = { 
  data: () => { allowEmpty: true }
 }

File: src/components/Table/GridTable.stories.tsx (hunk 2 above)
Comment (bdow) [reply src/components/Table/GridTable.stories.tsx:670]: I was originally thinking of opting in table by table, but maybe we want a setDefaultStyle on Blueprint. In my opinion I think Table best practices depend on the content of the table. I think there may be instances where showing the dashes would be incorrect, but maybe not?

File: src/components/Table/GridTable.test.tsx (hunk 3 above)
Comment (bdow) [reply src/components/Table/GridTable.test.tsx:1038]: Agreed. I think 0 should be shown if that is the value.

File: src/components/Table/GridTable.stories.tsx (hunk 2 above)
Comment (stephenh) [reply src/components/Table/GridTable.stories.tsx:671]: > In my opinion I think Table best practices depend on the content of the table

Yeah, I don't necessarily disagree, but at least what we're hearing from UX is that "this is for all tables", so I think it'd be interesting to start there, and then see how many exceptions are needed.

And then based on what those exceptions are / how many there are, we can work with UX to define names for the different patterns instead of just "default" or what not.

File: src/components/Table/GridTable.test.tsx [hunk 4]
```diff
@@ -1023,6 +1024,42 @@ describe("GridTable", () => {
...
+          { kind: "data", id: "1", name: "", value: null },
+          { kind: "data", id: "2", name: "a", value: undefined },
+          { kind: "data", id: "3", name: "c", value: 1 },
+        ]}
+      />,
+    );
+    // Then the cells with missing content have the `emptyCell` node applied.
+    expect(cell(r, 1, 0).textContent).toBe("empty");
+    expect(cell(r, 1, 1).textContent).toBe("empty");
+    expect(cell(r, 2, 1).textContent).toBe("empty");
```
stephenh, line 1046: Maybe go ahead and add assertions for the 2,0 and 3,0 and 3,1 cells?

Both as a pedantic sanity check that we don't literally output "empty" for _everything_ :-) but also I think it'd help the readability to have "a", "c" and "1" in here to guide the mental mapping of test input --> output assertions.

src/components/Table/GridTable.tsx
```diff
@@ -1027,6 +1038,11 @@ function isContentAndSettings(content: ReactNode | GridCellContent): content is
   return typeof content === "object" && !!content && "content" in content;
 }
 
+const emptyValues = ["", null, undefined] as any[];
```
stephenh, line 1041: Nice! I wonder how many cells this is going to actually change... :thinking: Will be interesting to see.

stephenh (approved):

PR #321, opened 2021-09-24
feat: Allow for actions to be placed on right hand side of Tabs list

bdow:


stephenh (approved):
Nice! Yeah, it's too bad we need to touch a context every time we want to do this... but... :shrug: 

...I guess, just talking about loud, what if we didn't need a brand new key each time we wanted to do this...

So like:

```
class BeamContext {
  const portalDivs: Map<string, div>;
}
```

So that tab actions could be something like:

```
const div = usePortalDiv("tabActions");
return createPortal(children, div);
```

And the tabs would look like:

```
const tabsRef = usePortalRef("tabActions");
...<div ref={tabsRef} />
```

And inside of `usePortalRef` it would do something almost exactly like:

https://github.com/homebound-team/internal-frontend/blob/main/src/hooks/usePortalDiv.tsx#L4

I.e.:

```
export function usePortalRef(name: string): [MutableRefObject<HTMLDivElement | null>] {
  const { portalDivs } = useBeamContext();
  const div = useMemo(() => {
    const div = document.createElement("div");
    portalDivs.put(name, div);
    return div;
  }, []);
  // ref is cheap to make so don't bother useMemo  
  const ref: MutableRefObject<HTMLDivElement | null> = {
    get current() {
      return null;
    },
    set current(current: HTMLDivElement | null) {
      if (current) {
        current.appendChild(div);
      }
    },
  };
  return ref;
}
```

PR #324, opened 2021-09-28
fix: Allow forcing tabs to show even if only one is enabled

bdow:
This is to support an existing requirement in Blueprint where Stage Tabs may be hiding the Planning stage, and the Construction stage may be disabled, making PreCon the only active stage. In this case, as part of Product requirements (which may be outdated) we want to show both the Pre-Con and disabled Construction tabs. Regardless if the requirements are outdated, I think allowing showing the single enabled tab is useful.

### DIFF
--- a/src/components/Tabs.tsx
+++ b/src/components/Tabs.tsx
@@ -31,7 +31,7 @@ export interface TabsProps<V extends string, X> {
   onChange: (value: V) => void;
   contentXss?: X;
   // Allow for showing the tabs even if there is only one enabled tab
-  forceShow?: boolean;
+  alwaysShowAllTabs?: boolean;
 }
 
 // Tabs can be rendered as Links (omit "onChange") and we'll use React-Router for matching (omit "selected")/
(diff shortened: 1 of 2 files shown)


src/components/Tabs.tsx
```diff
@@ -30,6 +30,8 @@ export interface TabsProps<V extends string, X> {
   tabs: Tab<V>[];
   onChange: (value: V) => void;
   contentXss?: X;
+  // Allow for showing the tabs even if there is only one enabled tab
+  forceShow?: boolean;
```
stephenh, line 34: Wdyt about `hideIfOnlyOneTabActive` and it defaults to `true`?

stephenh (approved):

ksk5280 (approved):

src/components/Tabs.tsx
bdow: Personally I prefer boolean/flag type props to be enabled by just passing them with a value, something like:
`<Tabs forceShow />` vs `<Tabs hideIfOnlyOneTabActive={false} />`.
I could go for a prop like: `disableHideTabs`, thoughts?

stephenh: Ah yeah, I guess... `disableHideTabs` has a double negative, maybe `showAllTabsEvenIfOnlyOneIsActive`? :-) 

I dunno, given the core feature is "hiding the one tab" and it defaults to "on", it's hard to avoid a double-negative of "don't hide" in the name to achieve the "setting to true" --> "inverts the default ... which is also true".

So, dunno, still kinda like the `=false` --> turning off the feature, but maybe some form of `showAllTabs`, `alwaysShowAllTabs`...

PR #330, opened 2021-10-05
fix: Allow Tab actions to show even if no tabs are being displayed

bdow:


src/components/Tabs.tsx
```diff
@@ -126,7 +126,9 @@ export function Tabs<V extends string>(props: TabsProps<V, {}> | RouteTabsProps<
   useEffect(() => setActive(selected), [selected]);
 
   useEffect(() => {
+    console.log("in effect", tabActionsRef, tabActionsDiv);
```
stephenh, line 129: Snuck in

src/components/Tabs.stories.tsx
```diff
@@ -161,6 +161,23 @@ export const TabsWithActions = () => {
   return <TabsWithContent tabs={testTabs} onChange={setSelectedTab} selected={selectedTab} ariaLabel="Sample Tabs" />;
 };
 
+export const OneTabWithActions = () => {
+  const [selectedTab, setSelectedTab] = useState("tab1");
+  const testTabs: Tab<TabValue>[] = [
+    {
+      name: "Tab 1",
+      value: "tab1",
```
stephenh, line 169: Should this tab be disabled to trigger the `hideTabs` line?

src/components/Tabs.tsx
bdow: Whoops, thanks!

src/components/Tabs.stories.tsx
bdow: No, as long as there is only one enabled tab, then the tabs will be hidden.

stephenh: Ah right! Thanks

stephenh (approved):

src/components/Tabs.tsx
```diff
@@ -157,10 +153,12 @@ export function Tabs<V extends string>(props: TabsProps<V, {}> | RouteTabsProps<
...
   }
 
-  // We also check this in TabsWithContent, but if someone is using Tabs standalone, check it here as well
+  const actionsDiv = <div css={Css.ml("auto").addIn("&>div", Css.df.aic.childGap1.$).$} ref={tabActionsRef} />;
+
+  // Check to see if we should hide the tabs. We will still return something so that the Tab Actions (if any) can render
   const hideTabs = props.alwaysShowAllTabs ? false : (props.tabs as any[]).filter((t) => !t.disabled).length === 1;
   if (hideTabs) {
-    return <></>;
+    return <div css={Css.df.aic.$}>{actionsDiv}</div>;
```
stephenh, line 161: I was thinking maybe we could sometimes leave this out, but I guess at this point we're not sure if a child will/will not end up using `TabActions` so we need to always include it. Makes sense.

PR #337, opened 2021-10-08
feat: Update DatePicker to utilize TextFieldBase and Popover

bdow:


src/inputs/TextFieldBase.tsx
```diff
@@ -53,12 +59,17 @@ export function TextFieldBase(props: TextFieldBaseProps) {
...
+    inlineLabel,
   } = props;
   const errorMessageId = `${inputProps.id}-error`;
   const labelSuffix = getLabelSuffix(required);
-
   const ElementType: React.ElementType = multiline ? "textarea" : "input";
   const tid = useTestIds(props, defaultTestId(label || "textField"));
+  const [isFocused, setIsFocused] = useState(false);
+  const { focusWithinProps } = useFocusWithin({
+    onFocusWithinChange: (isFocusedWithin) => setIsFocused(isFocusedWithin),
```
stephenh, line 71: This can probably be `onFocusWithinChange: setIsFocused`?

src/inputs/DateField.tsx
```diff
@@ -30,160 +29,146 @@ export interface DateFieldProps {
...
+  const { labelProps, inputProps } = useTextField(
+    {
+      ...textFieldProps,
+      // Open on focus of the input.
+      onFocus: (e) => {
+        state.toggle();
+        maybeCall(onFocus, e);
+      },
+      onBlur: (e) => {
+        // If we are interacting any other part of the inputWrap (such as the calendar button) return early
```
stephenh, line 77: Huh...so if it goes something like:

* focus input
* change focus to the calendar button
* this onBlur fires but early returns/does nothing b/c inputWrap.Ref.contains(related) returns true 
* change focus to some other field
* this onBlur isn't fired b/c it's hooked up directly to the `input` and not the `inputWrapper`

Will we miss the if-`!parseDate`-reset-the-date logic? How does that get called? Does it need to be?

PR #338, opened 2021-10-11
feat: CH-7223: DateFilter

bdow:
Creates DateFilter which allows for specifying an operation as well as a date. Provides an "Any" operation by default.
Adds tests and stories for DateFilter.

### DIFF
--- a/src/components/Filters/DateFilter.stories.tsx
+++ b/src/components/Filters/DateFilter.stories.tsx
@@ -1,6 +1,7 @@
 import { Meta } from "@storybook/react";
 import { taskDueFilter } from "src/components/Filters/testDomain";
 import { Filters } from "src/components/index";
+import { Css } from "src/Css";
 import { jan2 } from "src/forms/formStateDomain";
 
 export default {
@@ -11,7 +12,8 @@ export default {
 
 export function DateFilterInPage() {
   const filter = taskDueFilter("date");
-  return filter.render({ op: "BEFORE", date: jan2 }, () => {}, {}, false, false);
+  // Mimics the container of the "in page" filters to demonstrate the field can shrink in size.
+  return <div css={Css.df.aic.$}>{filter.render({ op: "BEFORE", date: jan2 }, () => {}, {}, false, false)}</div>;
 }
 
 export function DateFilterInModal() {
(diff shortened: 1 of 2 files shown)


src/inputs/TextFieldBase.tsx
```diff
@@ -149,7 +162,7 @@ export function TextFieldBase(props: TextFieldBaseProps) {
             }}
             {...tid}
           />
-          {!multiline && endAdornment && <span>{endAdornment}</span>}
+          {!multiline && endAdornment && <span css={Css.df.aic.fs0.$}>{endAdornment}</span>}
```
bdow, line 165: Snuck this in because the `endAdornment` didn't look great for compact fields.

src/components/Filters/DateFilter.tsx
```diff
@@ -0,0 +1,101 @@
...
+class DateFilter<O, V extends Key, DV extends DateFilterValue<V>>
+  extends BaseFilter<DV, DateFilterProps<O, V, DV>>
+  implements Filter<DV>
+{
+  render(value: DV, setValue: (value: DV | undefined) => void, tid: TestIds, inModal: boolean, vertical: boolean) {
+    const { label, operations, getOperationValue, getOperationLabel } = this.props;
+    const [focusedEl, setFocusedEl] = useState<"op" | "date" | undefined>();
+    const commonStyles = Css.df.aic.fs1.maxwPx(550).bt.bb.bGray300.$;
+
+    // TODO: Maybe make a `CompoundField` component, which could handle the `display: flex` and any sizing requirements per field and focus states.
```
chr1sjf0x, line 38: FYI - Schedules V2 has a compound single-select/text input. A beam ticket is likely to be inbound soon.

image

It also has that `x` to clear the field on the right, which doesn't currently exist 🤔

bdow: Hey @chr1sjf0x , thanks for the heads up! I actually chatted with Dara about this as well, and she showed me this mock. This led me in the direction of refactoring the SelectFieldInput to utilize the TextFieldBase for consistent styling between the two fields. Perhaps as part of the next ticket we can make a more generic "CompoundField" to accomplish what is happening in DateFilter here. 
And yes, we definitely do not have the "X"/clear button yet. Will be on the look out for tickets! Thanks for the review!

PR #341, opened 2021-10-13
fix: Utilize TextFieldBase for SelectFieldInput

bdow:
Additionally makes SelectField/MultiSelectField `label` prop required.

### DIFF
--- a/src/inputs/TextFieldBase.tsx
+++ b/src/inputs/TextFieldBase.tsx
@@ -84,23 +84,25 @@ export function TextFieldBase(props: TextFieldBaseProps) {
   const { hoverProps, isHovered } = useHover({});
   const { focusWithinProps } = useFocusWithin({ onFocusWithinChange: setIsFocused });
 
+  const maybeSmaller = borderless ? 2 : 0;
   const fieldHeight = 40;
   const compactFieldHeight = 32;
 
   const fieldStyles = {
     container: Css.df.fdc.w100.maxw(px(550)).$,
     inputWrapper: {
-      ...Css.sm.df.aic.br4.px1.w100.$,
+      ...Css.sm.df.aic.br4.px1.w100
+        .hPx(fieldHeight - maybeSmaller)
+        .if(compact)
+        .hPx(compactFieldHeight - maybeSmaller).$,
       ...Css.bgWhite.bGray300.gray900.if(contrast).bgGray700.bGray700.white.$,
-      // For borderless heights, remove the border from the element's height with the expectation the caller will apply a border if necessary (Example: DateFilter)
-      ...(borderless
-        ? Css.hPx(fieldHeight - 2)
-            .if(compact)
-            .hPx(compactFieldHeight - 2).$
-        : Css.ba.hPx(fieldHeight).if(compact).hPx(compactFieldHeight).$),
+      ...(!borderless ? Css.ba.$ : {}),
     },
     inputWrapperReadOnly: {
-      ...Css.sm.df.aic.w100.mhPx(fieldHeight).if(compact).mhPx(compactFieldHeight).$,
+      ...Css.sm.df.aic.w100
+        .mhPx(fieldHeight - maybeSmaller)
+        .if(compact)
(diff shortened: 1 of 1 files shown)


src/inputs/internal/SelectFieldInput.tsx
```diff
@@ -72,85 +62,64 @@ export function SelectFieldInput<O, V extends Value>(props: SelectFieldInputProp
...
-            ...hoverStyles,
-            ...disabledStyles,
-            ...readOnlyStyles,
-          }}
-          onKeyDown={(e) => {
+    <TextFieldBase
+      {...otherProps}
+      inputRef={inputRef}
+      inputWrapRef={inputWrapRef}
+      label={label || ""}
```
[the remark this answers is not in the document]
stephenh: Yep! That'd be great to do.

src/inputs/TextFieldBase.tsx
```diff
@@ -60,15 +71,49 @@ export function TextFieldBase(props: TextFieldBaseProps) {
...
-  const tid = useTestIds(props, defaultTestId(label || "textField"));
+  const tid = useTestIds(props, defaultTestId(label));
   const [isFocused, setIsFocused] = useState(false);
+  const { hoverProps, isHovered } = useHover({});
   const { focusWithinProps } = useFocusWithin({ onFocusWithinChange: setIsFocused });
 
+  const fieldHeight = 40;
+  const compactFieldHeight = 32;
+
+  const fieldStyles = {
```
stephenh, line 90: Hm, this kinda of de-inlines things vs. the usual tachyons/utility css approach, but ah I guess it's because we re-use the styles between the read-only vs. not-read-only side of things.

src/inputs/TextFieldBase.tsx
```diff
@@ -60,15 +71,49 @@ export function TextFieldBase(props: TextFieldBaseProps) {
...
+    container: Css.df.fdc.w100.maxw(px(550)).$,
+    inputWrapper: {
+      ...Css.sm.df.aic.br4.px1.w100.$,
+      ...Css.bgWhite.bGray300.gray900.if(contrast).bgGray700.bGray700.white.$,
+      // For borderless heights, remove the border from the element's height with the expectation the caller will apply a border if necessary (Example: DateFilter)
+      ...(borderless
+        ? Css.hPx(fieldHeight - 2)
+            .if(compact)
+            .hPx(compactFieldHeight - 2).$
+        : Css.ba.hPx(fieldHeight).if(compact).hPx(compactFieldHeight).$),
```
stephenh, line 100: Could maybe do `const maybeSmaller = borderless ? 2 : 0` and then `hPx(fieldHeight - maybeSmaller).if(compact(compactFieldHeight - maybeSmaller)`

src/inputs/TextFieldBase.tsx
```diff
@@ -60,15 +71,49 @@ export function TextFieldBase(props: TextFieldBaseProps) {
...
+    },
+    inputWrapperReadOnly: {
+      ...Css.sm.df.aic.w100.mhPx(fieldHeight).if(compact).mhPx(compactFieldHeight).$,
```
stephenh, line 103: Do we need the "borderless maybe smaller" logic here?

src/inputs/TextFieldBase.tsx
```diff
@@ -60,15 +71,49 @@ export function TextFieldBase(props: TextFieldBaseProps) {
...
+      ...Css.gray900.if(contrast).white.$,
+    },
+    input: {
+      ...Css.w100.mw0.outline0.br4.fg1.$,
+      // Not using Truss's inline `if` statement here because `addIn` properties do not respect the if statement.
```
stephenh, line 108: Ah interesting, sounds like a bug!

src/inputs/TextFieldBase.tsx
bdow: Yea, probably should. Would keep it consistent. Will update, thanks!

PR #365, opened 2021-10-30
fix: RichTextField to initialize based on trix supplied init event

bdow:
Removing the document.getElementById call to grab the editor element and replacing with a useRef. This wasn't safe as when using the RichTextField within a Portal (such as a Modal) it was possible that the element for the editor is not yet in the DOM.
Now listening to the 'trix-initialize' event in order to set the editor ref and call 'loadHTML'. This is not immediately available off of the element's ref because the 'editor' is attached to the 'trix-editor' web component using 'requestAnimationFrame', which waits for the event loop to clear before running its execution (similar to a setTimeout(..., 0))

### DIFF
--- a/src/inputs/RichTextField.tsx
+++ b/src/inputs/RichTextField.tsx
@@ -30,11 +30,6 @@ export interface RichTextFieldProps {
   readOnly?: boolean;
 }
 
-// There aren't types for trix, so add our own. For now `loadHTML` is all we call anyway.
-type Editor = {
-  loadHTML(html: string): void;
-};
-
 /**
  * Provides a simple rich text editor based on trix.
  *
@@ -47,8 +42,7 @@ export function RichTextField(props: RichTextFieldProps) {
 
   // We get a reference to the Editor instance after trix-init fires
   const editor = useRef<Editor | undefined>(undefined);
-  const editorElement = useRef<HTMLElement>(null);
-  const parentRef = useRef<HTMLElement>(null);
+  const editorElement = useRef<HTMLElement>();
 
   // Keep track of what we pass to onChange, so that we can make ourselves keep looking
   // like a controlled input, i.e. by only calling loadHTML if a new incoming `value` !== `currentHtml`,
@@ -67,11 +61,13 @@ export function RichTextField(props: RichTextFieldProps) {
   const id = useMemo(() => {
     if (readOnly) return;
 
-    const id = `${new Date().getTime()}-${Math.floor(Math.random() * 1000)}`;
+    const id = `trix-editor-${trixId}`;
+    trixId++;
 
     function onEditorInit(e: Event) {
-      const elId = (e.target as HTMLElement).id;
-      if (elId === `trix-editor-${id}`) {
+      const targetEl = e.target as HTMLElement;
+      if (targetEl.id === id) {
+        editorElement.current = targetEl;
         editor.current = (editorElement.current as any).editor;
         if (mergeTags !== undefined) {
           attachTributeJs(mergeTags, editorElement.current!);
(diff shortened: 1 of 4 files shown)


src/inputs/RichTextField.tsx
```diff
@@ -130,22 +119,63 @@ export function RichTextField(props: RichTextFieldProps) {
...
+      parentRef.current.addEventListener("trix-initialize", onEditorInit);
+      setShowTrix(true);
+
+      return () => {
+        if (parentRef.current) {
+          parentRef.current.removeEventListener("trix-initialize", onEditorInit);
+        }
+      };
+    }
+  }, [parentRef]);
```
stephenh, line 143: Does this need to be `parentRef.current`? I don't think `parentRef` itself will ever change, unless that is your intent? If so, maybe just use `[]`? ...I'm ~pretty sure that refs are ignored in the usual "check that variables you access in the lambda are in the deps list" rule.

src/inputs/RichTextField.tsx
```diff
@@ -130,22 +119,63 @@ export function RichTextField(props: RichTextFieldProps) {
...
+        editor.current = (editorElement.current as any).editor;
+        if (mergeTags !== undefined) {
+          attachTributeJs(mergeTags, editorElement.current!);
+        }
+
+        currentHtml.current = value;
+        editor.current!.loadHTML(value || "");
+      }
+
+      parentRef.current.addEventListener("trix-initialize", onEditorInit);
```
stephenh, line 134: What if, on the first render, we attached this directly to `window`? Like:

* Use `useMemo` with `deps=[]` to get a unique-ish id (to handle two RichTextFields initing during the same render)
* Use the same `useMemo` to hook up a `window.trix-initialize` callback 
* Use the unique-ish id in the call to `createElement`
* When the callback is fired, make sure the event is for our unique-ish id, save the editor to a ref, and unregister our `window.trix-initiatize` callback?

bdow: Would be better than the rerender solution I have now. Will give it a shot. Thanks for the suggestion.

bdow: Updated! Thanks for the suggestion

src/inputs/RichTextField.tsx
```diff
@@ -63,28 +63,35 @@ export function RichTextField(props: RichTextFieldProps) {
...
-      editor.current = (editorElement as any).editor;
-      if (!editor.current) {
-        throw new Error("editor not found");
-      }
-      if (mergeTags !== undefined) {
-        attachTributeJs(mergeTags, editorElement!);
+    function onEditorInit(e: Event) {
+      const elId = (e.target as HTMLElement).id;
+      if (elId === `trix-editor-${id}`) {
+        editor.current = (editorElement.current as any).editor;
```
stephenh, line 75: I'm kind surprised this works... Like I thought the "attaching `trix-initialize` to `window` in an useMemo" was we needed to attach the event listener _immediately_ because anything in `useEffect` (or anything that uses a ref) would miss the `trix-initialize` event? 

But if `editorElement.current` is correctly assigned here, that means it's ref was assigned before this was triggered...could we just attach `trix-initialize` directly to `editorElement.current` in a `useEffect`? 

Also, if that doesn't work, we could probably use `e.target === editorElement.current` as the "is this the `trix-initialize` event for my element" instead of the random id?...

bdow: Good question, and I wonder if the `strictMode` flag was messing up my testing for this...
I had put a log statement in the trix code when the `trix-initialize` is called, and that happens before the `useEffect` is executed.
But good question on "how is the ref available then?". Seems like this is a potential race condition...

stephenh: Maybe just `editorElement.current` directly from `e.target`, and skip the ref all together?

bdow: Just to make sure I thinking through it correctly - If I set the `editorElement.current` directly form `e.target`, then I'll want to keep that random id around as well to ensure I am setting the ref to the proper element.

stephenh: @bdow yeah, that sounds right

stephenh: Also, the "random" id really doesn't have to be random per se, just unique, i.e. you could do:

```
let trixId = 1;
```

As a top-level module const and just use `trixId++` for each new trix instance.

src/inputs/RichTextField.tsx
```diff
@@ -63,65 +57,59 @@ export function RichTextField(props: RichTextFieldProps) {
...
-        editorElement.removeEventListener("trix-change", trixChange as any);
-        editorElement.removeEventListener("trix-blur", trixBlur as any);
-        editorElement.removeEventListener("trix-focus", trixFocus as any);
-        window.removeEventListener("trix-file-accept", preventDefault);
-      };
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [readOnly],
-  );
+    // Attaching listener to the `window` to we're listening prior to render.
```
stephenh, line 109: Maybe throw in a comment that this is specifically because web component events might fire before our `useEffect`s run, particularly when we're rendered in a portal?

stephenh (approved):

PR #367, opened 2021-11-02
fix: Remove cloneElement from Tooltip to avoid

bdow:
The previous implementation to clone the element resulted in loss of functionality for the cloned element as the tooltip trigger props would overwrite pointer events tied to the element.
Now wrapping in our own span to provide the tooltip actions and keeps the existing children component/element as is.


src/components/Tooltip.tsx
```diff
@@ -28,7 +28,9 @@ export function Tooltip(props: TooltipProps) {
 
   return (
     <>
-      {cloneElement(children, { ref: triggerRef, ...triggerProps })}
+      <span ref={triggerRef} {...triggerProps}>
+        {children}
+      </span>
```
stephenh, line 33: Well that's a terribly simple fix, in terms of LOC. :-D

stephenh (approved):

PR #371, opened 2021-11-03
feat: Generic CompoundField component

bdow:
Utilizing CompoundField to create the BoundSelectAndTextField, expected for use on SchedulesV2 in Blueprint.

### DIFF
--- a/src/components/internal/CompoundField.tsx
+++ b/src/components/internal/CompoundField.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { cloneElement, useState } from "react";
 import { Css } from "src/Css";
 import { TextFieldInternalProps } from "src/interfaces";
 import { maybeCall } from "src/utils";
@@ -20,21 +20,17 @@ export function CompoundField({ children }: { children: JSX.Element[] }) {
           ...Css.bl.borderRadius("4px 0 0 4px").if(focusedEl === "left").bLightBlue700.$,
         }}
       >
-        {{
-          ...children[0],
-          props: {
-            ...children[0].props,
-            onFocus: () => {
-              maybeCall(children[0].props.onFocus);
-              setFocusedEl("left");
-            },
-            onBlur: () => {
-              maybeCall(children[0].props.onBlur);
-              setFocusedEl(undefined);
-            },
-            internalProps,
+        {cloneElement(children[0], {
+          onFocus: () => {
+            maybeCall(children[0].props.onFocus);
+            setFocusedEl("left");
           },
-        }}
+          onBlur: () => {
+            maybeCall(children[0].props.onBlur);
+            setFocusedEl(undefined);
+          },
+          internalProps,
+        })}
       </div>
       {/* Separation line */}
       <div css={Css.wPx(1).flexNone.bgGray300.if(focusedEl !== undefined).bgLightBlue700.$} />
@@ -45,21 +41,17 @@ export function CompoundField({ children }: { children: JSX.Element[] }) {
           ...Css.fg1.br.borderRadius("0 4px 4px 0").if(focusedEl === "right").bLightBlue700.$,
         }}
       >
-        {{
-          ...children[1],
-          props: {
-            ...children[1].props,
-            onFocus: () => {
-              maybeCall(children[1].props.onFocus);
-              setFocusedEl("right");
-            },
-            onBlur: () => {
-              maybeCall(children[1].props.onBlur);
-              setFocusedEl(undefined);
-            },
-            internalProps,
+        {cloneElement(children[1], {
+          onFocus: () => {
+            maybeCall(children[1].props.onFocus);
+            setFocusedEl("right");
           },
-        }}
+          onBlur: () => {
+            maybeCall(children[1].props.onBlur);
+            setFocusedEl(undefined);
+          },
+          internalProps,
+        })}
       </div>
     </div>
   );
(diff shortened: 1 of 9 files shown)


src/forms/BoundSelectAndTextField.test.tsx
```diff
@@ -0,0 +1,69 @@
+import { ObjectConfig, required, useFormState } from "@homebound/form-state";
+import { click } from "@homebound/rtl-utils";
+import { fireEvent } from "@testing-library/react";
+import { Observer } from "mobx-react";
+import { BoundSelectAndTextField } from "src";
+import { ScheduleTypes } from "src/components/Filters/testDomain";
+import { render, type } from "src/utils/rtl";
+
+describe("BoundSelectAndTextField", () => {
+  it("renders and updates values", async () => {
```
khalidwilliams, line 10: **Question:** is there any value to testing that clearing works when rendered as clearable? Might be redundant based on other coverage.

src/components/internal/CompoundField.tsx
```diff
@@ -0,0 +1,66 @@
+/** Internal component to help create compound fields */
```
stephenh, line 1: Looks like this snuck up to the top (happens with auto-add import sometimes)

src/components/internal/CompoundField.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <div css={Css.df.$}>
+      <div
+        css={{
+          ...commonStyles,
+          ...Css.bl.borderRadius("4px 0 0 4px").if(focusedEl === "left").bLightBlue700.$,
+        }}
+      >
+        {{
```
stephenh, line 23: I don't think I've seen this syntax...is this like a `cloneElement`?  Seems neat either way.

stephenh (approved):

src/forms/BoundSelectAndTextField.tsx
```diff
@@ -0,0 +1,37 @@
...
+): JSX.Element;
+export function BoundSelectAndTextField<O extends HasIdAndName<V>, V extends Value, X extends Only<TextFieldXss, X>>(
+  props: Omit<BoundSelectAndTextFieldProps<O, V, X>, "selectFieldProps"> & {
+    selectFieldProps: Optional<CompoundSelectFieldProps<O, V>, "getOptionLabel" | "getOptionValue">;
+  },
+): JSX.Element {
+  const { selectFieldProps, textFieldProps, compact = true } = props;
+  return (
+    <CompoundField>
+      <BoundSelectField {...selectFieldProps} sizeToContent compact={compact} />
```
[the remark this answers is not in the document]
bdow: Hmm.. yea, custom testids are a bit of a pain here. Though, the testid for these fields is based on the field's label, so you can still customize it.
I will play around to see how we might be able to better support custom testids here.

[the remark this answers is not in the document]
bdow: You _can_ do:
```
<BoundSelectAndTextField selectFieldProps={{ "data-testid": "customTestId" }} />
```
But typescript won't like that as "data-testid" isn't a BoundSelectFieldProp. But you could try instead:
```
<BoundSelectAndTextField 
  selectFieldProps={{ field: formState.type, options: types, ...{ "data-testid": "hey" } }}
/>
```
That extra spread makes Typescript happy, or really oblivious.

src/components/internal/CompoundField.tsx
bdow: Yea, hopefully this is cool to do. I assume it is like `cloneElement`, but not sure which is really better. I saw what was returned in the `children` prop, and saw I could just clone/expand the props and not the whole element.

src/components/internal/CompoundField.tsx
bdow: Thanks

src/forms/BoundSelectAndTextField.test.tsx
bdow: Good question. I don't think that test would be needed. I agree that it would be redundant based on the other test coverage.

src/forms/BoundSelectAndTextField.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+export function Example() {
+  const formState = useFormState({ config, init: { type: ScheduleTypes.Task } });
+  const types = [
+    { id: ScheduleTypes.Task, name: "Task" },
+    { id: ScheduleTypes.Milestone, name: "Milestone" },
+  ];
+
+  return (
+    <>
+      <BoundSelectAndTextField
```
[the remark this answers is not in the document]
bdow: Thanks!

src/forms/BoundSelectAndTextField.tsx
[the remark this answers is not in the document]
bdow: This has been updated to support custom test ids.
The two fields (SelectField and TextField) will internally set a `data-testid` based on the label of the `SelectField`, which, if not given will default the `field.key` property.
In addition, you can prefix the field `data-testid`s with a custom value passed to the component, example:
```
<BoundSelectAndTextField
  selectFieldProps={{ field: formState.type, options: types }}
  textFieldProps={{ field: formState.name, clearable: true, placeholder: "Add new" }}
  data-testid="customTestId"
/>
```
This would result in the SelectField having a `data-testid="customTestId_type"` and the TextField with `data-testid="customTestId_name"`.

src/components/internal/CompoundField.tsx
bdow: I switched over to using `cloneElement` after looking at the source code. We can accomplish the same thing, but appears using `cloneElement` would be safer. There are a couple of properties (`child._source` and `child._self`) that don't get copied over when just spreading. They are `undefined` in storybook, but I'm no sure how/when they're populated.

PR #372, opened 2021-11-04
feat: ChipTextField

bdow:
ChipTextField is expected to be used in SchedulesV2 when creating a new Subcategory. It is mostly a self contained form, which is why we handle an onEnter (for "Save") and listening to the "Escape" key to blur the field.

### DIFF
--- a/src/inputs/ChipTextField.tsx
+++ b/src/inputs/ChipTextField.tsx
@@ -1,5 +1,5 @@
-import { InputHTMLAttributes, KeyboardEvent, useRef, useState } from "react";
-import { useFocus, useTextField } from "react-aria";
+import { KeyboardEvent, useRef, useState } from "react";
+import { useFocus } from "react-aria";
 import { usePresentationContext } from "src/components/PresentationContext";
 import { Css } from "src/Css";
 import { maybeCall, useTestIds } from "src/utils";
@@ -11,8 +11,6 @@ interface ChipTextFieldProps {
   value?: string;
   // Whether the input is required. Adds aria-required attribute
   required?: boolean;
-  // Sets input's placeholder attribute
-  placeholder?: string;
   // Whether to call `onBlur` when the user presses the "Escape" key while focused on the input.
   // @default `true`
   blurOnEscape?: false;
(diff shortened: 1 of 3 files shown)


src/inputs/ChipTextField.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <input
+      {...(inputProps as InputHTMLAttributes<HTMLInputElement>)}
+      {...focusProps}
+      ref={inputRef}
+      css={{
+        ...Css[typeScale].dif.aic.br16.pl1.pxPx(10).pyPx(2).gray900.bgGray300.outline0.$,
+        ...(isFocused ? Css.bshFocus.$ : {}),
+      }}
+      {...tid}
+      size={props.value?.length ?? 1}
```
lucyconklin, line 71: **suggestion: (non-blooking):** This size prop looks like it's making some weird behavior. Is there a way to accomplish this with css instead?

Screen Shot 2021-11-04 at 11 58 08 AM
Screen Shot 2021-11-04 at 11 57 44 AM
But I tried some normal words, and it still looked a little off:
Screen Shot 2021-11-04 at 11 55 44 AM

lucyconklin (approved):

src/inputs/ChipTextField.tsx
bdow: Yea, this is a bummer but there really isn't a great way to do this when using an `<input />` field. It cannot be accomplished using CSS because the `value` is not actually in the DOM, but just an attribute on the input element.  Because it isn't in the DOM then we can't check against the width. 
Potential ways of fixing:
- Use a hidden element that we fill its contents with the save value as the input. Every time that changes, we'd measure the hidden element and use that width to set on the input.
- Use a `<span contenteditable />` rather than an `<input />`. Going this route we'd need to handle things a bit differently... like how we initially select the text when focused would need to use `createRange`. Though, this is a pretty simple input, maybe worth testing out if contenteditable can replace it?

bdow: @lucyconklin - I just tried this out with contenteditable. Let me know if you want to take a look.

PR #374, opened 2021-11-04
fix: Move typeScale to be top level PresentationProp

bdow:


src/components/PresentationContext.stories.tsx
```diff
@@ -49,7 +49,7 @@ export function PresentationFieldProps() {
 
       <div css={Css.pb2.mb2.bb.bGray300.$}>
         <h1 css={Css.xlEm.$}>Small Font size</h1>
-        <PresentationProvider fieldProps={{ typeScale: "xs", compact: true }}>
+        <PresentationProvider fieldProps={{ compact: true }} typeScale="xs">
```
stephenh, line 52: Hm, I guess looking at this concretely...it seems like setting this would be like setting a global type scale, which we don't really have? Or do we? Is all of our text `xs` in both form fields and just text-on-the-page/tables?

I just assumed they're different. And so something like `fieldProps={{ typeScale: xs }} tableProps={{ typeScale: sm }}` would be more necessary, i.e. for setting up the default sizes at the app-level.

...hm,  yeah, I think we've also talked about moving all of the `setGridDefault` things into `PresentationContext`, so that would probably get us the `tableProps` key? With both `typeScale` + various other table-specific things in it, like the default style...

I think I like that. And then like we could have `GridTable.typeScale` as a prop as well, but it defaults to coming from the `PresentationContext.tableProps.typeScale`?

bdow: Yea, its true.. I was thinking about adding in a wrapping element to PresentationProvider which would do:
```
<PresentationProvider>
  <div css={Css[typeScale].$}>{children}</div>
</PresentationProvider>
```
That would set a local typescale, but anything that defines the font size locally would override it, unless looking at the presentation prop.

stephenh: Hm, yeah that's interesting. Somewhat tangentially, I like the idea of a single top-level `PresentationProvider` when an app boots, but otherwise was kinda hoping this would be an internal impl detail and that most pages would only interact with it by happenstance by setting props on `FormLine` or `GridTable` or what not...

I kinda wonder if `PresentationProvider` itself could become public and instead an app would set its default settings as part of the existing `<BeamContext fieldProps=... tableProps... />`. Just to avoid yet-another-provider in the tree of top-level providers.

I dunno, maybe that's overly strict (making it private), but for some reason it just seems odd to change whole sections of the app by dropping in a  `PresentationProvider` ... where-as the more purpose-specific things like "oh yeah use `FormLine`s to set some things", "use `GridTable` to set some things" ... dunno, seems more kosher to me for some reason.

bdow: The good news, that is how BeamContext is set up right now. It accepts the PresentationProps and set it internally. 
https://github.com/homebound-team/beam/blob/main/src/components/BeamContext.tsx#L46

I see what you mean. Perhaps GridTable would also need to accept PresentationProps as we may have different tables with different requirements.

bdow:
Closing per comments where this doesn't really make sense at the moment. Will continue the discussion elsewhere.

PR #375, opened 2021-11-06
feat: ChipSelectField

bdow:
First pass a a ChipSelectField, utilizing our existing Listbox and Option components

### DIFF
--- a/src/components/internal/Popover.tsx
+++ b/src/components/internal/Popover.tsx
@@ -11,7 +11,7 @@ interface PopoverProps extends React.PropsWithChildren<any> {
 
 /** Popover is used for generating menus and list-boxes */
 export function Popover(props: PopoverProps) {
-  const { triggerRef, popoverRef, positionProps, children, onClose, isOpen } = props;
+  const { triggerRef, popoverRef, positionProps, children, onClose, isOpen, ...others } = props;
   const { overlayProps } = useOverlay(
     {
       onClose,
@@ -32,6 +32,7 @@ export function Popover(props: PopoverProps) {
         onClose();
         return false;
       },
+      ...others,
     },
     popoverRef,
   );
--- a/src/inputs/ChipSelectField.test.tsx
+++ b/src/inputs/ChipSelectField.test.tsx
@@ -43,6 +43,8 @@ describe("ChipSelectField", () => {
     click(r.chipSelectField_clearButton);
     // Then expect the default placeholder value to display
     expect(r.chipSelectField()).toHaveTextContent("Select an option");
+    // And the clear button is removed when no value is selected
+    expect(r.queryByTestId("chipSelectField_clearButton")).toBeFalsy();
     // And onSelect to be called
     expect(onSelect).toBeCalledWith([undefined, undefined]);
   });
--- a/src/inputs/ChipSelectField.test.tsx
+++ b/src/inputs/ChipSelectField.test.tsx
@@ -1,6 +1,6 @@
 import { fireEvent } from "@testing-library/react";
 import { useState } from "react";
-import { ChipSelectField, ChipSelectFieldProps } from "src/inputs/ChipSelectField";
+import { ChipSelectField, ChipSelectFieldProps } from "src";
 import { Value } from "src/inputs/Value";
 import { Optional } from "src/types";
 import { maybeCall } from "src/utils";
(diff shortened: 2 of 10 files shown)


chr1sjf0x (approved):
Looks great 👍 🍣

src/components/internal/Popover.tsx
```diff
@@ -11,7 +11,7 @@ interface PopoverProps extends React.PropsWithChildren<any> {
 
 /** Popover is used for generating menus and list-boxes */
 export function Popover(props: PopoverProps) {
```
KoltonG, line 13: **question:** Did we find out if we could get rid of React-Popper and use this? Does this handle changing positions when hitting the edges of the screen?

src/inputs/ChipSelectField.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const r = await render(
+      <TestComponent label="Label" value="s:2" options={sports} onFocus={onFocus} onBlur={onBlur} />,
+    );
+
+    // When focusing the field and opening the menu
+    click(r.chipSelectField);
+    // And onFocus should have been called
+    expect(onFocus).toBeCalledTimes(1);
+    // And when firing the blur event while the menu is opened,
+    fireEvent.blur(r.chipSelectField());
```
KoltonG, line 94: **question:** This is neat! How are you controlling the blur to not be propagated?

src/inputs/ChipSelectField.tsx
```diff
@@ -0,0 +1,194 @@
...
+    <>
+      <div
+        css={{
+          ...chipStyles,
+          ...Css.dif.p0.mwPx(32).if(!value).bgGray200.$,
+          ...(isFocused ? Css.bshFocus.$ : {}),
+        }}
+      >
+        <Label label={label} labelProps={labelProps} hidden {...tid.label} />
+        <button
```
KoltonG, line 141: Is this the rounded button?

src/inputs/ChipSelectField.tsx
```diff
@@ -0,0 +1,194 @@
...
+    shouldFlip: true,
+    scrollRef: listBoxRef,
+    isOpen: state.isOpen,
+    onClose: state.close,
+    placement: "bottom left",
+  });
+
+  positionProps.style = {
+    ...positionProps.style,
+    width: buttonRef?.current?.clientWidth,
```
KoltonG, line 126: Nice, this matches the button width!

src/inputs/ChipSelectField.tsx
```diff
@@ -0,0 +1,194 @@
...
+      {state.isOpen && (
+        <Popover
+          triggerRef={buttonRef}
+          popoverRef={popoverRef}
+          positionProps={positionProps}
+          onClose={state.close}
+          isOpen={state.isOpen}
+          shouldCloseOnBlur={true}
+        >
+          <ListBox
```
KoltonG, line 182: **question:** Is this the virtual one?

KoltonG (approved):
🍣

src/inputs/ChipSelectField.tsx
bdow: Yes, this is the virtualized one.

src/inputs/ChipSelectField.tsx
bdow: Yes. This button matches the designs for the Chips with just a different color to follow the SchedulesV2 design. This is a prototype component, so not forcing it to comply with existing styles (but it does for the most part).

src/inputs/ChipSelectField.test.tsx
bdow: We are able to conditionally call on blur based on whether the menu is open or not.

src/components/internal/Popover.tsx
bdow: Yes, this does handle changing positions. We could get rid of React-Popper, but we might want to generalize this component a bit more before then. Part of some tech debt.

src/inputs/ChipSelectField.test.tsx
KoltonG: Is that through React-Aria?

src/components/internal/Popover.tsx
KoltonG: Nice! Great find!

src/inputs/ChipSelectField.test.tsx
bdow: Not aria, but, React-Stately - same idea. We know when the menu is opened/closed.

PR #378, opened 2021-11-11
fix: All Chip components to only be one line of text

bdow:
If the text within a Chip exceeds the width allotted to it, then it will use an ellipsis. A title attribute is also tied to the Chip to display the full value of the chip on hover.


lucyconklin (approved):
💯  🍣  Nice

stephenh (approved):

bdow:
@lucyconklin - Thank you for reviewing the earlier PR. This has completely changed since you approved it according to new direction from the Design team. Rather than giving the option to have a chip truncate its text, we now do it by default. This behavior is done across the board for all Chip related components.

PR #379, opened 2021-11-16
feat: ChipSelectField supports creating new options

bdow:
New prop ChipSelectFieldProps.onCreateNew will append a "Create new" option. When "Create new" is chosen, then the SelectField is replaced with the ChipTextField for creating a new option.
New prop ChipSelectFieldProps.disabled supports showing Tooltip for displaying a disabled reason.

ChipSelectField now uses React-Stately's 'useListData' to manage a dynamic list of option.

Updates tests and stories

Other changes:
- React-Aria and React-Stately package version bumps.
- Move the "DismissButton" component to always be in Popover instead of each caller needing to add it.
- ChipTextField supports autoFocus


package.json
```diff
@@ -34,20 +34,20 @@
   },
   "dependencies": {
     "@homebound/form-state": "^2.1.5",
-    "@react-aria/utils": "^3.8.1",
+    "@react-aria/utils": "^3.9.0",
```
KoltonG, line 37: I always wondered why this wasn't part of the general package `react-aria/utils`

src/components/internal/Popover.tsx
```diff
@@ -41,6 +41,7 @@ export function Popover(props: PopoverProps) {
     <OverlayContainer>
       <div {...{ ...overlayProps, ...positionProps }} ref={popoverRef}>
         {children}
+        <DismissButton onDismiss={() => onClose()} />
```
KoltonG, line 44: **praise:** Nice!

src/inputs/internal/ListBox.tsx
```diff
@@ -84,7 +84,7 @@ export function ListBox<O, V extends Key>(props: ListBoxProps<O, V>) {
           // We don't really need to set this, but it's handy for tests, which would
           // otherwise render just 1 row. A better way to do this would be to jest.mock
           // out Virtuoso with an impl that just rendered everything, but doing this for now.
-          initialItemCount={3}
+          initialItemCount={5}
```
KoltonG, line 87: **question:** Why did this change?

src/inputs/ChipSelectField.stories.tsx
```diff
@@ -18,11 +19,12 @@ export function Example() {
     { id: "ts:1", code: TaskStatus.NotStarted, name: "Not Started" },
     { id: "ts:2", code: TaskStatus.InProgress, name: "In Progress" },
     { id: "ts:3", code: TaskStatus.Complete, name: "Complete" },
-    { id: "ts:5", code: TaskStatus.OnHold, name: "On Hold ".repeat(10) },
-    { id: "ts:6", code: TaskStatus.Delayed, name: "Delayed" },
+    { id: "ts:4", code: TaskStatus.OnHold, name: "On Hold ".repeat(10) },
+    { id: "ts:5", code: TaskStatus.Delayed, name: "Delayed" },
```
KoltonG, line 23: Ahh was the Virtual item number change for this?

src/inputs/ChipSelectField.test.tsx
```diff
@@ -105,6 +105,58 @@ describe("ChipSelectField", () => {
...
+    expect(r.chipSelectField_createNewField()).toBeTruthy();
+    expect(r.queryByTestId("chipSelectField")).not.toBeVisible();
+    // And when entering a new value
+    fireEvent.input(r.chipSelectField_createNewField(), { target: { textContent: newOpt.name } });
+    // And hitting the Enter key
+    fireEvent.keyDown(r.chipSelectField_createNewField(), { key: "Enter" });
+    // Wait for the async request to finish
+    await wait();
+    // Then expect the text field to be removed
+    expect(r.queryByTestId("chipSelectField_createNewField")).toBeFalsy();
```
KoltonG, line 142: **praise:** Neat tests!

src/inputs/ChipTextField.tsx
```diff
@@ -53,6 +54,10 @@ export function ChipTextField(props: ChipTextFieldProps) {
   const fieldRef = useRef<HTMLSpanElement>(null);
   const typeScale = fieldProps?.typeScale ?? "sm";
 
+  useEffect(() => {
+    autoFocus && fieldRef.current?.focus();
```
KoltonG, line 58: Nice! I thought React-Aria would give us a helper for this!

KoltonG (commented):
:sushi

src/inputs/internal/ListBox.tsx
bdow: Tests only render this number of items. The test I was working on needed access to more items in Virtuoso.

src/inputs/ChipSelectField.stories.tsx
bdow: Yes! You found it. Good 👁️

KoltonG:
Super nit but what if we create two items with the same name? Should it be like this?

CleanShot 2021-11-18 at 13 58 27@2x

KoltonG (approved):

bdow:
> Super nit but what if we create two items with the same name? Should it be like this?
> 
> CleanShot 2021-11-18 at 13 58 27@2x

Yea, great question. Right now we do not do any dedupe logic. As per these being prototype components, these sort of edge cases aren't being captured. I'm sure there will most likely be follow ups to address this sort of behavior, or even how to deal with typos? FWIW, the caller could potentially catch duplicates in the `onCreateNew` callback and throw an error or something.

PR #383, opened 2021-11-18
fix: Break mid word in Chips and Tags when truncated

bdow:
The result reveals more of the content of the Chip's/Tag's text, utilizing all of the space of the available to the Chip/Tag.

### DIFF
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -20,7 +20,7 @@ export function Chip<X extends Only<Xss<Margin>, X>>(props: ChipProps<X>) {
       {...tid}
       title={text}
     >
-      <span css={Css.lineClamp1.$}>{text}</span>
+      <span css={Css.lineClamp1.breakAll.$}>{text}</span>
     </span>
   );
 }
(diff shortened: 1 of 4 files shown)


src/components/Chip.tsx
```diff
@@ -20,7 +20,7 @@ export function Chip<X extends Only<Xss<Margin>, X>>(props: ChipProps<X>) {
       {...tid}
       title={text}
     >
-      <span css={Css.lineClamp1.$}>{text}</span>
+      <span css={Css.lineClamp1.breakAll.$}>{text}</span>
```
stephenh, line 23: Makes me think, we already snuck `textOverflow: ellipsis` into the `lineClamp` methods ... should we do that for `breakAll` as well? As a "surely you want this as well" sort of default?

stephenh (approved):

src/components/Chip.tsx
bdow: I was wondering that too. But my thought was maybe it makes sense when `lineClamp1`, but I dunno maybe not for clamping multiple lines.

stephenh: Ah yeah, I have no idea / haven't looked what about `lineClamp2`+ would look like with / without `breakAll` and whether it makes sense.

The way we're making the `lineClampN` rules, it would technically be possible to set `breakAll` only for `lineClamp1`, i.e. with a conditional include here

https://github.com/homebound-team/truss/pull/58/files#

Around `lineClamp.ts` line 13

PR #386, opened 2021-11-19
feat: Fields should blur on interactions

bdow:
SelectField, ChipSelectField, and DateField should blur whenever a user changes the value.
TextField, TextAreaField (with preventNewLines), NumberField, and DateField should blur when the user hits the Enter key

### DIFF
--- a/src/inputs/ChipSelectField.tsx
+++ b/src/inputs/ChipSelectField.tsx
@@ -142,10 +142,14 @@ export function ChipSelectField<O, V extends Value>(
       if (selectedItem) {
         onSelect(key as V, selectedItem);
       }
+      // Per UX, when a field is selected then we want to call our `onBlur` callback and remove the focus styles. The field _is_ still in focus, but that is only to retain tab position in the DOM.
+      setIsFocused(false);
+      maybeCall(onBlur);
     },
     onOpenChange: (isOpen) => {
       if (!isOpen && buttonRef.current) {
-        // When closing reset the focus to the button element.
+        // When closing reset the focus to the button element. This is to retain "tab position" in the document, allowing hte user to hit "Tab" and move to the next tabbable element.
+        // If the menu closed due to a user selecting an option, then the field will not visually appear focused.
         buttonRef.current.focus();
       }
     },
@@ -232,6 +236,7 @@ export function ChipSelectField<O, V extends Value>(
             }}
             onClick={() => {
               onSelect(undefined as any, undefined as any);
+              maybeCall(onBlur);
               setIsClearFocused(false);
             }}
             aria-label="Remove"
--- a/src/inputs/ChipSelectField.tsx
+++ b/src/inputs/ChipSelectField.tsx
@@ -59,7 +59,8 @@ export function ChipSelectField<O, V extends Value>(
   const isDisabled = !!disabled;
   const showClearButton = !disabled && clearable && !!value;
   const chipStyles = Css[typeScale].tl.bgGray300.gray900.br16.pxPx(10).pyPx(2).$;
-  const [isFocused, setIsFocused] = useState(false);
+  // Controls showing the focus border styles.
+  const [visualFocus, setVisualFocus] = useState(false);
   const [isClearFocused, setIsClearFocused] = useState(false);
   const { focusProps } = useFocus({
     onFocus: (e) => {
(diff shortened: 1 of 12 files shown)


src/inputs/ChipSelectField.tsx
```diff
@@ -142,10 +142,14 @@ export function ChipSelectField<O, V extends Value>(
       if (selectedItem) {
         onSelect(key as V, selectedItem);
       }
+      // Per UX, when a field is selected then we want to call our `onBlur` callback and remove the focus styles. The field _is_ still in focus, but that is only to retain tab position in the DOM.
+      setIsFocused(false);
```
stephenh, line 146: I almost want to add a new `useState` that is like:

```
const [pretendNotFocused, setPretendNotFocused] = useState()
```

And call `setPretendNotFocused` here, and use like a `const effectivelyFocused = isFocused && !pretendNotFocused`.

Just b/c it's going to make me really confused to 6 months from now wonder wtf `isFocused` is false when the browser still thinks its focused. :-)

That said, this is probably too safe; but maybe add a comment where we declare the `isFocused` `useState` to add a disclaimer that we maybe prematurely call this.

Also I wonder if this warrants a comment on the `onBlur` prop that this is more a logical blur than the physical blur.

src/inputs/TextField.tsx
```diff
@@ -34,7 +34,18 @@ export function TextField<X extends Only<TextFieldXss, X>>(props: TextFieldProps
...
   };
   const inputRef = useRef<HTMLInputElement | null>(null);
-  const { labelProps, inputProps } = useTextField(textFieldProps, inputRef);
+  const { labelProps, inputProps } = useTextField(
+    {
+      ...textFieldProps,
+      onKeyDown: (e) => {
+        if (e.key === "Enter") {
+          // Blur the field when the user hits the enter key - as if they are "committing" the value and done with the field
+          inputRef.current?.blur();
```
stephenh, line 43: And lazy ask/I think I already asked this, if they re-click the field, does the browser automatically re-focus us?

src/inputs/ChipSelectField.tsx
stephenh: ...the other components seem to just call `element.blur()`; maybe a comment about why this one is different?

src/inputs/TextField.tsx
bdow: It does not! Good catch.

src/inputs/ChipSelectField.tsx
bdow: Great questions around this, and great perspective. There is definitely a difference here,... we have the actual browser focus, and then our 'visual' focus state. I think separating those would be very helpful. I will work on this. Thanks!

bdow: The `isFocused` value we are using here was really to just represent a visual focus state, so I named the variables accordingly. Please take another look when you get a chance.

src/inputs/DateField.test.tsx
```diff
@@ -103,16 +103,21 @@ describe("DateField", () => {
...
+    const onBlur = jest.fn();
+    const r = await render(<DateField value={jan2} label="Date" onChange={onChange} onBlur={onBlur} />);
 
     // When triggering the Date Picker
     fireEvent.focus(r.date());
     // Then the Date Picker should be shown
     expect(r.date_datePicker()).toBeTruthy();
 
+    // Fire the blur event with the date picker as the related target. This should not fire `onBlur`, but will set the proper focus state internally.
+    expect(onBlur).toBeCalledTimes(0);
```
stephenh, line 118: Should the `toBeCalledTimes(0)` happen after the `fireEvent.blur`?

stephenh (approved):
Nice

src/inputs/DateField.test.tsx
bdow: Yes, it definitely should be. Good catch.

src/inputs/ChipSelectField.tsx
```diff
@@ -142,11 +149,19 @@ export function ChipSelectField<O, V extends Value>(
...
-      if (!isOpen && buttonRef.current) {
-        // When closing reset the focus to the button element.
-        buttonRef.current.focus();
+      if (!isOpen) {
+        // When closing, reset the focus to the button element. This is to retain "tab position" in the document, allowing hte user to hit "Tab" and move to the next tabbable element.
+        // If the menu closed due to a user selecting an option, then the field will not visually appear focused.
+        buttonRef.current?.focus();
+      } else {
+        // If opened, set visual focus to true. It is possible to be in a state where the browser focus is on this element, but we are not "visually" focused (see `onSelectionChange`). If the user opens the menu again, we should trigger the visual focus.
+        setVisualFocus(true);
```
stephenh, line 164: Just curious, but why does `SelectField` not need this same "separate `visualFocus`" tracking? I would have assumed the behavior between the two + the impl of "focus briefly moves to a menu" would be similar.

bdow: SelectField retains the actual focus in the DOM because we still allow the user to type in the input to filter the options. Under the hood it actually uses `useComboBox` which that different focus behavior vs `useSelect`.

PR #387, opened 2021-11-29
[POC] - Draggable & Add In Place POC

KoltonG:
This is a POC WIP of potentially how "Drag and Drop" and the "Add in Place" could be introduced into Beam GridTables. Thanks to @chr1sjf0x for the great pairing on this ideation! 

The general changes are as follow:

1. Create a new renderer that removes `display: contents` (this allows the use of the HTML attribute `draggable`)
2. Add spacer row to the last element in a card so that we can add the dragging and add in place functionality to the spacer row only.
3. Pass contextual information to the NestedCard class to help store the parent, above and below ID's for each row. This is what is used to pass positional information to the callback.

If you are interested in taking this for a spin, the Storybook SchedulesV2 example shows in the console relevant outputs when dragging any element over a spacer row.

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -258,6 +258,58 @@ export function NestedCardsThreeLevels() {
   );
 }
 
+export function NestedCardsThreeLevelsDraggable() {
+  const nameColumn: GridColumn<NestedRow> = {
+    header: () => "Name",
+    parent: (row) => ({
+      content: <div css={Css.base.$}>{row.name}</div>,
+      value: row.name,
+    }),
+    child: (row) => ({
+      content: <div css={Css.sm.$}>{row.name}</div>,
+      value: row.name,
+    }),
+    grandChild: (row) => ({
+      content: <div css={Css.xs.$}>{row.name}</div>,
+      value: row.name,
+    }),
+    add: () => "Add",
+  };
+  const actionColumn: GridColumn<NestedRow> = {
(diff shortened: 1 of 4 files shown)


src/components/Table/GridTable.stories.tsx
```diff
@@ -258,6 +258,58 @@ export function NestedCardsThreeLevels() {
...
+  };
+
+  return (
+    <div css={Css.df.fdc.vh100.$}>
+      <GridTable
+        as="virtualFixed"
+        columns={[nameColumn, nameColumn, actionColumn]}
+        {...{ rows }}
+        style={nestedStyle}
+        sorting={{ on: "client", initial: [0, "ASC"] }}
```
chr1sjf0x, line 307: question: Are you still thinking about how the API will look exactly for specifying which types can be dragged and where they can be dropped?

KoltonG: Yeah! Adding that to the design document!

src/components/Table/GridTable.tsx
```diff
@@ -558,6 +575,52 @@ function renderVirtual<R extends Kinded>(
   );
 }
 
+/**
+ * renderVirtual without the `display: contents`
+ */
+function renderVirtualFixed<R extends Kinded>(
```
KoltonG, line 581: @blambillotte we made it work by taking off the `display: contents` which now is removed completely for virtual tables. So no need to create a new `as` type and you can ignore the majority of this code.

PR #388, opened 2021-12-01
fix: Remove required response from ChipSelectFieldProps.onCreateNew

bdow:


src/inputs/ChipSelectField.tsx
```diff
@@ -202,8 +202,7 @@ export function ChipSelectField<O, V extends Value>(
           value={inputValue}
           onChange={setInputValue}
           onEnter={async () => {
-            const newOption = await onCreateNew(inputValue);
-            listData.insertBefore(createNewOpt.id, newOption);
```
stephenh, line 206: The assumption is that the parent page will create a new option and re-render us with that option listed in the regular options prop?

stephenh (approved):

src/inputs/ChipSelectField.tsx
bdow: Yup, thats correct. In the case of Schedules V2, creating a new option is mixed in with the `saveTask` mutation. They are using a single mutation to both create a new Subcategory and set Task's Subcategory.

PR #389, opened 2021-12-02
POC - Unsorted Table

KoltonG:
This is a very minimal exploration on adding to the table headers the ability on the 3rd click to unsort/return the sort back to the original state.

In the .gif below, it doesn't seem to be going back to the original state may be due to the `currentKey` not being reset... There is more to explore here for sure, just wanted to get something out to start a conversation 😄 

CleanShot 2021-12-02 at 10 59 56

*As you can see, when unsorting the value column the rows don't seem to go back to their original state.*


bdow:
An "unsorted" is now supported. Closing this PR.

PR #390, opened 2021-12-02
fix: Trigger blur after onSelect is called

bdow:


stephenh (approved):
This is probably hard to test? Seems nuanced so would be great to have regression coverage but fine if it's a PITA.

bdow:
> This is probably hard to test? Seems nuanced so would be great to have regression coverage but fine if it's a PITA.

Yea, it needs to be in conjunction with formState's auto save. Which we can do, but will probably do it in a follow up in order to get this fix out there... blew up bug bashing.

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -120,10 +120,13 @@ export function SelectFieldBase<O, V extends Value>(props: SelectFieldBaseProps<
...
         selectedOptions: firstSelectedOption ? [firstSelectedOption] : [],
       });
-      // When a single select menu item changes, then blur the field
-      inputRef.current?.blur();
     }
     selectionChanged && onSelect(([...keys.values()] as Key[]).map(keyToValue) as V[]);
+
+    if (!multiselect) {
+      // When a single select menu item changes, then blur the field AFTER `onSelect` has been called
+      inputRef.current?.blur();
```
KoltonG, line 128: Nice!

PR #391, opened 2021-12-02
chore: Add test to ensure SelectField's onBlur triggered via selection change is only called after onSelect

bdow:


src/forms/BoundSelectField.test.tsx
```diff
@@ -45,6 +45,23 @@ describe("BoundSelectField", () => {
...
+      { favoriteSport: "s:1" },
+      { onBlur: () => onBlur(author.favoriteSport.value) },
+    );
+    const r = await render(<BoundSelectField field={author.favoriteSport} options={sports} />);
+    // When changing the value
+    r.favoriteSport().focus();
+    click(r.getByRole("option", { name: "Soccer" }));
+
+    // Then formState has the latest value when onBlur is called
+    expect(onBlur).toBeCalledWith("s:2");
```
stephenh, line 63: Ah nice! Thanks!

Really nice how you were able to black-box test this, i.e. we don't need to worry too much about poking around internal implementation details/order of events/etc., and instead just observe "yes the `.value` was correct at time of blur".

stephenh (approved):

chr1sjf0x (approved):

PR #392, opened 2021-12-02
fix: Account for firstLastColumnWidth when calculating total GridTable width

blambillotte:


src/components/Table/GridTable.tsx
```diff
@@ -636,9 +636,12 @@ export function calcVirtualGridColumns(columns: GridColumn<any>[], firstLastColu
     { claimedPercentages: 0, claimedPixels: 0, totalFr: 0 },
```
stephenh, line 636: It might be too cute, but you could do `claimedPixels: firstLastColumnWidth ? firstLastColumnWidth * 2 : 0` to sneak it in here and avoid the separate `totalClaimedPixels` adjustment later

stephenh (approved):
Nice, makes sense!

chr1sjf0x (approved):
TY @blambillotte

bdow (approved):

PR #393, opened 2021-12-02
fix: Fix addCanClose not being removed on unmount.

stephenh:


src/components/Modal/useModal.tsx
```diff
@@ -13,6 +13,15 @@ export interface UseModalHook {
 export function useModal(): UseModalHook {
   const { modalState, modalCanCloseChecks } = useBeamContext();
   const lastCanClose = useRef<CheckFn | undefined>();
+  useEffect(() => {
+    // Capture the lastCanClose from when the `useEffect` runs, so that when our cleanup
+    // lambda is called later, we clean up the as-when-scheduled value, and not the super-latest
+    // value, because then we need up nuking the actually valid current value.
+    const { current } = lastCanClose;
+    return () => {
+      modalCanCloseChecks.current = [...modalCanCloseChecks.current.filter((c) => c !== current)];
```
bdow, line 22: **question:** Any reason to spread this? Could it just be:
```
modalCanCloseChecks.current = modalCanCloseChecks.current.filter((c) => c !== current);
```

bdow (approved):

src/components/Modal/useModal.tsx
stephenh: Good catch! I'd copy/pasted it from the code that was adding a new element after the `filter`. Removed.

PR #394, opened 2021-12-03
fix: ChipSelectField and ListBox updates

bdow:
ListBox now supports two sections, (1) Options, and (2) Persistent Actions. The Options section will scroll independently of the Persistent Actions, make the Actions always visible.
Sections will only be rendered if necessary, otherwise it will just be a flat list of options as it works today. If the onCreateNew prop is passed to ChipSelectField, then we'll create the two sections. Right now there are a lot of assumptions in the code that we'll only ever have at most two sections, and that the second section is the Persistent Actions, and there will only ever be a single persistent action. Prototype, period. :)

Bugs fixed:
- Auto scroll to focused item ONLY if using virutal focus (i.e. in SelectField where we use a Combobox and the browser focus stays within the input field). This fixes an issue where the ListBox was scrolling when trying to click on an option for the ChipSelectField. This was due to the focus moving as the user clicks and our auto-scroll repositioning the options. We now handle scrolling Virtuoso in the Option component itself if not using a virutal focus.
- Use 'offset' prop in 'useOverlayPosition' to properly position the ListBox. Our previous way of doing this would create issues when the menu opened above the trigger.
- Call 'onBlur' after user removes ChipTextField, regardless of whether they created a new option or not. If a new option was created, this should trigger formState's auto save behavior. Adds tests to ensure this is called with expected values.

Updates:
- Max height of Listbox has been set to 512px, per design request.

### DIFF
--- a/src/inputs/ChipSelectField.tsx
+++ b/src/inputs/ChipSelectField.tsx
@@ -1,6 +1,7 @@
+import { camelCase } from "change-case";
 import React, { Key, ReactNode, useCallback, useEffect, useRef, useState } from "react";
 import { mergeProps, useButton, useFocus, useOverlayPosition, useSelect } from "react-aria";
-import { Item, useListData, useSelectState } from "react-stately";
+import { Item, Section, useListData, useSelectState } from "react-stately";
 import { Icon, Tooltip } from "src/components";
 import { Popover } from "src/components/internal";
 import { Label } from "src/components/Label";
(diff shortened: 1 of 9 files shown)


src/inputs/internal/ListBoxSection.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface ListBoxSectionProps<O> {
+  section: Node<O>;
+  state: SelectState<O>;
+  contrast: boolean;
+  onListHeightChange: (height: number) => void;
+  popoverHeight: number;
+  scrollOnFocus?: boolean;
+}
+export function ListBoxSection<O>(props: ListBoxSectionProps<O>) {
```
chr1sjf0x, line 18: Thought: Probably worth a javadoc explaining how this is used?

chr1sjf0x (approved):
TY @bdow

chr1sjf0x:
@stephenh @DeanGilewicz @KoltonG @lucyconklin @ksk5280 - We'd like to fast track this if you would like to weigh in, please do.

src/inputs/ChipSelectField.tsx
```diff
@@ -190,7 +203,8 @@ export function ChipSelectField<O, V extends Value>(
   const removeCreateNewField = useCallback(() => {
     setShowInput(false);
     setInputValue("Add new");
-    buttonRef.current?.focus();
+    // Trigger onBlur to initiate any auto-saving behavior.
+    maybeCall(onBlur);
```
stephenh, line 207: Nice comment

src/inputs/internal/ListBoxSection.tsx
```diff
@@ -0,0 +1,51 @@
...
+      {isPersistentSection && <li {...separatorProps} css={Css.bt.bGray200.$} />}
+      <li {...itemProps} css={Css.if(!isPersistentSection).overflowAuto.$}>
+        {/* Styles assume only one Persistent Item is ever shown. Will need to adjust if that ever changes */}
+        <ul
+          css={
+            Css.listReset.if(!isPersistentSection).hPx(popoverHeight - sectionSeparatorHeight - persistentItemHeight).$
+          }
+          {...groupProps}
+        >
+          {isPersistentSection ? (
```
stephenh, line 36: Kinda wonder if separate `OptionSection` and `PersistentActionSection` functions would be simpler than the conditionals in this combined function? Maybe not though.

src/inputs/internal/Option.tsx
```diff
@@ -32,9 +33,34 @@ export function Option<O>(props: OptionProps<O>) {
     ref,
   );
 
+  const onKeyDown = useCallback(
+    (e: KeyboardEvent) => {
+      if (!scrollToIndex || !(e.key === "ArrowDown" || e.key === "ArrowUp")) {
```
stephenh, line 38: Naively, usually we get keyboard stuff for free with react-aria/stately; can you throw in an in-source comment about why we need our own here? I'm sure it makes sense/is good, just would be curious as a future maintainer.

stephenh (approved):

src/inputs/internal/ListBoxSection.tsx
bdow: Yea, I like the thought. I wanted to make a single generic ListBoxSection, that can be expanded upon in the future. But maybe we'd be better off being a bit more specific in the first go-round. I'm going to ponder that bit more and might create a follow up PR.

src/inputs/internal/Option.tsx
bdow: Yup, makes sense. 
Fwiw, this is just so we can properly scroll the React Virtuoso list... which React-Spectrum also has their own VirtualizedListBox component that I started glancing at. Appears there is a bit more set up to it than Virtuoso, but could be cool to try out their built in one and see if we get more of this scrolling behavior stuff for free.

PR #396, opened 2021-12-07
perf: no JSX in visitRows => 100x improvement

chr1sjf0x:
TLDR - GridTable is performing expensive JSX computations in 2 places which it shouldn't be (at-least/especially for virtualized lists):

1. vistRows
2. sortRows - NOTE: Not addressing this, since we aren't using sorting at the moment

Here is a Profile from main for a virtualized list with 1500 rows (2.1 seconds to visitRows):
image

Here is that same virtualized list on this branch (24 milliseconds to visitRows):
image

~~NOTE: Still working on getting the styling to be correct with this approach.~~ All good now 👍  🍣

### DIFF
--- a/src/components/Table/nestedCards.tsx
+++ b/src/components/Table/nestedCards.tsx
@@ -82,6 +82,7 @@ export class NestedCards {
  * row separate from the card's actual content.
  */
 export function makeOpenOrCloseCard(openCards: NestedCardStyle[], kind: "open" | "close"): Chrome {
+  const scopeCards = [...openCards];
   return () => {
     let div: any = <div />;
     const place = kind === "open" ? "Top" : "Bottom";
@@ -92,7 +93,7 @@ export function makeOpenOrCloseCard(openCards: NestedCardStyle[], kind: "open" |
     // | card1 | card2 / ... card3 ... \ card2 | card1 |
     // | card1 | card2 | ... card3 ... | card2 | card1 |
     //
-    [...openCards].reverse().forEach((card, i) => {
+    [...scopeCards].reverse().forEach((card, i) => {
       const first = i === 0;
       div = (
         <div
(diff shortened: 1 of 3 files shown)


src/components/Table/nestedCards.tsx
```diff
@@ -167,23 +177,33 @@ export function makeSpacer(height: number, openCards: NestedCardStyle[]) {
 export function maybeCreateChromeRow(
   columns: GridColumn<any>[],
   filteredRows: RowTuple<any>[],
-  chromeBuffer: JSX.Element[],
+  chromeBuffer: ChromeBuffer,
 ): void {
   if (chromeBuffer.length > 0) {
     filteredRows.push([
       undefined,
       // We add 2 to account for our dedicated open/close columns
```
stephenh, line 185: Nit, move this comment to where the `Css.gc` moved to

src/components/Table/nestedCards.tsx
```diff
@@ -167,23 +177,33 @@ export function makeSpacer(height: number, openCards: NestedCardStyle[]) {
...
-      <div css={Css.gc(`span ${columns.length + 2}`).$} data-chrome="true">
-        {chromeBuffer.map((c, i) => (
-          <Fragment key={i}>{c}</Fragment>
-        ))}
-      </div>,
+      <ChromeRow chromeBuffer={[...chromeBuffer]} columns={columns.length} />,
```
stephenh, line 186: Wow, I'm struggling to understand how this changed the perf, but it is nonetheless an amazing find...

So, what this does AFAICT is make `filteredRows` a homogeneous list of function components.

Previously it was `GridRow FC`, `JSX snippet`, `GridRow FC`, `JSX snippet`, like alternating (right?).

But, now it's all FC, like `GrowRow FC`, `Chrome FC`, `GridRow FC`, `Chrome FC`...

...and we're not even `React.memo`-ing the `ChromeRow`...

But, okay, I get here, we were still like doing a bunch of `Css.gc...` calls, and `React.createElement(div)`, and emotion `css=`, for ~600 chrome rows when we really only needed to render 10-20 of them.

Got it.

Really great find @chr1sjf0x !

stephenh (approved):

src/components/Table/nestedCards.tsx
chr1sjf0x: Thanks! You got the difference exactly. Defer the work until render time, so it's done only for the needed rows.

PR #397, opened 2021-12-07
fix: Performance improvements for ChipSelectField

bdow:
Only run the 'listData.update' within useEffect after first render
Memoize 'selectChildren' to only re-evaluate if the options/items change.
Separate "CreateNewField" to isolate its state management.
Separate ListBoxChip component.
getOptionValue/Label default values are stored as constants. These were changing identity unnecessarily.

Profiling against Schedules V2 on Blueprint:
### Initial render
Screen Shot 2021-12-07 at 3 52 16 PM

### Re-render
Screen Shot 2021-12-07 at 3 55 57 PM

## Perf Test
Created a Perf Test story for ChipSelectField that will be ignored by Chromatic snapshots. It is a single ChipSelectField with 100 options. See the below screenshots for change in performances recorded in the  Performance Storybook Addon. (Set at 1 copy and 100 samples)

#### Before
Screen Shot 2021-12-07 at 8 56 33 PM

#### After
Screen Shot 2021-12-07 at 8 55 12 PM

### DIFF
--- a/src/inputs/ChipSelectField.stories.tsx
+++ b/src/inputs/ChipSelectField.stories.tsx
@@ -1,11 +1,12 @@
 import { action } from "@storybook/addon-actions";
 import { Meta } from "@storybook/react";
-import { useState } from "react";
+import { useCallback, useState } from "react";
 import { ChipSelectField } from "src";
 import { PresentationProvider } from "src/components";
-import { TaskStatus } from "src/components/Filters/testDomain";
+import { InternalUser, TaskStatus } from "src/components/Filters/testDomain";
 import { Css } from "src/Css";
 import { noop } from "src/utils";
+import { zeroTo } from "src/utils/sb";
 
 export default {
   component: ChipSelectField,
@@ -142,3 +143,34 @@ export function Example() {
     </div>
   );
 }
+
+export function PerfTest() {
+  const [internalUsers, setInternalUsers] = useState<InternalUser[]>(
+    zeroTo(100).map((i) => ({ id: `iu:${i}`, name: `Internal User ${i}` })),
+  );
+  const [createValue, setCreateValue] = useState<string>(internalUsers[0].id);
+  const onFocus = useCallback(action("onFocus"), []);
+  const onBlur = useCallback(action("onBlur"), []);
+  return (
+    <ChipSelectField
+      clearable
+      label="Test"
+      placeholder="+ User"
+      options={internalUsers}
+      value={createValue}
+      onSelect={(v) => setCreateValue(v)}
+      onFocus={onFocus}
+      onBlur={onBlur}
+      onCreateNew={async (name) => {
+        const newOpt = { id: `ts:${internalUsers.length + 1}`, name };
+        // Requires user to update their local state.
+        setInternalUsers((prevState) => [...prevState, newOpt]);
+        setCreateValue(newOpt.id);
+      }}
+    />
+  );
+}
+
+PerfTest.parameters = {
+  chromatic: { disableSnapshot: true },
+};
--- a/src/inputs/ChipSelectField.tsx
+++ b/src/inputs/ChipSelectField.tsx
@@ -1,5 +1,5 @@
 import { camelCase } from "change-case";
-import React, { Key, ReactNode, useCallback, useEffect, useRef, useState } from "react";
+import React, { Key, ReactNode, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { mergeProps, useButton, useFocus, useOverlayPosition, useSelect } from "react-aria";
 import { Item, Section, useListData, useSelectState } from "react-stately";
 import { Icon, Tooltip } from "src/components";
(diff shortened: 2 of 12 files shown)


src/inputs/ChipSelectField.stories.tsx
```diff
@@ -142,3 +143,34 @@ export function Example() {
     </div>
   );
 }
+
+export function PerfTest() {
```
chr1sjf0x, line 147: praise: This is cool! I was wondering about options for failing the build on perf regressions to GridTable. Does this do that?

src/inputs/ChipSelectField.tsx
```diff
@@ -319,3 +323,29 @@ type ListBoxSection<O> = { title: string; options: O[]; isPersistent?: boolean }
...
+      value={value}
+      onChange={setValue}
+      onEnter={() => onEnter(value)}
+      onBlur={onBlur}
+      {...tid}
+    />
+  );
+}
+
+let firstRender = true;
```
chr1sjf0x, line 351: question: Is this down at the global scope? How does that work when there are more than one chip select rendered?

src/inputs/ChipSelectField.tsx
```diff
@@ -102,43 +104,49 @@ export function ChipSelectField<O, V extends Value>(
     getKey: (item) => (isListBoxSection(item) ? item.title : getOptionValue(item)),
   });
```
chr1sjf0x, line 106: question: Not sure I understand the difference/benefits of doing this approach with `!firstRender` vs:
```
const memoOptions = useMemo(() => { title: "Options", options }), [options]);
useEffect(() => listData.update("Options", memoOptions), [memoOptions]);
```

src/utils/options.ts
```diff
@@ -0,0 +1,2 @@
+export const defaultOptionValue = <O>(opt: O) => (opt as any).id;
+export const defaultOptionLabel = <O>(opt: O) => (opt as any).name;
```
chr1sjf0x, line 2: praise: Nice!

chr1sjf0x (commented):
Looks great @bdow TY!

I didn't want to `approve` due to my question about the `firstRender` variable, but otherwise this seems great!

src/inputs/ChipSelectField.tsx
bdow: Great question and thank you for flagging! I had assumed this would be in the module scope so that it'd be isolated per component. But you are right, this seems to bleed further out. I will update to be a ref internal to the component.

src/inputs/ChipSelectField.stories.tsx
bdow: I didn't see a way to store the results of the performance testing in the application so that we can fail on regressions. We can dig in further and see what we might get there.

src/inputs/ChipSelectField.tsx
chr1sjf0x: thanks for clarifying 👍

chr1sjf0x (approved):
👍

src/inputs/ChipSelectField.tsx
```diff
@@ -59,7 +62,7 @@ export function ChipSelectField<O, V extends Value>(
   const typeScale = fieldProps?.typeScale ?? "sm";
   const isDisabled = !!disabled;
   const showClearButton = !disabled && clearable && !!value;
-  const chipStyles = Css[typeScale].tl.bgGray300.gray900.br16.pxPx(10).pyPx(2).$;
+  const chipStyles = useMemo(() => Css[typeScale].tl.bgGray300.gray900.br16.pxPx(10).pyPx(2).$, [typeScale]);
```
stephenh, line 65: If it's too late that's fine, but if it's easy to do, I'd be interested in the before/after profile of just this one change. Doing `Css...$` everytime imo has a chance at actually being cheaper than `useMemo` adding the slight overhead of another hook to track.

stephenh (commented):
Nice! @bdow kinda wonder, could we skip initializing/building `listData` all together until the chip is actually interacted with? AFAIU we shouldn't need it until the listbox is popped open?

src/inputs/ChipSelectField.tsx
bdow: Not much of a difference, but here's the result. Granted, the perf tests locally for this jump between 6ms and 7ms.

# Without Memo
Without Memo

# With Memo
With Memo

bdow:
> Nice! @bdow kinda wonder, could we skip initializing/building `listData` all together until the chip is actually interacted with? AFAIU we shouldn't need it until the listbox is popped open?

@stephenh - Maybe,... it'd be interesting to see how we can configure it. Though, right now the way it is set up is to use the list data options for construction the `items` subsequent `children` props for the `useSelect` and `useSelectState` hooks. There are a lot of ways to skin this cat, and I'm sure we haven't found the most efficient yet, so I'd be happy to keep exploring.

stephenh:
Yeah, naively I wonder if pushing the `listData` logic into a separate component that we only render when the list needs to be open. Like you've already go `ListBox` or what not, so not re-create that, but like all of the `listData`/etc. hooks just copy/paste them into something that we conditionally include. Dunno, really just making stuff up. Totally defer to you / schedules v2 about whether it's worth spending more time on.

PR #398, opened 2021-12-08
feat: Add optional footer padding for virtual tables; add optional scrolling for tables

khalidwilliams:
This PR adds an additional GridTable prop to provide virtual tables with a footer. 

This uses the [Footer component](https://virtuoso.dev/footer/) from virtuoso.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -200,6 +200,8 @@ export interface GridTableProps<R extends Kinded, S, X> {
    */
   persistCollapse?: string;
   xss?: X;
+  /** Adds padding to a footer at the bottom of a GridTable that's rendered `as=virtual` */
+  virtualFooterPadding?: number;
 }
 
 /**
@@ -239,6 +241,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     setRowCount,
     observeRows,
     persistCollapse,
+    virtualFooterPadding,
   } = props;
 
   const [collapsedIds, toggleCollapsedId] = useToggleIds(rows, persistCollapse);
@@ -399,6 +402,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
         style.nestedCards?.firstLastColumnWidth,
         xss,
         virtuosoRef,
+        virtualFooterPadding,
       )}
     </PresentationProvider>
   );
@@ -423,6 +427,7 @@ function renderCssGrid<R extends Kinded>(
   firstLastColumnWidth: number | undefined,
   xss: any,
   virtuosoRef: MutableRefObject<VirtuosoHandle | null>,
+  _virtuosoFooterPadding?: number,
 ): ReactElement {
   return (
     <div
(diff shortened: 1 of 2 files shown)


chr1sjf0x (approved):
ty @khalidwilliams 

question: I get the impression there aren't good/easy ways to test this or have stories. What do you think?

src/components/Table/GridTable.tsx
```diff
@@ -200,6 +200,8 @@ export interface GridTableProps<R extends Kinded, S, X> {
    */
   persistCollapse?: string;
   xss?: X;
+  /** Adds padding to a footer at the bottom of a GridTable that's rendered `as=virtual` */
+  virtualFooterPadding?: number;
```
stephenh, line 204: Hrm...yeah, this seems a little "bag on the side-ish" in terms of ergonomics of the overall props API...

Could we work this into `GridStyle` somewhere?

Would it be too magical to do:

```
const style: GridStyle = {
  rootCss: { pb: 1 },
};
```

And then in:

```
const VirtualRoot = memoizeOne
```

We could like destructure the `pb` out of `style.rootCss` and pass it to `Footer` that way?

Or is this footer bottom padding different that what `rootCss.pb` would be?

stephenh: Was chatting with @bdow a little bit, and being more concrete, my thought was that this style of prop name could end up being like:

```
${virtual || nonvirtual} x ${header || footer || something else } x ${padding || margin || something else} = ...
```

Disclaimer I'm really guessing about "just use `pb` on `rootCss`", so defer a lot on the context of whether that makes sense...

khalidwilliams:
@chr1sjf0x The best way to demonstrate that it works is to color the background of the footer, or have another element in the story that goes below the table:
Screen Shot 2021-12-08 at 10 16 27 AM

But the weird thing is that in ScheduelsV2 story, adding `as=virtual` creates this weird padding between the footer and the table itself. I can try making a smaller table story for testing purposes (the padding goes away if I collapse one of the milestones), but I don't understand how to get that padding to go away otherwise. I also wasn't sure if it was worth the time to keep poking at it 🤷‍♂️ 

@bdow this is what I was referring to in stand up.

src/components/Table/GridTable.tsx
khalidwilliams: I don't think the bottom padding would be different than what the root padding would be, we're just addressing it for a virtualized table in BP right now.

khalidwilliams: Would destructuring `pb` our of rootCss look more like 
```
const { paddingBottom } = rootCss
```
?

chr1sjf0x:
@khalidwilliams - Thanks for answering. Just wanted to make sure it as something you were thinking about, but not a blocker for me given the small scope of the ticket, and other more important things to do.

src/components/Table/GridTable.tsx
bdow: @khalidwilliams - yes, that's how'd you pull it out of the GridStyle.rootCss. And I do believe this should work, then you can apply that paddingBottom to the footer div. FWIW, since rootCss is already of type `Properties`, then I'd apply the `paddingBottom` directly... something like: 
```
const { paddingBottom } = style.rootCss || {};
...
  Footer: () => <div css={{ paddingBottom }} />,
```

stephenh (approved):

PR #399, opened 2021-12-08
fix: Fix Virtuoso 'Zero-sized element' error

bdow:


chr1sjf0x (approved):
Amazing!! TY 🎉 

I wonder if this fixed the scrolling jank 🤔

bdow:
> Amazing!! TY 🎉
> 
> I wonder if this fixed the scrolling jank 🤔

It might!

stephenh (approved):
Nice!

FWIW I wonder if we should just straight-up remove the `display: contents` specifically for the virutalized rendering. I'd like to keep it for the non-virtualized for a little bit.

bdow:
> Nice!
> 
> FWIW I wonder if we should just straight-up remove the `display: contents` specifically for the virutalized rendering. I'd like to keep it for the non-virtualized for a little bit.

Oh yea, thats a good question. I'll see what happens...

PR #407, opened 2021-12-14
fix: Update for the collapse icon to correctly reflect the state of the row

chr1sjf0x:
I accidentally broke the collapse chevron behaviour with my perf fix the other day.

Before:
https://share.getcloudapp.com/DOumRXBZ

After:
https://share.getcloudapp.com/E0uoG2lq

### DIFF
--- a/src/components/Table/CollapseToggle.tsx
+++ b/src/components/Table/CollapseToggle.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useContext } from "react";
+import { useCallback, useContext, useState } from "react";
 import { GridCollapseContext, GridDataRow, IconButton } from "src/components";
 
 export interface GridTableCollapseToggleProps {
@@ -8,9 +8,15 @@ export interface GridTableCollapseToggleProps {
 export function CollapseToggle(props: GridTableCollapseToggleProps) {
   const { row } = props;
   const { isCollapsed, toggleCollapsed } = useContext(GridCollapseContext);
-  const toggleOnClick = useCallback(() => toggleCollapsed(row.id), [row.id, toggleCollapsed]);
-  const iconKey = isCollapsed(row.id) ? "chevronRight" : "chevronDown";
-  const headerIconKey = isCollapsed(row.id) ? "chevronsRight" : "chevronsDown";
+  const [, setTick] = useState(0);
+  const currentlyCollapsed = isCollapsed(row.id);
+  const toggleOnClick = useCallback(() => {
+    toggleCollapsed(row.id);
+    setTick(Date.now());
+  }, [row.id, currentlyCollapsed, toggleCollapsed]);
+  console.log({ id: row.id, currentlyCollapsed });
+  const iconKey = currentlyCollapsed ? "chevronRight" : "chevronDown";
+  const headerIconKey = currentlyCollapsed ? "chevronsRight" : "chevronsDown";
   const isHeader = row.kind === "header";
   if (!isHeader && (!props.row.children || props.row.children.length === 0)) {
     return null;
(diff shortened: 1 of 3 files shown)


src/components/Table/CollapseToggle.tsx
```diff
@@ -8,9 +8,15 @@ export interface GridTableCollapseToggleProps {
...
-  const toggleOnClick = useCallback(() => toggleCollapsed(row.id), [row.id, toggleCollapsed]);
-  const iconKey = isCollapsed(row.id) ? "chevronRight" : "chevronDown";
-  const headerIconKey = isCollapsed(row.id) ? "chevronsRight" : "chevronsDown";
+  const [, setTick] = useState(0);
+  const currentlyCollapsed = isCollapsed(row.id);
+  const toggleOnClick = useCallback(() => {
+    toggleCollapsed(row.id);
+    setTick(Date.now());
+  }, [row.id, currentlyCollapsed, toggleCollapsed]);
+  console.log({ id: row.id, currentlyCollapsed });
```
bdow, line 17: This snuck in.

chr1sjf0x: TY! Removed and build error fixed

src/components/Table/CollapseToggle.tsx
```diff
@@ -8,9 +8,14 @@ export interface GridTableCollapseToggleProps {
 export function CollapseToggle(props: GridTableCollapseToggleProps) {
   const { row } = props;
   const { isCollapsed, toggleCollapsed } = useContext(GridCollapseContext);
-  const toggleOnClick = useCallback(() => toggleCollapsed(row.id), [row.id, toggleCollapsed]);
-  const iconKey = isCollapsed(row.id) ? "chevronRight" : "chevronDown";
-  const headerIconKey = isCollapsed(row.id) ? "chevronsRight" : "chevronsDown";
+  const [, setTick] = useState(0);
```
stephenh, line 11: I'm a little worried that a row could be doing other custom `isCollapsed` logic and will not get re-rendered when just this local `setTick` is called.

I liked the `collapseAllContext` vs. `collapseRowContext` change, but is there a tldr of why we couldn't keep using `isCollapsed=true/false` as a boolean prop on `GridRow` and let that drive re-rendering of the whole row?

chr1sjf0x: We could probably go back to the old way of doing things. I think the main win here is to have the GridRow as a child of the CollapseContext, such that changes to the toggleCollapsed or isCollapsed do not trigger a rerender of the whole row, but just the toggle caret.

This change has the effect of:

Toggling all rows will rerender all toggle carets
Adding/removing rows will not trigger rerenders (except for the header row)

If we went back to a `isCollapsed: boolean`, then we would end up with a new context.value every time, which would always rerender all the carets. But maybe we shouldn't care, there aren't so many of them 🤷

stephenh: Personally I'm fine with row-level re-renders on collapse state change, as I think a thesis of the virtualized row approach is that ~20-30 rows at a time should still render very quickly. And, per original comment, I think eventually it'd be a feature that non-`ToggleCollapse` cells could observe the current state.

But I'm happy to defer to you on this PR; especially if the "did / did not render" behavior you want is covered by unit tests so that we can futz with things weeks/months down the road and not have regressions.

chr1sjf0x: Yea, given the perf I've seen (and setting lofty goals), I'm trying to have really fine grained rerenders 🤷 

I've [spun up another PR](https://github.com/homebound-team/beam/pull/410/files) to work on testing the rerender behaviour. Could we run this one through as-is?

bdow (approved):

PR #408, opened 2021-12-15
feat: [SC-10740] Disable days for DateField

blambillotte:
Forwards the `disabledDays` prop from `react-day-picker`


src/inputs/internal/DatePickerOverlay.tsx
```diff
@@ -40,6 +41,10 @@ export function DatePickerOverlay(props: DatePickerOverlayProps) {
         "& .DayPicker-Day:active": Css.bgGray400.$,
         // Make the month title, i.e. "May 2021", match figma; pyPx nudge matches the NavbarElement nudging
         "& .DayPicker-Caption > div": Css.base.pyPx(2).$,
+        // For days that are disabled via `disabledDays`, 
+        "& .DayPicker-Day--disabled": Css.cursorNotAllowed.$,
+        // Override `.DayPicker-Day:active` background when the day is disabled
+        "& .DayPicker-Day--disabled:active": Css.bgWhite.$,
```
stephenh, line 47: Eventually we need to upgrade storybook and use the new `play` feature to take screenshots of open widgets. If that is something you'd like to pick up, that'd be great :-D but fine to do as later infra

stephenh (approved):

src/inputs/DateField.tsx
```diff
@@ -32,6 +32,8 @@ export interface DateFieldProps
   placeholder?: string;
   format?: keyof typeof dateFormats;
   iconLeft?: boolean;
+  // Set custom logic for individual dates or date ranges to be disabled in the picker, exposed from `react-day-picker`
+  disabledDays?: Modifier;
```
ksk5280, line 36: **thought:** What do you think about adding a link to the docs for what a Modifier can be?

src/inputs/internal/DatePickerOverlay.tsx
KoltonG: I had no idea it was in GA release. See https://storybook.js.org/docs/react/writing-stories/play-function for more details.

Open to play around with this as well :)

src/inputs/internal/DatePickerOverlay.tsx
```diff
@@ -48,11 +53,13 @@ export function DatePickerOverlay(props: DatePickerOverlayProps) {
         weekdayElement={Weekday}
         selectedDays={[value]}
         initialMonth={value ?? new Date()}
-        onDayClick={(day) => {
+        onDayClick={(day, modifiers) => {
+          if (modifiers.disabled) return;
```
KoltonG, line 57: Interesting, is this per day?

KoltonG (approved):
🍣

src/inputs/internal/DatePickerOverlay.tsx
blambillotte: Very cool! I've had a lot of success in the past actually using Storybook as a test harness to mock different component scenarios then you point a tool like Cypress at storybook and run tests against it. The *dev UX* is great and you get a lot closer to testing the actual *user UX* ... think the `play` feature could open up even more interesting options for easily testing permutations! I put a note to see if there's any interest in this type of testing approach for an upcoming frontend guild.

As for this ticket, I'll punt on the upgrade for now since I need to get back to the schedules V2 ticket that prompted this one.

src/inputs/internal/DatePickerOverlay.tsx
blambillotte: yeah exactly - the callback gets passed the `day` that is clicked, any `modifies` on the actual day (in this case we've set the `disabled` modifier and lastly the `event` object, which we're not using
https://react-day-picker.js.org/api/DayPicker/#onDayClick

PR #409, opened 2021-12-15
fix: `firstNonHeaderRowCss` for Virtual GridTable

KoltonG:
This PR fixed the `firstNonHeaderRowCss` style for GridTables using the `as="virtual"` rendering engine. 

**Important Changes**:

- When using a `virtual` GridTable with `nestedCards` styles, we noticed that there was an extra Chrome row included at the start of the table list. This was due to the `VirtualRoot` not compensating for Chrome rows in this scenario. Resolved the Chrome rows to be pushed to both the `headerRows` and `filteredRows` arrays.

**Before**
CleanShot 2021-12-14 at 15 01 56@2x

**After**
CleanShot 2021-12-14 at 19 27 09@2x

- Conditionally add the `firstNonHeaderRowCss` for nestedCard styles situation to not add these styles to the header.
- Refactored nestedCard.tsx to more clearly expose the row buffer being updated.
- Fixed console.error from React regarding missing `key` props

# Performance

This was tested using the Storybook Performance addon using _1 copy and 10 samples_.

## Before

<table>
<tr>
	<th>Non-Virtual</th>
	<th>Virtual</th>
<tr>
	<td>
CleanShot 2022-01-19 at 13 28 48@2x
</td>
	<td> 
CleanShot 2022-01-19 at 13 30 15@2x
</td>

</table>

## After

<table>
<tr>
	<th>Non-Virtual</th>
	<th>Virtual</th>
<tr>
	<td>
CleanShot 2022-01-19 at 13 19 46@2x
</td>
	<td>
CleanShot 2022-01-19 at 13 20 15@2x
</td>
</table>

## Conclusion

There is a performance boost on the virtual tables since we are no longer rendering un-needed chrome rows but the total number of chrome rows are the same! Maybe the cleanup of the NestedCard component helped with this too.

There is a performance hit on the non-virtual table could be due to the extra few lines of computation that was added to the renderer?

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -301,7 +301,6 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
           value={row.kind === "header" ? collapseAllContext : collapseRowContext}
         >
           <RowComponent
-            key={`${row.kind}-${row.id}`}
             {...{
               as,
               columns,
@@ -368,7 +367,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
 
     // If nestedCards is set, we assume the top-level kind is a card, and so should add spacers between them
     visitRows(maybeSorted, !!nestedCards);
-    nestedCards && nestedCards.done(filteredRows);
+    nestedCards && nestedCards.done(maybeSorted[maybeSorted.length - 1], filteredRows);
 
     return [headerRows, filteredRows];
   }, [
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -683,9 +683,9 @@ const VirtualRoot = memoizeOne<
   return React.forwardRef(function VirtualRoot({ style, children }, ref) {
     // This VirtualRoot list represent the header when no styles are given. The
     // table list generally has styles to scroll the page for windowing.
-    const isList = Object.keys(style || {}).length !== 0;
+    const isHeader = Object.keys(style || {}).length === 0;
 
-    // This re-renders each time we have new children in the view port
+    // This re-renders each time we have new children in the viewport
     return (
       <div
         ref={ref}
@@ -694,11 +694,11 @@ const VirtualRoot = memoizeOne<
           // Add an extra `> div` due to Item + itemContent both having divs
           ...Css.addIn("& > div + div > div > *", gs.betweenRowsCss || {}).$,
           // Table list styles only
-          ...(isList
-            ? {
+          ...(isHeader
+            ? {}
+            : {
                 ...Css.addIn("& > div:first-of-type > *", gs.firstNonHeaderRowCss).$,
-              }
-            : {}),
+              }),
           ...gs.rootCss,
           ...xss,
         }}
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -460,19 +460,41 @@ function renderCssGrid<R extends Kinded>(
   // We must determine if the header is using nested card styles to account for
   // the opening and closing Chrome rows.
   const isNestedCardStyleHeader = !!style.nestedCards?.kinds["header"];
-  // Determine at what index is the first non header row.
-  // The +1 at the end is to compensate for CSS nth counter starting at 1 (vs 0 for JS)
-  const firstNonHeaderRowIndex = (!isNestedCardStyleHeader ? 0 : 3) + 1;
+  /*
+    Determine at what CSS selector index is the first non header row. Note that
+    CSS selectors are not zero-based, unlike JavaScript, so we must add 1 to
+    what we'd normally expect.
+
+    Ex: When we don't have nestedCard styled headers, then we don't have opening
+    and closing Chrome rows. Therefore, the first non-header row is the second
+    row and since CSS selectors are not zero-based, we are aiming for a value
+    of 2 (1 + 1).
+
+    Ex: When we have nestedCard styled headers, then we have opening and closing
+    Chrome rows. Therefore, the first non-header row is the fourth row because:
+    - Row 1 is the opening Chrome Row
+    - Row 2 is the header
+    - Row 3 is the closing Chrome Row
+    - Row 4 is the first non-header row
+    And since CSS selectors are not zero-based, we are aiming for a value of 4
+    (3 + 1).
+  */
+  const firstNonHeaderRowIndex = (!isNestedCardStyleHeader ? 1 : 3) + 1;
 
   return (
     <div
       css={{
         ...Css.dg.gtc(calcDivGridColumns(columns, firstLastColumnWidth)).$,
-        // Apply the between-row styling with `div + div > *` so that we don't have to have conditional
-        // `if !lastRow add border` CSS applied via JS that would mean the row can't be React.memo'd.
-        // The `div + div` is also the "owl operator", i.e. don't apply to the 1st row.
-        ...(style.betweenRowsCss ? Css.addIn("& > div + div > *", style.betweenRowsCss).$ : {}),
-        // removes border between header and second row
+        /*
+          Using n + (firstNonHeaderRowIndex + 1) here to target all rows that
+          are after the first non-header row. Since n starts at 0, we can use
+          the + operator as an offset.
+
+          Inspired by: https://stackoverflow.com/a/25005740/2551333
+        */
+        ...(style.betweenRowsCss
+          ? Css.addIn(`& > div:nth-of-type(n + ${firstNonHeaderRowIndex + 1}) > *`, style.betweenRowsCss).$
+          : {}),
         ...(style.firstNonHeaderRowCss
           ? Css.addIn(`& > div:nth-of-type(${firstNonHeaderRowIndex}) > *`, style.firstNonHeaderRowCss).$
           : {}),
@@ -566,10 +588,12 @@ function renderVirtual<R extends Kinded>(
   xss: any,
   virtuosoRef: MutableRefObject<VirtuosoHandle | null>,
 ): ReactElement {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
   const { footerStyle, listStyle } = useMemo(() => {
     const { paddingBottom, ...otherRootStyles } = style.rootCss ?? {};
     return { footerStyle: { paddingBottom }, listStyle: { ...style, rootCss: otherRootStyles } };
   }, [style]);
+
   return (
     <Virtuoso
       overscan={5}
@@ -649,8 +673,8 @@ const VirtualRoot = memoizeOne<
     firstLastColumnWidth: number | undefined,
     xss: any,
   ) => Components["List"]
->((gs, columns, id, firstLastColumnWidth, xss) => {
-  return React.forwardRef(function VirtualRooot({ style, children }, ref) => {
+>((gs, _columns, id, _firstLastColumnWidth, xss) => {
+  return React.forwardRef(function VirtualRoot({ style, children }, ref) {
     // This VirtualRoot list represent the header when no styles are given. The
     // table list generally has styles to scroll the page for windowing.
     const isList = Object.keys(style || {}).length !== 0;
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -354,10 +354,15 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
       const length = rows.length;
       rows.forEach((row, i) => {
         if (row.kind === "header") {
-          nestedCards && nestedCards.maybeOpenCard(row, headerRows);
+          // Flag to determine if the header has nested card styles.
+          // This will determine if we should include opening and closing
+          // Chrome rows.
+          const isHeaderNested = !!style.nestedCards?.kinds["header"];
+
+          isHeaderNested && nestedCards.maybeOpenCard(row, headerRows);
           headerRows.push([row, makeRowComponent(row)]);
-          nestedCards && nestedCards.closeCard();
-          nestedCards && nestedCards.done(headerRows);
+          isHeaderNested && nestedCards.closeCard();
+          isHeaderNested && nestedCards.done(row, headerRows);
           return;
         }
         visit(row);
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -359,10 +359,10 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
           // Chrome rows.
           const isHeaderNested = !!style.nestedCards?.kinds["header"];
 
-          isHeaderNested && nestedCards.maybeOpenCard(row, headerRows);
+          isHeaderNested && nestedCards && nestedCards.maybeOpenCard(row, headerRows);
           headerRows.push([row, makeRowComponent(row)]);
-          isHeaderNested && nestedCards.closeCard();
-          isHeaderNested && nestedCards.done(row, headerRows);
+          isHeaderNested && nestedCards && nestedCards.closeCard();
+          isHeaderNested && nestedCards && nestedCards.done(row, headerRows);
           return;
         }
         visit(row);
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -693,8 +693,6 @@ const VirtualRoot = memoizeOne<
         css={{
           // Add an extra `> div` due to Item + itemContent both having divs
           ...Css.addIn("& > div + div > div > *", gs.betweenRowsCss || {}).$,
-          // Add `display:contents` to Item to flatten it like we do GridRow
-          ...Css.addIn("& > div", Css.display("contents").$).$,
           // Table list styles only
           ...(isList
             ? {
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -49,7 +49,16 @@ export interface GridStyle {
   firstNonHeaderRowCss?: Properties;
   /** Applied to all cell divs (via a selector off the base div). */
   cellCss?: Properties;
-  /** Applied to the header (really first) row div. */
+  /**
+   * Applied to the header (really first) row div.
+   *
+   * NOTE: Adding margin bottom to a virtual table header will not result in the
+   * expected of adding space between the header and list. This will only add
+   * margin bottom to the cell itself and now the header row.
+   *
+   * To achieve space between the header and list, add margin top to the
+   * `firstNonHeaderRowCss` styles.
+   * */
   headerCellCss?: Properties;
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -49,7 +49,16 @@ export interface GridStyle {
...
+   * Applied to the header (really first) row div.
+   *
+   * NOTE: Adding margin bottom to a virtual table header will not result in the
+   * expected of adding space between the header and list. This will only add
+   * margin bottom to the cell itself and now the header row.
+   *
+   * To achieve space between the header and list, add margin top to the
+   * `firstNonHeaderRowCss` styles.
+   * */
   headerCellCss?: Properties;
```
KoltonG, line 62: Added this comment to help with common Virtual GridTable style issues where the expectation is that this will add space between the header and list. This will not since a virtual table has two separate adjacent `divs` for the header list and the list(non-header) list.

src/components/Table/GridTable.tsx
```diff
@@ -284,9 +293,11 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
       const RowComponent = observeRows ? ObservedGridRow : MemoizedGridRow;
 
       return (
-        <GridCollapseContext.Provider value={row.kind === "header" ? collapseAllContext : collapseRowContext}>
+        <GridCollapseContext.Provider
+          key={`${row.kind}-${row.id}`}
```
KoltonG, line 297: @chr1sjf0x just adding this here to resolve the React console error.

src/components/Table/GridTable.tsx
```diff
@@ -308,7 +319,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     const filteredRows: RowTuple<R>[] = [];
 
     // Misc state to track our nested card-ification, i.e. interleaved actual rows + chrome rows
-    const nestedCards = !!style.nestedCards && new NestedCards(columns, filteredRows, style);
+    const nestedCards = !!style.nestedCards && new NestedCards(columns, style);
```
KoltonG, line 322: Removed the ability to instantiate the NestedCards class with a buffer to add the Chrome rows to. Instead we are making it explicit on which buffer to add each Chrome row.

src/components/Table/GridTable.tsx
```diff
@@ -350,7 +368,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
 
     // If nestedCards is set, we assume the top-level kind is a card, and so should add spacers between them
     visitRows(maybeSorted, !!nestedCards);
-    nestedCards && nestedCards.done();
+    nestedCards && nestedCards.done(maybeSorted[maybeSorted.length - 1], filteredRows);
```
KoltonG, line 371: For Chrome row `key` support, passing a Row was needed to create the proper unique `key` props. For this case, we expect that the last parent is being closed.

src/components/Table/GridTable.tsx
```diff
@@ -558,36 +585,53 @@ function renderVirtual<R extends Kinded>(
...
+        return filteredRows[index][1];
       }}
       totalCount={(headerRows.length || 0) + (firstRowMessage ? 1 : 0) + (filteredRows.length || 0)}
     />
   );
 }
 
 /**
- * Customizes the `List` element that react-virtuoso renders, to have our css grid logic.
+ * A table might render two of these components to represent two virtual lists.
```
KoltonG, line 623: This was unclear when first trying to solve this problem. Adding this comment for future reference.

src/components/Table/GridTable.tsx
```diff
@@ -599,17 +643,28 @@ const VirtualRoot = memoizeOne<
   ) => Components["List"]
 >((gs, columns, id, firstLastColumnWidth, xss) => {
   return React.forwardRef(({ style, children }, ref) => {
+    // This VirtualRoot list represent the header when no styles are given. The
+    // table list generally has styles to scroll the page for windowing.
+    const isList = Object.keys(style || {}).length !== 0;
```
KoltonG, line 648: Sadly this is the only way to determine if the List being rendered is the "sticky" list or the non header list.

src/components/Table/SchedulesV2.stories.tsx
```diff
@@ -181,14 +181,14 @@ const buttonColumns = actionColumn<Row>({
 const spacing = { brPx: 8, pxPx: 16 };
 const style: GridStyle = {
   headerCellCss: Css.sm.gray700.py1.df.aic.$,
-  firstNonHeaderRowCss: Css.mt2.$,
+  firstNonHeaderRowCss: Css.mt1.$,
```
KoltonG, line 184: This is now properly showing for both Virtual and non Virtual tables.

src/components/Table/nestedCards.tsx
```diff
@@ -28,25 +28,41 @@ type Chrome = () => JSX.Element;
...
 
 export class NestedCards {
+  private columns: GridColumn<any>[];
   // A stack of the current cards we're showing
   private readonly openCards: Array<string> = [];
   // A buffer of the open/close/spacer rows we need between each content row.
   private readonly chromeBuffer: ChromeBuffer = [];
   private readonly styles: NestedCardsStyle;
 
-  constructor(private columns: GridColumn<any>[], private filteredRows: RowTuple<any>[], private style: GridStyle) {
```
KoltonG, line 37: Remove the `private` keyword in the constructor as it was not clear that this was a Class variable...

src/components/Table/GridTable.tsx
```diff
@@ -49,7 +49,16 @@ export interface GridStyle {
   firstNonHeaderRowCss?: Properties;
   /** Applied to all cell divs (via a selector off the base div). */
   cellCss?: Properties;
-  /** Applied to the header (really first) row div. */
+  /**
+   * Applied to the header (really first) row div.
+   *
+   * NOTE: Adding margin bottom to a virtual table header will not result in the
+   * expected of adding space between the header and list. This will only add
+   * margin bottom to the cell itself and now the header row.
```
chr1sjf0x, line 57: nit:
```suggestion
   * margin bottom to the cell itself and not the header row.
```

src/components/Table/GridTable.tsx
chr1sjf0x: TY 🚀

src/components/Table/GridTable.tsx
```diff
@@ -558,36 +585,53 @@ function renderVirtual<R extends Kinded>(
...
             return (
               <div css={Css.add("gridColumn", `${columns.length} span`).$}>
                 <div css={{ ...style.firstRowMessageCss }}>{firstRowMessage}</div>
               </div>
             );
           }
-          i -= 1;
+
+          // Shift index -1 when there is a firstRowMessage to not skip the
+          // first `filteredRow`
```
chr1sjf0x, line 610: praise: TY For all the comments in this PR

chr1sjf0x (approved):
Looks good to me @KoltonG, though I'm half asleep 😫  😆

chr1sjf0x:
@KoltonG - A couple of the stories in here have larger header rows which looks a little off/unexpected 🤔 
https://www.chromatic.com/test?appId=60466f7d43c37c0021788867&id=61b93f2cbfb4a1003af78926

KoltonG:
> @KoltonG - A couple of the stories in here have larger header rows which looks a little off/unexpected 🤔 [chromatic.com/test?appId=60466f7d43c37c0021788867&id=61b93f2cbfb4a1003af78926](https://www.chromatic.com/test?appId=60466f7d43c37c0021788867&id=61b93f2cbfb4a1003af78926)

@chr1sjf0x agree! I will walk through it with @bdow to see if I am missing some edge cases.

src/components/Table/GridTable.tsx
stephenh: Cool, makes sense

src/components/Table/GridTable.tsx
stephenh: Per other comment, but if we just use `chromeRow-${i++}` for keys, we wouldn't need to pass the previous row here. Not a big deal.

src/components/Table/nestedCards.tsx
```diff
@@ -28,25 +28,41 @@ type Chrome = () => JSX.Element;
...
     const card = this.styles.kinds[row.kind];
     // If this kind doesn't have a card defined, don't put it on the card stack
     if (card) {
       this.openCards.push(row.kind);
       this.chromeBuffer.push(makeOpenOrCloseCard(this.openCards, this.styles.kinds, "open"));
     }
     // But always close previous cards if needed
-    maybeCreateChromeRow(this.columns, this.filteredRows, this.chromeBuffer);
+    maybeCreateChromeRow({
+      key: `chrome-open-${row.kind}-${row.id}`,
```
stephenh, line 61: I was thinking about whether these keys should be called "chrome-open" or "chrome-closed" or what not...because a single chrome row can include both open chrome + spacer chrome + closed chromes, but the key would still be called `chrome-_open_...`...

Kinda thinking, instead of passing `key: ` to `maybeCreateChromeRow` and having that id be based on the row kind/row id we're on, what if the key was just `chrome-${chromeRowIndex++}` where `chromeRowIndex` is just a private variable that starts at 0 and we increment up.

Granted, pretty minor change...

(...I guess `maybeCreateChromeRow` is a top-level function, so we'd have to pass in the `chromeRowIndex`; if it was a method on the `NestedCards` class, it could access the `chromeRowIndex` field internally, which seems nice.)

src/components/Table/GridTable.tsx
stephenh: ...an alternative would be to make two `NestedCards`, i.e.

```
const rowCards  = !!style.nestedCards && new NestedCards(columns, filteredRows, style);
const headerCards = !!style.nestedCards && !!style.nestedCards["header"] && new NestedCards(columns, headerRows, style);
```

Then in the `if row.kind === header` case, use `headerCards` instead of `nestedCards`.

stephenh (approved):

src/components/Table/GridTable.tsx
```diff
@@ -435,6 +453,13 @@ function renderCssGrid<R extends Kinded>(
   xss: any,
   virtuosoRef: MutableRefObject<VirtuosoHandle | null>,
 ): ReactElement {
+  // We must determine if the header is using nested card styles to account for
+  // the opening and closing Chrome rows.
+  const isNestedCardStyleHeader = !!style.nestedCards?.kinds["header"];
+  // Determine at what index is the first non header row.
+  // The +1 at the end is to compensate for CSS nth counter starting at 1 (vs 0 for JS)
+  const firstNonHeaderRowIndex = (!isNestedCardStyleHeader ? 0 : 3) + 1;
```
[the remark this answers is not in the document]
KoltonG: Yeah, good observation. That was the issue causing style differences for non-virtual lists.

src/components/Table/nestedCards.tsx
```diff
@@ -176,25 +205,38 @@ export function makeSpacer(height: number, openCards: string[], styles: NestedCa
...
-): void {
+export function maybeCreateChromeRow({
+  key,
+  columns,
+  buffer,
+  chromeBuffer,
+}: {
+  key: Key;
+  columns: GridColumn<any>[];
+  buffer: RowTuple<any>[];
```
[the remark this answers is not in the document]
KoltonG: Make sense, will refactor.

src/components/Table/nestedCards.tsx
```diff
@@ -28,25 +28,41 @@ type Chrome = () => JSX.Element;
...
     if (card) {
       this.openCards.push(row.kind);
       this.chromeBuffer.push(makeOpenOrCloseCard(this.openCards, this.styles.kinds, "open"));
     }
     // But always close previous cards if needed
-    maybeCreateChromeRow(this.columns, this.filteredRows, this.chromeBuffer);
+    maybeCreateChromeRow({
+      key: `chrome-open-${row.kind}-${row.id}`,
+      columns: this.columns,
+      buffer,
```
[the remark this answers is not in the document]
KoltonG: You lost me on this comment! I think with the coming refactors this should be more clear since we will have two instances of the NestedCard class.

src/components/Table/nestedCards.tsx
KoltonG: Got it, good thought. Refactoring!

src/components/Table/GridTable.tsx
KoltonG: You got it!

src/components/Table/GridTable.tsx
KoltonG: Yeah! I did the cleanup :)

src/components/Table/nestedCards.tsx
KoltonG: I've refactored this a bit to use the same approach for the `this.styles` since the constructor was only a single line. Let me know what you think!

src/components/Table/nestedCards.tsx
```diff
@@ -176,25 +205,38 @@ export function makeSpacer(height: number, openCards: string[], styles: NestedCa
...
-  columns: GridColumn<any>[],
-  filteredRows: RowTuple<any>[],
-  chromeBuffer: ChromeBuffer,
-): void {
+export function maybeCreateChromeRow({
+  key,
+  columns,
+  buffer,
+  chromeBuffer,
+}: {
```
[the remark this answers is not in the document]
KoltonG: This is cleaned up now with moving this function inside the class.

chr1sjf0x (approved):
Looks good @KoltonG 

I'm curious, have you done a before/after perf test of this in schedule V2 to ensure there aren't any perf regressions?

Might be interesting to profile `initial load` and `add task` perf before/after.

src/components/Table/GridTable.tsx
```diff
@@ -327,37 +337,48 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
-      isCard && nestedCards && nestedCards.closeCard();
+      isCard && rowCards && rowCards.closeCard();
     }
 
     function visitRows(rows: GridDataRow<R>[], addSpacer: boolean): void {
       const length = rows.length;
       rows.forEach((row, i) => {
         if (row.kind === "header") {
-          nestedCards && nestedCards.maybeOpenCard(row);
+          // Flag to determine if the header has nested card styles.
```
stephenh, line 359: Just curious, does schedules v2 header have a nested card style? 

...I kinda wonder if @bdow 's maybe forthcoming/WIP refactoring of "for leaf-rows, use a real border" refactoring would clean this up, b/c we could assume that headers only ever have one row, hence are defacto leaf rows.

That musing aside, we should keep this approach for now/this PR

src/components/Table/GridTable.tsx
```diff
@@ -437,21 +458,52 @@ function renderCssGrid<R extends Kinded>(
...
-        ...(style.firstNonHeaderRowCss ? Css.addIn("& > div:nth-of-type(2) > *", style.firstNonHeaderRowCss).$ : {}),
+        /*
+          Using n + (firstNonHeaderRowIndex + 1) here to target all rows that
+          are after the first non-header row. Since n starts at 0, we can use
+          the + operator as an offset.
+
+          Inspired by: https://stackoverflow.com/a/25005740/2551333
+        */
+        ...(style.betweenRowsCss
+          ? Css.addIn(`& > div:nth-of-type(n + ${firstNonHeaderRowIndex + 1}) > *`, style.betweenRowsCss).$
```
stephenh, line 502: > here to target all rows that are after the first non-header row

Seems like this is going to apply `betweenRowsCss` to both regular rows as well as chrome rows?

Granted, it was doing that before...but I don't know that that was on purpose...

If I'm following, it seems like you've fixed not apply `betweenRowCss` and `firstNonheaderRowCss` to chrome rows within the header, but we've not yet fixed it (or even decided what it should really do?) for non-header regular vs. chrome rows?

src/components/Table/GridTable.tsx
```diff
@@ -608,16 +679,29 @@ const VirtualRoot = memoizeOne<
...
     return (
       <div
         ref={ref}
+        /** Styles from React-Virtuoso. You should probably not be touching this... */
         style={style}
         css={{
           // Add an extra `> div` due to Item + itemContent both having divs
           ...Css.addIn("& > div + div > div > *", gs.betweenRowsCss || {}).$,
+          // Add `display:contents` to Item to flatten it like we do GridRow
+          ...Css.addIn("& > div", Css.display("contents").$).$,
```
stephenh, line 698: Hasn't `display:contents` gone away for virtualized tables @bdow ? Fwiw this might have _just_ landed, but in theory we wouldn't want this?

src/components/Table/GridTable.tsx
stephenh: I kinda wonder if, when using nested cards, `betweenRowCss` should just target the chrome/space rows? I think we've got a `data-` attribute we could use.

src/components/Table/GridTable.tsx
bdow: Yes, this is accurate. Chatting with Kolton now about it.

src/components/Table/GridTable.tsx
KoltonG: SchedulesV2 header doesn't need to be nested cards but due to the left and right padding columns, we need it to spread the background colours, alignment and other things that we didn't fully figure out.

It would for sure benefit from @bdow refactor so don't need to call these open cards and deal with CSS selectors with interesting counts to compensate for open and close rows.

src/components/Table/GridTable.tsx
KoltonG: Yeah, this is a can of worms that I did not intend to touch but in that case, where we only target Chrome, spacers, this will get pretty complicated and weird for the user trying to understand what it does. 

This seems to be more related to a non-nested approach. Maybe even renaming it as @bdow suggested since the only usage we can see is adding a border-top style.

src/components/Table/GridTable.tsx
KoltonG: Great! This should be removed now

src/components/Table/GridTable.tsx
stephenh: Huh, that is an interesting point... you are right it's only used for border-tops, so maybe we could "dumb down" / remove `betweenRowCss` insofar as turning into a "betweenRowBorder" that would only a) have border color + size and b) be for non-card tables?

I.e. for card tables I think we assume the cards are serving the UX function of row delineation, so why bother?

Fwiw if we wanted to simplify all of this by just throwing a runtime error when both cards and `betweenRowCss` / `betweenRowBorder` are used, and sneak that into this PR, that'd be great; also fine as a todo/cleanup.

stephenh (approved):

src/components/Table/GridTable.tsx
KoltonG: Good idea!

PR #411, opened 2021-12-17
fix: Sometimes require GridCellContent.content to be a function.

stephenh:
When sorting a very large table, we use the `applyRowFn` to access the row's numerical/sort value (for all rows in the table), but we don't want that to also call a JSX snippet to be built (b/c we'll only need the JSX for the subset of the rows that will be rendered virutally).

This adds runtime enforcement that client-side sorted, virtualized tables need to use `content: () => ...` to build their JSX.

(Great find from @chr1sjf0x around another instance where we're unwittingly creating `ReactElement`s when we don't need to.)

(...technically just creating a `ReactElement` i.e. calling `React.createElement(type: div, props: { width: 1, height: 2 })` shouldn't be super expensive, b/c that in and of itself doesn't cause the in-memory `ReactElement` that it returns to be rendered in the DOM; except that in our case one of the props is `css=...emotion class...` so it does end up adding up.)


src/components/Table/GridTable.tsx
```diff
@@ -1069,8 +1093,8 @@ type GridCollapseContextProps = {
 
 export const GridCollapseContext = React.createContext<GridCollapseContextProps>({
   headerCollapsed: false,
-  isCollapsed: (id: string) => false,
-  toggleCollapsed: (id: string) => {},
+  isCollapsed: () => false,
+  toggleCollapsed: () => {},
```
chr1sjf0x, line 1097: question: I'm a little confused as to how you managed to remove the `id:string` args from here without any other code updates or breaking the build 🤔

chr1sjf0x (approved):
Looks good to me, aside from my confusion about the collapse context change.

src/components/Table/GridTable.tsx
stephenh: AH yeah, this is not the definition of `isCollapsed`, just the default impl, and functions are allowed to implement their spec w/o explicitly accepting all of the params if they don't use the params anyway.

src/components/Table/GridTable.stories.tsx
```diff
@@ -219,24 +190,26 @@ export function NestedCardsThreeLevels() {
...
-      Rendering {rows.length * 500} rows virtualized
-      <NestedCards rows={[header, ...zeroTo(100).flatMap(() => rows)]} as={"virtual"} />
+      Rendering {deepCount(rows)} rows virtualized
+      <NestedCards rows={rows} as="virtual" />
     </div>
   );
 }
 
 export function NestedCardsThreeLevelsVirtualizedAtScaleSorted() {
+  const rows = useMemo(() => [header, ...makeNestedRows(500)], []);
```
stephenh, line 208: @chr1sjf0x fyi I refactored these "at scale" stories while lazy-izing their content, and fwiw they're pretty snappy, like now they have ~5000 some rows and sorting is still ~instant. Pretty neat.

src/components/Table/GridTable.tsx
chr1sjf0x: Right 🤦‍♂️

src/components/Table/GridTable.stories.tsx
chr1sjf0x: Awesome 👌

src/components/Table/GridTable.tsx
stephenh: ...I need to remember that face palm emoji, I've needed it at least twice this morning and forgot what it was :-D

PR #412, opened 2021-12-18
perf: Fix regression caused by style change

chr1sjf0x:
On 2.95.1 - 165ms render of all table rows
image

With this fix - 70ms render of only the changed row
image

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -284,7 +284,10 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
       const RowComponent = observeRows ? ObservedGridRow : MemoizedGridRow;
 
       return (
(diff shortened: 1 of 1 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -597,7 +600,7 @@ const VirtualRoot = memoizeOne<
     firstLastColumnWidth: number | undefined,
     xss: any,
   ) => Components["List"]
->((gs, columns, id, firstLastColumnWidth, xss) => {
+>(function VirtualRoot(gs, columns, id, firstLastColumnWidth, xss) {
```
chr1sjf0x, line 603: Note: I was trying to get a better name for this in the profile than `Anonymous Memo`, but it didn't work :(

Might be worth trying on the forwardRef below instead 🤔

bdow (approved):
Wow. Crazy! Thanks for finding and fixing.

stephenh (approved):
Nice, good clean up!

Still an annoying game of whack-a-mole though... :-/

PR #414, opened 2021-12-21
feat: Update SelectFieldMock to accept disabledOptions

DeanGilewicz:
[Shortcut Story](https://app.shortcut.com/homebound-team/story/10875/update-selectfieldmock-to-accept-disabledoptions)

### DIFF
--- a/src/inputs/SelectField.mock.test.tsx
+++ b/src/inputs/SelectField.mock.test.tsx
@@ -72,7 +72,7 @@ describe("MockSelectField", () => {
       />,
     );
     // Then matching options are disabled
-    expect(r.getByText("two")).toHaveAttribute("disabled");
-    expect(r.getByText("thr")).toHaveAttribute("disabled");
+    expect(r.getByText("two")).toBeDisabled();
+    expect(r.getByText("thr")).toBeDisabled();
   });
 });
--- a/src/inputs/SelectField.mock.test.tsx
+++ b/src/inputs/SelectField.mock.test.tsx
@@ -49,4 +49,30 @@ describe("MockSelectField", () => {
     // Then onSelect was called with 3
     expect(onSelect).toHaveBeenCalledWith("3", { id: "3", name: "thr" });
   });
+
+  it("disables options that match disabledOptions", async () => {
+    const options = [
+      { id: "1", name: "one" },
+      { id: "2", name: "two" },
+      { id: "3", name: "thr" },
+      { id: "4", name: "four" },
+    ];
+    const value = "1";
+    const onSelect = jest.fn();
+    // When disabledOptions are provided
+    const r = await render(
+      <MockSelectField
+        label="test"
+        getOptionValue={(o) => o.id}
+        getOptionLabel={(o) => o.name}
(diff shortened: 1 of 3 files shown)


src/inputs/SelectField.mock.test.tsx
```diff
@@ -49,4 +49,30 @@ describe("MockSelectField", () => {
...
+        getOptionLabel={(o) => o.name}
+        value={value}
+        onSelect={onSelect}
+        options={options}
+        disabledOptions={["2", "3"]}
+      />,
+    );
+    // Then matching options are disabled
+    expect(r.getByText("two")).toHaveAttribute("disabled");
+    expect(r.getByText("thr")).toHaveAttribute("disabled");
```
bdow, line 76: **nit (non-blocking):** Just wanted to note that you could've also done: `expect().toBeDisabled()`. I sometimes like these methods better than checking against attributes... it'll do basically the same thing, but I tend to like using the specific methods when available.

src/inputs/SelectField.mock.tsx
```diff
@@ -15,6 +15,7 @@ export function SelectField<T extends object, V extends Key>(props: SelectFieldP
     onBlur,
     onFocus,
     disabled,
+    disabledOptions,
```
bdow, line 18: **suggestion:** Rather than having to check and see if the array is defined for each options as we loop through, we could default it to an empty array.
```
const { disabledOptions = [] } = props;

...
options.map((o, i) => {
  return <option disabled={disabledOptions.includes(getOptionValue(option).toString())}>
})
```

src/inputs/SelectField.mock.test.tsx
DeanGilewicz: good shout - I'll update

src/inputs/SelectField.mock.tsx
DeanGilewicz: I was thinking about implementing with a fallback empty array so I'll make that update. Cheers

src/inputs/SelectField.mock.tsx
```diff
@@ -45,8 +46,10 @@ export function SelectField<T extends object, V extends Key>(props: SelectFieldP
     >
       <option disabled value=""></option>
       {options.map((option, i) => {
+        const isDisabled =
+          disabledOptions && disabledOptions.includes(getOptionValue(option).toString()) ? true : false;
```
bdow, line 50: I was thinking you wouldn't need a ternary statement here as `isDisabled` would return `undefined | Boolean`, but maybe we can't use `undefined` as a value of `disabled`? If you use the suggestion of defaulting to an empty array, then you could omit the `disabledOptions && ` part of the conditional, then it'll always return a Boolean, at which point the `? true : false` is redundant.

DeanGilewicz: yeah you are right, it didn't like `undefined` as a value for `disabled` so put the ternary in there but yeah can remove it now with the other approach. Will make this update - thanks!

bdow (approved):

PR #420, opened 2022-01-04
fix: sc-10414 Increase overscan and update height calculations for better virtual table scrolling experience

khalidwilliams:
This pull request updates GridTable's renderVirtual function to:
1. Calculate element heights with `El.getBoundingClientRect().height` 
2. Add an overscan to crate a smoother scrolling experience 

Previously the ChromeRows on Virtual tables would round their `offsetHeights` somewhat unpredictably when zoomed in, leading to a jumpy scrolling experience. Furthermore `getBoundingClientRect` is the method that virtuoso uses by default for getting an element's height.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "react-router": "^5.2.0",
     "react-router-dom": "^5.2.0",
     "react-stately": "^3.9.0",
-    "react-virtuoso": "^1.9.0",
+    "react-virtuoso": "^2.3.1",
     "tinycolor2": "^1.4.2",
     "tributejs": "^5.1.3",
     "trix": "^1.3.1",
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -11,7 +11,7 @@ import React, {
   useState,
 } from "react";
 import { Link } from "react-router-dom";
-import { Components, LogLevel, Virtuoso, VirtuosoHandle } from "react-virtuoso";
+import { Components, Virtuoso, VirtuosoHandle } from "react-virtuoso";
 import { navLink } from "src/components/CssReset";
 import { PresentationProvider } from "src/components/PresentationContext";
 import { createRowLookup, GridRowLookup } from "src/components/Table/GridRowLookup";
@@ -544,13 +544,8 @@ function renderVirtual<R extends Kinded>(
   }, [style]);
   return (
     <Virtuoso
-      logLevel={LogLevel.DEBUG}
-      // overscan={{ main: 10, reverse: 10 }}
+      overscan={5}
       ref={virtuosoRef}
-      // scrollSeekConfiguration={{
-      //   enter: (velocity) => velocity > 50,
-      //   exit: (velocity) => velocity < 10,
-      // }}
       components={{
         List: VirtualRoot(listStyle, columns, id, firstLastColumnWidth, xss),
         Footer: () => <div css={footerStyle} />,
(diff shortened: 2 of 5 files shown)


ksk5280 (approved):

package.json
```diff
@@ -48,7 +48,7 @@
     "react-router": "^5.2.0",
     "react-router-dom": "^5.2.0",
     "react-stately": "^3.9.0",
-    "react-virtuoso": "^1.9.0",
+    "react-virtuoso": "^2.3.1",
```
KoltonG, line 51: **question:** Any neat perf updated with the major version change?

src/components/Table/GridTable.tsx
```diff
@@ -544,6 +544,7 @@ function renderVirtual<R extends Kinded>(
   }, [style]);
   return (
     <Virtuoso
+      overscan={5}
```
KoltonG, line 547: **question:** Is this above and below?

KoltonG: Maybe in the future if we need to update this we can expose a prop?

src/components/Table/GridTable.tsx
```diff
@@ -556,12 +557,12 @@ function renderVirtual<R extends Kinded>(
...
         // If it is a chrome row, then we are not using `display: contents;`, return the height of this element.
         if ("chrome" in maybeContentsDiv.dataset) {
-          return maybeContentsDiv.offsetHeight;
+          return maybeContentsDiv.getBoundingClientRect().height || 1;
         }
 
         // Both the `Item` and `itemContent` use `display: contents`, so their height is 0,
         // so instead drill into the 1st real content cell.
-        return (maybeContentsDiv.firstElementChild! as HTMLElement).offsetHeight;
+        return (maybeContentsDiv.firstElementChild! as HTMLElement).getBoundingClientRect().height;
```
KoltonG, line 565: **question:** Is the `...clientRect` returns a float vs an integer?

KoltonG (approved):
Looks great! Seems that Chromatic does not see any changes!

chr1sjf0x (approved):
TY @khalidwilliams 🚀

package.json
khalidwilliams: No, but the update provided the ability to use their logging / debugging, which was nice.

src/components/Table/GridTable.tsx
khalidwilliams: I believe it's above and below, but their docs don't use very clear language

src/components/Table/GridTable.tsx
khalidwilliams: Yep!

src/components/Table/GridTable.tsx
```diff
@@ -556,12 +557,12 @@ function renderVirtual<R extends Kinded>(
 
         // If it is a chrome row, then we are not using `display: contents;`, return the height of this element.
         if ("chrome" in maybeContentsDiv.dataset) {
-          return maybeContentsDiv.offsetHeight;
+          return maybeContentsDiv.getBoundingClientRect().height || 1;
```
stephenh, line 560: Just curious, but what's better about `getBoundingClientRect()` vs `offsetHeight`? Just more accurate?

stephenh: Ah gotcha, gotcha, I read the PR description; makes sense, I think this is great as-is then, i.e. not need for a follow up in-source comment.

PR #421, opened 2022-01-10
fix: SelectField field state should update inputValue, selectedKeys, and selectedOptions when options update

bdow:


src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -283,7 +283,22 @@ export function SelectFieldBase<O, V extends Value>(props: BeamSelectFieldBasePr
...
+        return {
+          ...prevState,
+          selectedKeys: selectedOptions?.map((o) => valueToKey(getOptionValue(o))) ?? [],
+          inputValue:
+            selectedOptions.length === 1
+              ? getOptionLabel(selectedOptions[0])
+              : multiselect && selectedOptions.length === 0
+              ? nothingSelectedText
+              : "",
+          selectedOptions: selectedOptions,
```
bdow, line 297: Missing these three fieldState properties caused the field not to reflect the correct value.

stephenh: Kinda wonder if we could have like an `updateFieldState(prevState, newValues)` function where `newValues` is a Partial or something, and then any change to `setFieldState`` actually goes through `setFieldState(updateFieldState(prevState, newValues))` so that we ensure all ~3-5 keys are more "reactive" / updated as a group.

src/inputs/SelectField.test.tsx
```diff
@@ -195,6 +196,26 @@ describe("SelectFieldTest", () => {
     expect(r.getAllByRole("option")).toHaveLength(3);
   });
 
+  it("can set value when options are loaded later", async () => {
```
stephenh, line 199: Cool, good test!

stephenh (approved):

src/inputs/internal/SelectFieldBase.tsx
bdow: Yea, I think its a good idea as there is some copy and pasted logic. I think we when decouple our single and multi select fields from `SelectFieldBase`, that'd be a good time to create that helper function. Hopefully that is coming soon.

PR #422, opened 2022-01-10
fix: Expose an onEnter callback for TextField

bdow:
We recently introduced that hitting the "Enter" on a TextField would immediately blur the field. Because we blur out of the field, then we no longer trigger default form submission behavior when a user hits the "Enter" key when the `<input />` element is wrapped in a `<form />`. It is unfortunate to have this be the default behavior in my opinion, but perhaps we can change this in the future? For now, providing an `onEnter` as another way to support submitting a form via the Enter key when within a TextField.

### DIFF
--- a/src/forms/BoundTextField.test.tsx
+++ b/src/forms/BoundTextField.test.tsx
@@ -1,5 +1,6 @@
 import { createObjectState, ObjectConfig, required } from "@homebound/form-state";
 import { render } from "@homebound/rtl-utils";
+import { fireEvent } from "@testing-library/react";
 import { BoundTextField } from "src/forms/BoundTextField";
 import { AuthorInput } from "src/forms/formStateDomain";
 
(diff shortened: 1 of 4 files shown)


stephenh:
@bdow just curious, but what code do we have that is relying on `form` tag `onSubmit`s? Wondering if we should just update the code to use `autoSave` or something of the like...

Granted, I know "relying on `form` / `input` / HTML semantics" is typically considered a good thing, but dunno, I think we (and most advanced form UX in general) so far down the path of custom, that just going all the way to "you don't even need a `form` tag" is IMO fine. ...granted, maybe that's too naive wrt accessibility...

bdow:
> @bdow just curious, but what code do we have that is relying on `form` tag `onSubmit`s? Wondering if we should just update the code to use `autoSave` or something of the like...
> 
> Granted, I know "relying on `form` / `input` / HTML semantics" is typically considered a good thing, but dunno, I think we (and most advanced form UX in general) so far down the path of custom, that just going all the way to "you don't even need a `form` tag" (not literally, but in practice relying on all of our own logic) is IMO fine. ...granted, maybe that's too naive wrt accessibility...

The quick add TODO form is the example I was working off from, which doesn't currently use form state. It was working properly until we introduced the "blur on enter" logic which short circuits the native form submission behavior. 
I think you're right that they could use formState's autoSave functionality to achieve this goal. I initially felt like this would be a bit much of a setup for such a small form (one text field), but probably worth it if not just for consistency sake across the app. 
FYI @jgujgu, see above. The `blur` event was introduced as a way to trigger Form State's autoSave, which is essentially what we're trying to achieve here. You'll need to convert to using Form State. See [ItemTemplateItemDetail](https://github.com/homebound-team/internal-frontend/blob/b1d9e04da332bbf5843c08ec64fb547257b2d13d/src/routes/settings/itemTemplates/ItemTemplateItemDetail.tsx#L58) as an example on how to use. Your `autoSave` method will look a lot like your current `onSubmit` callback.
Let me know if you have any questions, happy to chat through the changes more.

jgujgu:
@bdow I think auto-save on blur sounded reasonable, but it auto-saves on any blur at this point (https://www.loom.com/share/7133db04be584eecbf50f5cf7ed49544). I think auto-saving in this manner is just a bit too eager for the original intention of this feature.

bdow:
@jgujgu that makes sense.. didn't even think about that. 
@stephenh, in Jeff's example we really want to know when the user hits enter to submit the form vs a generic blur event. So they really don't want "auto save" as defined today (where blur trigger save), then just want the `onEnter` to trigger save.

stephenh:
@bdow / @jgujgu ah gotcha, yeah, agreed this use case isn't actually auto-save...

So... can we keep "enter fires blur" (for our existing auto-save fields) and then, exactly what this PR does, as an additional `onEnter` that the "add todo" use case will use, i.e. it will not use form-states auto-save, and will just directly listen for `onEnter`? Maybe that was your intent all along @bdow ? If so, apologies for taking awhile to get it.

bdow:
@stephenh, yes that was the case. Though, I hadn't thought of the option to use form state's auto save to get the "on enter triggers save" behavior. Sounded like a good idea at first, but didn't think of how eager it'd be in saving as Jeff demoed.

stephenh (approved):

PR #425, opened 2022-01-13
feat: GridTable 'as=virtual' to not use CSS grid or display contents

bdow:


src/components/Table/GridTable.tsx
```diff
@@ -968,13 +976,21 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
           ...getIndentationCss(style, rowStyle, columnIndex, maybeContent),
           // Then apply any header-specific override
           ...(isHeader && style.headerCellCss),
+          // Non-virtualized tables use h100 so that all cells are the same height when scrolled. Virtualized table rows use `display: flex;` and applying `h100` to the cells causes layout issues (cell only taking up 100% if its content height vs parent height).
```
stephenh, line 979: > "Non-virtualized tables use h100 so that all cells are the same height when scrolled"

Is the `.h100` specifically for "when scrolled"? I thought it was just in general if cell 1 is shorter than cell 2, but they both have like background colors, we want cell 1 to grow to be cell 2's height?

> "Virtualized table rows use `display: flex;` and applying `h100` to the cells causes layout issues (cell only
> taking up 100% if its content height vs parent height)."

Hm, the "and ..." clauses leaves me hanging a little bit, like I expected a follow up "but that's okay we don't need it" or "and so we'll just live with mis-matched cell heights for now".

Are either of my guesses there correct?

src/components/Table/GridTable.tsx
```diff
@@ -968,13 +976,21 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
...
           // If we're within a card, use its background color
           ...(currentCard && Css.bgColor(currentCard.bgColor).$),
           // Add in colspan css if needed
           ...(currentColspan > 1 ? Css.gc(`span ${currentColspan}`).$ : {}),
           // And finally the specific cell's css (if any from GridCellContent)
           ...rowStyleCellCss,
+          // For virtual tables we define the width of the column on each cell. Supports col spans.
+          ...(columnSizes && {
+            width: `calc(${columnSizes
+              .slice(maybeNestedCardColumnIndex, maybeNestedCardColumnIndex + currentColspan)
```
stephenh, line 991: Slick!

stephenh (approved):
Great!

src/components/Table/GridTable.tsx
bdow: You're right on point one, I mostly copied this part of the comment from another file. But you're correct that the reason for `h100` in this instance is so that all cells take up all height available to them in the row.

For point 2, We don't need `h100` because the flex children will be `align-self: stretch` by default, and that gives us what we need. If we apply `h100` to those flex children, then that `100%` is based on the content of the element and not its parent (as there is no height defined on the parent).

I'll update this comment.

PR #427, opened 2022-01-14
fix: Styling issues after removing css from from virtual GridTable

bdow:


src/components/Table/GridTable.tsx
```diff
@@ -617,7 +617,7 @@ const VirtualRoot = memoizeOne<
         style={style}
         css={{
           // Add an extra `> div` due to Item + itemContent both having divs
-          ...Css.addIn("& > div + div > div", gs.betweenRowsCss || {}).$,
+          ...Css.addIn("& > div + div > div > *", gs.betweenRowsCss || {}).$,
```
bdow, line 620: Changing this back to what is was before - setting the bottom border on the cell elements, instead of the row element. This wasn't working as expected when a row extended the width of the screen, causing a horizontal scrollbar. The border bottom only extended to the viewport's width and not across the whole row. Would like to figure out how to fix that, but for now just putting it back to how it was before.

src/components/Table/styles.tsx
```diff
@@ -11,7 +11,7 @@ export const defaultStyle: GridStyle = {
   lastCellCss: Css.$,
   indentOneCss: Css.pl4.$,
   indentTwoCss: Css.pl7.$,
-  headerCellCss: Css.asfe.nowrap.py1.bgGray100.aife.$,
+  headerCellCss: Css.nowrap.py1.bgGray100.aife.$,
```
bdow, line 14: We want our header cells to be `align-self: stretch`, which is the default.

PR #429, opened 2022-01-14
fix: Use CSS only solution to style focus state for CompoundField

blambillotte:
Fixes an issue where `onFocus` and `onBlur` are getting overwritten (a problem for `Bound` type fields that never expect `onFocus` or `onBlur` to be passed in externally, but due to using `...others` prop spread below, the existing logic gets clobbered).

### DIFF
--- a/src/components/internal/CompoundField.tsx
+++ b/src/components/internal/CompoundField.tsx
@@ -1,55 +1,44 @@
-import { cloneElement, useState } from "react";
+import { cloneElement } from "react";
 import { Css } from "src/Css";
 import { TextFieldInternalProps } from "src/interfaces";
-import { maybeCall } from "src/utils";
 
 /** Internal component to help create compound fields */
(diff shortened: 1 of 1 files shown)


chr1sjf0x (approved):
This looks good to me, but I'll defer to @bdow as to whether this has any unintended consequences.

bdow (approved):
I'm good with this change, but I also think our `Bound____Fields` should in turn call `onFocus` and `onBlur` when passed to the component. We have avoided using CSS to dictate states (I think it was `:active`, or `:focus`) because of inconsistencies between rendering in browsers. This happed to use with Buttons.
Would you mind making sure this renders as expected in Safari, Firefox and Chrome?

blambillotte:
Done - tested on Chrome, Firefox and Safari

blambillotte:
oops, the final commit message omitted `fix:` from the PR title - @bdow is there a way to manually kick off the release?

PR #431, opened 2022-01-18
feat: Reset to initial sort state for GridTable

bdow:
### [SC-11548](https://app.shortcut.com/homebound-team/story/11548/gridtable-to-reset-back-to-initial-sort-when-clicking-on-a-column-that-is-currently-in-descending-order)

GridTable sorting state now toggle between ASC -> DESC -> Initial Sort State. For client side sorting, the initial sort is defined either by the specified initial sort, or as the first sortable column.
For server side sorting we do not define an initially sorted column if it was not specified, (i.e. we do not look for a "first sortable column"). This gives the ability for server side sorting implementations to handle an `undefined` sort state. For use cases like Blueprint's Schedule V2, this allows the table to handle it's sorting reset via: 

```
<GridTable sorting={{ on: "server", value: undefined, onSort: (sortState) => { 
  if (sortState[0] === undefined) { 
    // reset to original list
  } else {
    // sort tasks locally
  }
}} />
```

Granted, this now requires Blueprint's Schedule V2 table to do their sorting locally (as it is currently client-side sorted), but this seemed like kind of a one-off. I did not think it worth maintaining a list of the initial order (especially when it comes to nested structures) for a "reset" look up. Happy to revisit if we find it a more common pattern.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -386,6 +386,87 @@ describe("GridTable", () => {
       expect(cell(r, 2, 0)).toHaveTextContent("b");
     });
 
+    it("reverts to initial sorted column", async () => {
+      // Given a table
+      const r = await render(
+        <GridTable
+          columns={[nameColumn, valueColumn]}
+          // And there is an initial sort defined
+          sorting={{ on: "client", initial: [nameColumn, "ASC"] }}
+          rows={[
+            simpleHeader,
+            { kind: "data", id: "1", name: "a", value: 2 },
+            { kind: "data", id: "2", name: "b", value: 3 },
+            { kind: "data", id: "3", name: "c", value: 1 },
+          ]}
+        />,
+      );
+      // When initializing sort on the "value" column
+      click(r.sortHeader_1);
+      // Then expect ASC order
+      expect(cell(r, 1, 1)).toHaveTextContent("1");
+      // And when clicking a 2nd time
+      click(r.sortHeader_1);
+      // Then expect DESC order
+      expect(cell(r, 1, 1)).toHaveTextContent("3");
+      // And when clicking a 3rd time
+      click(r.sortHeader_1);
+      // Then expect the order to have been reset
+      expect(cell(r, 1, 1)).toHaveTextContent("2");
+    });
+
+    it("initializes with undefined sort", async () => {
+      // Given a table
+      const r = await render(
+        <GridTable
+          columns={[nameColumn, valueColumn]}
+          // And the initial sort is explicitly set to `undefined`
+          sorting={{ on: "client", initial: undefined }}
+          rows={[
+            simpleHeader,
+            // And the data is initially unsorted
+            { kind: "data", id: "2", name: "b", value: 2 },
+            { kind: "data", id: "1", name: "a", value: 3 },
+            { kind: "data", id: "3", name: "c", value: 1 },
+          ]}
+        />,
+      );
+      // Then the data remains unsorted
+      expect(cell(r, 1, 0)).toHaveTextContent("b");
+      expect(cell(r, 2, 0)).toHaveTextContent("a");
+      expect(cell(r, 3, 0)).toHaveTextContent("c");
+    });
+
+    it("reverts to initially undefined sort", async () => {
+      // Given a table
+      const r = await render(
+        <GridTable
+          columns={[nameColumn, valueColumn]}
+          // And the initial sort is explicitly set to `undefined`
+          sorting={{ on: "client", initial: undefined }}
+          rows={[
+            simpleHeader,
+            // And the data is initially unsorted
+            { kind: "data", id: "2", name: "b", value: 2 },
+            { kind: "data", id: "1", name: "a", value: 3 },
+            { kind: "data", id: "3", name: "c", value: 1 },
+          ]}
+        />,
+      );
+      // When initializing sort on the "name" column
+      click(r.sortHeader_0);
+      // Then expect ASC order
+      expect(cell(r, 1, 0)).toHaveTextContent("a");
+      // And when clicking a 2nd time
+      click(r.sortHeader_0);
+      // Then expect DESC order
+      expect(cell(r, 1, 0)).toHaveTextContent("c");
+      // And when clicking a 3rd time
+      click(r.sortHeader_0);
+      // Then expect the order to have been reset
+      expect(cell(r, 1, 0)).toHaveTextContent("b");
+    });
+
     it("can sort nested rows", async () => {
       // Given a table with nested rows
       const r = await render(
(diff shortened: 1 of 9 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -150,7 +150,7 @@ export type GridSortConfig<S> =
       /** The current sort by value + direction (if server-side sorting). */
       value?: [S, Direction];
       /** Callback for when the column is sorted (if server-side sorting). */
-      onSort: (orderBy: S, direction: Direction) => void;
+      onSort: (orderBy: S | undefined, direction: Direction | undefined) => void;
```
stephenh, line 153: Maybe include an in-source comment that `undefined` is "return to the initial sort state"?

src/components/Table/useSortState.ts
```diff
@@ -33,21 +33,42 @@ export function useSortState<R extends Kinded, S>(
...
+
+      // If sorting on a new key -> ASC this key.
+      if (clickedKey !== currentKey) {
+        setSortState([clickedKey, ASC]);
+
+        if (sorting?.on === "server") {
+          sorting.onSort(clickedKey, ASC);
+        }
+        return;
+      }
```
stephenh, line 52: Could we have a helper method like:

```
const [newKey, newDirection] = deriveSortState(currentKey, currentDirection clickedKey);
setSortState([newKey, newDirection])
if (sorting?.on === "server') {
  sorting.onSort(newKey, newDirection);
```

I.e. then we could have one `sorting?.on` + `sorting.onSort` call, which I think each of these three conditions is currently doing on its own.

stephenh (commented):
I think we have several existing tests that test sorting behavior; add a new one/update existing that shows the cycle back to initial?

bdow:
> I think we have several existing tests that test sorting behavior; add a new one/update existing that shows the cycle back to initial?

@stephenh, Definitely. I should have put this out as a draft, though wanted to get eyes on it. This was more of a exploration to see if this was a good first approach and get some feedback. I'll update/add tests and address your comments. Thanks for looking it over.

stephenh:
>  For client side sorting, the initial sort is defined either by the specified initial sort, or as the first sortable column.

I think this makes sense, same with the "schedules v2 can use the _cough_ server-side sorting _cough_" escape hatch. :-)

That said, we could probably allow:

`sorting={{ on: "client", initial: undefined }}`

To enable the "initial sort is 'not sorted'" behavior, while still using client-side sorting. I.e. use key presence to turn on/off the current "pick the 1st column if initial is unset" behavior.

I'm a little tempted to suggest doing that in this PR just for completeness, but also would be fine with a TODO somewhere noting that we could implement it this way when/if needed later.

bdow:
> `sorting={{ on: "client", initial: undefined }}`
> 
> To enable the "initial sort is 'not sorted'" behavior, while still using client-side sorting. I.e. use key presence to turn on/off the current "pick the 1st column if initial is unset" behavior.

@stephenh - Right now `initial` is an optional property, so `undefined` kind of already has a meaning. Though, it could have an explicit vs implicit meaning. What do you think about doing `initial: null` instead? 🤔 Then we should probably make sure `on: server` handles `null` vs `undefined` in the same fashion, which could be confusing with existing instances? Maybe just an explicit vs. implicit `undefined`...

stephenh:
> some ideas on storing that initial order for resetting the client side sort

Ah yeah...we should probably just stop mutating the `rows` that the client passes in, which is kind of cheating / not kosher anyway.

I.e. when client-sorting is on, always make a copy of `rows`, and sort that, then on re-sort (or back to initial sort), just use the not-touched-by-us `rows` props as-is.

(We've been kinda lucky re mutating the `rows` prop, i.e. the apollo client technically gives us `Object.freeze`-n data that we shouldn't be allowed to sort anyway ... except that we always `.map`/copy it to add our `kind` flags, so are accidentally setting `rows` to an already-copied/mutable array.)

src/components/Table/GridTable.test.tsx
```diff
@@ -386,6 +386,87 @@ describe("GridTable", () => {
...
+          rows={[
+            simpleHeader,
+            // And the data is initially unsorted
+            { kind: "data", id: "2", name: "b", value: 2 },
+            { kind: "data", id: "1", name: "a", value: 3 },
+            { kind: "data", id: "3", name: "c", value: 1 },
+          ]}
+        />,
+      );
+      // Then the data remains unsorted
```
stephenh, line 434: Great test!

src/components/Table/sortRows.ts
```diff
@@ -9,37 +9,35 @@ import {
...
-  sortBatch(columns, rows, sortState);
+): GridDataRow<R>[] {
+  const sorted = sortBatch(columns, rows, sortState);
   // Recursively sort child rows
-  for (const row of rows) {
+  for (const row of sorted) {
     if (row.children) {
-      sortRows(columns, row.children, sortState);
+      // Replace the existing `children` prop with a sorted shallow copy
+      Object.assign(row, { children: sortRows(columns, row.children, sortState) });
```
stephenh, line 23: Oh damn...we are actually mutating `row` so we're losing original order here...hrm...

src/components/Table/useSortState.ts
```diff
@@ -19,35 +19,35 @@ export function useSortState<R extends Kinded, S>(
...
-          ? [currentKey, currentDirection === ASC ? DESC : ASC]
-          : // Otherwise, use the new key, and default to ascending
-            [clickedKey, ASC];
-      setSortState([newKey, newDirection]);
+      const newState = deriveSortState(sortState, clickedKey, initialSortState);
+      setSortState(newState);
+
       if (sorting?.on === "server") {
-        sorting.onSort(newKey, newDirection);
+        const [initialKey, initialDirection] = newState ? newState : [undefined, undefined];
```
stephenh, line 49: Could this be `newState ?? [undefined, undefined]`? 

Could `initialKey` / `initialDirection` be called `newKey` / `newDirection`?

src/components/Table/GridTable.tsx
```diff
@@ -150,14 +150,14 @@ export type GridSortConfig<S> =
...
-      initial?: [S | GridColumn<any>, Direction];
+      initial?: [S | GridColumn<any>, Direction] | undefined;
     }
   | {
       on: "server";
       /** The current sort by value + direction (if server-side sorting). */
       value?: [S, Direction];
-      /** Callback for when the column is sorted (if server-side sorting). */
-      onSort: (orderBy: S, direction: Direction) => void;
+      /** Callback for when the column is sorted (if server-side sorting). Parameters set to `undefined` is a signal to return to the initial sort state */
```
stephenh, line 159: "Weird" but yeah makes sense :-)

src/components/Table/useSortState.ts
```diff
@@ -56,3 +56,26 @@ export function useSortState<R extends Kinded, S>(
...
+  }
+
+  // Otherwise when clicking the current column, toggle through sort states
+  if (currentDirection === ASC) {
+    // if ASC -> go to desc
+    return [clickedKey, DESC];
+  }
+
+  // Else, direction is already DESC, so revert to original sort value or undefined.
+  return initialSortState ? [...initialSortState] : undefined;
```
stephenh, line 80: Do we need the copy here? Could this even be just `return initialSortState`?

src/components/Table/sortRows.ts
stephenh: I think this probably needs to be:

```
sorted.forEach((row, i) => {
  if (row.children) {
    sorted[i] = { ...sorted[i], children: sortRows(...) };
  }
});
```

That way we'll leave the original `row` that is in the original `rows` untouched.

bdow: Yea, great point. I'm kind of surprised the tests didn't catch this... 🤔

src/components/Table/useSortState.ts
bdow: Yes and Yes.

src/components/Table/useSortState.ts
bdow: 💯 - I was slimming down from my initial logic and guess I stopped too early! Thanks for the thorough feedback as always.

stephenh (approved):

src/components/Table/sortRows.ts
bdow: Ahh.. I didn't write a test for reverting sort on a nested structure. Will add before merging.

PR #434, opened 2022-01-24
feat: Remove use of CSS grid from GridTable

bdow:
[SC-11802](https://app.shortcut.com/homebound-team/story/11802/remove-use-of-css-grid-in-gridtable)

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -159,7 +159,7 @@ export function NestedRows() {
     child: (row) => <CollapseToggle row={row} />,
     grandChild: () => "",
     add: () => "",
-    w: 0,
+    w: "60px",
   });
   const nameColumn: GridColumn<NestedRow> = {
     header: () => "Name",
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -283,8 +283,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     // Break up "foo bar" into `[foo, bar]` and a row must match both `foo` and `bar`
     const filters = (filter && filter.split(/ +/)) || [];
 
-    const columnSizes =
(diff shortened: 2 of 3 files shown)


src/components/Table/GridTable.stories.tsx
```diff
@@ -159,7 +159,7 @@ export function NestedRows() {
     child: (row) => <CollapseToggle row={row} />,
     grandChild: () => "",
     add: () => "",
-    w: 0,
+    w: "60px",
```
KoltonG, line 162: **question:** Since we are no longer using CSS Grid, are we losing out on responsiveness?

src/components/Table/GridTable.tsx
```diff
@@ -1177,11 +1124,11 @@ export function applyRowFn<R extends Kinded>(column: GridColumn<R>, row: GridDat
 
 /** Renders our default cell element, i.e. if no row links and no custom renderCell are used. */
 const defaultRenderFn: (as: RenderAs) => RenderCellFn<any> = (as: RenderAs) => (key, css, content) => {
-  const Row = as === "table" ? "td" : "div";
+  const Cell = as === "table" ? "td" : "div";
```
stephenh, line 1127: Ah yeah, nice

src/components/Table/GridTable.stories.tsx
KoltonG: I left some comments in Chromatic!

stephenh (approved):

src/components/Table/GridTable.stories.tsx
stephenh: Right, so I think we can be responsive to screen width (fr) but not like content-within-the-table (minmaxs, etc.). Right? 

And our rationale for not allowing responsive-to-content-within-the-table is that it makes cross-row column alignment hard i.e. have to use css-grid/table. And even in virtual tables, that is not enough b/c all content is not known at time of render.

bdow: 100% - well put. "response to the screen width, not the content within the table"

PR #436, opened 2022-01-27
fix: Adds generic decimal support for NumberField and fixes parsing logic

bdow:


src/forms/BoundNumberField.test.tsx
```diff
@@ -34,9 +34,9 @@ describe("BoundNumberField", () => {
     // When the user focuses
     fireEvent.focus(heightInInches());
     // And types an invalid, WIP value
-    fireEvent.input(heightInInches(), { target: { value: "11b" } });
+    fireEvent.input(heightInInches(), { target: { value: "11," } });
```
bdow, line 37: NumberField does not allow a user to type in non-numeric characters, such as "b". Instead, changing this to another NaN value, but with an approved NumberField character.

src/inputs/NumberField.tsx
```diff
@@ -148,14 +155,10 @@ export function NumberField(props: NumberFieldProps) {
...
-  if (isNaN(wip)) return;
-
-  // For percentage values we need to initially divide by 100 in order to get their "number value" ("4%" = .04) for the factor multiplier to be accurate.
-  // For example, if the using basisPoints and the user enters "4.31%", then we would expect the response to be 431 basisPoints. If only basing off the `factor` value, then 4.31 * 10000 = 43100, which would not be correct.
-  const value = type === "percent" || type === "basisPoints" ? wip / 100 : wip;
-  // Since the values returned is exactly what is in the field
-  return factor !== 1 ? Math.round(value * factor) : value;
+export function formatValue(value: number, factor: number, numFractionDigits: number | undefined): number | undefined {
+  // `value` for percentage style inputs will be in a number format, i.e. if input value is 4%, the `value` param will equal `.04`
+  // We allow for decimal places
```
stephenh, line 160: ...okay, it took me awhile to understand what, but I think maybe replace this comment with:

```
// We treat percents & cents as (mostly) integers, while useNumberField wants decimals, so
// undo that via `* factor` and `Math.round`, but also keep any specifically-requested `numFractionDigits`,
// i.e. for `type=percent value=12.34`, `value` will be `0.1234` that we want turn into `12.34`.
```

I.e. for awhile I was confused about why we need `maybeAdjustForDecimals` when we already had the `* factor` to "turn 0.04 --> 4". But, right, we have a Math.round, so we need to keep more than what `Math.round` would throw away.

stephenh (commented):
Nice!

PR #440, opened 2022-01-28
fix: 'Bound' fields to use maybeAutoSave

bdow:
Removes hacky use of blurring field just to trigger FormState's auto save.
For text input fields (DateField, TextField, TextAreaField, NumberField), we introduce an 'onEnter' callback used by each field's corresponding "Bound" version to explicitly call 'maybeAutoSave'. These fields will still 'blur' the field on "Enter" press, as that continues to give us the UX feedback we want (when hitting Enter, blurring the field makes it feel like it was committed).

For fields with menu selection  (DateField, SelectField, MultiSelectField, ChipSelectField), or transaction fields (RadioGroup, Checkbox), their corresponding "Bound" versions trigger 'maybeAutoSave' on change.

### DIFF
--- a/src/forms/BoundTextField.tsx
+++ b/src/forms/BoundTextField.tsx
@@ -39,8 +39,7 @@ export function BoundTextField<X extends Only<TextFieldXss, X>>(props: BoundText
           onFocus={() => field.focus()}
           onEnter={() => {
             maybeCall(onEnter);
-            // Blur the field when the user hits the enter key - as if they are "committing" the value and done with the field
-            field.blur();
+            field.maybeAutoSave();
           }}
           {...testId}
           {...others}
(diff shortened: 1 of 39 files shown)


src/inputs/SelectField.test.tsx
```diff
@@ -31,9 +31,6 @@ describe("SelectFieldTest", () => {
     click(r.getByRole("option", { name: "Three" }));
     // Then onSelect was called
     expect(onSelect).toHaveBeenCalledWith("3");
-    // And the field is no longer in focus
-    expect(r.age()).not.toHaveFocus();
-    expect(onBlur).toHaveBeenCalledTimes(1);
```
stephenh, line 36: Should we leave these in, but flipped, i.e. that it still has focus + onBlur was not called?

src/forms/BoundTextField.tsx
```diff
@@ -39,8 +39,7 @@ export function BoundTextField<X extends Only<TextFieldXss, X>>(props: BoundText
           onFocus={() => field.focus()}
           onEnter={() => {
             maybeCall(onEnter);
-            // Blur the field when the user hits the enter key - as if they are "committing" the value and done with the field
-            field.blur();
```
stephenh, line 43: Ah okay, the `TextField` does this for us now, that makes sense!

stephenh (approved):
Great!

PR #444, opened 2022-02-04
fix: Nested GridTable leaf cards draw their own borders

bdow:
Perf Test results on the "Nested Cards Three Levels" story:
## Before
Before

## After
After

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -16,7 +16,7 @@ import { navLink } from "src/components/CssReset";
 import { PresentationProvider } from "src/components/PresentationContext";
 import { createRowLookup, GridRowLookup } from "src/components/Table/GridRowLookup";
 import { GridSortContext, GridSortContextProps } from "src/components/Table/GridSortContext";
-import { isLeafRow, maybeWrapCard, NestedCards } from "src/components/Table/nestedCards";
+import { getNestedCardStyles, isLeafRow, NestedCards, wrapCard } from "src/components/Table/nestedCards";
 import { SortHeader } from "src/components/Table/SortHeader";
 import { ensureClientSideSortValueIsSortable, sortRows } from "src/components/Table/sortRows";
 import { SortState, useSortState } from "src/components/Table/useSortState";
@@ -898,14 +898,6 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
           .filter((style) => style)
       : undefined;
 
-  const leafCardStyles = isLeafRow(row) ? style.nestedCards?.kinds[row.kind] : undefined;
-  // Calculate the horizontal space already allocated by the open cards (paddings and borders)
-  const openCardWidth = openCardStyles ? openCardStyles.reduce((acc, o) => acc + o.pxPx + (o.bColor ? 1 : 0), 0) : 0;
-  // Subtract the openCardWidth from the `firstLastColumnWidth` to determine the amount of padding to add to this card.
-  // Also if it is a leaf card, then we need to additionally subtract out the border width to have it properly line up with the chrome rows
-  const currentCardPaddingWidth =
-    (style.nestedCards?.firstLastColumnWidth ?? 0) - openCardWidth - (leafCardStyles?.bColor ? 1 : 0);
-
   const rowStyleCellCss = maybeApplyFunction(row, rowStyle?.cellCss);
   const rowCss = {
     // For virtual tables use `display: flex` to keep all cells on the same row. For each cell in the row use `flexNone` to ensure they stay their defined widths
@@ -918,20 +910,7 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
     ...maybeApplyFunction(row, rowStyle?.rowCss),
     // Maybe add the sticky header styles
     ...(isHeader && stickyHeader ? Css.sticky.top(stickyOffset).z1.$ : undefined),
-    // Card styles apply a calculated padding to ensure the content lines up properly across all columns
-    ...(currentCardPaddingWidth ? Css.pxPx(currentCardPaddingWidth).$ : undefined),
-    // Leaf cards define their own borders + padding
-    ...(leafCardStyles
-      ? // We can have versions of the same card as a leaf and not as a leaf.
-        // When it is not a leaf then it has chrome rows that create the top and bottom "padding" based on border-radius size. (brPx = "chrome" row height)
-        // When it is a leaf, then we need to apply the brPx to the row to ensure consistent spacing between leaf & non-leaf renders
-        // Additionally, if the leaf card has a border, then subtract the 1px border width from the padding to keep consistent with the "chrome" row
-        Css.pyPx(leafCardStyles.brPx - (leafCardStyles.bColor ? 1 : 0))
-          .borderRadius(`${leafCardStyles.brPx}px`)
-          .bgColor(leafCardStyles.bgColor)
-          .if(!!leafCardStyles.bColor)
-          .bc(leafCardStyles.bColor).ba.$
-      : undefined),
+    ...getNestedCardStyles(row, openCardStyles, style),
   };
 
   let currentColspan = 1;
@@ -1010,7 +989,7 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
     </Row>
   );
 
-  return openCardStyles ? maybeWrapCard(openCardStyles, rowNode) : rowNode;
+  return openCardStyles && openCardStyles.length > 0 ? wrapCard(openCardStyles, rowNode) : rowNode;
 }
 
 // Fix to work with generics, see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37087#issuecomment-656596623
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -16,7 +16,7 @@ import { navLink } from "src/components/CssReset";
 import { PresentationProvider } from "src/components/PresentationContext";
 import { createRowLookup, GridRowLookup } from "src/components/Table/GridRowLookup";
 import { GridSortContext, GridSortContextProps } from "src/components/Table/GridSortContext";
-import { maybeAddCardPadding, NestedCards } from "src/components/Table/nestedCards";
+import { isLeafRow, maybeWrapCard, NestedCards } from "src/components/Table/nestedCards";
 import { SortHeader } from "src/components/Table/SortHeader";
 import { ensureClientSideSortValueIsSortable, sortRows } from "src/components/Table/sortRows";
 import { SortState, useSortState } from "src/components/Table/useSortState";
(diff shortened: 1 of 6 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -607,18 +563,6 @@ function renderVirtual<R extends Kinded>(
       }}
       // Pin/sticky both the header row(s) + firstRowMessage to the top
       topItemCount={(stickyHeader ? headerRows.length : 0) + (firstRowMessage ? 1 : 0)}
-      itemSize={(el) => {
```
stephenh, line 610: This goes away b/c we don't use `display: contents` anymore right?

bdow: Yup, 100%.

src/components/Table/nestedCards.tsx
```diff
@@ -189,6 +189,28 @@ export function maybeAddCardPadding(openCards: NestedCardStyle[], column: "first
   );
 }
 
+/**
+ * Wraps a row within its parent cards. Creates a wrapping div to add the card padding.
+ * Example:
+ * <div parent> <div child> <div grandchild /> </div> </div>
+ */
+export function maybeWrapCard(openCards: NestedCardStyle[], row: ReactNode): any {
```
stephenh, line 197: Maybe call this `maybeWrapLeafCard`? And maybe call the `row` param `leaf`? 

Or can we remove the `maybe`, and just call it `wrapLeafCard`? From looking at the call site, I think it already only calls this method if we know there are open cards.

src/components/Table/GridTable.tsx
```diff
@@ -944,40 +888,56 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
...
   let currentColspan = 1;
-  const maybeStickyHeaderStyles = isHeader && stickyHeader ? Css.sticky.top(stickyOffset).z1.$ : undefined;
 
-  return (
-    <Row css={rowCss} {...others}>
-      {openCardStyles &&
-        maybeAddCardPadding(openCardStyles, "first", {
-          ...maybeStickyHeaderStyles,
-          ...(columnSizes ? Css.w(columnSizes[0]).$ : {}),
-        })}
```
stephenh, line 980: Huh, that's neat that this can go away, and `maybeWrapCard` does it instead. Ah, I guess the css grid restriction going away means we don't need the cells to be exactly at the 2nd level of the DOM anymore. Nice!

...can `mayeAddCardPadding` being deleted?

src/components/Table/GridTable.tsx
```diff
@@ -944,40 +888,56 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
...
+          .filter((style) => style)
+      : undefined;
+
+  const leafCardStyles = isLeafRow(row) ? style.nestedCards?.kinds[row.kind] : undefined;
+  // Calculate the horizontal space already allocated by the open cards (paddings and borders)
+  const openCardWidth = openCardStyles ? openCardStyles.reduce((acc, o) => acc + o.pxPx + (o.bColor ? 1 : 0), 0) : 0;
+  // Subtract the openCardWidth from the `firstLastColumnWidth` to determine the amount of padding to add to this card.
+  // Also if it is a leaf card, then we need to additionally subtract out the border width to have it properly line up with the chrome rows
+  const currentCardPaddingWidth =
+    (style.nestedCards?.firstLastColumnWidth ?? 0) - openCardWidth - (leafCardStyles?.bColor ? 1 : 0);
```
stephenh, line 907: Do you think we could extract this to a `getNestedCardStyles(openCardStyles, row, style)` that lived in `nestedCards.tsx` and we could potentially even unit test?

And then lines 921 to 934 would live in this method, and line 921 would turn into `...getNestedCardStyles(openCardStyles, row, style)` or something like that. Maybe do that conditionally, but then ideally it's one/maybe two conditionals here in `GridRow`, and most of the card/leaf-specific local variables and other conditionals are hidden instead of `getNestedCardStyles`...

src/utils/rtl.tsx
```diff
@@ -84,10 +81,8 @@ export function cellAnd(r: RenderResult, row: number, column: number, testId: st
 }
 
 export function row(r: RenderResult, row: number, tableTestId: string = "gridTable"): HTMLElement {
-  const nonChromeRows = Array.from(r.getByTestId(tableTestId).childNodes).filter(
-    (node) => !("chrome" in (node as HTMLElement).dataset),
-  );
-  return nonChromeRows[row] as HTMLElement;
+  const dataRows = Array.from(r.getByTestId(tableTestId).querySelectorAll("[data-gridrow]"));
```
stephenh, line 84: Nice!

stephenh (approved):
This is great. I'm surprised it was this easy; I was really thinking getting all of the x / y padding / border calcs to line up was going to be a pita (or maybe it _was_ a pita to get here :-) ).

One ask to maybe extract some of the inline css out of `GridRow` but otherwise looks great!

src/components/Table/GridTable.tsx
```diff
@@ -736,7 +680,7 @@ export function calcColumnSizes(columns: GridColumn<any>[], firstLastColumnWidth
         throw new Error("Beam Table column width definition only supports px, percentage, or fr units");
       }
     },
-    { claimedPercentages: 0, claimedPixels: firstLastColumnWidth ? firstLastColumnWidth * 2 : 0, totalFr: 0 },
```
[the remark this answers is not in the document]
bdow: The GridDataRow element no longer takes up 100% of the table width as we now nest this element within the padding, so we do not need to account for the `firstLastColumnWidth` as part of that 100%. Maybe better explained with some examples:

Before the element containing the cells included padding elements, so we needed to account for that `firstLastColumnWidth` when coming up with these column sizes.
```
<div id=table>
  <div gridrow (width 100%)>
    <div padding />
    <div cell-1 />
    <div cell-2 />
    <div padding />
  </div>
</div>
```

Now the element containing our cell data takes up 100% of the remaining space of the table after the nested padding elements
```
<div id=table>
  <div parent-wrapper>
    <div grid-row (width 100%, which already is non-inclusive of the padding added by the parent wrapper)>
      <div cell-1 />
      <div cell-2 />
    </div>
  </div>
</div>
```

src/components/Table/GridTable.tsx
bdow: Yea, I meant to delete `maybeAddCardPadding`. Thanks for the reminder.

src/components/Table/nestedCards.tsx
bdow: I started out with that naming, but this isn't just used for leaf rows. This function is called for all GridRows where `openCardStyles !== undefined`. Though, probably could call this only when `openCardStyles !== undefined && openCardStyles.length > 0`. 
If we ever need to "add card padding" then this is called. In the example of `Parent -> Child -> GrandChild`, then this wrapping function should be executed for both `Child` and `GrandChild` to make sure the GridRow element is contained within each `openCardStyles`.

src/components/Table/GridTable.tsx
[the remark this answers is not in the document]
stephenh: Nice, good explanation! I'm kinda surprised... do you think this is going to cause alignment issues if a child is shown as different levels? E.g. a task three levels deep:

```
<div id=table>
  <div parent-wrapper>
  <div parent-wrapper>
  <div parent-wrapper>
    <div grid-row (width 100%, which already is non-inclusive of the padding added by the parent wrapper)>
      <div cell-1 />
      <div cell-2 />
    </div>
  </div>
  </div>
  </div>
</div>
```

Vs. one that is two levels deep:

```
```
<div id=table>
  <div parent-wrapper>
  <div parent-wrapper>
    <div grid-row (width 100%, which already is non-inclusive of the padding added by the parent wrapper)>
      <div cell-1 />
      <div cell-2 />
    </div>
  </div>
  </div>
</div>
```

Won't that inner `grid-row` have different widths between the 1st and 2nd DOMs, and so surely the cell borders are going to be thrown slightly off...

I.e. can our leaf rows can be different widths now? ... or I'm just forgetting ... in the css-grid days (of like two weeks ago:-)), the 1st and last columns were fixed down the whole table ... do we still do that?

File: src/components/Table/nestedCards.tsx (hunk 2 above)
Comment (stephenh) [reply src/components/Table/nestedCards.tsx:197]: Ah right right, that makes sense.

Comment (bdow) [reply src/components/Table/nestedCards.tsx:197]: I'll update that conditional for calling this function and then rename it to just `wrapCard`.

File: src/components/Table/GridTable.tsx (hunk 6 above)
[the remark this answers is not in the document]
Comment (stephenh) [reply src/components/Table/GridTable.tsx:739]: Oh, maybe that is what this does:

```
  const currentCardPaddingWidth =
    (style.nestedCards?.firstLastColumnWidth ?? 0) ... 
```

?

src/components/Table/nestedCards.tsx
stephenh: > In the example of Parent -> Child -> GrandChild, then this wrapping function should be
> executed for both Child and GrandChild to make sure the GridRow element is contained
> within each openCardStyles.

I think where I'm getting lost in ^ example is, for Child, it is not a leaf node, so we don't draw it's border in the new leaf-only GridRow code...so it needs to be drawn by `wrapCard` right? E.g. `wrapCard(Child)` would add both the parent padding/border and child's own padding/border? ...and it does that b/c when calling `wrapCard(Child)`, the child itself is already considered an open card?

src/components/Table/GridTable.tsx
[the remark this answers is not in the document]
bdow: Yes, the inner grid row would be 100% of the remaining width. But we account for that padding when drawing our row.
Screen Shot 2022-02-08 at 9 48 41 AM

src/components/Table/nestedCards.tsx
bdow: Yes, in that instance `parent` and `child` have their padding and border-left and border-right drawn by `wrapCard`, because the `parent` and `child` are both currently opened cards.

bdow: Maybe another name for the function would help?

stephenh: Gotcha. Maybe `addNonLeafPaddingAndBorders`? :-) `addNotLeafWrappers`... Or just keeping it as `wrapCard` and having a jsdoc that mentions something like "adds non-leaf padding & borders, e.g. if the current row is a non-leaf, it will already be in open cards"...

src/components/Table/GridTable.tsx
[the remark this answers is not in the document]
bdow: I think it'd be more difficult to do in the `wrapCard` function, because we want this `currentCardPaddingWidth` to be applied to the `leaf` card, and not on any of the open/wrapping cards.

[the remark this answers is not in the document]
bdow: I think it'd be more difficult to there because we want to add this padding to the row element built by the GridRow component and not applied to the wrapping elements. Right now I pass the `row: JSX.Element` element to the `wrapCard` function. I'm not sure if there is a good way to modify that element at this point, 🤷

PR #447, opened 2022-02-09
fix: Allow sortable column with init sort direction of DESC to change to ASC

bdow:


src/components/Table/useSortState.test.ts
```diff
@@ -10,11 +10,15 @@ describe("useSortState", () => {
   });
   it("can derive the next sort state when clicking on a currently ascending column", () => {
     expect(deriveSortState([1, ASC], 1, undefined)).toEqual([1, DESC]);
+    // With `initialSortState` the same as the currentSortState Ascending
+    expect(deriveSortState([1, ASC], 1, [1, ASC])).toEqual([1, DESC]);
   });
```
stephenh, line 15: Minor but could add new lines between the `it`s

stephenh (approved):
Nice, makes sense, thanks!

PR #448, opened 2022-02-11
feat: Beam Table Styles

bdow:
Includes a first pass at the following styles:
- Fixed: All rows have a fixed height and content will overflow with an ellipsis if the column is not wide enough to contain it. Provides a 'title' attribute on cells that only contain a string so the browser can provide the built in 'tooltip'.
- Flexible: Content within cells can wrap and will have 16px of padding on top and bottom.

- Nested: Provides a 'groupRowStyle' for providing parent when there is for a nesting depth of 2. Working off of the Budget table example. This can be used along side the Fixed or Flexible row styles.

### DIFF
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { usePresentationContext } from "src/components/PresentationContext";
 import { Css, Margin, Only, Xss } from "src/Css";
 import { useTestIds } from "src/utils/useTestIds";
 
@@ -10,11 +11,13 @@ export interface ChipProps<X> {
 /** Kinda like a chip, but read-only, so no `onClick` or `hover`. */
 export function Chip<X extends Only<Xss<Margin>, X>>(props: ChipProps<X>) {
   const { text, xss = {} } = props;
+  const { fieldProps } = usePresentationContext();
+  const typeScale = fieldProps?.typeScale ?? "sm";
   const tid = useTestIds(props, "chip");
   return (
     <span
       css={{
-        ...Css.dif.aic.br16.sm.pl1.px1.pyPx(2).gray900.bgGray200.$,
+        ...Css[typeScale].dif.aic.br16.pl1.px1.pyPx(2).gray900.bgGray200.$,
         ...xss,
       }}
       {...tid}
--- a/src/components/Chips.tsx
+++ b/src/components/Chips.tsx
@@ -12,12 +12,16 @@ export interface ChipsProps<X> {
 
 /** Renders a list of `Chip`s, with wrapping & appropriate margin between each `Chip`. */
 export function Chips<X extends Only<ChipsXss, X>>(props: ChipsProps<X>) {
-  const { tableStyle } = usePresentationContext();
+  const { wrap } = usePresentationContext();
   const { values, xss = {} } = props;
-  // Do not allow Chips to wrap if within a `fixed` style table
-  const wrap = tableStyle !== "fixed";
   return (
-    <div css={{ ...Css.df.gap1.whiteSpace("normal").$, ...(wrap ? Css.add({ flexWrap: "wrap" }).$ : {}), ...xss }}>
+    <div
+      css={{
+        ...Css.df.gap1.whiteSpace("normal").$,
+        ...(wrap !== false ? Css.add({ flexWrap: "wrap" }).$ : {}),
+        ...xss,
+      }}
+    >
       {values.map((value, i) => (
         <Chip key={i} text={value} />
       ))}
--- a/src/components/Chips.tsx
+++ b/src/components/Chips.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import { Chip } from "src/components/Chip";
+import { usePresentationContext } from "src/components/PresentationContext";
 import { Css, Margin, Only, Xss } from "src/Css";
 
 type ChipsXss = Xss<Margin>;
@@ -11,11 +12,14 @@ export interface ChipsProps<X> {
 
 /** Renders a list of `Chip`s, with wrapping & appropriate margin between each `Chip`. */
 export function Chips<X extends Only<ChipsXss, X>>(props: ChipsProps<X>) {
+  const { tableStyle } = usePresentationContext();
   const { values, xss = {} } = props;
+  // Do not allow Chips to wrap if within a `fixed` style table
+  const wrap = tableStyle !== "fixed";
   return (
-    <div css={{ ...Css.df.add({ flexWrap: "wrap" }).my1.$, ...xss }}>
+    <div css={{ ...Css.df.gap1.whiteSpace("normal").$, ...(wrap ? Css.add({ flexWrap: "wrap" }).$ : {}), ...xss }}>
       {values.map((value, i) => (
-        <Chip key={i} text={value} xss={Css.mr1.mb1.$} />
+        <Chip key={i} text={value} />
       ))}
     </div>
   );
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -13,12 +13,13 @@ import React, {
 import { Link } from "react-router-dom";
 import { Components, Virtuoso, VirtuosoHandle } from "react-virtuoso";
 import { navLink } from "src/components/CssReset";
-import { PresentationProvider } from "src/components/PresentationContext";
+import { PresentationFieldProps, PresentationProvider } from "src/components/PresentationContext";
 import { createRowLookup, GridRowLookup } from "src/components/Table/GridRowLookup";
 import { GridSortContext, GridSortContextProps } from "src/components/Table/GridSortContext";
 import { getNestedCardStyles, isLeafRow, NestedCards, wrapCard } from "src/components/Table/nestedCards";
 import { SortHeader } from "src/components/Table/SortHeader";
 import { ensureClientSideSortValueIsSortable, sortRows } from "src/components/Table/sortRows";
+import { beamFixedStyle, beamFlexibleStyle } from "src/components/Table/styles";
 import { SortState, useSortState } from "src/components/Table/useSortState";
 import { Css, Margin, Only, Properties, Xss } from "src/Css";
 import tinycolor from "tinycolor2";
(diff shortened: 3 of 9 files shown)


src/components/Table/Table.qa.stories.tsx
```diff
@@ -0,0 +1,332 @@
+import { Meta } from "@storybook/react";
```
bdow, line 1: **Thoughts:** My direction with the naming for this file is a couple things:
1. "Table" - At some point I want to rename the current GridTable component to just "Table". There's no "Grid" about it any longer.
2. ".qa.stories.tsx" - The ".qa" part is to signal that this Storybook file is for Design to do Quality Assurance against. For more information on the proposed Storybook structure for Beam see: https://docs.google.com/document/d/1WTnj11UwwIEgvaJLpkYdqqR7dJPyvTGde6WTfA6tn74/edit?usp=sharing

src/components/Table/Table.qa.stories.tsx
```diff
@@ -0,0 +1,332 @@
...
+  const rowStyles = useMemo(() => ({ parent: { cellCss: beamGroupRowStyle } }), []);
+  return (
+    <div css={Css.mw("fit-content").$}>
+      <GridTable<BeamNestedRow>
+        style={beamFixedStyle}
+        columns={beamNestedColumns}
+        rows={beamTotalsRows}
+        rowStyles={totalsRowStyles}
+      />
+      <GridTable<BeamNestedRow>
```
bdow, line 119: Using two "GridTables" to get the Totals bar on top and ensure it lines up with the Table below. They use the same column definitions, but a different set of `row` and rowStyles. 
Perhaps at some point we can teach the GridTable component about a non-header row being above the "header" kind, if it becomes a common pattern.
For now, wrapping the two tables in a `min-width: fit-content" keeps the columns aligned between the two.

src/components/Table/GridTable.tsx
```diff
@@ -1044,6 +1060,12 @@ function toContent(
   }
   if (content && typeof content === "string" && isHeader && canSortColumn) {
     return <SortHeader content={content} iconOnLeft={alignment === "right"} />;
+  } else if (content && typeof content === "string" && style === beamFixedStyle) {
+    return (
+      <span css={Css.mw0.truncate.$} title={content}>
+        {content}
+      </span>
+    );
```
bdow, line 1068: Allows text to overflow when using the 'fixed' style

src/components/Chip.tsx
```diff
@@ -10,11 +11,13 @@ export interface ChipProps<X> {
...
 export function Chip<X extends Only<Xss<Margin>, X>>(props: ChipProps<X>) {
   const { text, xss = {} } = props;
+  const { fieldProps } = usePresentationContext();
+  const typeScale = fieldProps?.typeScale ?? "sm";
   const tid = useTestIds(props, "chip");
   return (
     <span
       css={{
-        ...Css.dif.aic.br16.sm.pl1.px1.pyPx(2).gray900.bgGray200.$,
+        ...Css[typeScale].dif.aic.br16.pl1.px1.pyPx(2).gray900.bgGray200.$,
```
stephenh, line 20: Neat

src/components/Chips.tsx
```diff
@@ -11,11 +12,14 @@ export interface ChipsProps<X> {
 
 /** Renders a list of `Chip`s, with wrapping & appropriate margin between each `Chip`. */
 export function Chips<X extends Only<ChipsXss, X>>(props: ChipsProps<X>) {
+  const { tableStyle } = usePresentationContext();
   const { values, xss = {} } = props;
+  // Do not allow Chips to wrap if within a `fixed` style table
+  const wrap = tableStyle !== "fixed";
```
stephenh, line 18: Hrm, I'm not a huge fan of components being coupled to "am I in a table or not", but we do have precedent of "use no labels / use compact" being set by the grid table.

Is there some attribute ... like "wrap" ... something ... that we could use/set here instead of just "I'm in a table"? 

I wonder if eventually our GridStyles will learn about presentation context settings, e.g. `const style1: GridStyle = ...` and `const style2: GridStyle = ...` would each have their own like `presentationContext: ...` sub-keys to set things like no labels/etc.

src/components/Table/GridTable.tsx
```diff
@@ -404,13 +408,25 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   const firstRowMessage =
     (noData && fallbackMessage) || (tooManyClientSideRows && "Hiding some rows, use filter...") || infoMessage;
 
+  const borderless = style === beamFixedStyle;
+  const tableStyle = style === beamFixedStyle ? "fixed" : style === beamFlexibleStyle ? "flexible" : undefined;
```
stephenh, line 412: This seems really hard-coded... is there a way we can move this away from "we have two global `const`s that are special"?

Seems like at least we could do `GridStyle.style = "fixed" | "flexible"`...

Although I was hoping that "beam fixed" and "beam flexible" would just be presets that set lower-level `GridStyle` settings...

So, like `GridStyle.borderless: boolean` seems like one?... Maybe also `GridStyle.typeScale`?

src/components/Table/GridTable.tsx
```diff
@@ -1044,6 +1060,12 @@ function toContent(
   }
   if (content && typeof content === "string" && isHeader && canSortColumn) {
     return <SortHeader content={content} iconOnLeft={alignment === "right"} />;
+  } else if (content && typeof content === "string" && style === beamFixedStyle) {
+    return (
+      <span css={Css.mw0.truncate.$} title={content}>
```
stephenh, line 1065: Is this like `GridStyle.truncate: boolean` instead of `=== beamFixedStyle`? 

Fwiw I'm assuming that each of the flags I'm suggesting, `truncate`, `borderless`, `typeScale`, are for the most part orthogonal and so mix-and-matchable...if we have something that just can't work that way, please call it out...

src/components/Chips.tsx
stephenh: Maybe even just `PresentationContext.wrap` if that's the most direct name for what this is controlling?

src/components/Table/Table.qa.stories.tsx
stephenh: Yeah, or maybe a new `TableWithTotals` component that has `rows` / `columns` props and then two `totalStyles` / `rowStyles` or something like that. And it'd just make two GridTables like you have here, with the wrapper.

src/components/Table/GridTable.tsx
bdow: Yea, I was trying to think of another way of doing this. I did want to eventually have this be completely dictated by GridStyle, but was hoping for feedback on an approach here, so thanks!
When I first looked at GridStyle then I kind of saw it as mostly css/`Properties`, and moved away from that. But I like your suggestion and will update to use this approach. Then the `beamFixedStyles` would be something like:
```
const beamFixedStyle = { borderless: true, typeScale: "xs", ... } 
```

src/components/Table/GridTable.tsx
bdow: I can't think of anything off hand, so this approach may work just fine. Will definitely call it out if I think of one.

src/components/Table/GridTable.tsx
stephenh: Cool! ...for `borderless`, I wonder if we should repeat some of those `PresentationContext` props within `GridStyle` itself, e.g. have `GridStyle.borderless`, or we should have like `GridStyle.presentationContext`? 

My only disclaimer about the 2nd approach is that I think we probably still want to provide some "all tables get it for free" default presentation settings, e.g. borderless, hide labels, etc., although we would probably use like a base presentation settings for all tables, and then just override those with the current table's `style.presentationSettings`? Maybe...

src/components/Table/Table.qa.stories.tsx
```diff
@@ -0,0 +1,332 @@
...
+  status: string;
+  commitmentName: string;
+  tradeCategories: string[];
+  location: string;
+  date: string;
+  priceInCents: number;
+};
+type BeamRow = SimpleHeaderAndDataWith<BeamData>;
+
+export function Fixed() {
```
stephenh, line 35: These stories look great!

src/components/Table/Table.qa.stories.tsx
```diff
@@ -0,0 +1,332 @@
...
+  costToComplete?: number;
+  children?: BeamChildRow[];
+};
+type HeaderRow = { kind: "header" };
+type BeamTotalsRow = { kind: "totals"; id: string } & BeamBudgetData;
+type BeamParentRow = { kind: "parent"; id: string } & BeamBudgetData;
+type BeamChildRow = { kind: "child"; id: string } & BeamBudgetData;
+type BeamNestedRow = BeamTotalsRow | HeaderRow | BeamParentRow | BeamChildRow;
+
+export function Nested() {
```
stephenh, line 108: Maybe go ahead and have flexible nested and fixed nested stories? Afaiu nested could be either fixed or flexible?

src/components/Table/Table.qa.stories.tsx
```diff
@@ -0,0 +1,332 @@
...
+  const totalsRowStyles = useMemo(() => ({ totals: { cellCss: beamTotalsRowStyle } }), []);
+  const rowStyles = useMemo(() => ({ parent: { cellCss: beamGroupRowStyle } }), []);
```
stephenh, line 110: I'm wondering if there is a way we could avoid a per-file `cellCss` override for these...

I guess `totalsRowStyles` I'm more okay with being one-off...well, or that I think it could eventually move away from a per-file `cellCss` b/c of a pattern like `TableWithTotals rowStyle=fixedStyle totalStyle=totalStyle`, b/c `totalStyle` (with a better name), could define what's currently in `beamTotalsRowStyle` but inside of a dedicated `GridStyle`...

But for the `parent: ...`, I dunno, maybe we could have `GridStyle` know about depth?

Like `GridStyle.rowStyle[0]` is its own `GridStyle` for 1st-level parents, `GridStyle.parentStyle[1]` would be 2nd-level parents....dunno...

Or maybe it's not a whole nested `GridStyle`, but just a few key properties, like:

```
interface GridStyle {
  // existing
  cellCss: ...
  // new
  level: Record<number, { cellCss: Properties, somethingElse: ... }>
}
```

...we could probably also remove `indentOneCss` and `indentTwoCss`, and replace it with this "level 1 has this cellCss, level 2 has that cellCss". Granted, this new approach means a row's level is dictated by whether it's in a `row.children` or not, instead of just deciding to set `indent: 1` out of the blue, but I think that's probably fine. I don't think we ever actually used that flexibility, and it seems like a non-feature.

bdow: 100%. I think there is a ton of clean up we can do with the GridTable styling. Especially now that our rowCss can accomplish a lot more since we removed `display: contents;`. For instance, right now all of our default stylings for rows are applied at each cell. This is fine, but we don't need to do it any longer.
My initial thought/direction was to touch the fewest amount of existing styles as possible, migrate all the existing tables, make the new styles the default styling, and then start trimming off the unused styles. For now I just wanted to get the new styles out there and available.

stephenh: Ah yeah, that's def true for cell css not being as necessary now. Makes sense to minimize churn to get these out.

src/components/Table/GridTable.tsx
bdow: I think this is done now per your suggestions. Thanks again for the insightful and helpful feedback!

stephenh (approved):

PR #450, opened 2022-02-11
fix: Remove 'omitExternalIcon' prop in favor of 'download' prop

bdow:
Adding the 'download' prop will:
1. Default to the 'download' icon
2. Use an anchor tag with the 'download' prop
3. Omit the 'linkExternal' icon that is typically supplied with absolute URLs

The 'omitExternalIcon' prop was meant to give the option for remove this default behavior for a download sort of link. Though, this resulted in opening a new tab to trigger a download. Since there are multiple behaviors we want changed when we have a 'download' type of link, the we are creating a single prop to signal those modifications.


stephenh (approved):
Great, love the call on using a more purpose specific name

DeanGilewicz (approved):
Nice one! - yeah agree this makes sense to define a dedicated prop to handle behavior

PR #451, opened 2022-02-15
feat: ButtonMenu supports DatePicker

bdow:
ButtonMenu can now accept DatePicker props or MenuItem props in order to render a different either the DatePicker, or the existing MenuItems.
Moves the MenuItem state creation to the Menu component, which should hopefully be an improvement.
Moves DatePickerOverlay and DatePicker out of the 'fields' directory and into the 'components' directory as they are no longer only form field components.


src/components/ButtonMenu.tsx
```diff
@@ -30,8 +29,11 @@ interface ButtonMenuProps {
   tooltip?: ReactNode;
 }
 
-export function ButtonMenu(props: ButtonMenuProps) {
-  const { trigger, items, placement, persistentItems, defaultOpen, disabled, tooltip } = props;
+interface ButtonMenuDatePickerProps extends ButtonMenuBaseProps, DatePickerProps {}
+interface ButtonMenuItemsProps extends ButtonMenuBaseProps, MenuItemsProps {}
```
bdow, line 33: Not 100% sure I like this approach of using the `DatePickerProps` directly as it is not super clear that doing `<ButtonMenu onSelect={...} />` is going to give you a DatePicker rather than a Menu. I am contemplating instead doing something like:
```
interface ButtonMenuDatePickerProps extends ButtonMenuBaseProps, DatePickerProps {
  as: "datePicker"
}
```
Basically the same thing, but forcing the user to add `as="datePicker"` in the props... not sure if necessary. Open to other ideas.

stephenh: I've not looked at the code at all, but is this better done as a new component? Like ButtonDateMenu? Seems like a pretty specialized use case.

bdow: It'd basically be a copy & paste of ButtonMenu with a different component shown in the overlay. It could be its own component, but if felt like such a copy and paste that I wanted to improve the existing ButtonMenu to be more flexible about the overlay. This also included a good refactor of the ButtonMenu to remove the Menu specific logic into the Menu component. I'll probably keep that part of the refactoring around.

bdow: Sounds good. I've updated this PR to create a new `ButtonDatePicker` component and an `OverlayTrigger` component that includes the sharable pieces between ButtonDatePicker and ButtonMenu.

src/components/ButtonDatePicker.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const { defaultOpen, disabled, trigger } = props;
+  const state = useMenuTriggerState({ isOpen: defaultOpen });
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const { menuTriggerProps, menuProps } = useMenuTrigger({ isDisabled: !!disabled }, state, buttonRef);
+  const tid = useTestIds(props, isTextButton(trigger) ? defaultTestId(trigger.label) : trigger.icon);
+
+  return (
+    <OverlayTrigger {...props} menuTriggerProps={menuTriggerProps} state={state} buttonRef={buttonRef} {...tid}>
+      <DatePickerOverlay {...props} state={state} overlayProps={menuProps} {...tid.datePicker} />
+    </OverlayTrigger>
```
stephenh, line 26: Wow this is really short. Looks great!

src/components/DatePicker.tsx
```diff
@@ -1,32 +1,24 @@
...
 }
 
-export function DatePickerOverlay(props: DatePickerOverlayProps) {
-  const { value, state, positionProps, onChange, disabledDays } = props;
-  // We define some spacing between the Calendar overlay and the trigger element, and depending on where the overlay renders (above or below) we need to adjust the spacing.
-  // We can determine if the position was flipped based on what style is defined, `top` (for positioned below the trigger), and `bottom` (for above the trigger).
-  // The above assumption regarding `top` and `bottom` is true as long as we use `bottom` as our default `OverlayPosition.placement` (set in DateField).
-  // The reason the placement may not be on bottom even though we set `bottom` is because also set `shouldFlip: true`
-  const isPositionedAbove = !positionProps.style?.top;
+export function DatePicker(props: DatePickerProps) {
```
stephenh, line 14: Maybe include a jsdoc like "Renders a DatePicker inline in the current DOM, see ButtonDatePicker and DateField."

...do we expect users to use this `src/components/DatePicker`? Fwiw after pulling up `ButtonDatePicker` in storybook, I almost wonder if that component should be the "public `DatePicker` component", and this one is an internal `InlineDatePicker`?

Or maybe not, I'm not really sure, so feel free to leave as-is if this `src/components/DatePicker` is actually a useful/standalone public component.

bdow: My gut feeling is that it might be worth being a standalone component. Will move it to be an internal component for now.

PR #452, opened 2022-02-16
feat: Update DatePicker disabledDays type to include Modifier[]

ksk5280:
Updating `DatePicker` `disabledDays` prop to accept an array of Modifiers to fix TS error in this PR: https://github.com/homebound-team/internal-frontend/pull/1794/files#diff-b5da248aa7ad502a7ca5b33777e5a23e60a778ec6084a912bf80471f381480e1R61

### DIFF
--- a/src/components/internal/DatePicker.tsx
+++ b/src/components/internal/DatePicker.tsx
@@ -8,7 +8,7 @@ import "./DatePicker.css";
 export interface DatePickerProps {
   value?: Date;
   onSelect: (value: Date) => void;
-  disabledDays?: Modifier;
+  disabledDays?: Modifier | Modifier[];
 }
 
 export function DatePicker(props: DatePickerProps) {
--- a/src/components/internal/DatePicker.tsx
+++ b/src/components/internal/DatePicker.tsx
@@ -8,7 +8,7 @@ import "./DatePicker.css";
 export interface DatePickerProps {
   value?: Date;
   onSelect: (value: Date) => void;
-  disabledDays?: Modifier | Modifier[];
+  disabledDays?: Modifier[];
 }
 
 export function DatePicker(props: DatePickerProps) {
--- a/src/components/internal/DatePicker.tsx
+++ b/src/components/internal/DatePicker.tsx
@@ -8,7 +8,7 @@ import "./DatePicker.css";
 export interface DatePickerProps {
   value?: Date;
   onSelect: (value: Date) => void;
-  disabledDays?: Modifier[];
+  disabledDays?: Modifier | Modifier[];
 }
 
 export function DatePicker(props: DatePickerProps) {
--- a/src/inputs/DateField.mock.test.tsx
+++ b/src/inputs/DateField.mock.test.tsx
@@ -36,11 +36,14 @@ describe("MockDateField", () => {
 
   it("forwards disabledDays prop as stringified object", async () => {
     const testDate = new Date(2020, 0, 1);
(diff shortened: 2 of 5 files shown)


src/components/internal/DatePicker.tsx
```diff
@@ -8,7 +8,7 @@ import "./DatePicker.css";
 export interface DatePickerProps {
   value?: Date;
   onSelect: (value: Date) => void;
-  disabledDays?: Modifier;
+  disabledDays?: Modifier | Modifier[];
```
bdow, line 11: **thought:** Could we make it only accept an array? We'd need to update existing implementations of this, but I think SchedulesV2's `DateCellField` is the only user of this at the moment.

ksk5280: 👍 Updated!

src/components/internal/DatePicker.tsx
```diff
@@ -8,7 +8,7 @@ import "./DatePicker.css";
 export interface DatePickerProps {
   value?: Date;
   onSelect: (value: Date) => void;
-  disabledDays?: Modifier;
+  disabledDays?: Modifier[];
```
blambillotte, line 11: What do you think about `Modifier | Modifier[]` to allow for either?

src/inputs/DateField.mock.test.tsx
```diff
@@ -35,10 +35,13 @@ describe("MockDateField", () => {
   });
 
   it("forwards disabledDays prop as stringified object", async () => {
-    const testDate = new Date(2020, 0, 1)
-    const r = await render(<MockDateField label="Start Date" value={testDate} disabledDays={{ before: testDate }} onChange={() => {}} />);
+    const testDate = new Date(2020, 0, 1);
+    const disabledDays = [{ before: testDate }, () => false];
```
blambillotte, line 39: I'm not sure I understand this noop `() => false` as a `Modifier`? it that some special case?

src/components/internal/DatePicker.tsx
ksk5280: I had that initially, but Brandon suggested updating to only accept an array.

src/inputs/DateField.mock.test.tsx
ksk5280: It's to represent a FunctionModifier, which I'm using to [disable days for the scheduleSettings](https://github.com/homebound-team/internal-frontend/pull/1794/files#diff-b5da248aa7ad502a7ca5b33777e5a23e60a778ec6084a912bf80471f381480e1R61).

src/components/internal/DatePicker.tsx
bdow: Haha, I had just asked to see if we can only do the array type. I'm not sold on one way or the other, so happy to move back to what you originally had @ksk5280.

bdow (approved):

src/components/internal/DatePicker.tsx
blambillotte: Ha I don't have a strong opinion - my original intention was to just forward this prop and not deviate from `react-day-picker`'s implementation so the docs would still make sense with our wrapper... so this was a miss from me that I made it more restrictive

ksk5280: I think it seems good to keep it flexible, so I reverted back to `Modifier | Modifier[]`

bdow (approved):
Looks great. Thanks for adding the stories

PR #455, opened 2022-02-17
fix: Allow Tooltip children to be ReactNode, which includes strings

bdow:
The `children` prop was previously ReactElement was was require because we used 'cloneElement'. We no longer clone the element and can now allow for more types to be passed along such as strings.

Additionally adds the 'title' prop to the wrapping element when the tooltip is not showing and only if the title is of type string. This is to help with testing purposes and follows a similar pattern found from MUI.

### DIFF
--- a/src/components/Tooltip.test.tsx
+++ b/src/components/Tooltip.test.tsx
@@ -1,4 +1,5 @@
-import { resolveTooltip } from "src/components/Tooltip";
+import { resolveTooltip, Tooltip } from "src/components/Tooltip";
+import { render } from "src/utils/rtl";
 
 describe("Tooltip", () => {
   it("can resolve tooltip text", () => {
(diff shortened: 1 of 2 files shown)


src/components/Tooltip.test.tsx
```diff
@@ -26,4 +27,9 @@ describe("Tooltip", () => {
     expect(resolveTooltip(undefined, "Test 2", true)).toBe("Test 2");
     expect(resolveTooltip(undefined, "Test 2", "")).toBe("Test 2");
   });
+
+  it("renders with title prop", async () => {
+    const r = await render(<Tooltip title="Test tooltip">Text</Tooltip>);
+    expect(r.tooltip()).toHaveAttribute("title", "Test tooltip");
```
stephenh, line 33: Should we do `data-title` or is `title` a valid attribute for spans?

stephenh (approved):

src/components/Tooltip.test.tsx
bdow: `title` is a valid attribute.

PR #456, opened 2022-02-18
fix: Support adding typeScale to a specific kind's column cells

bdow:
Also updates group styles typography to be xsEm instead of smEm. Updates Storybook example to use `typeScale` prop for setting a specific row kind's column typography. This allowed for changing the font size on a single cell without having to nest it with additional elements. The nesting can circumvent the `PresentationContext.wrap`/truncation styles applied.

Also adds support for min-width definition on GridColumn

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -778,6 +778,8 @@ export type GridColumn<R extends Kinded, S = {}> = {
    * - The default value is `1fr`
    */
   w?: number | string;
+  /** The minimum width the column can shrink to */
+  mw?: string;
   /** The column's default alignment for each cell. */
   align?: GridCellAlignment;
   /** Whether the column can be sorted (if client-side sorting). Defaults to true if sorting client-side. */
@@ -1002,6 +1004,7 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
             width: `calc(${columnSizes
               .slice(maybeNestedCardColumnIndex, maybeNestedCardColumnIndex + currentColspan)
               .join(" + ")})`,
+            ...(column.mw ? Css.mw(column.mw).$ : {}),
           }),
         };
 
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -24,7 +24,7 @@ import { getNestedCardStyles, isLeafRow, NestedCards, wrapCard } from "src/compo
 import { SortHeader } from "src/components/Table/SortHeader";
 import { ensureClientSideSortValueIsSortable, sortRows } from "src/components/Table/sortRows";
 import { SortState, useSortState } from "src/components/Table/useSortState";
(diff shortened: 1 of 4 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -854,6 +854,7 @@ export type GridCellContent = {
   /** Whether to indent the cell. */
   indent?: 1 | 2;
   colspan?: number;
+  css?: Properties;
```
stephenh, line 859: Ah yeah; I think historically I would have pushed to keep this as just conditional logic in the `GridRowStyle[kind].cellCss(row)` lambda...

If anything to kind of purposefully nudge the user to "just use the provided GridStyle and don't do customize UI work" ... this almost makes it too easy. :-D 

Just curious, but is this solely for text sizing? Wdyt about adding like a `typeScale` instead of a completely open-ended `css` prop? Just thinking by staying more restricted on what we allow here (and have the user declare intent vs. raw css), it will give us more flexibility to change the internals...

That said, if you already have "more than just text" that you need this for, and yeah this is admittedly already possible with `GridRowStyle.cellCss`, just a different/easier way to skin the cat, I'm good with it too.

src/components/Table/GridTable.tsx
```diff
@@ -778,6 +778,8 @@ export type GridColumn<R extends Kinded, S = {}> = {
    * - The default value is `1fr`
    */
   w?: number | string;
+  /** The minimum width the column can shrink to */
+  mw?: string;
```
stephenh, line 782: Should we enforce that this can only be px or percent?

src/components/Table/GridTable.tsx
bdow: I don't believe this is actually possible with the `GridRowStyle.cellCss`, as that targets _every_ cell in the row, where I only want to target a specific column's cell for this row kind.

And yes, I was just looking to do typeScale for the moment, so I could limit to just that for now. Though, allowing this could also remove the need for the `alignment` and `indent` props, as a user could specify that sort of thing via CSS and have more flexibility on it. Something to think about.

src/components/Table/GridTable.tsx
bdow: Yea, probably should. Otherwise if we use something like `fit-content` then it allows for cells in a column to be drawn at different widths. Will add that in.

src/components/Table/GridTable.tsx
stephenh: > as that targets every cell in the row

Oh shoot, you're right, I was getting columns & kinds mixed up. Gotcha!

bdow:
@stephenh - Your feedback has been addressed. `GridCellContent.css` has changed to `GridCellContent.typeScale`. `GridColumn.mw` now throws error if the definition is not a `px` or `%` value.

stephenh:
@bdow great, thanks!

stephenh (approved):
Er, thought I'd already approved...

PR #459, opened 2022-02-25
fix: MultiFilter does not show chips if number of options is zero

bdow:
By rendering the MultiSelectField for the MultiFilter when the options.length is zero, then we allow users to provide additional UX such as disabling with a tooltip or helper text to communicate why the filter is currently disabled


stephenh (approved):

ksk5280 (approved):
🎉

PR #460, opened 2022-03-02
chore: SC-13496 update NPM contexts

blimmer:
https://app.shortcut.com/homebound-team/story/13496/regenerate-ci-npm-tokens

Our previous NPM token was quite old (2 years) and likely widely exposed before https://github.com/homebound-team/graphql-service/pull/2101 was merged.

This PR updates the CircleCI config to use two distinct `npm` contexts (read-only and publish). This reduces the risk of accidentally publishing in a step that should not. Additionally, it uses a new automation-type token that allows us to enable 2-Factor Auth for login but not for the publishing token.

Once I transition our projects, I will delete the old `npm` context.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  node: circleci/node@5.0.0
+
 node-image: &node-image
   image: cimg/node:16.4.0
   auth:
@@ -12,12 +15,24 @@ workflows:
     jobs:
       - build:
           context:
-            - npm
+            - npm-readonly
             - github
+            - dockerhub
+      - publish:
+          context:
+            - npm-publish
+            - github
+            - dockerhub
+          requires:
+            - build
+          filters:
+            branches:
+              only:
+                - main
       - chromatic:
           context:
+            - npm-readonly
             - dockerhub
-            - npm
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -12,12 +15,24 @@ workflows:
     jobs:
       - build:
           context:
-            - npm
+            - npm-readonly
+            - github
+            - dockerhub
+      - publish:
```
blimmer, line 21: I broke `publish` into a separate step that will run after `build` completes on `main`. This allows branches to use the read-only context, and the publish step to use the publisher context.

.circleci/config.yml
```diff
@@ -26,15 +41,21 @@ jobs:
       - <<: *node-image
     steps:
       - checkout
-      - run: 'echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > .npmrc'
-      - run: yarn install
+      - run: 'echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc'
+      - node/install-packages:
+          pkg-manager: yarn
```
blimmer, line 46: This will start saving the `yarn` cache, which will speed up subsequent builds.

stephenh (approved):

release.config.js
```diff
@@ -3,6 +3,12 @@ module.exports = {
   plugins: [
     ["@semantic-release/commit-analyzer", { preset: "conventionalcommits" }],
     ["@semantic-release/release-notes-generator", { preset: "conventionalcommits" }],
+    [
+      "@semantic-release/exec",
+      {
+        prepareCmd: "yarn build",
```
blimmer, line 9: This is nice because now when we run `yarn semantic-release` it'll ensure the app is built before releasing.

PR #462, opened 2022-03-07
fix: GridTable column 'fr' width values calculated to pixels

bdow:
When the GridTable is first rendered then we use the existing width calculations for 'fr' units. Once the table is rendered, then the 'fr' units are calculated as pixel values. This is very similar to the logic in Blueprint's SchedulesV2. This properly allows for text-overflow to work and keeps all columns aligned, which was a recent issue with the new Beam "fixed" styling.
Uses a ResizeObserver for handling width recalculations in screensize changes.
Allows for having the resize observer listen to another element instead of the GridTable container (also required by SchedulesV2).
Also adds support for a minimum table width (as required by SchedulesV2).


src/components/Table/GridTable.tsx
```diff
@@ -80,6 +82,8 @@ export interface GridStyle {
   emptyCell?: ReactNode;
   presentationSettings?: Pick<PresentationFieldProps, "borderless" | "typeScale"> &
     Pick<PresentationContextProps, "wrap">;
+  /** Minimum table width in pixels. Used when calculating columns sizes */
+  minWidth?: number;
```
bdow, line 86: I thought about either having this defined directly here, or grabbing the minWidth from `GridStyle.rootCss.minWidth`. I decided to go this route in order to define that the property must be a `number`/pixels. Otherwise we'd need to do more checks to ensure that we only accept pixel or percentage values. Though, I don't think a use case for use a percentage minWidth exists, as the default percentage minWidth would be 100% and that seems sufficient. This value will always override any value defined in `rootCss`.

src/components/Table/GridTable.tsx
```diff
@@ -217,6 +221,8 @@ export interface GridTableProps<R extends Kinded, S, X> {
   xss?: X;
   /** Experimental API allowing one to scroll to a table index. Primarily intended for stories at the moment */
   api?: MutableRefObject<GridTableApi | undefined>;
+  /** Specify the element in which the table should resize its columns against. If not set, the table will resize columns based on its owns container's width */
+  resizeTarget?: MutableRefObject<HTMLElement | null>;
```
bdow, line 225: This is required by SchedulesV2. When the Task Details pane opens, then the table container (`tableRef`) will resize. But we don't actually want to base the table width off of this value. Instead we want the Schedule table to stay its same width but introduce a horizontal scrollbar. In this instance the resizeTarget is the container of both the Schedule Table and the Task Details pane.

src/components/Table/GridTable.tsx
```diff
@@ -284,11 +292,26 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
+  const [columnSizes, setColumnSizes] = useState<string[]>(
+    calcColumnSizes(columns, style.nestedCards?.firstLastColumnWidth, undefined, style.minWidth),
   );
 
+  const [tableWidth, setTableWidth] = useState<number | undefined>();
+
+  const debouncedTableWidth = useDebouncedCallback((width) => {
+    setTableWidth(width);
+    setColumnSizes(calcColumnSizes(columns, style.nestedCards?.firstLastColumnWidth, width, style.minWidth));
+  }, 100);
```
bdow, line 304: Chose 100, as it feels pretty snappy. People taking longer than 100ms to resize their screen may cause multiple re-renders. I'd rather wait to see if that is an issue before introducing more lag.

src/components/Table/GridTable.tsx
```diff
@@ -469,12 +494,12 @@ function renderDiv<R extends Kinded>(
...
   xss: any,
   _virtuosoRef: MutableRefObject<VirtuosoHandle | null>,
+  tableRef: MutableRefObject<HTMLElement | null>,
 ): ReactElement {
   return (
     <div
+      ref={tableRef as any}
       css={{
-        // Ensure all rows extend the same width
-        ...Css.mw("fit-content").$,
```
bdow, line 477: 🎉 This can go away!

src/components/Table/GridTable.tsx
```diff
@@ -582,8 +612,12 @@ function renderVirtual<R extends Kinded>(
     <Virtuoso
       overscan={5}
       ref={virtuosoRef}
-      // Add `minWidth: fit-content` to ensure a sticky header and the virtualized table body maintain same width
-      style={{ minWidth: "fit-content" }}
```
bdow, line 586: 🎉 !!

src/components/Table/GridTable.tsx
```diff
@@ -712,6 +752,16 @@ export function calcColumnSizes(columns: GridColumn<any>[], firstLastColumnWidth
...
   function fr(myFr: number): string {
+    // If the tableWidth, then return a pixel value
+    if (tableWidth) {
+      const widthBasis = tableWidth > tableMinWidthPx ? tableWidth : tableMinWidthPx;
+      // When the tableWidth is defined, then we need to account for the `firstLastColumnWidth`s.
+      return `(${
+        (widthBasis - (claimedPercentages / 100) * widthBasis - claimedPixels - (firstLastColumnWidth ?? 0) * 2) *
+        (myFr / totalFr)
+      }px)`;
+    }
```
bdow, line 763: This was mostly stolen from SchedulesV2. Added in handling for "claimedPercentages" and `firstLastColumnWidth`

src/components/Table/GridTable.tsx
```diff
@@ -893,8 +939,7 @@ interface GridRowProps<R extends Kinded, S> {
   setSortKey?: (value: S) => void;
   // NOTE: openCards is a string of colon separated open kinds, so that the result is stable across renders.
   openCards: string | undefined;
-  // Required for setting column widths in a virtualized table
-  columnSizes: string[] | undefined;
+  columnSizes: string[];
```
bdow, line 942: Now always defined no matter the table type.

src/components/Table/GridTable.tsx
stephenh: This seems like too much of a hack / non-feature imo, and I've been hoping we could figure out a way to do this without the table being aware of the side panel being open next to it.

I.e. could we accomplish this with some sort of SidePanelContainer component? That manages the left child to always be 100% of the main width, and then when the side panel opens, "do something" to achieve the look & feel w/o whatever content is in the left child (i.e. the table in this instance) being none the wiser?

Even if SidePanelContainer had to have its own resize observer, i.e. it used the same trick of "capture width as a pixel and fix my left child's width to _always_ be that same px width, even after the right panel opens", I think I'd prefer to have the two separate resize observers (one in SidePanelContainer and one in GridTable) b/c they just seem like orthogonal concerns.

If that seems doable, could we leave resizeTarget out of GridTable and let Schedules V2 keep doing it's local hack until we have a SidePanelContainer (or whatever a better name is) solution? Or do we really need to do `resizeTarget` now and we could follow up with removing it when/if we have a dedicated SidePanelContainer solution?

src/components/Table/GridTable.tsx
stephenh: Yeah, seems like 1-re-render-every-100ms while they're dragging things around is really pretty reasonable. Definitely a lot better than 1-re-render-every-pixel-move-of-the-mouse. :-)

src/components/Table/GridTable.tsx
```diff
@@ -284,11 +292,26 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     return rows;
   }, [columns, rows, sorting, sortState]);
 
-  const columnSizes = useMemo(
-    () => calcColumnSizes(columns, style.nestedCards?.firstLastColumnWidth),
-    [columns, style.nestedCards?.firstLastColumnWidth],
+  const [columnSizes, setColumnSizes] = useState<string[]>(
```
stephenh, line 295: Can we in-source comment this with like `// Calc our initial/first render sizes where we won't have a width yet`?

src/components/Table/GridTable.tsx
```diff
@@ -284,11 +292,26 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
+    calcColumnSizes(columns, style.nestedCards?.firstLastColumnWidth, undefined, style.minWidth),
   );
 
+  const [tableWidth, setTableWidth] = useState<number | undefined>();
```
stephenh, line 299: Maybe put this before `const [columnSizes, setColumnSzies] = ` so that we could pass `tableWidth`, even it's initial `undefined` value, to `calcColumnSizes`? Really minor just thinking it might be nice to always see that 3rd param is `tableWidth`.

src/components/Table/GridTable.tsx
```diff
@@ -284,11 +292,26 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
+
+  const debouncedTableWidth = useDebouncedCallback((width) => {
```
stephenh, line 301: Maybe call this `setTableWidthDebounced`? Pretty neat, I just didn't get it was a setter.

src/components/Table/GridTable.tsx
```diff
@@ -682,7 +717,12 @@ const VirtualRoot = memoizeOne<
  * Calculates column widths using a flexible `calc()` definition that allows for consistent column alignment without the use of `<table />`, CSS Grid, etc layouts.
  * Enforces only fixed-sized units (% and px)
  */
-export function calcColumnSizes(columns: GridColumn<any>[], firstLastColumnWidth: number | undefined): string[] {
+export function calcColumnSizes(
+  columns: GridColumn<any>[],
+  firstLastColumnWidth: number | undefined,
+  tableWidth?: number,
```
stephenh, line 723: Small preference for `tableWidth: number | undefined` since I think we do want callers to always pass `tableWidth`, it just may be `undefined`

src/components/Table/nestedCards.tsx
```diff
@@ -243,12 +240,11 @@ export function makeSpacer(height: number, openCards: string[], styles: NestedCa
 
 interface ChromeRowProps {
   chromeBuffer: ChromeBuffer;
-  columns: number;
 }
-export function ChromeRow({ chromeBuffer, columns }: ChromeRowProps) {
+export function ChromeRow({ chromeBuffer }: ChromeRowProps) {
   return (
     // We add 2 to account for our dedicated open/close columns
```
stephenh, line 246: Can remove the comment?

src/components/Table/GridTable.tsx
```diff
@@ -712,6 +752,16 @@ export function calcColumnSizes(columns: GridColumn<any>[], firstLastColumnWidth
 
   // This is our "fake but for some reason it lines up better" fr calc
   function fr(myFr: number): string {
+    // If the tableWidth, then return a pixel value
+    if (tableWidth) {
+      const widthBasis = tableWidth > tableMinWidthPx ? tableWidth : tableMinWidthPx;
```
stephenh, line 757: Maybe `const widthBasis = Math.max(tableWidth, tableMinWidthPx)`

src/components/Table/GridTable.tsx
stephenh: Also, just super-naive question, but if on the 1st render, we just output a dummy div, and used it to just measure the width we have available, and then immediately re-render, would the user really notice? And would that be simpler than having a dual "we don't know the width" / "actually know we do know the width"? Or is that sort of bookkeeping net/net more complex?

src/components/Table/GridTable.tsx
stephenh: (All ^ said, I'm fine merging this as-is to GSD, and keeping it on a todo list to clean up if my nudge towards a SidePanelContainer is actually doable/wouldn't be that bad.)

stephenh (approved):
Nice! Pretty small changes for the big lift we get in staying w/current approach.

src/components/Table/GridTable.tsx
bdow: Yea, I tried that "dummy div" idea quickly, but was running into errors. I definitely think there is room for improvement to not render the whole table just to get the width.

src/components/Table/GridTable.tsx
bdow: Yea, I haven't thought too much about including a SidePanelContainer directly in GridTable, but definitely something that we might want. The Documents page in Blueprint has a persistent drawer open, which could benefit from this. Though, the big difference between that Documents and the Schedule Table drawers would be the Schedule's table needs to keep its width when the drawer opens, while the Documents table would try and avoid a horizontal scrollbar while the drawer is open. 
I would prefer to keep this as is for now and I'll update the comment to discourage people from using it.

src/components/Table/GridTable.tsx
bdow: The error happens when using Virtuoso: 
```
Rendered more hooks than during the previous render.
```
Haven't dug in any deeper.

src/components/Table/GridTable.tsx
stephenh: > SidePanelContainer directly in GridTable

Just checking, but "in GridTable" --> "in Beam" is what I was thinking, such that GridTable wouldn't have to know whether it is/is not in a side panel-based UX.

> Schedule's table needs to keep its width when the drawer opens, while the
> Documents table would try and avoid a horizontal scrollbar while the drawer is open.

Hopefully we could get UX to just pick one, or yeah we end up with a `SidePanelContainer` that has a flag about which style of UX it uses, and then either way each page's main GridTable is blithely unaware. As the goal anyway.

> I would prefer to keep this as is for now and I'll update the comment
> to discourage people from using it.

Cool, sgtm.

bdow: Gotcha - I was assuming you were referencing a side panel that GridTable is aware of / the side pane was an option of GridTable. That is how its been poised from Design, at least on the Documents page: "Table with persistent drawer". Though, I agree with you and would prefer to keep them separate.

PR #464, opened 2022-03-07
fix: Do not debounce the column calculations on the first render

bdow:


src/components/Table/GridTable.tsx
```diff
@@ -292,24 +292,41 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     return rows;
   }, [columns, rows, sorting, sortState]);
 
+  const calculateImmediately = useRef<boolean>(true);
```
stephenh, line 295: I think the rationale of why we're doing this would be pretty opaque to me as a maintainer; could you in-source document something along the lines of "storybook/mockdate/etc"

Also, does this mean we won't be able to test / futz with scroll / resizing behavior in storybook? Or I guess we'd have to remember to turn off mock date while doing so? That's not too bad, once maintainers remember "oh right, comment this out" to specifically test resizing, which should hopefully be pretty rare, which is good.

stephenh (approved):

src/components/Table/GridTable.tsx
bdow: Yes, anything using `new Date()`, or `Date.now()` would need to turn off mock date.

PR #465, opened 2022-03-10
feat: Add beamNestedFixedStyle.

stephenh:
The new `beamFixedStyle` didn't work with our old `indent: ...` way of applying "out-of-the-box" parent/child styling, and instead used manually applied row styles like:

```
const rowStyles = useMemo(() => ({ parent: { cellCss: beamGroupRowStyle } }), []);
<GridTable<BeamNestedRow>
  style={beamFlexibleStyle}
  sorting={{ on: "client" }}
  columns={beamNestedColumns}
  rows={beamNestedRows}
  rowStyles={rowStyles}
  stickyHeader
/>
```

This was fine, but I wanted something more out-of-the-box, so this adds a specific `beamNestedFixedStyle` that uses a new `GridStyle.levels` map of level to style, so the usage is now:

```
<GridTable<BeamNestedRow>
  style={beamNestedFlexibleStyle}
  sorting={{ on: "client" }}
  columns={beamNestedColumns}
  rows={beamNestedRows}
  stickyHeader
/>
```

I.e. the extra `rowStyles` prop and `useMemo` lines went away.

Granted, we only do two-levels of style at the moment, but I imagine that's what most of our use cases will be; use cases with three levels can define their own `GridStyle.levels` entries.

I did briefly consider having only a single `beamFixedStyle` with `levels` that were _conditionally_ used based on the presence of "does this table have parent/children rows in it? if so, use the levels, if not just ignore it".

I still kinda like the automatic-ness of that, but for now going with the more explicit option that isn't based on "are there parent/child rows being passed to us?" and instead requires an explicit `style=beamNestedFixedStyle`.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -84,6 +84,8 @@ export interface GridStyle {
     Pick<PresentationContextProps, "wrap">;
   /** Minimum table width in pixels. Used when calculating columns sizes */
   minWidthPx?: number;
+  /** Css to apply at each level of a parent/child nested table. */
+  levels?: Record<number, { cellCss: Properties }>;
 }
 
 export type NestedCardStyleByKind = Record<string, NestedCardStyle>;
@@ -337,7 +339,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     // Break up "foo bar" into `[foo, bar]` and a row must match both `foo` and `bar`
     const filters = (filter && filter.split(/ +/)) || [];
 
-    function makeRowComponent(row: GridDataRow<R>): JSX.Element {
+    function makeRowComponent(row: GridDataRow<R>, level: number): JSX.Element {
       // We only pass sortState to header rows, b/c non-headers rows shouldn't have to re-render on sorting
       // changes, and so by not passing the sortProps, it means the data rows' React.memo will still cache them.
       const sortProps = row.kind === "header" ? { sorting, sortState, setSortKey } : { sorting };
@@ -359,6 +361,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
               stickyOffset,
               openCards: nestedCards ? nestedCards.currentOpenCards() : undefined,
               columnSizes,
+              level,
               ...sortProps,
             }}
           />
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -891,6 +892,11 @@ function getIndentationCss<R extends Kinded>(
 ): Properties {
   // Look for cell-specific indent or row-specific indent (row-specific is only one the first column)
   const indent = (isGridCellContent(maybeContent) && maybeContent.indent) || (columnIndex === 0 && rowStyle?.indent);
+  if (typeof indent === "number" && style.levels !== undefined) {
```
stephenh, line 895: I included this as a reminder to future me of what style to use instead of `indent` :-D

src/components/Table/Table.qa.stories.tsx
```diff
@@ -106,45 +113,29 @@ type BeamChildRow = { kind: "child"; id: string } & BeamBudgetData;
 type BeamNestedRow = BeamTotalsRow | HeaderRow | BeamParentRow | BeamChildRow;
 
 export function NestedFixed() {
-  const totalsRowStyles = useMemo(() => ({ totals: { cellCss: beamTotalsRowStyle } }), []);
-  const rowStyles = useMemo(() => ({ parent: { cellCss: beamGroupRowStyle } }), []);
   return (
     <div css={Css.mw("fit-content").$}>
+      <GridTable<BeamNestedRow> style={beamTotalsFixedStyle} columns={beamNestedColumns} rows={beamTotalsRows} />
```
stephenh, line 118: @Levi-Huynh this story is a good example of what we'll need to follow for the commitments page...

In particular, for the header we used to embed totals directly into the headers row, but now they want to pull headers out into their own like "pre-header" row.

Which is fine, but just b/c of the way our table internals work, we have to do this thing of "okay, that means two `GridTable`s, one with just the total row, and then one with everything else" (and wrapped in a fit-content div so they line up).

I think eventually we'll either fix this, so the total row can be included directly in the GridTable, or we'll make like a `TableWithTotal` component that just wraps the two separate GridTables internally...

But, for now, this isn't too bad to just copy/paste...

Levi-Huynh (approved):
👍  thanks for the quick update to this

PR #469, opened 2022-03-11
chore: Extract columnSizes.

stephenh:
Starting to poke at refactoring the collapse state, to also support selected state, and wanted to do this clean up first.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -1,4 +1,3 @@
-import { useResizeObserver } from "@react-aria/utils";
 import memoizeOne from "memoize-one";
 import { Observer } from "mobx-react";
 import React, {
@@ -19,6 +18,7 @@ import {
   PresentationFieldProps,
   PresentationProvider,
 } from "src/components/PresentationContext";
+import { useSetupColumnSizes } from "src/components/Table/columnSizes";
 import { createRowLookup, GridRowLookup } from "src/components/Table/GridRowLookup";
 import { GridSortContext, GridSortContextProps } from "src/components/Table/GridSortContext";
 import { getNestedCardStyles, isLeafRow, NestedCards, wrapCard } from "src/components/Table/nestedCards";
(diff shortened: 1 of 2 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -294,45 +293,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     return rows;
   }, [columns, rows, sorting, sortState]);
 
-  // Calculate the column sizes immediately rather than via the `debounce` method.
-  // We do this for Storybook integrations that may use MockDate. MockDate changes the behavior of `new Date()`, which is used by `useDebounce` and essentially turns off the callback.
-  const calculateImmediately = useRef<boolean>(true);
```
stephenh, line 299: I moved all of this code, unchanged, into the `SetupColumnSizes` helper method.

See the comment disclaimer on `SetupColumnSizes` but eslint wants to enforce "hooks can always be called from components", and so we're pretending that `SetupColumnSizes` is a component, vs. one-by-one adding eslint ignore rules to each hook usage in the method.

src/components/Table/columnSizes.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * So we've just got with essentially fixed/deterministic widths, i.e. `px` or `percent` or
+ * `fr`.
+ *
+ * Disclaimer that we roll our own `fr` b/c we're not in CSS grid anymore.
+ *
+ * Also, disclaimer that the capital-S `SetupColumnSizes` name is terrible, but it placates
+ * the eslint rule about "hooks can only be used in components", which has good intentions,
+ * but because this is deterministically called from GridTable, we still follow the rules
+ * of hooks.
```
TylerR909, line 28: Why not just call it useSetupColumnSize()? Looks like a custom hook. Smells like a custom hook. Just call it a custom hook 🤔

stephenh: Lol, b/c I forgot about that. Yay code previews :-D

stephenh: @TylerR909 fixed!

DeanGilewicz (approved):

PR #471, opened 2022-03-11
chore: Fix wonkiness in initial RowState PR.

stephenh:
See inline PR comments.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -44,7 +44,13 @@ type ChildRow = { kind: "child"; id: string; name: string };
 type GrandChildRow = { kind: "grandChild"; id: string; name: string };
 type NestedRow = HeaderRow | ParentRow | ChildRow | GrandChildRow;
 
+// I tried https://github.com/keiya01/react-performance-testing#count-renders but
+// it didn't work with our fake timers, so this is easier for now.
+let renderedNameColumn: string[] = [];
+beforeEach(() => (renderedNameColumn = []));
+
 // And two columns for NestedRow
+// TODO Move this to the bottom of the file in it's own PR
 const nestedColumns: GridColumn<NestedRow>[] = [
   {
     header: (row) => <Collapse id={row.id} />,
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.test.tsx
```diff
@@ -44,7 +44,13 @@ type ChildRow = { kind: "child"; id: string; name: string };
 type GrandChildRow = { kind: "grandChild"; id: string; name: string };
 type NestedRow = HeaderRow | ParentRow | ChildRow | GrandChildRow;
 
+// I tried https://github.com/keiya01/react-performance-testing#count-renders but
+// it didn't work with our fake timers, so this is easier for now.
+let renderedNameColumn: string[] = [];
+beforeEach(() => (renderedNameColumn = []));
```
stephenh, line 50: Adding some super-basic re-render tracking, so we can assert against re-rendering behavior.

Granted, this assumes we can instrument our components with code that manually writes to a global variable, which we probably shouldn't do in production code per se, but for this it works just fine.

src/components/Table/GridTable.tsx
```diff
@@ -292,14 +292,14 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
 
   const columnSizes = useSetupColumnSizes(style, columns, tableRef, resizeTarget);
 
+  // Make a single copy of our current collapsed state, so we'll have a single observer.
+  const collapsedIds = useComputed(() => rowState.collapsedIds, [rowState]);
```
stephenh, line 296: The wonkiness fix is moving this to it's own `useComputed`, and moving the other `useComputed` back to a `useMemo`.

The issue was that, just the way `useComputed` works today, when mobx detects its value is dirty, i.e. it's observer is invoked, `useComputed`'s lambda is immediately run ... where ever that observer was triggered, i.e. in some browser "onclick" callback for collapsing the rows.

Usually that's fine, but the current `useComputed`-now-back-to-`useMemo` does quite a bit of stuff, in particular instantiation of JSX `<GridRow... />` lines that expect to only be done within the context of a React render pass so that things like context propagation works properly.

In particular, the wonkiness was that the `useComputed` as before would actually get invoked, but the updated state, i.e. the new filtered rows, wouldn't actually make their way into the DOM, so it looked like the click got dropped.

PR #472, opened 2022-03-11
feat: Add selected row support to GridTable.

stephenh:
Leverages the new mobx-based `RowState` to add a map of `rowId` --> `checked | unchecked | partial`.

Uses the `Checkbox` `indeterminate` support to show partially-selected parents as indeterminate.

Handles clicking a parent --> make all children equally selected/unselected.

Handles clicking a child --> recalc it's parents "are you all/none/partially selected", recursively.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1,4 +1,5 @@
 import React, { MutableRefObject, useContext } from "react";
+import { selectColumn } from "src/components/Table/columns";
 import { GridRowLookup } from "src/components/Table/GridRowLookup";
 import {
   calcColumnSizes,
@@ -8,6 +9,7 @@ import {
   GridRowStyles,
   GridStyle,
   GridTable,
+  GridTableApi,
   matchesFilter,
   setRunningInJest,
 } from "src/components/Table/GridTable";
@@ -21,6 +23,7 @@ import {
 } from "src/components/Table/simpleHelpers";
 import { Css, Palette } from "src/Css";
 import { useComputed } from "src/hooks";
+import { Checkbox } from "src/inputs";
 import { cell, cellAnd, cellOf, click, render, row, rowAnd } from "src/utils/rtl";
 
 // Most of our tests use this simple Row and 2 columns
@@ -58,6 +61,12 @@ const nestedColumns: GridColumn<NestedRow>[] = [
     child: (row) => <Collapse id={row.id} />,
     grandChild: () => "",
   },
+  selectColumn<NestedRow>({
+    header: (row) => ({ content: <Select id={row.id} /> }),
+    parent: (row) => ({ content: <Select id={row.id} /> }),
+    child: (row) => ({ content: <Select id={row.id} /> }),
+    grandChild: (row) => ({ content: <Select id={row.id} /> }),
+  }),
   {
     header: () => "Name",
     parent: (row) => ({
@@ -913,9 +922,9 @@ describe("GridTable", () => {
     // When we filter by 'foo'
     const r = await render(<GridTable filter="foo" columns={nestedColumns} rows={rows} />);
     // Then we show the header
-    expect(cell(r, 0, 1)).toHaveTextContent("Name");
+    expect(cell(r, 0, 2)).toHaveTextContent("Name");
     // And the child that matched
-    expect(cell(r, 1, 1)).toHaveTextContent("foo");
+    expect(cell(r, 1, 2)).toHaveTextContent("foo");
     // And that's it
     expect(row(r, 2)).toBeUndefined();
   });
@@ -935,13 +944,13 @@ describe("GridTable", () => {
     ];
     const r = await render(<GridTable<NestedRow> columns={nestedColumns} rows={rows} />);
     // And all three rows are initially rendered
-    expect(cell(r, 0, 1)).toHaveTextContent("parent 1");
-    expect(cell(r, 1, 1)).toHaveTextContent("child p1c1");
-    expect(cell(r, 2, 1)).toHaveTextContent("grandchild p1c1g1");
+    expect(cell(r, 0, 2)).toHaveTextContent("parent 1");
+    expect(cell(r, 1, 2)).toHaveTextContent("child p1c1");
+    expect(cell(r, 2, 2)).toHaveTextContent("grandchild p1c1g1");
     // When the parent is collapsed
     click(cell(r, 0, 0).children[0] as any);
     // Then only the parent row is still shown
-    expect(cell(r, 0, 1)).toHaveTextContent("parent 1");
+    expect(cell(r, 0, 2)).toHaveTextContent("parent 1");
     expect(row(r, 1)).toBeUndefined();
   });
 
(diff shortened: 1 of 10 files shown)


src/components/Table/GridTable.test.tsx
```diff
@@ -913,9 +922,9 @@ describe("GridTable", () => {
     // When we filter by 'foo'
     const r = await render(<GridTable filter="foo" columns={nestedColumns} rows={rows} />);
     // Then we show the header
-    expect(cell(r, 0, 1)).toHaveTextContent("Name");
+    expect(cell(r, 0, 2)).toHaveTextContent("Name");
```
stephenh, line 925: With a new column for testing "select", some existing assertions moved their column over by 1.

src/components/Table/GridTable.test.tsx
```diff
@@ -1036,6 +1045,48 @@ describe("GridTable", () => {
...
+        ...{ kind: "parent", id: "p1", name: "parent 1" },
+        children: [
+          {
+            ...{ kind: "child", id: "p1c1", name: "child p1c1" },
+            children: [{ kind: "grandChild", id: "p1c1g1", name: "grandchild p1c1g1" }],
+          },
+        ],
+      },
+    ];
+    const api: MutableRefObject<GridTableApi | undefined> = { current: undefined };
```
stephenh, line 1062: In production code, this `api` prop would be a `useRef` that the page passes into `GridTable` to "get back" a pointer to API methods it can call to ask/interrogate `GridTable` about various things, i.e. what is your current list of selected rows.

Note that I don't have an `onRowsSelected` callback yet...should be easy to add...will probably do as a separate PR...

src/components/Table/GridTable.tsx
```diff
@@ -959,7 +966,7 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
           .filter((style) => style)
       : undefined;
 
-  const rowStyleCellCss = maybeApplyFunction(row, rowStyle?.cellCss);
+  const rowStyleCellCss = maybeApplyFunction(row as any, rowStyle?.cellCss);
```
stephenh, line 969: The `IfAny` I did to "fix" `GridDataRow<any>` broke these two other spots, just doing an `as any` on them for now.

src/inputs/Checkbox.tsx
```diff
@@ -6,17 +6,12 @@ import { CheckboxBase } from "src/inputs/CheckboxBase";
 export interface CheckboxProps {
   label: string;
   checkboxOnly?: boolean;
+  selected: boolean | "indeterminate";
```
stephenh, line 9: So, dunno, I always thought it was kind of weird that "indeterminate" was considered orthogonal to selected, b/c afaiu a checkbox really can be only 1 of "selected yes", "selected no", or "indeterminate".

So I nudged it that way, by removing the `indeterminate` prop and instead having callers pass `selected="indeterminate"` to highlight it doesn't really mean anything to be "selected _and_ indeterminate".

TylerR909: Doesn't really sit right with me to have a `boolean | "someString"` type but it doesn't appear that an Enum actually simplifies anything.

src/inputs/Checkbox.tsx
```diff
@@ -26,18 +21,14 @@ export interface CheckboxProps {
...
-    label,
-    indeterminate: isIndeterminate = false,
-    disabled: isDisabled = false,
-    selected,
-    ...otherProps
-  } = props;
-  const ariaProps = { isSelected: selected, isDisabled, isIndeterminate, ...otherProps };
+  const { label, disabled: isDisabled = false, selected, ...otherProps } = props;
+  // Treat indeterminate as false so that clicking on indeterminate always goes --> true.
+  const isSelected = selected === "indeterminate" ? false : selected;
```
TylerR909, line 26: Same as `selected === true`?

src/inputs/Checkbox.stories.tsx
```diff
@@ -16,14 +16,16 @@ export function Checkboxes() {
...
             onBlur={action("onBlur")}
             label="Default"
           />
-          <Checkbox onChange={action("onChange")} selected label="Selected" />
-          <Checkbox onChange={action("onChange")} indeterminate label="Indeterminate" />
-          <Checkbox onChange={action("onChange")} disabled label="Disabled" />
+          <Checkbox onChange={action("onChange")} selected={false} label="Selected" />
+          <Checkbox onChange={action("onChange")} selected="indeterminate" label="Indeterminate" />
+          <Checkbox onChange={action("onChange")} selected={false} disabled label="Disabled while unselected" />
+          <Checkbox onChange={action("onChange")} selected={true} disabled label="Disabled while selected" />
```
TylerR909, line 28: Where do we expect `selected` will be coming from? A database field? Manually set as a prop by a developer? Form State?

Just wondering how we programmatically get from some value to `true | false | "indeterminate"`

DeanGilewicz (approved):
Looks good - I think a nice use of `indeterminate` in this case to represent "partially" selected (some but not all children selected).

src/components/Table/RowState.ts
```diff
@@ -37,6 +35,82 @@ export class RowState {
...
+      // Everything here & down is deterministically on/off
+      const map = new Map<string, SelectedState>();
+      const todo = [curr.row];
+      while (todo.length > 0) {
+        const row = todo.pop()!;
+        map.set(row.id, selected ? "checked" : "unchecked");
+        if (row.children) {
+          todo.push(...row.children);
+        }
+      }
```
TylerR909, line 94: Dang. If only we could tell Array.prototype.flat() what key to flatten on this could be so much more succinct. 

```js
const map = new Map<string, SelectedState>();
curr.row.flatIshMap((row) => map.set(row.id, selected ? "checked" : "unchecked"));
```

Not generally a fan of while/for loops as they're more imperative than declarative. But, they work, so can't really fuss too much.

I do feel like a helper function could assist?

```js
function traverseWithChildren(items, fn, parent) {
  items.forEach((item) => {
    fn(item);
    if (item.children) traverseWithChildren(item.children, fn, item);
  });
}

traverseWithChildren(curr.row,
  (item) => map.set(row.id, selected ? "checked" : "unchecked"),
);
```

TylerR909 (approved):

src/components/Table/RowState.ts
stephenh: Ah nice, good call!

When writing the 2nd `while`, I thought of that, but swear that it was "different from the other `while`" at first; but turns out it either never was, or at least is not anymore.

So I added a `visit` method giving the "visitor pattern" is exactly what this sort of "please traverse this complex (okay, for us just a tree) data structure for me, and just call my function as necessary kthx" pattern is typically called:

```
function visit(rows: GridDataRow<any>[], fn: (row: GridDataRow<any>) => void): void {
  const todo = [...rows];
  while (todo.length > 0) {
    const row = todo.pop()!;
    fn(row);
    if (row.children) {
      todo.push(...row.children);
    }
  }
}
```

Pushed up.

src/inputs/Checkbox.stories.tsx
stephenh: Yeah, it will nearly always come from a database boolean field of like "is this enabled yes/no".

I.e. I don't think I've ever seen indeterminate used for anything other than "partially selected rows in a table of parent/child selectable rows" :-) so I'm not too worried about how it would show up in other "just a regular CRUD form" places.

src/inputs/Checkbox.tsx
stephenh: Ah yeah, I admit "mixing separate primitive types in a single union type" is maybe slightly esoteric functional programming sort of a thing, but agreed it's net/net what an enum would be, but generally same (if not more) simplicity, especially for the ~90% case of "it is really just either true/false".

src/inputs/Checkbox.tsx
stephenh: Yep, will change

PR #473, opened 2022-03-12
feat: SC-12556 add line to modal

aTrujilloHomeBound:


src/components/Modal/Modal.tsx
```diff
@@ -39,7 +41,7 @@ export type ModalApi = {
  * Provides underlay, modal container, and header. Will disable scrolling of page under the modal.
  */
 export function Modal(props: ModalProps) {
-  const { size = "md", content, forceScrolling, api } = props;
+  const { size = "md", content, forceScrolling, api, drawHeaderBorder} = props;
```
bdow, line 44: **nit:** You could default this to `..., drawHeaderBorder = false } = props; `. Then your below if statement doesn't need to do the double negation operator in `if(!!drawHeaderBorder).bb.bGray200.$`

aTrujilloHomeBound:
@bdow 

I believe Implemented this correctly please let me know if you have any thoughts, feels, concerns

src/components/Modal/Modal.stories.tsx
```diff
@@ -59,6 +60,25 @@ export const OpenModalTest = () => {
   );
 };
 
+export const drawHeaderBorderTest = () => {
```
bdow, line 63: Since you have `WithDrawHeaderBorder` above, I'd suggest we remove this `drawHeaderBorderTest` story. They are basically validating the same thing.

bdow (approved):

PR #474, opened 2022-03-12
fix(rich-text-field): async init had issues

TylerR909:
## Live Bug
Working example of the issue:

* Spam refresh on this page and the text will nondeterministically pop in and out of the textboxes https://blueprint.dev-homebound.com/development-commitments/dc:197?filter=%7B%7D&tab=scheduleAndScope

## Context/Issue
Some combination of Async Initialization and stale closures were leading to cases where, more often than not, either `editor.current` was undefined, or `value` was undefined.

* The useEffect would bail due to an undefined `editor.current` (After triggering from an external `value` update)
* And `value` would be `undefined` during the onEditorInit block

## The Chosen Fix
This change switches `editor` from a Ref to State so that **onEditorInit can trigger a rerender**, at which point the useEffect can do its job and re-sync the `value`

## Other Possible Fixes
We're abusing useMemo, which is intended to calculate a value, not induce side-effects. `onEditorInit` should be broken out into its own slice of code, potentially with a [useEventListener](https://usehooks.com/useEventListener/) hook or a useEffect.

### DIFF
--- a/src/inputs/RichTextField.test.tsx
+++ b/src/inputs/RichTextField.test.tsx
@@ -1,0 +1,16 @@
+import { render } from "@homebound/rtl-utils";
+import { noop } from "src/utils";
+import { RichTextField } from "./RichTextField";
+
+describe(RichTextField, () => {
+  it("renders", async () => {
+    const r = await render(<RichTextField value="<div>test</div>" onChange={noop} />);
+    expect(r.getByText("test")).toBeInTheDocument();
+  });
+
+  it("rehydrates if data populates", async () => {
+    const r = await render(<RichTextField value={""} onChange={noop} />);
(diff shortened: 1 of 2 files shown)


stephenh (approved):
Nice!

src/inputs/RichTextField.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { render } from "@homebound/rtl-utils";
+import { noop } from "src/utils";
+import { RichTextField } from "./RichTextField";
+
+describe(RichTextField, () => {
```
stephenh, line 5: Huh, I didn't realize you could pass a non-string to this; that's neat

PR #476, opened 2022-03-16
feat: Export SelectToggle, make selectColumn provide by default.

stephenh:
Export `SelectToggle` so that it can be used directly downstream in Blueprint.

Also make the `selectColumn` smarter such that, unless users need to override specific row kinds (i.e. for a totals row that is not selectable), then merely `selectColumn()` will do the right thing and provide a `SelectToggle` for each of the table's row kinds.

...

I also did the same thing for `collapseColumn`.

And also added a `GridTableApi.getSelectedRows` (in addition to the current `GridTableApi.getSelectedRowIds`).

### DIFF
--- a/src/components/Table/SelectToggle.tsx
+++ b/src/components/Table/SelectToggle.tsx
@@ -1,0 +1,21 @@
+import React, { useContext } from "react";
+import { RowStateContext } from "src/components/Table/RowState";
+import { useComputed } from "src/hooks/index";
+import { Checkbox } from "src/inputs/index";
+
+/** Provides a checkbox to show/drive this row's selected state. */
+export function SelectToggle({ id }: { id: string }) {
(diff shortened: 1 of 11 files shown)


src/components/Table/SelectToggle.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { useContext } from "react";
+import { RowStateContext } from "src/components/Table/RowState";
+import { useComputed } from "src/hooks/index";
+import { Checkbox } from "src/inputs/index";
+
+/** Provides a checkbox to show/drive this row's selected state. */
+export function SelectToggle({ id }: { id: string }) {
```
stephenh, line 7: Extracted from the Table.qa storybook

Levi-Huynh (approved):
looks good!

PR #477, opened 2022-03-22
feat: GridTable support sticky columns and cells

bdow:


src/components/Table/GridTable.tsx
```diff
@@ -1045,6 +1055,33 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
         ensureClientSideSortValueIsSortable(sorting, isHeader, column, columnIndex, maybeContent);
         const maybeNestedCardColumnIndex = columnIndex + (style.nestedCards ? 1 : 0);
 
+        const maybeSticky = ((isGridCellContent(maybeContent) && maybeContent.sticky) || column.sticky) ?? undefined;
+        const maybeStickyColumnStyles =
+          maybeSticky && columnSizes
+            ? {
+                ...Css.sticky.z1.$,
```
stephenh, line 1062: Some of the "sticky columns _and_ headers" examples I'd seen get kinda complicated in their `zIndex` tracking; do we need to do something similar, like a storybook with sticky columns and sticky header? Or does it just work?

src/components/Table/nestedCards.tsx
```diff
@@ -214,6 +214,27 @@ export function getNestedCardStyles(
...
+ * Returns the background color for the row.
+ * Either based on the current row's card having a defined background color, or the last opened card's background color
+ * Defaults to White.
+ */
+export function getCurrentBgColor(
+  row: GridDataRow<any>,
+  openCardStyles: NestedCardStyle[] | undefined,
+  nestedKinds: NestedCardStyleByKind,
+): string {
+  if (nestedKinds[row.kind]) {
```
stephenh, line 227: Huh, is it possible to use nested cards w/o having a `nestedKind` entry for _every_ kind?

bdow: Yes it is.

src/components/Table/GridTable.tsx
bdow: Yea, that makes sense. I will add a storybook example with a stick column and sticky header. Perhaps can use the "play" functionality to scroll to the bottom or something like that to ensure we see the sticky-on-sticky in action.

stephenh (approved):

src/components/Table/GridTable.tsx
bdow: Done and done. We had to update the z-index of the sticky header to z2, so thanks for that call out! Added a storybook snapshot for this as well.

stephenh: Sweet!

PR #479, opened 2022-03-22
feat: SC-13751: Compact IconButton style

bdow:


src/components/IconButton.tsx
```diff
@@ -31,6 +33,7 @@ export function IconButton(props: IconButtonProps) {
     tooltip,
     menuTriggerProps,
     openInNew,
+    compact = false,
```
stephenh, line 36: Should we check the presentation settings, i.e. icons in grid tables auto-compact? Or is that not desirable? I guess probably not...

stephenh (approved):

src/components/IconButton.tsx
bdow: Good question. I don't know of a case right now where the button size is dependent on the presentation settings or where its located. I'd add presentation setting support once we need it, or see it is necessary.

stephenh: Yep, sounds good

aTrujilloHomeBound (approved):
great work!

PR #481, opened 2022-03-23
feat: GridTable support active row styling

bdow:


src/components/Table/GridTable.tsx
```diff
@@ -225,6 +225,12 @@ export interface GridTableProps<R extends Kinded, S, X> {
   api?: MutableRefObject<GridTableApi<R> | undefined>;
   /** Experimental, expecting to be removed - Specify the element in which the table should resize its columns against. If not set, the table will resize columns based on its owns container's width */
   resizeTarget?: MutableRefObject<HTMLElement | null>;
+  /**
+   * Defines which row in the table should be provided with an "active" styling.
+   * Expected format is `${row.kind}_${row.id}`. This helps avoid id conflicts between rows of different types/kinds that may have the same id.
+   * Example "data_123"
+   */
+  activeRowId?: string;
```
stephenh, line 233: Yeah, makes sense re `${kind}_${id}`; fwiw I've been a little lazy about that, i.e. the `getSelectedRowIds` is just the `.id` and not the `kind+id` tuple... And like eventually internally I think I'm assuming the ids are effectively unique, which we, right, should probably not do...

I was wondering if we should have a specific name for the kind+id pair, but maybe that term would actually be "row id", so +1.

---

I'd eventually really like to have this driven by GridTable itself, similar to selected/collapse state, and be something that the user observes from `GridTableApi.activeRow` instead of driving by hand + we can eventually hook up keyboard j/k + up/down things / etc.

And when if they do want to drive it by hand, they could write to `GridTableApi.activeRow = rowThatJustGotClickedOn`...

Or maybe we could use one of our existing contexts to expose like a `const { setActive } = useGridContext()` sort of thing, similar to how a user can toggle selected if they have a custom collapse component they want to use...

That said, unless you like really love the `setActive`-via-a-context approach / want to do that now, I'm good with this for now too.

src/components/Table/GridTable.tsx
```diff
@@ -1109,6 +1120,8 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
           ...(!isHeader && !!style.levels && style.levels[level]?.cellCss),
           // The specific cell's css (if any from GridCellContent)
           ...rowStyleCellCss,
+          // Apply active row styling for non-nested card styles.
+          ...(style.nestedCards === undefined && isActive ? Css.bgLightBlue50.$ : {}),
```
stephenh, line 1124: Maybe use a `GridStyle.activeBgColor` just to pre-emptively make this style-aware/changeable?

stephenh (approved):

src/components/Table/GridTable.tsx
bdow: Just thinking through the context approach more.. it'd the content of our components that utilize the `useGridContext` to be React components and not just return JSX...
So instead of...
```
function columns(setActive: (id: string) => void) {
  return [{ data: (row) => <Button onClick={() => setActive(row.id)} /> }, ....]
}
```
To:
```
const columns = [{ data: (row) => <ButtonWithContext row={row} /> }, ...];

function ButtonWithContext({ row }: { row: Row }) {
  const { setActive } = useGridContext();
  return <Button onClick={() => setActive(row.id)} />;
}
```

I don't really have a strong opinion on either approach. One one hand, it is nice to be able to just return some JSX, but on the other it is also nice not having to pass around StateHook setters.

stephenh: Ah yeah, that's great point re context...we could pass it as an additional param to row functions, i.e.:

```
const columns = [{
  data: (row, api) => <Button onClick=() => api.activeRow = row.id />,
]}
```

That might be nice for other things in `api` as well like `.scrollToWhatever`...?

bdow: Gave this a quick shot, as we chatted about on Slack. Ended up running into perf issues causing all rows to rerender any time the 'active' row changed, even if it was unrelated to the current row. Going to punt on this for now and come back to it after I get some more free time.

bdow: No longer punting on it.. these changes are now available in the latest version of the PR.

src/components/Table/GridTable.stories.tsx
```diff
@@ -950,3 +951,80 @@ export function StickyColumnsAndHeader() {
     </div>
   );
 }
+
+export function ActiveRow() {
+  const api = useRef<GridTableApi<Row> | undefined>();
```
stephenh, line 956: You could remove this `useRef` and use the `data` params?

stephenh: Oh wait, this is a row click, gotcha...

bdow: Yeah, I could. I did it that way for the NestedActiveRow story and wanted to see how it'd work with the `rowStyles.onClick` -> Perhaps we should pass the `api` back in that function too?

stephenh (approved):
Yep, lgtm!

PR #482, opened 2022-03-23
feat: GridTable hover color now constant. No longer support 'GridStyle.rowHoverColor'

bdow:


src/components/Table/GridTable.tsx
```diff
@@ -1006,11 +1003,10 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
...
     // For virtual tables use `display: flex` to keep all cells on the same row. For each cell in the row use `flexNone` to ensure they stay their defined widths
     ...(as === "table" ? {} : Css.relative.df.fg1.fs1.addIn("&>*", Css.flexNone.$).$),
-    ...((rowStyle?.rowLink || rowStyle?.onClick) &&
-      style.rowHoverColor && {
-        // Even though backgroundColor is set on the cellCss (due to display: content), the hover target is the row.
-        "&:hover > *": Css.cursorPointer.bgColor(maybeDarken(rowStyleCellCss?.backgroundColor, style.rowHoverColor)).$,
-      }),
+    ...((rowStyle?.rowLink || rowStyle?.onClick) && {
+      // Even though backgroundColor is set on the cellCss, the hover target is the row.
+      "&:hover > *": Css.cursorPointer.bgLightBlue100.$,
```
stephenh, line 1008: Removing `maybeDarken` is fine, but started to inline-hard-code colors seems like a step backwards imo...

Granted, a ton of our non-table components like `TextField`/etc. hard-color colors all over the place / are not themeable, but, dunno, just given the complexity of grid table + how often "let's prototype this" comes up with tables, I'd like to keep as much as possible in the `GridStyle`.

I.e. we really use `GridTable` for basically "anything that remotely looks like a table", b/c not only to get out-of-the-box UX, but also to just get the good API / code structure for "here's some rows w/data + here's some columns w/UX --> please glue them together", i.e. for use cases where design would not all be thinking "this is a beam fixed table" or "this is a beam flexible table", and so, for whatever reason, want a different color...

src/components/Table/GridTable.tsx
```diff
@@ -1349,10 +1345,6 @@ export function matchesFilter(maybeContent: ReactNode | GridCellContent, filter:
   return false;
 }
 
-function maybeDarken(color: string | undefined, defaultColor: string): string {
```
stephenh, line 1352: I'm personally a little skeptical that a single hover color will end up working long-term, and I thought `maybeDarken` was pretty cute :-), but I get that we'd probably use like a configured-per-level style going forward, so +1.

src/components/Table/GridTable.tsx
bdow: Okay, I'm fine with bringing back the `rowHoverColor`. I just got a little deletion happy.

PR #484, opened 2022-03-25
chore(grid-table): fallback message styling

TylerR909:
Updates fallback message styling 

<details><summary>From...</summary>
image

</details>

To this:

image

I also wanted to update the Fallback message from `string` to `ReactNode` but that started to cascade and I didn't feel it was worth the trouble right now.


stephenh (approved):
:1st_place_medal:

PR #487, opened 2022-03-29
fix(grid-table): correct style util fallback styling

TylerR909:
The updates provided by #484 were getting overridden when used in conjunction with `beamFlexibleStyle`. This fixes that, and slightly shrinks the padding for the Condensed style as well.

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -3,7 +3,6 @@ import { observable } from "mobx";
 import { Fragment, useEffect, useMemo, useRef, useState } from "react";
 import {
   actionColumn,
-  beamFlexibleStyle,
   Button,
   cardStyle,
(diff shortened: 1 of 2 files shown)


src/components/Table/GridTable.stories.tsx
```diff
@@ -3,6 +3,7 @@ import { observable } from "mobx";
 import { Fragment, useEffect, useMemo, useRef, useState } from "react";
 import {
   actionColumn,
+  beamFlexibleStyle,
```
stephenh, line 6: Might be unused?

stephenh (approved):
Thanks!

src/components/Table/GridTable.stories.tsx
TylerR909: Heh, it was while testing. Forgot to autoformat 😅

PR #492, opened 2022-04-04
chore: Bump truss to config-based version.

stephenh:
This follows @KoltonG 's suggestion to use a more standard "just put this in your devDependencies" setup.

There is a slight downside that now `@homebound/truss` brings in a dependency on `ts-node`, so that it can TS-to-JS the `truss-config.ts` on it's own, but realistically most of our projects have `ts-node` anyway, so it really doesn't change anything.

And it does make for a much slicker setup experience than "download this zip file" :-). Thanks @KoltonG !


package.json
```diff
@@ -20,8 +20,7 @@
   "scripts": {
     "start": "yarn storybook",
     "build": "yarn copy && ttsc",
-    "build:truss": "cd ./truss && npm run generate",
-    "watch:truss": "cd ./truss && watch 'npm run generate' ./",
+    "build:truss": "truss",
```
KoltonG, line 23: This is great! So we just need to call truss with a file named `truss-config.ts` and it works!?

KoltonG (approved):
This makes implementing it on the RN side much easier! Thank you!

package.json
stephenh: @KoltonG yep

PR #493, opened 2022-04-05
feat: GridTable improvements

bdow:
Updates include:
- Selected state for parent rows is rederived when child rows change (either by filter or adding new)
- Selecting a parent element only selects visible child rows
- When filtering, parent rows remain in the filtered list if any of their children match the filter. This keeps the nested structure intact.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -972,10 +972,12 @@ describe("GridTable", () => {
     const r = await render(<GridTable filter="foo" columns={nestedColumns} rows={rows} />);
     // Then we show the header
     expect(cell(r, 0, 2)).toHaveTextContent("Name");
+    // And the parent that had a child that matched
+    expect(cell(r, 1, 2)).toHaveTextContent("p1");
     // And the child that matched
-    expect(cell(r, 1, 2)).toHaveTextContent("foo");
+    expect(cell(r, 2, 2)).toHaveTextContent("foo");
     // And that's it
-    expect(row(r, 2)).toBeUndefined();
+    expect(row(r, 3)).toBeUndefined();
   });
 
   it("can collapse parent rows", async () => {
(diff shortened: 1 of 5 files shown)


src/components/Table/RowState.ts
```diff
@@ -161,6 +177,30 @@ export class RowState {
...
+    children: GridDataRow<any>[],
+    map: Map<string, SelectedState>,
+  ): SelectedState[] {
+    return children
+      .filter((row) => row.id !== "header" && this.visibleRows.has(row.id))
+      .map((row) => map.get(row.id) || this.getSelected(row.id));
+  }
+
+  // Recursively traverse through rows to determine selected state of parent rows based on children
+  private setNestedSelectedStates(row: GridDataRow<any>, map: Map<string, SelectedState>): SelectedState[] {
```
bdow, line 191: I thought about trying to consolidate this function and the verbosely named `getVisibleImmediateChildrenSelectedStates` function, but there are times when we don't need to recursively look through children, so I thought its better to keep them separate and only use this recursive function when needed.

src/components/Table/GridTable.test.tsx
```diff
@@ -1158,8 +1160,79 @@ describe("GridTable", () => {
     // And when applying a filter
     type(r.filter, "p1c1g2");
     // Then expect only the visible rows selected are returned
-    expect(api.current!.getSelectedRows()).toEqual([rows[1].children![0].children![1]]);
-    expect(api.current!.getSelectedRowIds()).toEqual(["p1c1g2"]);
+    expect(api.current!.getSelectedRowIds()).toEqual(["p1", "p1c1", "p1c1g2"]);
+    expect(api.current!.getSelectedRows()).toEqual([rows[1], rows[1].children![0], rows[1].children![0].children![1]]);
+  });
+
+  it("re-derives parent row selected state", async () => {
```
stephenh, line 1167: Great tests

src/components/Table/GridTable.tsx
```diff
@@ -359,16 +365,38 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
-  const collapsedIds = useComputed(() => rowState.collapsedIds, [rowState]);
+      if (matches) {
+        return acc.concat([[row, row.children?.reduce(filterRows, []) ?? []]]);
+      }
 
-  // Filter + flatten + component-ize the sorted rows.
-  let [headerRows, filteredRows]: [RowTuple<R>[], RowTuple<R>[]] = useMemo(() => {
-    // Break up "foo bar" into `[foo, bar]` and a row must match both `foo` and `bar`
-    const filters = (filter && filter.split(/ +/)) || [];
+      const isCollapsed = collapsedIds.includes(row.id);
```
stephenh, line 385: Maybe put this chunk into an `else` of the `if matches`? I just initially was getting concerned we were adding `row` to `acc` twice, b/c we do it on line 382 as well as 389, but was missing "oh right", line 382 is an early return.

src/components/Table/GridTable.tsx
```diff
@@ -404,45 +432,34 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
-    // Depth-first to filter
-    function visit(row: GridDataRow<R>, level: number): void {
-      const matches =
-        filters.length === 0 ||
-        row.pin ||
-        filters.every((filter) =>
-          columns.map((c) => applyRowFn(c, row, api)).some((maybeContent) => matchesFilter(maybeContent, filter)),
-        );
-      let isCard = false;
+    function visit(row: ParentChildrenTuple<R>, level: number): void {
```
stephenh, line 435: You could do `[row, children]: ParentChildrenTuple<R>` to clean up some of the `[0]` / `[1]` later

src/components/Table/RowState.ts
```diff
@@ -45,6 +45,23 @@ export class RowState {
...
+      () => {
+        const map = new Map<string, SelectedState>();
+        // go through the rows and re-derive all visible parent states.
+        this.rows.current.forEach((row) => this.setNestedSelectedStates(row, map));
+
+        // Finally re-derive the 'header's state.
+        map.set("header", deriveParentSelected(this.getVisibleImmediateChildrenSelectedStates(this.rows.current, map)));
+
+        // And finally merge the changes back into the selected rows state
+        this.selectedRows.merge(map);
```
stephenh, line 61: Nice! This ended up being pretty short and sweet.

src/components/Table/RowState.ts
```diff
@@ -45,6 +45,23 @@ export class RowState {
...
+    // Whenever our `visibleRows` change (i.e. via filtering) then we need to re-derive header and parent rows' selected state.
+    reaction(
+      () => [...this.visibleRows.values()].sort(),
+      () => {
+        const map = new Map<string, SelectedState>();
+        // go through the rows and re-derive all visible parent states.
+        this.rows.current.forEach((row) => this.setNestedSelectedStates(row, map));
+
+        // Finally re-derive the 'header's state.
+        map.set("header", deriveParentSelected(this.getVisibleImmediateChildrenSelectedStates(this.rows.current, map)));
```
stephenh, line 58: Is this the same as:

```suggestion
        map.set("header", deriveParentSelected(this.rows.current.flatMap((row) => this.setNestedSelectedStates(row, map)));
```

src/components/Table/RowState.ts
stephenh: Yeah, I'm fine keeping two separate functions.

Maybe add an in-source comment that "we return a `SelectedState[]` so that hidden rows don't "have a vote" about the full/partial checked-ness of their parent.

An alternative to returning `SelectedState[]` would be `SelectedState | undefined`, but then we'd have to filter out `undefined` in `derivedParentSelected`.

Or add like an addition `SelectedState` value of `"hidden"` and again teach `derivedParentSelected` to ignore `hidden` rows... like:

```
const allVisibleChecked = children.every(child => child === "hidden" || child === "unchecked');
const allVisibleUnchecked = children.every(child => child === "hidden" || child === "unchecked");
```

...I kind of like the explicitness of adding "hidden", but the current `SelectedState[]` is fine too.

src/components/Table/RowState.ts
```diff
@@ -197,5 +237,5 @@ function findRow(rows: GridDataRow<any>[], id: string): FoundRow | undefined {
 function deriveParentSelected(children: SelectedState[]): SelectedState {
   const allChecked = children.every((child) => child === "checked");
   const allUnchecked = children.every((child) => child === "unchecked");
-  return allChecked ? "checked" : allUnchecked ? "unchecked" : "partial";
+  return children.length === 0 ? "unchecked" : allChecked ? "checked" : allUnchecked ? "unchecked" : "partial";
```
stephenh, line 240: Ah yeah, makes sense

src/components/Table/RowState.ts
```diff
@@ -161,6 +177,30 @@ export class RowState {
...
+    return children
+      .filter((row) => row.id !== "header" && this.visibleRows.has(row.id))
+      .map((row) => map.get(row.id) || this.getSelected(row.id));
+  }
+
+  // Recursively traverse through rows to determine selected state of parent rows based on children
+  private setNestedSelectedStates(row: GridDataRow<any>, map: Map<string, SelectedState>): SelectedState[] {
+    if (this.visibleRows.has(row.id)) {
+      if (!row.children) {
+        return [map.get(row.id) || this.getSelected(row.id)];
```
stephenh, line 194: This could probably be just `return [this.getSelected(row.id)]` b/c I don't think we never put leaf rows into `map`?

src/components/Table/RowState.ts
```diff
@@ -161,6 +177,30 @@ export class RowState {
       localStorage.setItem(this.persistCollapse, JSON.stringify(collapsedIds));
     }
   }
+
+  private getVisibleImmediateChildrenSelectedStates(
```
stephenh, line 181: Fwiw I'd be fine with `getVisibleChildrenStates` just to nudge it a little shorter

stephenh (approved):
Great demo/UX improvements.

src/components/Table/RowState.ts
bdow: Yes, you're right. I will update.

src/components/Table/RowState.ts
bdow: Yea! That works. Thanks for the suggestion.

PR #494, opened 2022-04-06
fix: useComputed object comparison

Levi-Huynh:


bdow (approved):

src/hooks/useComputed.ts
```diff
@@ -1,5 +1,6 @@
 import { autorun, IReactionDisposer } from "mobx";
 import { useMemo, useRef, useState } from "react";
+import equal from "fast-deep-equal";
```
stephenh, line 3: Even though we already have it in `node_modules`, can you add `fast-deep-equal` to `package.json` just to make it explicit that our code directly calls it?

Historically this didn't matter as much, but newer bundlers like vite can get confused w/o this being correct.

Levi-Huynh:
Accidentally used. npm install instead of yarn install to add fast-deep-equal causing the ci/circle build to blow up. closing this and using yarn to install the package

PR #496, opened 2022-04-06
fix: Fix sorting prop may not be stable.

stephenh:
We allow setting `sorting` to a non-stable / inline hash:

```
<GridTable
  columns={columns}
  // And there is no initial sort given
  sorting={{ on: "client" }}
  rows={rows}
/>
```

But had been using `sorting` in several deps arrays.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1,4 +1,4 @@
-import React, { MutableRefObject, useContext, useState } from "react";
+import React, { MutableRefObject, useContext, useMemo, useState } from "react";
 import { selectColumn } from "src/components/Table/columns";
 import { GridRowLookup } from "src/components/Table/GridRowLookup";
 import {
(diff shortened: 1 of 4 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -356,14 +356,14 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   // Make a single copy of our current collapsed state, so we'll have a single observer.
   const collapsedIds = useComputed(() => rowState.collapsedIds, [rowState]);
 
-  const [sortState, setSortKey] = useSortState<R, S>(columns, sorting);
+  const [sortState, setSortKey, sortOn] = useSortState<R, S>(columns, props.sorting);
```
stephenh, line 359: Most places that were depending on `sorting` really just needed `sortOn`, which is a primitive of `"client" | "server" | undefined`.

src/components/Table/GridTable.test.tsx
```diff
@@ -559,6 +572,18 @@ describe("GridTable", () => {
       expect(row(r, 4).getAttribute("data-render")).toEqual("1");
       expect(row(r, 5).getAttribute("data-render")).toEqual("1");
       expect(row(r, 6).getAttribute("data-render")).toEqual("1");
+
+      // And the table re-renders for some other reason
+      click(r.rerenderParent);
+
+      // Then memoization did not break
+      expect(row(r, 0).getAttribute("data-render")).toEqual("2");
```
bdow, line 580: question: Why did the header rerender?

stephenh: `sortState` changed, and right now we don't have cell-level re-rendering of the sort headers. Just the whole row re-renders.

bdow (approved):

PR #497, opened 2022-04-06
fix: Fix visible rows chatter

stephenh:


src/components/Table/Table.qa.stories.tsx
```diff
@@ -103,14 +104,37 @@ export function NestedFixed() {
 }
 
 export function NestedFlexible() {
+  const [api, setApi] = useState<GridTableApi<BeamNestedRow> | null>();
+  const selectedCount = useComputed(() => {
+    return api?.getSelectedRows().map((r) => r.id);
+  }, [api]);
+
+  // const api = useRef<GridTableApi<BeamNestedRow>>();
```
stephenh, line 112: This commented out code was actually not working, b/c when `GridTable` would run and set `api.current = api`, nothing tells the parent component "hey you should re-run".

Which, in the canonical usage of refs, where this is a DOM element, that makes sense b/c you need a `useEffect` to wait until post-1st-render to access the DOM element, and then your `useEffect` lambda can access the element but not _necessarily_ cause a re-render of the whole component.

For our `useComputed` usage, which is using the api as a dep, we actually do want to be told "hey the api is set now, you should re-render"...

Which makes me think we should move away from:

```
const api = useRef<GridTableApi<BeamNestedRow>>();
```

And just use:

```
  const [api, setApi] = useState<GridTableApi<BeamNestedRow> | null>();
```

Everywhere instead.

bdow: Can we move these changes of the story into the `Filterable` story? There is already a "get selected" sort of behavior there and I'd like to keep the `.qa.stories` as free of eng testing functionality as possible.

bdow (approved):

PR #500, opened 2022-04-07
feat: Add useGridTableApi hook as a better ref

stephenh:
We have a chicken-vs-egg condition with our current GridTable ref-based approach:

```
const api = useRef();
// can't actually use the api yet...
const count = useComputed(() => ..., [api]);
// now api will be set, but we'll only know about it after either
// a useEffect or a re-render.
return <GridTable api={api} />;
```

The "refs are only set after DOM elements are available" makes a lot of sense for DOM elements.

But we want the function to have immediate access to the GridTable API, to set up reactivity on things like which rows are selected.

To provide this, we just let the caller instantiate the api directly:

```
const api = useGridTableApi();
// Can immediately use it
const count = useComputed(() => api.selectedRows.length, [api]);
// Pass it to GridTable
return <GridTable api={api} />
```

And have GridTable internally realize, before it goes off and makes "an api" or "a row state", to check to see if the caller passed one in, and use that instead.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -9,10 +9,10 @@ import {
   GridRowStyles,
   GridStyle,
   GridTable,
-  GridTableApi,
   matchesFilter,
   setRunningInJest,
 } from "src/components/Table/GridTable";
+import { GridTableApi, useGridTableApi } from "src/components/Table/GridTableApi";
 import { RowStateContext } from "src/components/Table/RowState";
 import {
   simpleDataRows,
@@ -1136,7 +1136,12 @@ describe("GridTable", () => {
       },
     ];
     const api: MutableRefObject<GridTableApi<NestedRow> | undefined> = { current: undefined };
-    const r = await render(<GridTable<NestedRow> api={api} columns={nestedColumns} rows={rows} />);
+    function Test() {
+      const _api = useGridTableApi<NestedRow>();
+      api.current = _api;
+      return <GridTable<NestedRow> api={_api} columns={nestedColumns} rows={rows} />;
+    }
+    const r = await render(<Test />);
     // And all three rows are initially rendered
     expect(cell(r, 1, 2)).toHaveTextContent("parent 1");
     expect(cell(r, 2, 2)).toHaveTextContent("child p1c1");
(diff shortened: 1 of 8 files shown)


src/components/Table/GridTable.test.tsx
```diff
@@ -1136,7 +1136,12 @@ describe("GridTable", () => {
       },
     ];
     const api: MutableRefObject<GridTableApi<NestedRow> | undefined> = { current: undefined };
```
stephenh, line 1138: Even those pages don't have to use the MutableRef anymore, these tests do so that they can export the api out of the test component and up into the test itself (the test can't do `const api = useGridTableApi()` b/c hooks aren't allowed to be instantiated outside of a component)

src/components/Table/GridTable.tsx
```diff
@@ -248,22 +247,6 @@ export interface GridTableProps<R extends Kinded, S, X> {
   activeRowId?: string;
 }
 
-/** NOTE: This API is experimental and primarily intended for story and testing purposes */
```
stephenh, line 251: Moved to `GridTableApi`

src/components/Table/GridTable.tsx
```diff
@@ -300,46 +283,23 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
     persistCollapse,
-    api: callerApi,
     resizeTarget,
     activeRowId,
   } = props;
 
-  // Create a ref that always contains the latest rows, for our effectively-singleton RowState to use
-  const rowsRef = useRef(rows);
-  rowsRef.current = rows;
-  const [rowState] = useState(() => new RowState(rowsRef, persistCollapse, activeRowId));
```
stephenh, line 311: The `RowState` is now managed as an internal impl detail of `GridTableApi`, so that when the caller creates an api immediately with `const api = useGridTableApi()`, it will already have a `useComputed`-able `RowState` living inside it.

src/components/Table/Table.qa.stories.tsx
```diff
@@ -122,9 +116,8 @@ export function NestedFlexible() {
 // 2. Select individual "Project Items". Note parent is "checked"
 // 3. Remove filter, additional project items under parent are "unchecked", though parent still shows "checked" instead of what it should show as "indeterminate".
 export function Filterable() {
-  // Using a useState so that useComputed works
-  const [api, setApi] = useState<GridTableApi<BeamNestedRow> | null>();
-  const selectedIds = useComputed(() => api?.getSelectedRows().map((r) => r.id) || [], [api]);
+  const api = useGridTableApi<BeamNestedRow>();
+  const selectedIds = useComputed(() => api.getSelectedRows().map((r) => r.id), [api]);
```
stephenh, line 120: This is the usage this approach unlocks: the page can immediately bind a `useComputed` against the `api` and not worry about "api is not available yet" / "I need to trigger a re-render just to see it's set value".

bdow (approved):

PR #503, opened 2022-04-11
fix: Setting activeRowId to be reactive

bdow:


src/components/Table/GridTable.test.tsx
```diff
@@ -1592,6 +1593,29 @@ describe("GridTable", () => {
     // But the 2nd added row did not
     expect(row(r, 2).getAttribute("data-render")).toEqual("1");
   });
+
+  it("reacts to setting activeRowId", async () => {
+    const activeRowIdRowStyles: GridRowStyles<Row> = {
+      data: {
+        onClick: (row, api) => {
+          api.setActiveRowId(`${row.kind}`);
```
stephenh, line 1601: Kinda expect to see row.id in here?

bdow: oh yea.. i wonder why this worked....

bdow: Ahh, because the second color is the same as the first... whoops!

bdow: Okay, all fixed.

stephenh (approved):

PR #508, opened 2022-04-14
fix: Update input field styles in tables

bdow:
Changes include:
- Input fields no longer use box-shadow for focus state when 'borderless'
- New presentation prop for fields: 'errorInTooltip'. Allows the error message to display in a tooltip instead of beneath the field.
- 'borderless' presentation prop is defined for Flexible table styles as well.
- Adds 'internalProps' for TextField to allow for forcing hover and focus states for snapshot purposes.

Adds story for fields in a table in their various states.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -474,6 +474,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
       hideLabel: true,
       numberAlignment: "right",
       compact: true,
+      errorInTooltip: true,
       // Avoid passing `undefined` as it will unset existing PresentationContext settings
       ...(borderless !== undefined ? { borderless } : {}),
       ...(typeScale !== undefined ? { typeScale } : {}),
--- a/src/components/Table/Table.qa.stories.tsx
+++ b/src/components/Table/Table.qa.stories.tsx
@@ -1,3 +1,4 @@
+import { ObjectConfig, ObjectState, required, useFormStates } from "@homebound/form-state";
 import { Meta } from "@storybook/react";
 import { default as React, ReactNode, useMemo, useState } from "react";
 import { Chips } from "src/components/Chips";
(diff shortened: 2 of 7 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -474,6 +474,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
       hideLabel: true,
       numberAlignment: "right",
       compact: true,
+      errorInTooltip: true,
```
stephenh, line 477: Nice

src/components/Table/Table.qa.stories.tsx
```diff
@@ -378,3 +388,188 @@ const flatRows: GridDataRow<BeamRow>[] = [
...
+          const os = getFormState(data);
+          return <BoundTextField field={os.firstName} {...applyStateProps(row.id)} />;
+        },
+      }),
+    }),
+    column<InputFieldRows>({
+      header: "Biography",
+      data: (data, row) => ({
+        content: () => {
+          const os = getFormState(data);
```
stephenh, line 443: Wonder if we should eventually have `textAreaCell` type things that return the content + value `GridCellContents`... Nothing for now, just talking out loud.

src/components/Table/Table.qa.stories.tsx
```diff
@@ -378,3 +388,188 @@ const flatRows: GridDataRow<BeamRow>[] = [
...
+          return flexible ? (
+            <BoundTextAreaField field={os.bio} preventNewLines {...applyStateProps(row.id)} />
```
stephenh, line 445: Can you think of any cute ways to infer `preventNewLines` from `PresentationContext`? Np if not...

stephenh (approved):

src/components/Table/Table.qa.stories.tsx
bdow: We do have a `PresentationSettings.wrap` property that we use, but I don't think thats really all too clear. I'm not sure exactly when else we'd want `preventNewLines` to be applied. Right now it only happens in schedulesV2. It's for when we really just want a text field, but also want the value to wrap. Hmm... was thinking maybe if a user has a `TextField` component within a presentation setting of `wrap: true`, then we'd instead display the `TextAreaField` with `preventNewLines` enabled. But I don't think that is super clear. Not sure if the `wrap` setting is all that flushed out yet and don't really want to pile things on it at the moment. I think the only consumer of `wrap` is the Chips component.

stephenh: Agreed that maybe someday `wrap` would drive this if we end up doing it a lot, but good to hold off for now.

PR #510, opened 2022-04-18
feat: Update to nesting styles for Table component

bdow:
Includes UX for a third level of nesting in GridTable.
Changes include:
- Collapse Toggle to use 'compact' variant if already nested
- Apply 'tiny' type scale for first non-action column for further visual difference between 2nd and 3rd level of nesting.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -335,7 +335,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
         filters.length === 0 ||
         !!row.pin ||
         filters.every((f) =>
-          columns.map((c) => applyRowFn(c, row, api)).some((maybeContent) => matchesFilter(maybeContent, f)),
+          columns.map((c) => applyRowFn(c, row, api, 0)).some((maybeContent) => matchesFilter(maybeContent, f)),
         );
 
       // If the row matches, add it in
@@ -848,8 +848,13 @@ export type GridColumn<R extends Kinded, S = {}> = {
     | string
     | GridCellContent
     | (DiscriminateUnion<R, "kind", K> extends { data: infer D }
-        ? (data: D, row: GridRowKind<R, K>, api: GridTableApi<R>) => ReactNode | GridCellContent
-        : (data: undefined, row: GridRowKind<R, K>, api: GridTableApi<R>) => ReactNode | GridCellContent);
+        ? (data: D, row: GridRowKind<R, K>, api: GridTableApi<R>, level: number) => ReactNode | GridCellContent
+        : (
+            data: undefined,
+            row: GridRowKind<R, K>,
+            api: GridTableApi<R>,
+            level: number,
+          ) => ReactNode | GridCellContent);
 } & {
   /**
    * The column's width.
@@ -870,6 +875,8 @@ export type GridColumn<R extends Kinded, S = {}> = {
   sticky?: "left" | "right";
   /** Prevent column from supporting RowStyle.onClick/rowLink in order to avoid nested interactivity. Defaults to true */
   wrapAction?: false;
+  /** Used as a signal to defer adding the 'indent' styling */
+  isAction?: true;
 };
 
 export const nonKindGridColumnKeys = ["w", "mw", "align", "clientSideSort", "serverSideSortKey"];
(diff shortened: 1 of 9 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -847,8 +847,13 @@ export type GridColumn<R extends Kinded, S = {}> = {
     | string
     | GridCellContent
     | (DiscriminateUnion<R, "kind", K> extends { data: infer D }
-        ? (data: D, row: GridRowKind<R, K>, api: GridTableApi<R>) => ReactNode | GridCellContent
-        : (data: undefined, row: GridRowKind<R, K>, api: GridTableApi<R>) => ReactNode | GridCellContent);
+        ? (data: D, row: GridRowKind<R, K>, api: GridTableApi<R>, level: number) => ReactNode | GridCellContent
```
stephenh, line 850: Kinda wonder if we should use `opts` for the 2nd param:

```
(data: D, opts: { row, api, level }) => ..
```

So that it's easier for callers to get whatever extra opt they want...

Granted, we just did a big API change, but I guess while we're at it? :-)

Wdyt?

src/components/Table/GridTable.tsx
```diff
@@ -905,16 +912,11 @@ export interface RowStyle<R extends Kinded> {
...
   rowStyle: RowStyle<R> | undefined,
-  columnIndex: number,
   maybeContent: ReactNode | GridCellContent,
+  applyRowIndent: boolean,
 ): Properties {
-  // Look for cell-specific indent or row-specific indent (row-specific is only one the first column)
-  const indent = (isGridCellContent(maybeContent) && maybeContent.indent) || (columnIndex === 0 && rowStyle?.indent);
-  if (typeof indent === "number" && style.levels !== undefined) {
-    throw new Error(
-      "The indent param is deprecated for new beam fixed & flexible styles, use beamNestedFixedStyle or beamNestedFlexibleStyle",
```
stephenh, line 915: Hm, I dunno, maybe we could stay with having this defined via `style.levels`?

...I think the idea is that we have three ways of specifying behavior:

* `GridCellContent` --> very cell specific
* `RowStyle` --> based on the kind or data within the row
* `Style.levels` --> "this level is always ..."

And it kinda seems like `Style.levels` would be most what we want here, i.e.:

```
export const beamNestedFixedStyle: GridStyle = {
  ...beamFixedStyle,
  levels: { 3: { indentCss: Css.pl3 } },
};
```

So that a page doesn't have to define their own `RowStyle`s b/c the behavior really isn't based on rows/content.

I guess maybe we want to support all three, for more complex pages like Schedules V2 that are going to intermix regular rows with special behavior rows...

src/components/Table/GridTable.tsx
bdow: Yea, I like that idea. Might was well get it done.

src/components/Table/GridTable.tsx
bdow: Hmm. I get what you're saying. Yea, schedules V2 has different row types at the same `level`, which could hinder this sort of thing, though for that they can always just use rowStyles as is done today to get around it, even without the `indent` property. 
I'm okay with updating this to be something more like your example. Though, rather than calling it `indentCss`, I wonder if it'd be something like, `firstNonActionColumn` and we'd apply both padding & type scale,  `Css.pl3.tiny.$. What do you think?

stephenh: Yeah agreed it'd be great to do `Css.pl3.tiny.$`... Maybe `firstContentColumn`?

src/components/Table/GridTable.tsx
bdow: And done!

src/components/Table/GridTable.tsx
bdow: Updated to follow the suggestion. I think it's much cleaner. Thank you!

src/components/Table/GridTable.tsx
```diff
@@ -1048,10 +1056,17 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
...
   const rowNode = (
     <Row css={rowCss} {...others} data-gridrow {...getCount(row.id)}>
       {columns.map((column, columnIndex) => {
-        const { wrapAction = true } = column;
+        const { wrapAction = true, isAction = false } = column;
+
+        const applyFirstContentColumnStyles = !isAction && !firstContentColumnStylesApplied;
+        if (!firstContentColumnStylesApplied && applyFirstContentColumnStyles) {
+          firstContentColumnStylesApplied = true;
+        }
```
stephenh, line 1069: This might be too cute, but I think:

```
firstContentColumnStyledApplied ||= applyFirstContentColumnStyles;
```

Would replace the `if` statement

src/components/Table/GridTable.tsx
```diff
@@ -1239,12 +1256,13 @@ export function applyRowFn<R extends Kinded>(
   column: GridColumn<R>,
   row: GridDataRow<R>,
   api: GridTableApi<R>,
+  level?: number,
```
stephenh, line 1259: Maybe make this required to make sure we don't have callers that we've forgotten to update?

stephenh (approved):
Cool!

src/components/Table/GridTable.tsx
bdow: Nice! I don't think I've ever used that operator.

PR #511, opened 2022-04-20
fix: TextFieldBase readonly styles to include padding when borderless presentaiton prop is set

bdow:


src/inputs/TextFieldBase.tsx
```diff
@@ -120,6 +120,7 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
         .if(compact)
         .mhPx(compactFieldHeight - maybeSmaller).$,
       ...Css.gray900.if(contrast).white.$,
+      ...(borderless ? Css.px1.$ : {}),
```
stephenh, line 123: Maybe inline comment `// Make read-only fields vertically line up with editable fields in tables`?

stephenh (approved):

PR #513, opened 2022-04-24
fix: Memo on GridDataRow.data instead of the row itself.

stephenh:
This should dramatically increase the effectiveness of our per-row
memoization, given that otherwise basically any re-calc of the
GridTable.rows prop (due to new rows/different group by/etc)
would invalidate every single row, and not just the ones with
changed `data` props.

Note that this is now only possible b/c of our recent PR to double
down on `GridDataRow.data` as required for all tables, and not
interleaving "misc data" directly into the `GridDataRow`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@emotion/jest": "^11.3.0",
     "@emotion/react": "^11.1.5",
     "@homebound/rtl-react-router-utils": "^1.0.3",
-    "@homebound/rtl-utils": "^2.51.0",
+    "@homebound/rtl-utils": "^2.59.3",
     "@homebound/truss": "^1.111.3",
     "@homebound/tsconfig": "^1.0.3",
     "@semantic-release/exec": "^6.0.3",
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1588,6 +1588,18 @@ describe("GridTable", () => {
     expect(row(r, 2).getAttribute("data-render")).toEqual("1");
   });
 
+  it("memoizes rows based on the data attribute", async () => {
+    const [header, row1, row2] = rows;
+    const columns = [nameColumn];
+    // Given a table is initially rendered with 2 rows
+    const r = await render(<GridTable key="a" columns={columns} rows={[header, row1, row2]} />);
+    // When we render with new rows but unchanged data prows
+    r.rerender(<GridTable key="a" columns={columns} rows={[header, { ...row1 }, row2]} />);
+    // Then neither row was re-rendered
+    expect(row(r, 1).getAttribute("data-render")).toEqual("1");
+    expect(row(r, 2).getAttribute("data-render")).toEqual("1");
+  });
+
   it("reacts to setting activeRowId", async () => {
     const activeRowIdRowStyles: GridRowStyles<Row> = {
       data: {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -77,7 +77,7 @@
     "@emotion/jest": "^11.3.0",
     "@emotion/react": "^11.1.5",
     "@homebound/rtl-react-router-utils": "^1.0.3",
-    "@homebound/rtl-utils": "^2.51.0",
+    "@homebound/rtl-utils": "^2.59.3",
```
stephenh, line 80: Had to bump for a fix to `rerender` that we made recently, that was otherwise making each `r.rerender` look like a brand new render.

src/components/Table/GridTable.test.tsx
```diff
@@ -1588,6 +1588,18 @@ describe("GridTable", () => {
...
+  it("memoizes rows based on the data attribute", async () => {
+    const [header, row1, row2] = rows;
+    const columns = [nameColumn];
+    // Given a table is initially rendered with 2 rows
+    const r = await render(<GridTable key="a" columns={columns} rows={[header, row1, row2]} />);
+    // When we render with new rows but unchanged data values
+    r.rerender(<GridTable key="a" columns={columns} rows={[header, { ...row1 }, { ...row2 }]} />);
+    // Then neither row was re-rendered
+    expect(row(r, 1).getAttribute("data-render")).toEqual("1");
+    expect(row(r, 2).getAttribute("data-render")).toEqual("1");
```
bdow, line 1600: whoa, this is awesome!

src/components/Table/GridTable.tsx
```diff
@@ -1180,8 +1181,23 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
...
+ * updated) and technically creates new row instances, but a row's `row.data` points to the
+ * same/unchanged Apollo cache fragment, then we won't re-render that row.
+ *
+ * Note that if you're using virtualization, it can be surprising how unnoticeable broken row
+ * memoization is.
+ */
+// Declared as a const + `as typeof GridRow` to work with generics, see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37087#issuecomment-656596623
+const MemoizedGridRow = React.memo(GridRow, (one, two) => {
+  const { row: row1, ...others1 } = one;
+  const { row: row2, ...others2 } = two;
```
bdow, line 1198: **question:** Where is it that we are memoizing on `data`? From what I can tell here, we are still memoizing on the whole `GridDataRow`, not just `GridDataRow.data` when we destructure `row1` and `row2` out.

stephenh: `shallowEqual` compares keys+values within an object (kinda like props).

So we're treating `row1` (old) and `row2` (new) as "mini-props", and so `shallowEqual` effectively checks, for each key/value in `row1`/`row2`, are they the same? So is `id` the same, `data` the same, etc.

So even w/o an explicit `data1 === data2`, `shallowEqual` is effectively doing that for us, plus checking any other keys within each `GridDataRow` just in case something else changed.

I.e. you're right that we "memoize on the whole `GridDataRow`"--except now that we're memoizing on the key/values _within_ the row, instead of the row's hash/identity itself (which is what happens by default when it was just a `GridRow` prop with all the other `GridRow` props).

bdow: Makes sense. Just the title of the PR threw me off, I guess. But I get what we're doing now. Thanks!

bdow (approved):

src/components/Table/GridTable.tsx
stephenh: Yep! Fwiw I had originally written it to `shallowEquals(row1.data, row2.data)` b/c that does seem the most intuitive, but rewrote it to `shallowEquals(row1, row2)` when I realized that'd work too, and also sanity check that some other row key (...pinned? something like that...) changes on us.

PR #515, opened 2022-04-26
feat: Add ability to show indicators on Date Picker

bdow:


src/components/internal/DatePicker.stories.tsx
```diff
@@ -10,5 +10,5 @@ export default {
 
 export function Default() {
   const [date, setDate] = useState(jan1);
-  return <DatePicker value={date} onSelect={setDate} />;
+  return <DatePicker value={date} onSelect={setDate} indicators={[jan1, jan2, jan29]} />;
```
stephenh, line 13: Wdyt of having a callback sort of UI like `isIndicated(day) => boolean` so we don't have to materialize a full list of dates? Disclaimer I'm not really sure what this is used for, like will this list be small / medium / large? If we know it's always small, this is a great.

src/components/internal/DatePicker.tsx
```diff
@@ -9,10 +9,11 @@ export interface DatePickerProps {
   value?: Date;
   onSelect: (value: Date) => void;
   disabledDays?: Modifier | Modifier[];
+  indicators?: Modifier | Modifier[];
```
stephenh, line 12: Huh, looking at the story, I expected this to be `indicators: DateOnly[]` because we're passing `indicators=[{jan1, jan2]}`...what is a `Modifier`?

bdow: Modifier comes straight from React-Day-Picker. We could have this just be `Date[]` (I don't think Beam knows, or should care about `DateOnly`) and as I don't believe we'd need to have any special modifiers that are provided by React Date Picker outside of a date. 
Modifiers allow special syntax for doing things like a range of dates, specific days of the week, etc...

src/components/internal/DatePicker.stories.tsx
bdow: This was an ask for the Job Logs page. When looking at the calendar, they want this indicator to show which days already have a Job Log attached. So this list could get pretty large as we expect to have a Job Log for _every_ working day.

Great question regarding how this should be used from Blueprint side of things as we probably wouldn't want to front load every date that has a job log. Since the Date picker displays month by month, maybe thats how we can do the API?

```
const [indicatedDates, setDates] = useState(data.logs.map((log) => log.date));
<DatePicker 
  onMonthChange={async (d: Date) => {
    // add some sort of caching to not query again if we already have this month's data.
    const result = await getJobLogsByMonth(); 
    setDates((prev) => prev.concat(result.data.logs.map((log) => log.date)));
  }};
  indicators={indicatedDates}
/>
```
Something like that? The list could still get really large, but hopefully not initially.

src/components/internal/DatePicker.tsx
stephenh: > I don't think Beam knows, or should care about DateOnly

Yeah-ish; fwiw eventually when the new time APIs ship, `PlainDate` would be great to use here in Beam / and in Blueprint.

I guess for now I'd prefer doing `Date[]`s, b/c it keeps the API more "ours" than leaking `react-day-picker`'s types into our API.

Also can we pick a different name than `indicator`? Like...`highlightDays`? `underscoreDays`? Some semantic notion of what we're indicating?

src/components/internal/DatePicker.stories.tsx
stephenh: @zgavin just checking, does this "each time the calendar width's month changes, we'll load the X days within that month have already have a job log"?

Or were you just going to load them all up-front anyway b/c :shrug: it'll be a few hundred and not a big deal...

zgavin: right now there is no pagination and we are just uploading them all up front.

stephenh: Cool, we could ship `indicators?: Date[]` today then, and later follow up with an `indicators?: Date[] | (month: Month) => MaybePromise<Date[]>` overload when/if we need it.

stephenh (approved):

PR #517, opened 2022-04-28
fix: Virutalized table sticky header to apply zIndex of 2

bdow:
This fixes an issue where sticky columns were layering on top of the sticky header when virtualized.
Uses the Virtuoso's 'TopItemList' component to apply styling directly to the TopItemList vs the rest of the list.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -654,6 +654,7 @@ function renderVirtual<R extends Kinded>(
         }
       }}
       components={{
+        TopItemList: VirtualHeader(listStyle, columns, id, firstLastColumnWidth, xss),
         List: VirtualRoot(listStyle, columns, id, firstLastColumnWidth, xss),
         Footer: () => <div css={footerStyle} />,
       }}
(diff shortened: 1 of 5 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -730,12 +727,35 @@ const VirtualRoot = memoizeOne<
...
+        }}
+        data-testid={id}
+      >
+        {children}
+      </div>
+    );
+  });
+});
+
+const VirtualHeader = memoizeOne<
```
bdow, line 744: @stephenh - I stole this `memoizeOne` from the `VirtualRoot` definition, though I am wondering if this is necessary here, or even still necessary in the VirtualRoot? Seems like it was to help with Virtuosos rendering.. but not sure I completely follow what's going on here.

stephenh: Fwiw if they had two props, like topItemComponent and topItemProps where we could pass in extra props to our header/list components we wouldn't need this, but I don't think they have those?

bdow: No, they just have you pass your props in directly...
```
components={{
  TopItemList: React.forwardRef((props, ref) => <CustomComponent customProp={prop} {...props} ref={ref} />);
}}
```

stephenh: ...huh, and that doesn't break their memorization? I mean, if it doesn't and I was just making that up, then 🤷 , yeah sounds like we could remove the memoizeOnes.

bdow: haha. Do you recall what was breaking memoization? Was it on filtering, scrolling, etc... ?

bdow: I am testing it out now by just scrolling and see how it memoizes/rerenders. Will report back soon.

bdow: @stephenh - When scrolling the table to force the VirtualRoot to rerender, I am not really seeing a difference between using the `memoizeOne` and not using it.
I think for now I'm going to keep it as is. If we want to change the I'd prefer to do it in an isolated PR in the event it really breaks something and we have to revert.
Though, I have made some changes to the `TopItemList` component, because turns out it is just for the wrapping element. `TopItemList` just wraps the `List`/`VirutalRoot` component, so it is much simpler than I originally made it.

PR #519, opened 2022-04-28
feat: Get table styles via function call using config options

bdow:


src/components/Table/styles.tsx
```diff
@@ -41,47 +41,45 @@ export const cardStyle: GridStyle = {
...
-  emptyCell: "-",
-  presentationSettings: { borderless: true, typeScale: "xs", wrap: false },
-  firstRowMessageCss: Css.tc.py3.$,
-  // Included as a hacky "treat indent as deprecated for this table" hint to GridTable
-  levels: {},
-};
+export interface GridStyleDef {
+  form?: boolean;
+  grouped?: boolean;
+  flexible?: boolean;
```
stephenh, line 47: Maybe `flexibleHeight`? `dynamicRowHeight`? Or alternatively `fixedRowHeight`?

Just as-is I think I'd assume this is like "flexible columns" or something.

Maybe `fixedRowHeight` b/c I assume our default is flexible, so then users would pass `fixedRowHeight: true`? 

I kinda like `fixedRowHeight` the best, as being most descriptive, although I get that like, from a design language perspective, they call this "just fixed" or "just flexible"...

Maybe I'm throwing out too many options, but we could also do like `rowHeight: "fixed" | "flexible"`

Wdyt?

bdow: Yea, I think we could do `rowHeight: "fixed" | "flexible"`. I think I might have `flexible` be the default then, if that sounds right to you?

src/components/Table/styles.tsx
```diff
@@ -41,47 +41,45 @@ export const cardStyle: GridStyle = {
...
-  headerCellCss: Css.gray700.xsEm.bgGray200.aic.nowrap.pxPx(12).hPx(40).$,
-  cellCss: Css.gray900.xs.bgWhite.aic.nowrap.pxPx(12).hPx(36).boxShadow(`inset 0 -1px 0 ${Palette.Gray200}`).$,
-  emptyCell: "-",
-  presentationSettings: { borderless: true, typeScale: "xs", wrap: false },
-  firstRowMessageCss: Css.tc.py3.$,
-  // Included as a hacky "treat indent as deprecated for this table" hint to GridTable
-  levels: {},
-};
+export interface GridStyleDef {
+  form?: boolean;
```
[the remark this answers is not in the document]
bdow: Yea, `inlineEditing` sounds fine to me. Will update

src/components/Table/styles.tsx
stephenh: Yep!

stephenh (approved):
Nice! Lgtm!

PR #523, opened 2022-05-02
feat: Support Totals row in GridTable

bdow:


src/components/Table/GridTable.tsx
```diff
@@ -213,7 +215,7 @@ export interface GridTableProps<R extends Kinded, S, X> {
   rowLookup?: MutableRefObject<GridRowLookup<R> | undefined>;
   /** Whether the header row should be sticky. */
   stickyHeader?: boolean;
-  stickyOffset?: string;
+  stickyOffset?: number;
```
stephenh, line 218: Do we still need this? I thought I grepped internal-frontend the other day and we don't use it anymore, has been replaced by ScrollableParent

stephenh (approved):
Nice!

src/components/Table/GridTable.tsx
bdow: It's not being used, but I'd kind of like to keep it for now.

PR #526, opened 2022-05-05
fix: Shallow watch rows to trigger getSelectedRows computeds.

stephenh:


bdow (approved):
Awesome! Thanks for fixing. Nice that wasn't so bad to address. Thanks mobx!

PR #530, opened 2022-05-16
feat: Support contrast variant of Buttons

bdow:
**Note:** Design only provided primary and tertiary variants of buttons for 'contrast'/dark mode. Since the "text" variant is extremely similar to "tertiary", then I followed suite on those stylings. Secondary, danger made some slight modifications, but didn't want to spend too much guess work on it until design comes back with updates for these.

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -3,7 +3,7 @@ import { ButtonHTMLAttributes, ReactNode, RefObject, useMemo, useRef } from "rea
 import { useButton, useFocusRing, useHover } from "react-aria";
 import { Link } from "react-router-dom";
 import { Icon, IconProps, maybeTooltip, navLink, resolveTooltip } from "src/components";
-import { Css } from "src/Css";
+import { Css, Palette } from "src/Css";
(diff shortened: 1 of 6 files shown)


src/components/Button.tsx
```diff
@@ -115,54 +127,58 @@ export function Button(props: ButtonProps) {
...
-const variantStyles: Record<ButtonVariant, { baseStyles: {}; hoverStyles: {}; disabledStyles: {}; pressedStyles: {} }> =
-  {
-    primary: {
-      baseStyles: Css.bgLightBlue700.white.$,
-      hoverStyles: Css.bgLightBlue900.$,
-      pressedStyles: Css.bgLightBlue500.$,
-      disabledStyles: Css.bgLightBlue200.$,
-    },
+const variantStyles: (
+  contrast?: boolean,
```
stephenh, line 143: Do we want `contrast` to be an optional params that callers can forget to pass?

stephenh (approved):

src/components/Button.tsx
bdow: yea, probably not necessary to be optional here. Thanks for pointing out.

PR #531, opened 2022-05-17
feat: Add GridCellContent.onClick property to easily wrap content in a link/button component

bdow:
By utilizing the new prop, then GridCells can properly truncate text when the rowHeight is fixed. This prevents users having to ensure they add proper truncation styles locally when wanting to wrap the content of the cell in an interactive element

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -38,6 +38,7 @@ import { Css, Margin, Only, Palette, Properties, Typography, Xss } from "src/Css
 import { useComputed } from "src/hooks";
 import { useRenderCount } from "src/hooks/useRenderCount";
 import { safeKeys } from "src/utils";
+import { getButtonOrLink } from "src/utils/getInteractiveElement";
 import { shallowEqual } from "src/utils/shallowEqual";
 import { defaultStyle } from ".";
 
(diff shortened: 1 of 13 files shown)


src/interfaces.ts
```diff
@@ -18,7 +18,7 @@ export interface BeamButtonProps {
    */
   disabled?: boolean | ReactNode;
   /** If function, then it is the handler that is called when the press is released over the target. Otherwise if string, it is the URL path for the link */
-  onClick?: ((e: PressEvent) => void) | string;
+  onClick: ((e: PressEvent) => void) | string;
```
bdow, line 21: I made this required, which ended up causing a bunch of updates needed to supply the `onClick` in many stories/tests that didn't include it yet.

src/components/Table/GridTable.tsx
```diff
@@ -1284,11 +1287,21 @@ function toContent(
...
+          css: Css.maxw100.lightBlue700.if(style?.presentationSettings?.wrap === false).truncate.$,
+        })
+      : content;
+
   if (content && typeof content === "string" && isHeader && canSortColumn) {
     return <SortHeader content={content} iconOnLeft={alignment === "right"} />;
-  } else if (content && typeof content === "string" && style?.presentationSettings?.wrap === false) {
+  } else if (content && style?.presentationSettings?.wrap === false && typeof content === "string") {
+    // In order to truncate the text properly, then we need to wrap it in another element
+    // as our cell element is a flex container, which don't allow for applying truncation styles directly on it.
```
stephenh, line 1302: Cool, great comment

src/interfaces.ts
stephenh: +1 to the change

stephenh (approved):
Nice, I like just handling this for callers.

PR #532, opened 2022-05-17
feat: Client side Table sorting to be case insensitive by default

bdow:
Adds a 'caseSensitive' flag that can be added to change the default behavior.

### DIFF
--- a/src/components/Table/sortRows.ts
+++ b/src/components/Table/sortRows.ts
@@ -7,12 +7,13 @@ export function sortRows<R extends Kinded>(
   columns: GridColumn<R>[],
   rows: GridDataRow<R>[],
   sortState: SortState<number>,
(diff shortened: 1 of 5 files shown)


src/components/Table/sortRows.ts
```diff
@@ -68,7 +70,9 @@ function sortValue(value: ReactNode | GridCellContent): any {
   if (maybeFn instanceof Function) {
     return maybeFn();
```
stephenh, line 71: Do we need to lower case here as well? Maybe not early return an do `maybeFn = maybeFn()`?

stephenh (approved):

src/components/Table/sortRows.ts
bdow: Makes sense. Thanks!

PR #535, opened 2022-05-19
fix: Bug fix for toggling collapse state

bdow:
Code had assumed that the 'header' was always in the first position of the 'collapsedIds' array. However, we always just pushed the 'header' into the array, so if the header was not the first row to be collapsed then the logic would break. To fix we now check if the array includes 'header' rather than expecting it in the first position.

Additionally updated the header's collapse toggle behavior to update depending on the state of every top level collapsible row.

Adds tests for the bug fix and for the new header behavior.

### DIFF
--- a/src/components/Table/RowState.ts
+++ b/src/components/Table/RowState.ts
@@ -156,7 +156,7 @@ export class RowState {
         // Expand all means keep `collapsedIds` empty
         collapsedIds.splice(0, collapsedIds.length);
       } else {
-        // Otherwise splice `header` on the front of the list as a hint that we're in the collapsed-all state
+        // Otherwise push `header` to the list as a hint that we're in the collapsed-all state
         collapsedIds.push("header");
         // Find all non-leaf rows so that toggling "all collapsed" -> "all not collapsed" opens
         // the parent rows of any level.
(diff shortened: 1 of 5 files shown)


src/components/Table/RowState.ts
```diff
@@ -180,6 +180,16 @@ export class RowState {
       } else {
         collapsedIds.splice(i, 1);
       }
+
+      // If all rows have been expanded individually, but the 'header' was collapsed, then remove the header from the collapsedIds so it reverts to the expanded state
+      if (collapsedIds.length === 1 && collapsedIds[0] === "header") {
+        collapsedIds.splice(0, 1);
+      } else {
+        // If every top level child has been collapsed, then splice the header to the front of the array to be considered collapsed as well.
```
stephenh, line 188: Nit that comment says splice, probably written before you fixed the "header must be first" vs. ".includes("header")" line.

bdow: Ah yes... I had spliced it in at first and kept the check looking for 'header' as the first in the array, but afterwards felt like checking if it was anywhere in the array was a bit safer. Will update the comment.

PR #536, opened 2022-05-20
fix: Sort BoundCheckboxGroupField values so formState can properly check if dirty

bdow:


src/forms/BoundCheckboxGroupField.tsx
```diff
@@ -27,9 +27,10 @@ export function BoundCheckboxGroupField(props: BoundCheckboxGroupFieldProps) {
       {() => (
         <CheckboxGroup
           label={label}
-          values={field.value || []}
+          // Sort the values so that the array's contents can properly assert if the field is dirty no matter which order options are added/removed
+          values={field.value?.sort() || []}
```
stephenh, line 31: Hm, does sorting the `values` change the order they are rendered in? I could see a page wanting to show checkboxes in a very specific order? 

Good call on "this should not be dirty", but I wonder if we should solve this more directly inside of form-state, i.e. somehow teach value fields to not be dirty if a list order is different...

Granted, I could see this actually mattering sometimes, like for some page an order of `[a, b, c]` vs. `[c, b, a]` should actually be dirty; so we might have to pick a default behavior (does order matter? yes/no) and then allow the user to override it maybe via the formConfig, like a `colors: { type: "value", orderMatters: false }` or something...

bdow: `options` is what denotes the order of the checkboxes.

And yea, I think order could matter in some instances, so probably best to handle locally instead. Unless we had some sort of hint in FormState to have it not care about order when evaluating whether it is dirty or not.

bdow:
After more discussion outside of the PR, we have decided this to support order agnostic equality checks in FormState: https://github.com/homebound-team/form-state/pull/66
With that update to FormState this PR is no longer relevant, so closing!

PR #539, opened 2022-05-20
feat: Bump form state and demonstrate strictOrder property

bdow:


stephenh (approved):
wtf, I swore I shipped this earlier, sorry about that!

PR #540, opened 2022-05-20
fix: Base resizing on a specifically styled div rather than table element

bdow:


src/components/Table/columnSizes.ts
```diff
@@ -25,8 +25,8 @@ import { useDebouncedCallback } from "use-debounce";
 export function useSetupColumnSizes(
   style: GridStyle,
   columns: GridColumn<any>[],
-  tableRef: MutableRefObject<HTMLElement | null>,
-  resizeTarget: MutableRefObject<HTMLElement | null> | undefined,
+  resizeRef: MutableRefObject<HTMLDivElement | null>,
+  externalResizeTarget: MutableRefObject<HTMLElement | null> | undefined,
```
stephenh, line 29: Hm, should we have just 1 `resizeRef` param here, and have line 325 be:

```
const columnSizes = useSetupColumnSizes(style, columns, resizeRef ?? resizeTarget);
```

?

AFAICT we only ever use 1 or the other in this method?

bdow: True, will update

src/components/Table/GridTable.tsx
```diff
@@ -297,6 +297,8 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   // We only use this in as=virtual mode, but keep this here for rowLookup to use
   const virtuosoRef = useRef<VirtuosoHandle | null>(null);
   const tableRef = useRef<HTMLElement>(null);
```
stephenh, line 299: Looks like maybe we don't need `tableRef` anymore?

src/components/Table/GridTable.tsx
```diff
@@ -513,6 +515,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   return (
     <RowStateContext.Provider value={rowStateContext}>
       <PresentationProvider fieldProps={fieldProps} wrap={style?.presentationSettings?.wrap}>
+        <div ref={resizeRef} css={Css.w100.$} />
```
stephenh, line 518: This makes a lot of sense!

stephenh (approved):
Nice! Will be really interested to see if this fixes the chromatic issues.

src/components/Table/GridTable.tsx
bdow: Yea, I don't think so... wasn't sure if I should delete it yet though. I guess I can.

bdow: Alrighty - since you took the time to see if it was okay to delete, I deleted it.

stephenh: :100: :-)

PR #541, opened 2022-05-26
feat: Support setting group rows as initially collasped

bdow:
Adds new property on GridDataRow, 'initCollapsed'. By setting this to true it will signal to the table to initially render this row as collapsed. As the name describes, this is only applied/read on initial render. If the user has also set up the 'persistCollapse' then any values available in the localStorage will trump those set on GridDataRow.

Includes tests

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -309,7 +309,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   const style = resolveStyles(maybeStyle);
 
   const { rowState } = api;
-  rowState.rows = rows;
+  rowState.setRows(rows);
 
   useEffect(() => {
     rowState.activeRowId = activeRowId;
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -301,7 +301,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
 
   const api = useMemo<GridTableApiImpl<R>>(() => {
     const api = (props.api as GridTableApiImpl<R>) ?? new GridTableApiImpl();
-    api.init(persistCollapse, virtuosoRef);
+    api.init(persistCollapse, virtuosoRef, rows);
     api.setActiveRowId(activeRowId);
     return api;
   }, [props.api]);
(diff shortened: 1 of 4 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -301,7 +301,7 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
 
   const api = useMemo<GridTableApiImpl<R>>(() => {
     const api = (props.api as GridTableApiImpl<R>) ?? new GridTableApiImpl();
-    api.init(persistCollapse, virtuosoRef);
+    api.init(persistCollapse, virtuosoRef, rows);
```
stephenh, line 304: Wdyt about handling rows that show up after the initial render? Or if the initial render has an empty rows array b/c we're waiting on a gql response? Like do we need this to be more about "the first time we see a row"?

...part of me wonders if super drawing everything seems our persistent collapse state feature isn't necessary anymore.

bdow: I think keeping persistCollapse is worth it for larger workflows that take the user away from a page and then they return to it. I agree if everything time you dive into a row then it opens a super drawer, then we shouldn't need it. But that is too much of an assumption to me.

bdow: Good question about the rows that are loaded for the first time... I'm not sure how we'd handle that. I think thats a good idea, especially for tables that are rendered but the data is still in flight, such as done via QueryTable. Not a use case that I thought about, so thanks for bringing it up.
Do you have any thoughts about how we can determine whether this row is new vs just re-rendered?

stephenh: I forget, do we use session storage for persisting, or local? Like does it get reset if I restart my browser?

bdow: It is localStorage, so it would persist across instances/tabs.

stephenh: Still talking out loud, but maybe tables that use initCollapse assert that they "know better"/have better defaults than "everything is open" and opt-out of persist collapse?

bdow: I like the idea of being able to handle both, like for schedules when grouped by status, we can initially say that Completed are collapsed. Though, if a user is working through completed tasks for whatever reason, then they would probably get annoyed if everytime they came back to the page they had to re-open the completed group. Maybe instead we just move to using `sessionStorage` to get around the "if you come back 3 days later" sort of thing. I think that'd be a good middle ground.

stephenh: Yeah, I like sessionStorage too. So, something like:

* If you have persisted collapsed state, that always wins
* If you have zero rows, we'll "wait" to determine "does persisted collapsed state win or initCollapsed win"?

bdow: It got a bit more complicated than just that. I have pushed up a new commit with the updates and new tests

TylerR909 (approved):

PR #543, opened 2022-05-26
feat: Loading Spinner component

bdow:


stephenh (approved):
Neat!

src/components/CssReset.tsx
```diff
@@ -43,6 +43,15 @@ const ourReset = css`
       background-color: ${Palette.Gray500};
     }
   }
+
+  @keyframes spin {
```
lucyconklin, line 47: Out of curiosity I added an animation to a truss-config like this:
```
const colorChangeBgKeyframes = "@keyframes color-change-background {0% {background-color: #F1B62B} 11% {background-color: #E37C22} 23% {background-color: #DA0B36} 37% {background-color: #AF0B58} 50% {background-color: #7B3A88} 64% {background-color: #3F51B5} 77% {background-color: #1CA176} 91% {background-color: #7DB759} 100% {background-color: #F1B62B}}"
const colorChangeFillKeyframes = "@keyframes color-change-fill {0% {fill: #F1B62B} 11% {fill: #E37C22} 23% {fill: #DA0B36} 37% {fill: #AF0B58} 50% {fill: #7B3A88} 64% {fill: #3F51B5} 77% {fill: #1CA176} 91% {fill: #7DB759} 100% {fill: #F1B62B}}"
const sections = {
  animation: () => [
    newMethod("colorChangeBg", { animation: `color-change-background 10s linear infinite; ${colorChangeBgKeyframes}` }),
    newMethod("colorChangeFill", { animation: `color-change-fill 10s linear infinite; ${colorChangeFillKeyframes}`})
  ],
};
```

Just in case you wanted to add this in there instead of this CssReset.

lucyconklin (commented):
Very cool!

src/components/CssReset.tsx
bdow: Interesting. I _think_ what would happen here is that you could end up having multiple `@keyframes color-change-fill` definitions if the animation method, `colorChangeFill`, is used multiple times in different variations of a truss style. For instance,
```
<div css={Css.colorChangeFill.mb1.$} />
<div css={Css.colorChangeFill.mb2.$} />
<div css={Css.colorChangeFill.mb3.$} />
```
Each of those will generate a unique class in order to group both the `animation` and the `margin-bottom` definitions together. And each time the animation is generated then it will also include the `@keyframes` definition. 

I decided to try and validate what I thought might happen, and it unfortunately does. The animations all still work fine.


Would be cool for Truss to handle animation keyframes and maybe generate them as part of a CssReset file.

PR #544, opened 2022-05-31
fix: RowLookup's 'getKinds' ignores action columns

bdow:


chr1sjf0x (approved):
praise: TY @bdow - great fix/cleanup 🚀  🍣

PR #548, opened 2022-06-01
fix: Modal should not close if interacting with another dialog ontop of it - such as a Lightbox

bdow:
Blueprint's To Do modal includes a list of attachments. These attachments can be images that trigger a 3rd party lightbox component, that was closing both the modal and the lightbox component if you interacted with the lightbox.
This PR allows for keeping the modal open if the user is interacting with another dialog on top of it.

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -52,8 +52,8 @@ export function Modal(props: ModalProps) {
       onClose: closeModal,
       isDismissable: true,
       shouldCloseOnInteractOutside: (el) => {
-        // Do not close the Modal if the user is interacting with the Tribute mentions dropdown (via RichTextField).
-        return !el.closest(".tribute-container");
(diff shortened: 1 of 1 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -52,8 +52,8 @@ export function Modal(props: ModalProps) {
       onClose: closeModal,
       isDismissable: true,
       shouldCloseOnInteractOutside: (el) => {
-        // Do not close the Modal if the user is interacting with the Tribute mentions dropdown (via RichTextField).
-        return !el.closest(".tribute-container");
+        // Do not close the Modal if the user is interacting with the Tribute mentions dropdown (via RichTextField) or with another dialog on top of it.
```
stephenh, line 55: Maybe "another 3rd party dialog like lightbox" to differentiate its not a beam dialog.

stephenh (approved):
Cool

PR #549, opened 2022-06-02
fix: Add onClick to button element in getInteractiveElement

bdow:


stephenh (approved):
Lgtm, although I'm not completely following what broke...like `Button onClick={lambda}` was ... entirely broken? Is this something we could cover in a unit test, if it's more than just a visual regression afaiu?

bdow:
I'm going to add another test from GridTable to ensure this prop is handled correctly.

PR #550, opened 2022-06-03
feat: Sc 13020/dev auto save notification component

TylerR909:
[SC-13020](https://app.shortcut.com/homebound-team/story/13020/dev-auto-save-notification-component)

Adds `<AutoSaveIndicator />` component

A bit more embellished than the comps: https://www.figma.com/file/aWUE4pPeUTgrYZ4vaTYZQU/%E2%9C%A8Beam-Design-System?node-id=12035%3A11589

Kapture 2022-06-03 at 14 51 34

### DIFF
--- a/src/components/AutoSaveIndicator.stories.tsx
+++ b/src/components/AutoSaveIndicator.stories.tsx
@@ -1,34 +1,56 @@
-import { AutoSaveStatusContext } from "@homebound/form-state";
-import { AutoSaveStatus } from "@homebound/form-state/dist/AutoSaveStatus/AutoSaveStatusProvider";
+import { AutoSaveStatus, AutoSaveStatusContext, AutoSaveStatusProvider } from "@homebound/form-state";
 import { Meta } from "@storybook/react";
+import { Css } from "..";
 import { AutoSaveIndicator } from "./AutoSaveIndicator";
+import { Button } from "./Button";
 
 export default {
   component: AutoSaveIndicator,
   title: "Components/Auto Save Indicator",
 } as Meta;
 
-const renderForStatus = (status: AutoSaveStatus) => (
-  <AutoSaveStatusContext.Provider
-    value={{
-      status,
-      resetStatus() {},
-      errors: [],
-      resolveAutoSave() {},
-      triggerAutoSave() {},
-    }}
-  >
-    <AutoSaveIndicator />
-  </AutoSaveStatusContext.Provider>
+const autoSaveProviderValue = (status: AutoSaveStatus) => ({
+  status,
+  resetStatus() {},
+  errors: [],
+  resolveAutoSave() {},
+  triggerAutoSave() {},
+});
+
+export const AutoSaveIndicatorStatuses = () => (
+  <>
+    <AutoSaveStatusContext.Provider value={autoSaveProviderValue(AutoSaveStatus.IDLE)}>
+      <AutoSaveIndicator />
+    </AutoSaveStatusContext.Provider>
+    <AutoSaveStatusContext.Provider value={autoSaveProviderValue(AutoSaveStatus.SAVING)}>
+      <AutoSaveIndicator />
+    </AutoSaveStatusContext.Provider>
+    <AutoSaveStatusContext.Provider value={autoSaveProviderValue(AutoSaveStatus.DONE)}>
+      <AutoSaveIndicator />
+    </AutoSaveStatusContext.Provider>
+    <AutoSaveStatusContext.Provider
+      value={{ ...autoSaveProviderValue(AutoSaveStatus.ERROR), errors: [new Error("Some issue happened saving")] }}
+    >
+      <AutoSaveIndicator />
+    </AutoSaveStatusContext.Provider>
+  </>
 );
 
-export const AutoSaveIndicatorStatuses = () => {
-  return (
-    <>
-      {renderForStatus(AutoSaveStatus.IDLE)}
-      {renderForStatus(AutoSaveStatus.SAVING)}
-      {renderForStatus(AutoSaveStatus.DONE)}
-      {renderForStatus(AutoSaveStatus.ERROR)}
-    </>
-  );
-};
+export const AutoSaveIndicatorInAction = () => (
+  <AutoSaveStatusProvider resetToIdleTimeout={2000}>
+    <AutoSaveIndicator />
+    <AutoSaveStatusContext.Consumer>
+      {({ triggerAutoSave, resolveAutoSave }) => (
+        <Button
+          label="Do a Save"
+          variant="secondary"
+          onClick={() => {
+            triggerAutoSave();
+            setTimeout(resolveAutoSave, 1500);
+          }}
+          css={Css.buttonBase.$}
+        />
+      )}
+    </AutoSaveStatusContext.Consumer>
+  </AutoSaveStatusProvider>
+);
--- a/src/components/AutoSaveIndicator.stories.tsx
+++ b/src/components/AutoSaveIndicator.stories.tsx
@@ -1,0 +1,34 @@
+import { AutoSaveStatusContext } from "@homebound/form-state";
+import { AutoSaveStatus } from "@homebound/form-state/dist/AutoSaveStatus/AutoSaveStatusProvider";
+import { Meta } from "@storybook/react";
+import { AutoSaveIndicator } from "./AutoSaveIndicator";
+
+export default {
+  component: AutoSaveIndicator,
+  title: "Components/Auto Save Indicator",
+} as Meta;
+
+const renderForStatus = (status: AutoSaveStatus) => (
+  <AutoSaveStatusContext.Provider
+    value={{
+      status,
+      resetStatus() {},
+      errors: [],
+      resolveAutoSave() {},
+      triggerAutoSave() {},
+    }}
+  >
+    <AutoSaveIndicator />
+  </AutoSaveStatusContext.Provider>
+);
+
+export const AutoSaveIndicatorStatuses = () => {
+  return (
+    <>
+      {renderForStatus(AutoSaveStatus.IDLE)}
+      {renderForStatus(AutoSaveStatus.SAVING)}
+      {renderForStatus(AutoSaveStatus.DONE)}
+      {renderForStatus(AutoSaveStatus.ERROR)}
+    </>
+  );
+};
--- a/src/components/AutoSaveIndicator.stories.tsx
+++ b/src/components/AutoSaveIndicator.stories.tsx
@@ -9,14 +9,6 @@ export default {
   title: "Components/Auto Save Indicator",
 } as Meta;
 
-const autoSaveProviderValue = (status: AutoSaveStatus) => ({
-  status,
-  resetStatus() {},
-  errors: [],
-  resolveAutoSave() {},
-  triggerAutoSave() {},
-});
-
 export const AutoSaveIndicatorStatuses = () => (
   <>
     <AutoSaveStatusContext.Provider value={autoSaveProviderValue(AutoSaveStatus.IDLE)}>
@@ -54,3 +46,11 @@ export const AutoSaveIndicatorInAction = () => (
     </AutoSaveStatusContext.Consumer>
   </AutoSaveStatusProvider>
 );
+
+const autoSaveProviderValue = (status: AutoSaveStatus) => ({
+  status,
+  resetStatus() {},
+  errors: [],
+  resolveAutoSave() {},
+  triggerAutoSave() {},
+});
--- a/src/components/AutoSaveIndicator.stories.tsx
+++ b/src/components/AutoSaveIndicator.stories.tsx
@@ -12,7 +12,7 @@ export default {
 export const AutoSaveIndicatorStatuses = () => (
   <>
     <AutoSaveStatusContext.Provider value={autoSaveProviderValue(AutoSaveStatus.IDLE)}>
-      <AutoSaveIndicator />
+      <AutoSaveIndicator showOnIdle />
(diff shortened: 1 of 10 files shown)


src/components/AutoSaveIndicator.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Css } from "..";
+import { AutoSaveIndicator } from "./AutoSaveIndicator";
+import { Button } from "./Button";
+
+export default {
+  component: AutoSaveIndicator,
+  title: "Components/Auto Save Indicator",
+} as Meta;
+
+const autoSaveProviderValue = (status: AutoSaveStatus) => ({
```
stephenh, line 12: Maybe `newAutoSaveProvider` and move it to the bottom?

src/components/AutoSaveIndicator.test.tsx
```diff
@@ -0,0 +1,80 @@
+import { AutoSaveStatus, AutoSaveStatusContext } from "@homebound/form-state";
+import React from "react";
+import { noop } from "src/utils";
+import { render } from "src/utils/rtl";
+import { AutoSaveIndicator } from "./AutoSaveIndicator";
+
+const MockAutoSaveProvider = ({
```
stephenh, line 7: Nit, move to the bottom, `function`...

src/components/AutoSaveIndicator.tsx
```diff
@@ -0,0 +1,34 @@
+import { AutoSaveStatus, useAutoSaveStatus } from "@homebound/form-state";
+import { assertNever } from "@homebound/form-state/dist/utils";
```
stephenh, line 2: Hm, do we have an `assertNever` here locally in Beam? Would be nice to avoid the dist import.

src/components/AutoSaveIndicator.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface AutoSaveIndicatorProps {
+  hideOnIdle?: boolean;
+}
+
+export function AutoSaveIndicator({ hideOnIdle }: AutoSaveIndicatorProps) {
+  const { status, resetStatus, errors } = useAutoSaveStatus();
+
+  /** On Dismount, reset back to Idle so new pages don't imply something "saved" */
+  useEffect(() => () => resetStatus(), []);
```
stephenh, line 16: thought: Seems like this shouldn't be the indicators responsibility, but the responsibility of anyone who's called `triggerAutoSave`, that has an outstanding save when they're unmounted, to resolve it / not leaving it dangling?

I.e. could `useAutoSaveStatus` just do this? Recognize when its component has called `triggerAutoSave` --> no resolve --> its component is dismounted, it auto-resolves it? 

(Like I could see a component hierarchy where the AutoSaveIndicator dismounts, but some sibling component still has valid/outstanding saves, live in the shared AutoSaveContext. Granted, this setup seems a little weird, insofar as I think the assumption you have here of "the AutoSaveIndicator lives above all other auto save components in this AutoSaveContext, so the indicator unmounting --> defacto the others have unmounted" is intuitive/probably how it'd be setup, but not necessarily _has_ to be / will be guaranteed afaiu.)

src/components/Icon.tsx
```diff
@@ -11,24 +11,28 @@ export interface IconProps extends AriaAttributes, DOMProps {
...
-      height={size}
-      viewBox="0 0 24 24"
-      xmlns="http://www.w3.org/2000/svg"
-      css={{ "path, rect": Css.fill(color).$, ...xss }}
-      data-icon={icon}
-      {...other}
-    >
-      {Icons[icon]}
-    </svg>
+    <div css={Css.df.gap1.color(color).$}>
```
stephenh, line 21: thought: I'm curious if the extra div is going to throw off existing usages that don't expect the root element to be a flex container instead of an svg.

Granted, we try to drive home the "callers can't couple themselves to DOM internals of a component", but sometimes especially for the "oh hey the child is flex" change then I _think_ there can be browser/layout-ism changes? @bdow ?

I'm good with merging and just assuming/hoping it will be fine; just thinking that we might have to conditionally wrap the div if `helperText` is set, as we've done that in a few other places for similar sort of reasons.

stephenh (approved):

src/components/Icon.tsx
bdow: I agree. I think I'd rather the Icon button stay as returning only the svg. If a component needs to add a label next to it or decorate it in some fashion, then I think that component should import the Icon element as an svg and style any containers needed for the Icon themselves. Icon is one of our components that is dead simple, and I really like that about it.

TylerR909: The only different Chromatic noted was this, and it really wasn't a difference: https://www.chromatic.com/test?appId=60466f7d43c37c0021788867&id=629a76013d6416003afee83e

But I'm not tethered to this approach.

src/components/AutoSaveIndicator.tsx
TylerR909: We do not. And this is the only one intellisense is picking up.

src/components/AutoSaveIndicator.tsx
TylerR909: I'd argue that there are plenty of components that _could_ call `useAutoSave()` just to observe behavior and they wouldn't want "side effects" from calling that hook. Form's only interest is Triggering/Resolving. It does not care if things are reset or what the active state of the button is. Just trigger and resolve.

AutoSaveIndicator, however, cares a lot about Status. It displays based on it. If * it * dismounts, it's almost definitely for a determinable reason (closed modal, page navigation, etc) and given that it's the only component that actually cares about what the Status is, it also adopts the responsibility to reset it at will.

stephenh: > wouldn't want "side effects" from calling that hook

Of course, my suggestion is only for components like forms who've explicitly `triggerAutoSave`-d.

The contract would be "if you trigger auto save, it's on you to either resolve it explicitly or implicitly on unmount".

> given that it's the only component that actually cares about what the Status is,
> it also adopts the responsibility to reset it at will.

This articulation just seems wrong. This component being the only one that cares about _reading_ status (to display) IMO doesn't follow that it gets to decide to _write_ status as well.

src/components/AutoSaveIndicator.tsx
bdow: Can we add one in to Beam, please?

src/components/Icon.tsx
bdow: Great, let's move this wrapping element into the `AutoSaveIndicator` component. And maybe we wrap it with a `display: inline-flex;` instead. This will help prevent the tooltip from getting placed far away from the component.
Screen Shot 2022-06-06 at 9 27 45 AM

src/components/BeamContext.tsx
```diff
@@ -86,13 +87,15 @@ export function BeamProvider({ children, ...presentationProps }: BeamProviderPro
...
       <PresentationProvider {...presentationProps}>
         <SnackbarProvider>
-          {/* OverlayProvider is required for Modals generated via React-Aria */}
-          <OverlayProvider>
-            {children}
-            {/*If the drawer is open, assume it will show modal content internally. */}
-            {modalRef.current && drawerContentStackRef.current.length === 0 && <Modal {...modalRef.current} />}
-          </OverlayProvider>
-          <SuperDrawer />
+          <AutoSaveStatusProvider>
```
bdow, line 90: What do you think about wrapping the SnackbarProvider with AutoSaveStatusProvider? Probably not necessary, but just thinking out loud if we'd ever want access to that status/errors from within the snackbar. Probably not...

TylerR909: I'm happy to do it but I don't think we gain anything from it.

_other things_ call `triggerNotice()` and they'd have access. Snackbar itself... not sure why it'd need it but it won't hurt, so sure 😄

src/components/Icon.tsx
TylerR909: Man, I was trying so many things to get that tooltip to be placed correctly. Thanks 😃

src/components/AutoSaveIndicator.tsx
```diff
@@ -0,0 +1,34 @@
...
+      return hideOnIdle ? null : <Icon icon="cloudSave" />;
+    case AutoSaveStatus.SAVING:
+      return <Icon icon="refresh" helperText="Saving..." color={Palette.LightBlue500} />;
+    case AutoSaveStatus.DONE:
+      return <Icon icon="cloudSave" helperText="Saved" color={Palette.Green600} />;
+    case AutoSaveStatus.ERROR:
+      return (
+        <Tooltip title={String(errors)} placement="bottom">
+          <Icon icon="error" helperText="Error saving" color={Palette.Red500} />
+        </Tooltip>
```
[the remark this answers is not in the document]
bdow: Was he referring to both the Icon and the text color, or just the Icon. I would guess just the icon, but want to double check.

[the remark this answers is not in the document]
TylerR909: It'd look pretty awful if icon/text were different colors.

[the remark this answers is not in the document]
TylerR909: > They should both be the same color at any given time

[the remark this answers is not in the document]
TylerR909: Kinda feel like it's a downgrade, but I made my case 🙂 

image

Specifically, "Saving" being the same color as Idle is very jarring to me. Not enough feedback that anything is happening. c'est la vie

src/components/AutoSaveIndicator.tsx
TylerR909: Can we find a middleground here, please: https://github.com/homebound-team/beam/pull/550/commits/4a21ef1f8b2d53645575227d323bd1a5d704b490

bdow: Has this behavior regarding the status needing to be reset vs persist across pages about been discussed with our UX/Design team? I think I've mentioned this before, but maybe just to you @TylerR909, so I'll say it again here... If that status persists, I don't think thats a terrible thing. With SPAs we kind of tend to want less 'reset everything' as the user navigates, and make it feel more like a cohesive experience. Though, the placement of the indicator could make it confusing if you were on Homeowner Contracts and then switched to Commitments as it'd show that you "saved" changes in Homeowner Contracts, then "saved" may still show next to the Commitments title - and that doesn't seem right. But if you were in Commitments and you switch between the Line Items tab to the Overview tab and the "Saved" indication persisted there, I think that's fine. It'll still show that you "saved" something related to Commitments.
Thinking through those use cases, I'd tend to want the caller to dictate when to reset the status. In the instance of switching between the tabs, then maybe the page that builds the tabs would have a `useEffect` that only resets the status on unmount. And the content within the tabs could trigger the autosave. 
I guess to distill it down/TLDR, I think if the content around the AutoSaveIndicator changes - like the title of the page/modal/drawer, then we'd want to reset it. But if the title remains and we change tabs within the page, I think it would be a-okay to persist it.
Maybe the `useEffect` for reset on unmount something we just do in the PageHeader, ModalHeader, Drawer header components? 
Anyways, these sort of behaviors in the application should be a discussion with UX/Design if it hasn't been already. Present the different use cases, pros & cons, and come to consensus.

TylerR909: Jet had a 10-minute with Jeffrey and we stepped through some pages & drawers thinking when it should vs shouldn't reset.

Long story short: Any time this title changes, it should reset:
image

1. Title doesn't change in "Tabbed" pages like the Commitments Editor
2. Title does change, and should reset, any time you click up and down the Project Navbar
3. A drawer opening (and a new AutoSaveIndicator mounting) should NOT indicate anything in the drawer has saved yet (i.e. should reset on mount)
3b. He doesn't care if a drawer opening resets an AutoSave in the background
3c. When you close the drawer it may also reset

To me, this effectively boils down to any time it Mounts or Dismounts, reset.

It's confusing to users that they'd go to a new page and BluePrint would be telling them something on that page or in a newly-opened Drawer has been "Saved" before they've even touched anything.

stephenh: Drawers would have their own AutoSaveContext, so that makes sense (an action auto-saving in the drawer shouldn't make my greyed-out page header AutoSaveIndicator start spinning).

Actually probably same with pages (seems like Jeff is asking for them conceptually to have their own individual AutoSaveContexts), but we could probably have a global one and call it good enough, with using this heuristic of "however high up the component tree you put the AutoSaveIndicator, you're probably meaning to have an AutoSaveContext basically mapped 1-1 to that context, so AutoSaveIndicator unmount --> probably means 'sure a new/reset context'".

I.e., thinking through the alternative, if we really did have a single global AutoSaveIndicator up in the top-left, our pages would defacto share an AutoSaveContext, b/c moving from page 1 (with a save in flight) to page 2, I assume, shouldn't flash-/immediate-reset the global AutoSaveIndicator. We'd want it to finish it's current spinning w/o jarring away.

So anyway, with that clarification/articulation in my that, that personally "when do I reset" should not really be a question/operation, and instead "where do you want resets?" is more about "where do you want to draw context lines in your app", and then you would explicitly have a 1-1 mapping between AutoSaveContext <-> AutoSaveIndicator, and it's kind of a non-issue.

_But_, if we want to roll with a single global AutoSaveContext, and use the heuristic of AutoSaveIndicator unmounts as a hint to "we're effectively moving to a new context", I'm good with that.

stephenh (approved):

src/components/AutoSaveIndicator.tsx
bdow: Great! Thanks for taking the time to work through the behavior with Jeff.

PR #551, opened 2022-06-06
feat: Build in 'unset' option for SelectField

bdow:


src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -178,7 +197,11 @@ export function SelectFieldBase<O, V extends Value>(props: BeamSelectFieldBasePr
   async function maybeInitLoad() {
     if (!Array.isArray(maybeOptions)) {
       setFieldState((prevState) => ({ ...prevState, optionsLoading: true }));
-      const { options } = await maybeOptions.load();
+      const loadedOptions = (await maybeOptions.load()).options;
+      // Ensure the `unset` option is prepended to the top of the list if `unsetLabel` was provided
+      const options = !unsetLabel
+        ? loadedOptions
+        : getOptionsWithUnset(unsetLabel, (await maybeOptions.load()).options);
```
stephenh, line 204: Do we need a 2nd `await` here, or just use `loadedOptions`?

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -400,3 +425,23 @@ function getInputValue<O>(
...
+  }
+
+  return { ...options, initial: getOptionsWithUnset(unsetLabel, options.initial) };
+}
+
+function getOptionsWithUnset<O>(unsetLabel: string, options: O[]): O[] {
+  return [{ id: undefined, name: unsetLabel } as unknown as O, ...options];
+}
+
+function isUnsetOption<O>(o: UnsetOption | O): o is UnsetOption {
```
stephenh, line 445: We could potentially replace the `{ id: undefined, name: unsetLabel }` hash with just at top-level `const` like:

```
const unsetOption = {};
```

And then `isUnsetOption` may not be necessary because we'd just do `o === unsetOption`?

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -61,19 +63,36 @@ export function SelectFieldBase<O, V extends Value>(props: BeamSelectFieldBasePr
...
     onSelect,
-    options: maybeOptions,
+    options,
     multiselect = false,
-    getOptionLabel,
-    getOptionValue,
-    getOptionMenuLabel = getOptionLabel,
+    getOptionLabel: gol,
+    getOptionValue: gov,
+    getOptionMenuLabel: goml,
```
stephenh, line 70: Hm, maybe just don't destructure these, and where we do `gol`, just do `props.getOptionLabel`?

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -61,19 +63,36 @@ export function SelectFieldBase<O, V extends Value>(props: BeamSelectFieldBasePr
...
 
+  // Call `initializeOptions` to prepend the `unset` option if the `unsetLabel` was provided.
+  const maybeOptions = useMemo(() => initializeOptions(options, unsetLabel), [options, unsetLabel]);
+  // Memoize the callback functions and handle the `unset` option if provided.
+  const getOptionLabel = useCallback(
+    (o: O) => (unsetLabel && isUnsetOption(o) ? unsetLabel : gol(o)),
+    [gol, unsetLabel],
+  );
+  const getOptionValue = useCallback(
+    (o: O) => (unsetLabel && isUnsetOption(o) ? (o.id as V) : gov(o)),
```
stephenh, line 88: Suggestion: The `o.id as V` could probably just be `undefined`?

src/inputs/SelectField.test.tsx
```diff
@@ -216,7 +216,102 @@ describe("SelectFieldTest", () => {
...
+    // Given a Select Field providing the `getOptionMenuLabel`
+    const r = await render(
+      <TestSelectField
+        label="Age"
+        value={undefined}
+        unsetLabel="None"
+        options={labelValueOptions}
+        getOptionLabel={(o) => o.label}
+        getOptionValue={(o) => o.value}
+        getOptionMenuLabel={(o, isUnsetOpt) => (isUnsetOpt ? "Custom" : o.label)}
```
stephenh, line 287: Nit: Maybe use "Unset Label" instead of "Custom"?

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -45,6 +45,8 @@ export interface BeamSelectFieldBaseProps<O, V extends Value> extends BeamFocusa
   contrast?: boolean;
   /** Placeholder content */
   placeholder?: string;
+  /** Only used for Single Select Fields. If set, prepends an option with a `undefined` value at the top of the list */
+  unsetLabel?: string;
```
stephenh, line 49: Question: Currently our `onSelect` has `values: V[]`...if a user uses `unsetLabel`, what do we call `onSelect` with? It looks like `undefined`... should we update `onSelect` to have `values: V[] | undefined`? 

Or should it be an empty list?

Fwiw getting this typing right, i.e. if you want an `unsetLabel` your V needs to `extends undefined` _but_ if you don't set `unsetLabel`, then :shrug: you don't have to accept `undefined`, has been primarily why we haven't done this before imo/iirc.

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -45,6 +45,8 @@ export interface BeamSelectFieldBaseProps<O, V extends Value> extends BeamFocusa
   contrast?: boolean;
   /** Placeholder content */
   placeholder?: string;
+  /** Only used for Single Select Fields. If set, prepends an option with a `undefined` value at the top of the list */
```
stephenh, line 48: Thought: Should we exclude it from `MultiSelectField`'s props then?

src/inputs/internal/SelectFieldBase.tsx
bdow: You're right. This was a copy & paste mistake.

src/inputs/internal/SelectFieldBase.tsx
bdow: We'd need to do `undefined as V`, I believe. Otherwise we'd get issues like "Type 'undefined' is not assignable to type 'V'."

src/inputs/internal/SelectFieldBase.tsx
bdow: We'd currently call `onSelect` with `values = [undefined]`

bdow: I'm not recalling our whole `keyToValue`/`valueToKey` stuff... will see how I can better work within those bounds.

src/inputs/internal/SelectFieldBase.tsx
bdow: Makes sense to me.

src/inputs/internal/SelectFieldBase.tsx
bdow: Okay, great. So this whole rewriting of the `getOptionValue` is unnecessary because we do the `keyToValue` and `valueToKey` conversions in this component. Only thing we need to overwrite is the label helpers.

stephenh: ...I only recall that those functions exist :-D will have to go refresh my memory.

src/inputs/internal/SelectFieldBase.tsx
bdow: This rewrite ended up not being necessary as we already handled values set to `undefined`.

src/inputs/internal/SelectFieldBase.tsx
stephenh: Hm, I'm not quite following...

* As a caller, I pass `options=[{id:1}, {id:2}, {id:3}]`
* I have `getOptionValue=o => o.id`
* I pass `unsetLabel=None`
* We inject `options=[unsetOption, {id:1}, {id:2}, ...]`

We're going to call `getOptionValue({})` and the caller is going to do `{}.id`...

I guess technically this will work for probably 90% of our `getOptionValue`s, but if a `getOptionValue` was `o => o.id.toString()`, that would compile as valid TS, but then fail at runtime...

So, I dunno, I think we still want to intercept `getOptionValue` and not call the user's `getOptionValue` with an empty `{}` that they are not suspecting.

src/inputs/SelectField.test.tsx
```diff
@@ -216,7 +216,102 @@ describe("SelectFieldTest", () => {
...
+        getOptionLabel={(o) => o.label}
+        getOptionValue={(o) => o.value}
+      />,
+    );
+    // When we focus the field to open the menu
+    r.age().focus();
+    // And we select the 'unset' option
+    click(r.getByRole("option", { name: "None" }));
+    // Then onSelect was called
+    expect(onSelect).toHaveBeenCalledWith(undefined);
```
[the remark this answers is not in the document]
bdow: Hmmm. I think I agree with `opt: 0 | undefined`, but `V extends Value`, and Value is:
```
export type Value = string | number | null | undefined | boolean;
```
This already includes `undefined`, which I think should be okay, right? And actually would `opt` ever be `undefined`? the `opt` in this case would end up being the `const unsetOption = {};`, right?
I might be missing something.. but I think this is fine. I'll pull the build to Blueprint locally and do a type-check to see what happens.

src/inputs/internal/SelectFieldBase.tsx
bdow: Great point. I will re-introduce overwriting the `getOptionValue` locally as well. Thanks for the thorough and thoughtful feedback.

src/inputs/SelectField.test.tsx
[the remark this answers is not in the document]
stephenh: @bdow `V extends Value` means it _could_ be `string | number | null | undefined | boolean`; but it could also be _just_ `string`.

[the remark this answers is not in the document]
bdow: I get what you're saying. Sorry it took me a bit to come around. It is less about what `V` could be, but what `V` is instantiated as by the caller. 10-4. Will update.
I will fix it, but fwiw, `type-check` didn't find any issues in `internal-frontend`. 

So do you think in the `onSelect` we'd do something more like:
```
onSelect={(values, options) => {
  if (values.length > 0 && options.length > 0) {
     const option = options[0];
     onSelect(values[0], option === unsetOption ? undefined : option);
  }
}}
```
Because `values[0]` will be `undefined`, but `options[0]` could be an empty object. So do you think we should return `undefined` or would an empty object be okay in this case?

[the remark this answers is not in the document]
stephenh: > I will fix it, but fwiw, type-check didn't find any issues in internal-frontend.

I was initially really surprised, but I guess in retrospect we probably ~95% of the time use GraphQL inputs where are defacto `someFoo: string | null | undefined`...

> So do you think in the onSelect we'd do something more like:

Right, I think that makes sense, as well at update the TypeScript definition of `onSelect` to be `value: V | undefined`; that is really what I'm nudging for, is that we don't we don't want the client to think "I'm using `SelectField<string>`" and suddenly get `value: undefined` passed to their `onSelect` method.

If all of our blueprint callers are already great with `V | undefined`, that's great, will make it easy to roll out then...

> I really want to break out the logic of the multi select and single select

Nice! +1

src/components/Filters/Filters.tsx
```diff
@@ -60,7 +60,7 @@ function Filters<F, G extends Value = string>(props: FilterProps<F, G>) {
         getOptionValue={(o) => o.id}
         getOptionLabel={(o) => o.name}
         value={groupBy.value}
-        onSelect={(g) => groupBy.setValue(g)}
+        onSelect={(g) => g && groupBy.setValue(g)}
```
stephenh, line 63: Cool, fwiw this is the sort of usage that I was worried/anticipating would be tripped by "now the `value` might be `undefined`".

Granted, we are not setting `unsetLabel` here.

The really fancy version of this PR would try to take these lines:

```
export function BoundSelectField<T, V extends Value>(props: BoundSelectFieldProps<T, V>): JSX.Element;
export function BoundSelectField<T extends HasIdAndName<V>, V extends Value>(
  props: Optional<BoundSelectFieldProps<T, V>, "getOptionLabel" | "getOptionValue">,
): JSX.Element;
```

Which are providing overloads of `SelectField` that saying "oh, your options already have id and name? np I'll implement getOptionLabel/Value for you" and do something similar like: "oh, you _don't_ set `unsetLabel`?, then you can do just `onSelect(value: V)`".

So then callers like this one, who aren't opting in to the `unsetLabel` "extra option", don't need to know about the "extra option" being passed to their `onSelect`. (This imo is part of why the current approach had been working out, in that sure the caller had to know "inject an extra option", but it kept a strict mapping of "your onSelect(V) and options/values always line up 1-1").

So the fancier typing is...probably...doable, but more complex that I really want to chase down at this point, so just mentioning, but this is fine.

stephenh (approved):
Getting the typing perfect here is hard, which is why it was the way it was ("it's just easier if the caller that wants 'none' injects their own none options"), but I think we don't have to make perfect the enemy of good, especially since you've ran this against Blueprint and hardly anyone wants the `Filters.tsx`-esque behavior of binding literally to a `string`-that-cannot-be-`undefined` field. So, sgtm.

src/components/Filters/Filters.tsx
bdow: I see what you mean. That would be nice to infer the user didn't include the `unsetLabel` and change the typing of `onSelect`. Again, I appreciate the thorough and thoughtful feedback.

PR #552, opened 2022-06-07
fix: GridTable - Collapsed rows are included in determining selected rows

bdow:


src/components/Table/GridTable.tsx
```diff
@@ -405,25 +410,29 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     // Split out the header rows from the data rows so that we can put an `infoMessage` in between them (if needed).
     const headerRows: RowTuple<R>[] = [];
     const totalsRows: RowTuple<R>[] = [];
-    const filteredRows: RowTuple<R>[] = [];
+    const visibleDataRows: RowTuple<R>[] = [];
+    const matchedRowIds: string[] = [];
```
stephenh, line 414: Nit/thought: Maybe call this `filteredRowIds`? I like the new `visibleDataRows` is "filtered && not collapsed".

...or maybe the name is good, and just in-source comment that this means something like "rows either explicitly matched by a filter or implicitly matched by a parent that matches the filter".

src/components/Table/GridTable.tsx
```diff
@@ -405,25 +410,29 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
-      filteredRows.push([row, makeRowComponent(row, level)]);
+    function visit([row, children]: ParentChildrenTuple<R>, level: number, visible: boolean): void {
+      let isCard = visible && nestedCards && nestedCards.maybeOpenCard(row);
 
-      const isCollapsed = collapsedIds.includes(row.id);
-      if (!isCollapsed && children.length) {
-        nestedCards && nestedCards.addSpacer();
-        visitRows(children, isCard, level + 1);
+      visible && visibleDataRows.push([row, makeRowComponent(row, level)]);
+      matchedRowIds.push(row.id);
```
stephenh, line 423: Suggestion: I'm thinking maybe an in-source comment here, like:

```
// This row may be invisible (because it's parent is collapsed), but we still want
// to consider it matched if it or it's parent matched a filter.
```

?

src/components/Table/GridTable.test.tsx
```diff
@@ -1445,6 +1445,99 @@ describe("GridTable", () => {
...
+    expect(api.current!.getSelectedRowIds()).toEqual(["p3", "p3c1", "p2", "p2c2", "p2c1", "p1", "p1c2", "p1c1"]);
+
+    // When applying a filter
+    type(r.filter, "filter");
+    // Then each group rows selected state should reflect the children that match the filter
+    expect(cellAnd(r, 1, 1, "select")).toBeChecked();
+    expect(cellAnd(r, 2, 1, "select")).toBeChecked();
+    expect(cellAnd(r, 3, 1, "select")).not.toBeChecked();
+    // And the API reflects the expected selected states of rows that match the filter
+    expect(api.current!.getSelectedRowIds()).toEqual(["p2", "p2c1", "p1", "p1c2", "p1c1"]);
```
stephenh, line 1499: Question: Doesn't `p3` have "filter" in the name "filter parent 3", and so should show up here?

bdow: It isn't selected though, the parent rows selection state is based on the it's matched children. No children match, so it isn't considered selected.

stephenh: Huh, yeah, I guess... Just thinking that p2 is there, even though p2 doesn't have _all_ of its children.

Do we have a case where a filtered & selected parent (both of which seem imo like "...surely this would be considered selected") shouldn't be returned b/c of no matched children? Like a specific page/etc...


Dunno, not a huge deal, just surprised.

stephenh (approved):

src/components/Table/GridTable.test.tsx
bdow: I think in Specs and Selections where our group rows just are roll ups, we don't actually do anything with those rows if they're selected. We just use the select box on those group rows as a way to quickly select all its children. I think that is true for most, if not all, of the grouped tables in Blueprint.
Thinking through a use case where I want to delete line items, should I be able to delete the whole group if it is selected, even if some children under it are hidden by a filter? I am not sure... I'll follow up with UX on that one to tease it out. I could play devil's advocate for either side of that argument.
This is at least a step in the right direction, imo. Thank you for questioning other use cases, though.

PR #553, opened 2022-06-09
fix: SC-11297: Reduce space between label and value for readOnly fields

bdow:


stephenh (approved):
Are we just assuming we won't have any more flip read/edit/read/edit use cases?

bdow:
@stephenh - Right. We're hoping for more of a _always_ auto save experience. But even if we do have a flip from read-only to editable, then we are okay with the content jumping a bit.

PR #554, opened 2022-06-09
fix: Update StaticField to match readOnly input field styles

bdow:


src/forms/StaticField.tsx
```diff
@@ -12,11 +13,13 @@ interface StaticFieldProps {
 export function StaticField(props: StaticFieldProps) {
   const { label, value, children } = props;
   const tid = useTestIds(props, defaultTestId(label));
+  const id = useId();
```
stephenh, line 16: Curious why the `id`?

stephenh (approved):

src/forms/StaticField.tsx
bdow: For accessibility purposes `<label>` elements should be tied to a value via a `for` -> `id` pairing.

stephenh: Ah yeah, that makes sense, I completely missed the `htmlFor={id}` part of the diff. Cool!

PR #557, opened 2022-06-10
fix: auto save status provider

TylerR909:
Adds AutoSaveStatusProvider back in, backed by `@homebound/form-state` upgrade

Already did the `npm run watch-beam` and `yarn ttsc -w` stuff to test it on BluePrint ✅

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "format": "prettier --loglevel warn --write \"**/*.{ts,tsx,css,md}\""
   },
   "dependencies": {
-    "@homebound/form-state": "2.15.1",
+    "@homebound/form-state": "^2.15.2",
     "@internationalized/number": "^3.0.3",
     "@react-aria/utils": "^3.11.3",
     "@react-hook/resize-observer": "^1.2.2",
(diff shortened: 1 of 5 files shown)


Levi-Huynh (approved):

stephenh:
Huh, so @TylerR909 what was wrong with the previous PR? The only difference is this one bumped form-state?

TylerR909:
I honestly couldn't tell you. I spent several hours making direct edits to `node_modules/@homebound/form-state/...` before finally doing [this](https://github.com/homebound-team/form-state/pull/69) (locally) and suddenly the 1 failing test passed.

I don't know why. None of AutoSaveStatus's functions were triggered, Form AutoSave wasn't even being attempted, the other 1630ish tests (including forms, modals, modalforms, etc) all worked. This _one_ test, _uniquely_, somehow was simply dropping all content when wrapped with an AutoSaveStatusProvider, even though there's no dependency, nothing was throwing, and no interactions were being attempted.

TylerR909:
Ah, found some old `r.debug()` in my terminal.

It was rendering _a_ component... but all the values were empty (emdash) for... some... reason...

image

Memoizing a Provider's values 40 steps away, that it doesn't depend on, that nothing is calling, that nobody cares about, was the difference between that and a fully hydrated Row:

image

bdow:
The two rows you've provided screenshots for are different rows. The first is the "totals" row that includes a Add Line Item button and a bunch of dashes in the totals because no line items are in that table for it to calculate yet. 
The second screenshot is the expected row to have been rendered in the test (via the add line item modal).

bdow:
The "nullish" content you're referring to is just the content of the initial render. To me it appeared the component was maybe being rerendered with its initial state, or not rerendering at all. I'm still very much so interested in understanding the root cause. It worked as expected in the app, just not the test.

PR #558, opened 2022-06-10
feat: Stepper Progress Bar Update

DeanGilewicz:
Updates progress bar to be calculated based upon provided `step`'s `status` to unblock Lot Config / Lot Release work - [Lot Config / Lot Release PR](https://github.com/homebound-team/internal-frontend/pull/2195)

- Progress calculated in accordance to provided `step`s completion status:
   - if provided `step`'s 3rd step is the last step to have its `state === "complete` then progress bar is set to it's index, e.g. 2
   - if no step has a `state === "complete"` then progress is reset (set to `-1`)

### DIFF
--- a/src/components/Stepper.tsx
+++ b/src/components/Stepper.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from "react";
+import { useRef } from "react";
 import { useButton, useFocusRing, useHover } from "react-aria";
 import { Icon } from "src/components/Icon";
 import { Css } from "src/Css";
@@ -15,31 +15,15 @@ interface StepperBarProps {
   // The 'value' of the Step that should be displayed
   currentStep: Step["value"];
   onChange: (stepValue: string) => void;
-  resetProgress?: boolean;
 }
 
-export function Stepper({ steps, currentStep, onChange, resetProgress }: StepperBarProps) {
+export function Stepper({ steps, currentStep, onChange }: StepperBarProps) {
   if (steps.length === 0) {
     throw new Error("Stepper must be initialized with at least one step");
   }
 
-  const [progressBarStep, setProgressBarStep] = useState(-1);
-
-  useEffect(() => {
-    if (resetProgress) {
-      return setProgressBarStep(-1);
-    }
-
-    // calc step progress based index of last step that is completed
-    const stepProgress = steps.reduce((acc, step, idx) => {
-      if (step.state === "complete") {
-        acc = idx;
-      }
-      return acc;
-    }, -1);
-
-    setProgressBarStep(stepProgress);
-  }, [currentStep, progressBarStep, resetProgress, steps]);
+  // calc progress based on last completed step - return -1 when no steps completed
+  const progressBarStep = steps.reduce((acc, step, idx) => (step.state === "complete" ? idx : acc), -1);
 
   return (
     <nav aria-label="steps" css={Css.df.fdc.$}>
--- a/src/components/Stepper.tsx
+++ b/src/components/Stepper.tsx
@@ -15,20 +15,31 @@ interface StepperBarProps {
   // The 'value' of the Step that should be displayed
   currentStep: Step["value"];
   onChange: (stepValue: string) => void;
+  resetProgress?: boolean;
 }
 
-export function Stepper({ steps, currentStep, onChange }: StepperBarProps) {
+export function Stepper({ steps, currentStep, onChange, resetProgress }: StepperBarProps) {
   if (steps.length === 0) {
     throw new Error("Stepper must be initialized with at least one step");
   }
 
-  const [progressBarStep, setProgressBarStep] = useState(0);
+  const [progressBarStep, setProgressBarStep] = useState(-1);
+
   useEffect(() => {
-    const currentStepIndex = steps.findIndex((s: Step) => s.value === currentStep) || 0;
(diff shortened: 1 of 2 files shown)


src/components/Stepper.tsx
```diff
@@ -15,20 +15,31 @@ interface StepperBarProps {
...
+
+    // calc step progress based index of last step that is completed
+    const stepProgress = steps.reduce((acc, step, idx) => {
+      if (step.state === "complete") {
+        acc = idx;
+      }
+      return acc;
+    }, -1);
+
+    setProgressBarStep(stepProgress);
```
TylerR909, line 41: Probably just as easy to do this all directly.

```
setProgressBarStep(
  steps.reduce((acc, step, idx) => step.state === "complete" ? idx : acc, -1),
);
```

src/components/Stepper.tsx
```diff
@@ -15,20 +15,31 @@ interface StepperBarProps {
...
 
-  const [progressBarStep, setProgressBarStep] = useState(0);
+  const [progressBarStep, setProgressBarStep] = useState(-1);
+
   useEffect(() => {
-    const currentStepIndex = steps.findIndex((s: Step) => s.value === currentStep) || 0;
-    if (currentStepIndex > progressBarStep) {
-      setProgressBarStep(currentStepIndex);
+    if (resetProgress) {
+      return setProgressBarStep(-1);
```
TylerR909, line 30: So how is this going to work? Whoever implements `<Stepper />` is going to flash ⚡  a `resetProgress` prop for a moment to reset the progress?

DeanGilewicz: I'm not sold on this imp but the reason is so that when a user navigates to the `lots` route, the stepper can render without a progress bar being "filled" in since the "step" is no longer associated with a project so we don't know what is complete ... and therefore, where the max progress bar "fill" should be.

This prop was added to allow the implementor to "reset" the progress bar for instances where the user is on say "options" (or something past "lots") then navigates to "lots" and we need to "clear" the progress bar.

Quick and maybe a little dirty but seems to work for the time being

stephenh: I agree flashing a reset prop is odd; re-reading the code, I'm wondering if we need `progressBarStep` as a separate `useState` at all, and instead should just infer/calc the `progressBarStep` as "iterate through `steps`, and until we hit an `incomplete` step...

Seems like that would handle both "racheting up" as the steps in the `steps` prop become progressively `state: complete`, as well as racheting back down when they go back to `state: incomplete`.

DeanGilewicz: yeah good point - `useState` doesn't seem necessary and I've updated to remove the `reset` prop and just recalc

src/components/Stepper.tsx
```diff
@@ -22,13 +22,8 @@ export function Stepper({ steps, currentStep, onChange }: StepperBarProps) {
...
 
-  const [progressBarStep, setProgressBarStep] = useState(0);
-  useEffect(() => {
-    const currentStepIndex = steps.findIndex((s: Step) => s.value === currentStep) || 0;
-    if (currentStepIndex > progressBarStep) {
-      setProgressBarStep(currentStepIndex);
-    }
-  }, [currentStep]);
+  // calc progress based on last completed step - return -1 when no steps completed
+  const progressBarStep = steps.reduce((acc, step, idx) => (step.state === "complete" ? idx : acc), -1);
```
stephenh, line 26: Question: I can't quite tell, but is this `steps.findIndex(s => s.state === "complete")`?

stephenh (approved):
Nice, this a much smaller diff; if this is all the modifications we need, that's pretty amazing. Thanks!

src/components/Stepper.tsx
bdow, line 26: Could we rename this to `lastCompletedStep` then? Will this prevent a user from going back steps?

stephenh: Will defer to Dean, but I believe the idea is that they could go back, but the progress bar locks at the most-completed-so-far step.

bdow: ahh, okay. Been a while since I looked at this. That makes sense.

bdow: It doesn't want the first index, it wants the last index of `complete`, so maybe:
```
steps.reverse().findIndex((s) => s.state === "complete")
```

bdow: It doesn't want the first index, it wants the last index of `complete`, so maybe:
```
steps.map((s) => s.state).lastIndexOf("complete");
```
?

bdow, line 26: If I understand this right, then the progress bar will show no progress when the user is on the first step. The progress bar is no longer reflecting what step the user is on, but which steps the user has completed, correct? This is a pretty big difference from the original specs for this component. I'm okay with it, just want to make sure that is intended.

bdow (approved):

src/components/Stepper.tsx
DeanGilewicz: Happy to rename variable to `lastCompletedStep`. This only impacts the progress bar "visuals" in terms of how much "fill" it has. The user can still navigate to different steps as before.

DeanGilewicz: yep nice suggestion - will update to use `lastIndexOf`. Much easier to comprehend the intention of the code then the `reduce` :)

PR #566, opened 2022-06-22
feat: add NavLink side contrast variant

0ces:
As part of the design of [SC-17566](https://app.shortcut.com/homebound-team/story/17566/new-development-level-navigation) there's need to add a contrast prop to NavLink side variant.

imagen

### DIFF
--- a/src/components/NavLink.stories.tsx
+++ b/src/components/NavLink.stories.tsx
@@ -26,12 +26,13 @@ export default {
 
 export function BaseStates() {
   const sideNavStyles = getNavLinkStyles("side");
+  const sideContrastNavStyles = getNavLinkStyles("side", true);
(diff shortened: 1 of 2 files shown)


aTrujilloHomeBound:
In the nav link on the story book when I toggle the contrast boolean, I don't see any changes, I'm not sure if this is an issue with storyboook or beam however 
Screen Shot 2022-06-23 at 12 45 39 PM

0ces:
That is because in the NavLink's storybook we render a non interactive version of the components, it's just a showcase for each variant.

aTrujilloHomeBound:
nod nod well looks good to me, I'd approve but I have no power here

bdow (approved):

PR #567, opened 2022-06-28
fix: NavLink contrast prop not working

0ces:
The prop wasn't working as expected as Alex told me in this comment https://github.com/homebound-team/beam/pull/566#issuecomment-1164750843, I was checking on the wrong storybook, this is now fixed with this PR and works as expected.

### DIFF
--- a/src/components/NavLink.tsx
+++ b/src/components/NavLink.tsx
@@ -23,7 +23,7 @@ export interface NavLinkProps extends BeamFocusableProps {
 type NavLinkVariant = "side" | "global";
 
 export function NavLink(props: NavLinkProps) {
-  const { disabled: isDisabled, label, openInNew, ...otherProps } = props;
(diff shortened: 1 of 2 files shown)


src/components/NavLink.tsx
```diff
@@ -81,8 +81,8 @@ export function NavLink(props: NavLinkProps) {
   );
 }
 
-export function getNavLinkStyles(variant: NavLinkVariant, contrast: boolean = false) {
-  return navLinkVariantStyles(contrast)[variant];
+export function getNavLinkStyles(variant: NavLinkVariant, contrast?: boolean) {
+  return navLinkVariantStyles(!!contrast)[variant];
```
bdow, line 85: **thought:** Can we make `contrast` a required param? We can default the value to `false` when we destructure the NavLinkProps on line 26.

0ces: Sure, right away!

PR #574, opened 2022-07-05
feat: [sc-18440] - Snackbar Bottom Offset

DeanGilewicz:
- Add prop via provider to set `bottomOffset` of snackbar
- Add `z-index` to snackbar to have it render above other fixed elements in the viewport


TylerR909 (approved):
Looks great, Dean ;)

src/components/Snackbar/SnackbarContext.tsx
```diff
@@ -6,13 +6,17 @@ export type SnackbarContextProps = { setNotices: React.Dispatch<React.SetStateAc
 
 export const SnackbarContext = createContext<SnackbarContextProps>({ setNotices: () => {} });
 
-export function SnackbarProvider(props: PropsWithChildren<{}>) {
+export function SnackbarProvider(
+  props: PropsWithChildren<{
+    bottomOffset?: number;
```
stephenh, line 11: Question: Isn't this pretty hidden from the app (on purpose) b/c `SnackbarProvider` is only instantiated inside of `BeamProvider`?

I assume this is to get the snackbar to show up above the progress bar, but how are we going to create a bespoke/extra `SnackbarProvider` around the just the lot config steps?

I don't necessarily have a better idea; thinking of either maybe `triggerNotice` taking the `bottomOffset` or maybe the `Stepper` using some sort of beam-internal API to tell `SnackbarProvider` "please put all snackbars about 80px up while I'm mounted".

Assuming I'm guessing the use case correctly.

DeanGilewicz: Yep you are correct - the current `SnackbarProvider` is instantiated inside of `BeamProvider`. I was looking for a way to "one-time" set the `bottomOffset` of the snackbar for x number of child pages (such as our Lot Steps) to position it above the also "fixed to the bottom of the viewport" stepper so render above it.

You are also correct in terms of implementation in `internal-frontend` would be something like:
```
<SnackbarProvider bottomOffset={args.bottomOffset}>
   <StepperContextProvider>
      ....
   </StepperContextProvider>
</SnackbarProvider>
```
Where we use this provider instead of the default one provided by Beam (only difference bottomOffset)

Other idea was to update `triggerNotice` in someway but didn't want to have to set it every time `useSnackbar.tsx` custom hook was used in a component.

stephenh: @TylerR909 wdyt wrt to having multiple `SnackbarProvider`s? 

I can imagine an admittedly unlikely scenario where we have a component hierarchy like:

* BeamContext
  * SomeComponentA
    * Stepper w/new provider
      * SomeComponentB

And `SomeComponentB`s triggers would be pushed up, but if someone "above us" in the hierarchy for some reason triggered a notice while the lot config page was active, it would a) not be pushed up, and b) potentially overlap/conflict with the 2nd provider.

Granted, I get this is pretty unlikely, so I'm fine shipping as-is, unless Tyler agrees with nudging away from multiple providers now, and that it's a pretty simple fix to have some sort of useState/internal state (gotta be a `useRef`) for Stepper to tell Snackbar to push things up...

...I suppose even then I'm assuming that the Stepper is mounted to the bottom of the page, which it may not be (not really a fan of that placement for a number of reasons but :shrug: ).

Yeah, maybe just do this...

TylerR909: Lol, this was my idea. 

`triggerNotice({ bottomOffset })` was our first pass but there were tradeoffs we didn't like, such as every page under LotConfig having to remember to pass one in.

This felt much cleaner to us, and I think it's useful to distinguish between us Convenience and Requirements. Conveniently, we get a SnackbarProvider through BeamContext, but that's not a requirement. AutoSaveStatusProvider provides an instance for free, but Modals or Drawers that need custom behavior are encouraged to use a nested provider for their use cases. This falls in line with that imo. i.e. nothing about it being provided for free via BeamProvider should dissuade us from using it hands-on in the cases we need to, like this.

SomeComponentA/SomeComponentB both calling triggerNotices to alternate providers is a useful callout, but in a game of tradeoffs this one felt the cleanest.

> Stepper to tell Snackbar to push things up...

This, however, is a very interesting idea.

stephenh (approved):

src/components/Snackbar/SnackbarContext.tsx
TylerR909: Mmk here's an experimental branch enabling "Stepper notifies SnackbarProvider of an offset."

https://github.com/homebound-team/beam/compare/experimental/snackbar-offset

Essentially, BluePrint's StepperContextProvider ("SCP") for LotConfig would call the following, where that destructure is an adhoc useEffect:

```jsx
const { useSnackbarOffset } = useSnackbar();
useSnackbarOffset({ bottom: 48 });
```

While the SCP is in the DOM the stepper would be bumped up a bit. The moment it exits the DOM, the useEffect's cleanup runs and returns it back to undefined values.

TylerR909: You can see, whenever TriggerNotice is mounted its offset is active. When it dismounts, offset returns. And whether it's mounted or not, there's only 1 stack of Notifications so no dual-provider concerns.

Kapture 2022-07-05 at 19 04 32

Thought if 2 components were each calling offset... I'm not sure how that fight would go, or what would happen if one of them dismounted.

DeanGilewicz: Yep I like this too. Keeps config / set up to one place while removing the need for an additional provider. Thanks all for the feedback and Tyler for proving this out. @TylerR909 do you just want to go ahead and create the PR? any other cases you want to consider / test for? I've just played around with an example with more than one component being mounted / unmounted with a defined (and different) offset so can connect with you on that if there's value on adding that test case in

TylerR909: Mmk Dean and I just chatted and I'm going to throw up the useSnackbarOffset approach into a PR to succeed this PR. I need to do some quick cleanup/testing.

We tested how 2 components calling it would work together, and basically last-render won the fight (and first-to-dismount unset everything), but it's such an edge case we're going to leave it on the table for now.

DeanGilewicz: Closing this PR in favor of most recent approach - https://github.com/homebound-team/beam/pull/575

DeanGilewicz:
Closing this PR in favor of most recent approach - https://github.com/homebound-team/beam/pull/575

PR #578, opened 2022-07-06
fix: Use label for default test id.

stephenh:


src/forms/BoundCheckboxField.tsx
```diff
@@ -21,7 +22,7 @@ export function BoundCheckboxField(props: BoundCheckboxFieldProps) {
     onBlur,
     ...others
   } = props;
-  const testId = useTestIds(props, field.key);
+  const testId = useTestIds(props, defaultTestId(label));
```
bdow, line 25: Doesn't UseTestIds do this defaulTestId() call?

stephenh: Ah yep!

stephenh: @bdow fixed!

bdow (approved):

PR #580, opened 2022-07-07
Sc 17513/showing nested childrens when filtering parent rows in gridtable

Grandlouis:
[sc-17513] Updating the filterRows function to also return children rows if the parent matches the
 provided filter, updating and adding test cases.

The test case updated is due to now filter is returning all childrens of matched parent, the selected childrens will remain selected even if they don't match the filter provided but the parent does.


src/components/Table/GridTable.tsx
```diff
@@ -1474,24 +1474,25 @@ function filterRows<R extends Kinded>(
...
-  function filterFn(acc: ParentChildrenTuple<R>[], row: GridDataRow<R>): ParentChildrenTuple<R>[] {
+  function filterFn(acc: ParentChildrenTuple<R>[], row: GridDataRow<R>, parentMatches?: boolean): ParentChildrenTuple<R>[] {
     // Break up "foo bar" into `[foo, bar]` and a row must match both `foo` and `bar`
     const filters = (filter && filter.split(/ +/)) || [];
     const matches =
       row.kind === "header" ||
       row.kind === "totals" ||
       filters.length === 0 ||
       !!row.pin ||
+      (filters.length > 0 && parentMatches) || //matching childrens if parent matches to display the nested records on searches
```
stephenh, line 1495: Question: Can this row be just `parentMatches ||`, i.e. do we need the `filters.length > 0` check?

src/components/Table/GridTable.tsx
```diff
@@ -1474,24 +1474,25 @@ function filterRows<R extends Kinded>(
   filter: string | undefined,
 ): ParentChildrenTuple<R>[] {
   // Make a function to do recursion
-  function filterFn(acc: ParentChildrenTuple<R>[], row: GridDataRow<R>): ParentChildrenTuple<R>[] {
+  function filterFn(acc: ParentChildrenTuple<R>[], row: GridDataRow<R>, parentMatches?: boolean): ParentChildrenTuple<R>[] {
```
stephenh, line 1487: Thought: I'm wondering if it'd be slightly cleaner to keep `filterFn` as it was, and then add a new function above this like:

```
function acceptAll(acc: ..., row: ...): ... {
  acc.concat([row, row.children.reduce(acceptAll, []) ?? []);
}
```

And then maybe change the logic to something like:

```
if (matches) {
  return acc.concat([row, row.children?.reduce(acceptAll. []));
} else {
  const matchedChildren = row.children?.reduce(filterFn, []);
  if (matchedChildren.length) {
     return acc.concat([row, matchedChildren]);
  }
  return acc;
}
```

Dunno, maybe it's not a big difference, wdyt?

stephenh (approved):
Nice!

PR #581, opened 2022-07-07
feat: Sort within pinned rows.

stephenh:
I.e. to put all of the favorited documents first, but then within
those still sort by the given column.

This basically becomes a way to add an "always on" first sort.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -712,27 +712,27 @@ describe("GridTable", () => {
           sorting={{ on: "client", initial: [1, "ASC"] }}
           rows={[
             simpleHeader,
-            // And the 1st row is pinned first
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.test.tsx
```diff
@@ -712,27 +712,27 @@ describe("GridTable", () => {
           sorting={{ on: "client", initial: [1, "ASC"] }}
           rows={[
             simpleHeader,
-            // And the 1st row is pinned first
-            { kind: "data", id: "1", pin: "first", data: { name: "a", value: 11 } },
-            { kind: "data", id: "2", pin: "first", data: { name: "b", value: 10 } },
```
stephenh, line 717: I just updated this test case to put the "pin: first" rows at the bottom, and the "pin: last" rows at the top, to make extra sure they were getting proactively moved to the right spot.

PR #582, opened 2022-07-07
fix: Support virtualized content in ModalBody

bdow:
Changes include:
- ModalBody portal div adds 'height: 100%'.
- Consolidated the 'modalBodyRef' and 'contentRef'. Doesn't appear 'contentRef' was actually needed.
- Adjusts styles to ModalBody wrapping element when 'virtualized' prop is defined.
- Adjusts GridTable's resize reference div to accommodate virtuoso's scrollbar.

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -1,12 +1,12 @@
 import { AutoSaveStatusProvider } from "@homebound/form-state";
 import useResizeObserver from "@react-hook/resize-observer";
-import { MutableRefObject, ReactNode, useEffect, useRef, useState } from "react";
+import { MutableRefObject, PropsWithChildren, ReactNode, useEffect, useRef, useState } from "react";
 import { FocusScope, OverlayContainer, useDialog, useModal, useOverlay, usePreventScroll } from "react-aria";
 import { createPortal } from "react-dom";
 import { useBeamContext } from "src/components/BeamContext";
 import { IconButton } from "src/components/IconButton";
 import { useModal as ourUseModal } from "src/components/Modal/useModal";
-import { Css, Only, px, Xss } from "src/Css";
+import { Css, Only, Xss } from "src/Css";
 import { useTestIds } from "src/utils";
 
 export type ModalSize = "sm" | "md" | "lg" | "xl";
@@ -62,7 +62,6 @@ export function Modal(props: ModalProps) {
   const { modalProps } = useModal();
   const { dialogProps, titleProps } = useDialog({ role: "dialog" }, ref);
   const [[width, height], setSize] = useState(getSize(size));
-  const contentRef = useRef<HTMLDivElement>(null);
   const modalBodyRef = useRef<HTMLDivElement | null>(null);
   const modalFooterRef = useRef<HTMLDivElement | null>(null);
   const modalHeaderRef = useRef<HTMLHeadingElement | null>(null);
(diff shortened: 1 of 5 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -121,12 +120,11 @@ export function Modal(props: ModalProps) {
                 </span>
               </header>
               <main
-                ref={contentRef}
+                ref={modalBodyRef}
                 css={Css.fg1.overflowYAuto.if(hasScroll).bb.bGray200.if(!!forceScrolling).overflowYScroll.$}
               >
                 {/* We'll include content here, but we expect ModalBody and ModalFooter to use their respective portals. */}
                 {content}
-                <div ref={modalBodyRef} />
```
bdow, line 129: Consolidated these refs. Didn't really seem necessary to need both. This also made it just simpler to have one less div of nesting for the ModalBody content.

src/components/Modal/Modal.tsx
```diff
@@ -145,11 +143,15 @@ export function ModalHeader({ children }: { children: ReactNode }): JSX.Element
 }
 
 /** Provides consistent styling and the scrolling behavior for a modal's primary content. */
-export function ModalBody({ children }: { children: ReactNode }): JSX.Element {
+export function ModalBody({
+  children,
+  virtualized = false,
```
stephenh, line 148: Thought: It's probably too cute, but iirc this is our 2nd "just know to set `virtualized` based on what you're putting in a parent" ... might be nice if there was something that would let a child tell it's parent "hey i'm virtualized".

Like in gridtable:

```
function GridTable(props) {
  const { setVirtualized } = useBeamContext();

   setVirtualized(props.virtualized);
}
```

And then "special" containers like `ScrollableParent` (?) and `ModalBody` would have providers listening for "ah yeah, my child is actually virtualized" and use that to drive their `virtualized` state/behavior instead of an explicit prop.

@bdow wdyt?

src/components/Modal/Modal.tsx
stephenh: Nice

src/components/Table/GridTable.tsx
```diff
@@ -483,7 +483,9 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   return (
     <RowStateContext.Provider value={rowStateContext}>
       <PresentationProvider fieldProps={fieldProps} wrap={style?.presentationSettings?.wrap}>
-        <div ref={resizeRef} css={Css.w100.$} />
+        {/* If virtualized take some pixels off the width to accommodate when virtuoso's scrollbar is introduced. */}
+        {/* Otherwise a horizontal scrollbar will _always_ appear once the vertical scrollbar is needed */}
```
stephenh, line 487: Great comment

src/components/Table/GridTable.tsx
```diff
@@ -483,7 +483,9 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
+        <div ref={resizeRef} css={Css.w100.if(as === "virtual").w("calc(100% - 20px)").$} />
```
bdow, line 488: I had noticed that a lot of tables in Blueprint are adding a horizontal scrollbar for very little content. The reason is that this `resizeRef` which we use to determine to measure the available space is outside of the Virtuoso scroller. So when Virtuoso adds in a scrollbar, the size of the table isn't adjusted and instead it requires the horizontal scrollbar to see all of the content. 
Writing this out makes me wonder if we should instead conditionally use `resizeRef` or the virtuoso scroller ref to do these calculations from.. Probably that would be a better way of going instead of just saying "remove 20px, maybe? that sounds good"

PR #585, opened 2022-07-08
feat: have persistent rows

aTrujilloHomeBound:


src/components/Table/sortRows.ts
```diff
@@ -11,6 +11,9 @@ export function sortRows<R extends Kinded>(
 ): GridDataRow<R>[] {
   const sorted = sortBatch(columns, rows, sortState, caseSensitive);
   // Recursively sort child rows
+
+  // check for persistent columns her move to top
```
stephenh, line 15: Fwiw my 1st guess would be to just "sort twice".

I.e. somehow get ahold of the `persistent` key in here (from `sortState`?) and, if it's set, do a 1st pass sort on `sortBatch` by somehow tell it to "actually use the persistent column", and then do a 2nd pass / 2nd `sortBatch` to sort on the current.

Like this line:

```
  const [value, direction] = sortState;
  const column = columns[value];
```

Would on the 1st pass use the persistent column index for `value`, and on the 2nd pass use actual `value` / current column index.

That will probably be the least terrible way of making this sort, albeit at the "cost" of doing two passes.

A fancier way of doing it would be to only do 1 pass, but teach the sort to use `persistent` first, i.e. around here:

```
    const v1 = sortValue(applyRowFn(column, a, {} as any, 0), caseSensitive);
    const v2 = sortValue(applyRowFn(column, b, {} as any, 0), caseSensitive);
```

Add something like:

```
    const v1p = sortValue(applyRowFn(persistentColumn, a, {} as any, 0), caseSensitive);
    const v2p = sortValue(applyRowFn(persistentColumn, b, {} as any, 0), caseSensitive);
```

And then work `v1p` and `v2p` into the various expressions below that. Which is doable but seems hard.

src/components/Table/GridTable.tsx
```diff
@@ -174,6 +174,7 @@ export type GridSortConfig<S> =
       /** The optional initial column (index in columns) and direction to sort. */
       initial?: [S | GridColumn<any>, Direction] | undefined;
       caseSensitive?: boolean;
+      persistent?: S | undefined;
```
stephenh, line 177: I think technically would want to do:

```
[S | GridColumn<any>, Direction]
```

So that you could say `[1, "ASC"]` i.e. to put favorites first (or last or what not).

aTrujilloHomeBound: I went back and forth on this,

it is now working with the desired, without allowing the persistent column to be moved to the bottom.

There is not currently a use case for putting favorites last, however this could be added, just more cycles on this.

src/components/Table/sortRows.ts
```diff
@@ -27,21 +27,47 @@ function sortBatch<R extends Kinded>(
...
+    if (persistentColumn) {
+      p1 = sortValue(applyRowFn(persistentColumn, a, {} as any, 0), caseSensitive);
+      p2 = sortValue(applyRowFn(persistentColumn, b, {} as any, 0), caseSensitive);
+    }
     if ((a.pin || b.pin) && !(a.pin === b.pin)) {
       const ap = a.pin === "first" ? -1 : a.pin === "last" ? 1 : 0;
       const bp = b.pin === "first" ? -1 : b.pin === "last" ? 1 : 0;
       return ap === bp ? 0 : ap < bp ? -1 : 1;
-    } else if ((v1e && v2e) || v1 === v2) {
+    } else if (p1 && p2){
```
stephenh, line 57: Suggestion: This is checking truthiness, but if one row is a favorite (p1=1/true) and the other row is now (p2=0/false), then we probably still want to go into this loop and tell the sort that row 1 should come before row 2?

I.e. seems like this should maybe use `else if (persistentColumn)`, like any time we have a persistent sort in play, we should compare p1 and p2 of both rows.

Not just if they're both set, or both truthy.

stephenh: (Fwiw this is the sort of thing where, unless I'm missing something, I'm missing that it looks like sorting is working on the test data that we happen to be testing with, but with various random combinations, it is actually not fully working.)

src/components/Table/GridTable.tsx
```diff
@@ -174,6 +174,7 @@ export type GridSortConfig<S> =
       /** The optional initial column (index in columns) and direction to sort. */
       initial?: [S | GridColumn<any>, Direction] | undefined;
       caseSensitive?: boolean;
+      persistent?: S | GridColumn<any> | undefined;
```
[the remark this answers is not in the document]
aTrujilloHomeBound: I'd like to chat more about this,  

There isn't currently a need for it, however if there is a use case or you think it would just be better to keep it consistent I respect your wisdom.

currently I'm struggling to get this working as is, (working in beam and not in internal frontend)

I will circle back to this once I get things settled, unless you insist otherwise :)

src/components/Table/sortRows.ts
```diff
@@ -27,21 +27,47 @@ function sortBatch<R extends Kinded>(
...
+      if ((v1e && v2e) || v1 === v2) {
```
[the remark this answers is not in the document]
aTrujilloHomeBound: my thought :)

use favoriting for context:
we have several items favorited. then we sort a list based on a different value, we would still like to see those items favorited at the top however sorted based on the sort we selected 

here sorted based off name, with favorites at the top

Screen Shot 2022-07-12 at 2 44 54 PM

src/components/Table/sortRows.ts
aTrujilloHomeBound: nod nod,

below this statement we have 

```
    else if (p1) {
      return -1; 
    } else if (p2) {
      return 1;
    }
```
    
in the case that only one of the persistent values is `true` it will always be sorted above a non persistent value. Furthermore I believe this is also the location to modify if we would like the ability to support sorting for persistent values to the top or bottom of the list .

as for test cases I think there can definitely be more and will try to address these corner cases. 

This is however effectively demonstrated in the `GridTable.stories.tsx:favoritingMore` ..... .terrible name

src/components/Table/sortRows.ts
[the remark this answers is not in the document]
stephenh: I see why this works now; we're assuming p1/p2 are boolean values, i.e. this only handles favorites.

I don't really like that assumption, as it's just creating work/technical debt for the next person who is going to want to persistent sort by a non-boolean column, but I won't block the PR if that's what we decide to do.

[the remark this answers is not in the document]
stephenh: Fwiw maybe extra a `compare` function and then invoke it twice:

```
function sortBatch<R extends Kinded>(
  columns: GridColumn<R>[],
  batch: GridDataRow<R>[],
  sortState: SortState<number>,
  caseSensitive: boolean,
): GridDataRow<R>[] {
  // When client-side sort, the sort value is the column index
  const [value, direction] = sortState;
  const column = columns[value];
  const invert = direction === "DESC";

  // Make a shallow copy for sorting to avoid mutating the original list
  return [...batch].sort((a, b) => {
    if ((a.pin || b.pin) && !(a.pin === b.pin)) {
      const ap = a.pin === "first" ? -1 : a.pin === "last" ? 1 : 0;
      const bp = b.pin === "first" ? -1 : b.pin === "last" ? 1 : 0;
      return ap === bp ? 0 : ap < bp ? -1 : 1;
    } else if (persistentColumn) {
      const pv = compare(persistentColumn, a, b, false, caseSensitive);
      const v = compare(column, a, b, invert, caseSensitive);
      return pv ?? v;
    } else {
      return compare(persistentColumn, a, b, invert, caseSensitive);
    }
    return 0;
  });
}

function compare(column: any, a: GridDataRow<any>, b: GridDataRow<any>, invert: boolean, caseSensitive: boolean) {
  const v1 = sortValue(applyRowFn(column, a, {} as any, 0), caseSensitive);
  const v2 = sortValue(applyRowFn(column, b, {} as any, 0), caseSensitive);
  const v1e = v1 === null || v1 === undefined;
  const v2e = v2 === null || v2 === undefined;
  if ((v1e && v2e) || v1 === v2) {
    return 0;
  } else if (v1e || v1 < v2) {
    return invert ? 1 : -1;
  } else if (v2e || v1 > v2) {
    return invert ? -1 : 1;
  } else {
    return 0;
  }
}
```

src/components/Table/GridTable.stories.tsx
```diff
@@ -1180,6 +1180,76 @@ export function TruncatingCells() {
...
+        // And the data is initially unsorted
+        { kind: "data", id: "2", data: { name: "b", value: 2, favorite: true } },
+        { kind: "data", id: "1", data: { name: "a", value: 3, favorite: false } },
+        { kind: "data", id: "3", data: { name: "c", value: 1, favorite: false } },
+      ]}
+    />
+  );
+}
+
+export function PrimaryColumnSorting() {
```
stephenh, line 1212: Question: What's the difference between these two stories, `favoriting` and `PrimaryColumnSorting`? Can we pick just one?

src/components/Table/GridTable.test.tsx
```diff
@@ -379,6 +379,134 @@ describe("GridTable", () => {
...
+      // Then the `value: 2' row is first
+      expect(cell(r, 1, 0)).toHaveTextContent("b");
+
+      // And the rows were memoized so didn't re-render
+      expect(row(r, 1).getAttribute("data-render")).toEqual("1");
+      expect(row(r, 2).getAttribute("data-render")).toEqual("1");
+      expect(row(r, 3).getAttribute("data-render")).toEqual("1");
+    });
+
+    it("can sort priamry rows", async () => {
```
stephenh, line 423: Similar question, what's the difference between this test and the one on line 382?

src/components/Table/sortRows.ts
```diff
@@ -27,31 +27,47 @@ function sortBatch<R extends Kinded>(
...
+      if (primaryCompare !== 0) {
+        return primaryCompare;
+      }
     }
-    return 0;
+    return compare(column, a, b, invert, caseSensitive);
   });
 }
 
+function compare(column: any, a: GridDataRow<any>, b: GridDataRow<any>, invert: boolean, caseSensitive: boolean) {
```
stephenh, line 55: This `any` was me being lazy in slack and typing psuedo-code, we should be able to do:

```
column: GridColumn<R>
```

Where `R` is a new generic on `compare`.

src/components/Table/GridTable.test.tsx
aTrujilloHomeBound: Old habit over providing the simplest implementation of a feature,  will remove 👍

src/components/Table/sortRows.ts
```diff
@@ -27,31 +27,47 @@ function sortBatch<R extends Kinded>(
...
+      if (primaryCompare !== 0) {
+        return primaryCompare;
+      }
     }
-    return 0;
+    return compare(column, a, b, invert, caseSensitive);
   });
 }
 
+function compare<R extends Kinded>(column: GridColumn<R> , a: GridDataRow<R>, b: GridDataRow<R>, invert: boolean, caseSensitive: boolean) {
```
stephenh, line 55: Nit: run formatter to fix the `,`.

(Fwiw either vscode or webstorm can pretty easily setup "run prettier on every file save")

stephenh (approved):

PR #593, opened 2022-07-15
feat: Button to smartly handle disabling when onClick response is a promise

bdow:


src/components/Button.tsx
```diff
@@ -176,3 +189,7 @@ const iconStyles: Record<ButtonSize, IconProps["xss"]> = {
 
 export type ButtonSize = "sm" | "md" | "lg";
 export type ButtonVariant = "primary" | "secondary" | "tertiary" | "danger" | "text";
+
+function isPromise(obj: void | Promise<void>): obj is Promise<void> {
+  return (typeof obj === "function" || typeof obj === "object") && "then" in obj && typeof obj.then === "function";
```
stephenh, line 194: Thought: I don't think the `typeof obj === function` is necessary? I.e. afaiu the "does `then` exist" is the canonical "is this a promise" check.

src/components/Button.tsx
```diff
@@ -53,8 +57,17 @@ export function Button(props: ButtonProps) {
...
       ...ariaProps,
-      onPress: typeof onPress === "string" ? noop : onPress,
-      elementType: typeof onPress === "string" ? "a" : "button",
+      onPress: asLink
+        ? noop
+        : (e) => {
+            const result = onPress(e);
+            if (isPromise(result)) {
+              setIsDisabled(true);
+              result.finally(() => setIsDisabled(false));
```
stephenh, line 66: Praise: Nice! I forgot about `.finally`

stephenh (approved):

src/components/Button.tsx
stephenh: @bdow I'm reading this late, but I wonder if we need to be careful about "stomping" the user's `disabled` prop...

Like maybe after the call is done, the user starts passing in `disabled=false`, but since our `setIsDisabled` runs after the `useEffect` (I think it might be a race condition about if it does/does not), then we might overwrite it...

So maybe this `setIsDisabled true/false` should be a dedicated `useState` like "setInFlight true/false`, and then we treat disabled as an `or` of the prop or in flight?

TylerR909: lol @stephenh... I was going to make that _exact_ comment then decided to just work on it myself.

bdow: Reverting this change, will bring back and update this in another PR

PR #595, opened 2022-07-15
fix: Add optional param to DateFieldBase to hide calendar icon

apattersonATX-HB:
[SC-18436](https://app.shortcut.com/homebound-team/story/18436/remove-calendar-icon-from-schedule)

Straightforward implementation of optional bool property. 

- `DateFieldBase` will not pass `calendarButton` to start/end adornment prop in any way if this value is true. This will eliminate any extra padding in the input field.
- Added a simple test and story view.

### DIFF
--- a/src/inputs/DateFields/DateFieldBase.test.tsx
+++ b/src/inputs/DateFields/DateFieldBase.test.tsx
@@ -37,14 +37,14 @@ describe(DateFieldBase, () => {
   it("renders correctly without calendar icon button", async () => {
     // Given a DateField without a calendar icon
     const r = await render(
-        <DateFieldBase
-            mode="single"
-            value={undefined}
-            label="Date"
-            onChange={noop}
-            placeholder="Select a date"
-            hideCalendarIcon
-        />,
+      <DateFieldBase
+        mode="single"
+        value={undefined}
+        label="Date"
+        onChange={noop}
+        placeholder="Select a date"
+        hideCalendarIcon
+      />,
     );
     // Calendar icon isn't rendered
     expect(r.queryByTestId("date_calendarButton")).toBeNull()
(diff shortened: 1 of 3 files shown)


src/inputs/DateFields/DateFieldBase.test.tsx
```diff
@@ -34,6 +34,24 @@ describe(DateFieldBase, () => {
     expect(r.date_helperText().textContent).toBe("Helper text");
   });
 
+  it("renders correctly without calendar icon button", async () => {
+    // Given a DateField without a calendar icon
+    const r = await render(
+        <DateFieldBase
```
stephenh, line 40: Nit: Run prettier? This looks 4-space indented and I think prettier will nudge it down to 2-space indented.

stephenh (approved):

src/inputs/DateFields/DateFieldBase.test.tsx
apattersonATX-HB: I appreciate the nitpick! Fixed.

src/inputs/DateFields/DateFieldBase.test.tsx
```diff
@@ -34,6 +34,24 @@ describe(DateFieldBase, () => {
...
+      <DateFieldBase
+        mode="single"
+        value={undefined}
+        label="Date"
+        onChange={noop}
+        placeholder="Select a date"
+        hideCalendarIcon
+      />,
+    );
+    // Calendar icon isn't rendered
```
stephenh, line 49: Nit suggestion: I really like the inline comments here, so +1; even better would be to double down on the Given/When/Then prefixes from BDD and do "// Then the calendar icon isn't render" and "// And the placeholder is still there".

PR #596, opened 2022-07-15
fix: Revert "feat: Button to smartly handle disabling when onClick re…

bdow:
…sponse is a promise (#593)"

This reverts commit 444729f1757f9396f02d66e9c5b56ec717d42f9b.


stephenh (approved):
Makes sense, given context we have a few tests to update in Blueprint and this is blocking other Beam upgrades.

PR #597, opened 2022-07-18
feat: Button to smartly handle disabling when onClick response is a promise

bdow:


src/components/Button.tsx
```diff
@@ -37,9 +37,11 @@ export function Button(props: ButtonProps) {
     contrast = false,
     ...otherProps
   } = props;
-  const showExternalLinkIcon = (typeof onPress === "string" && isAbsoluteUrl(onPress)) || openInNew;
+  const asLink = typeof onPress === "string";
```
stephenh, line 40: Nice asLink var

src/components/Button.tsx
```diff
@@ -53,8 +55,17 @@ export function Button(props: ButtonProps) {
   const { buttonProps, isPressed } = useButton(
     {
       ...ariaProps,
-      onPress: typeof onPress === "string" ? noop : onPress,
-      elementType: typeof onPress === "string" ? "a" : "button",
+      onPress: asLink
```
stephenh, line 58: Just curious, but if the user mashes onPress even while asyncInProgress is true, will this fire again?

stephenh: Oh, I see it; n/m, reviewing on mobile :-)

bdow: No, once the first `onPress` is invoked, then the button immediately becomes disabled stopping the user from double clicking, or mashing on it.

PR #602, opened 2022-07-20
feat: Add destructive styling to MenuItems

khalidwilliams:
SC-18697


bdow (approved):
Nice!

stephenh:
@khalidwilliams fyi since you only have 1 commit, GitHub won't make a combined "squash commit", so the semantic release logic will use your commit's commit message as-is, so `feat` will need to be "git amended" into the commit and `git push -f`d.

PR #604, opened 2022-07-21
feat(bound-select-field): dirty indicator

TylerR909:
I started doing this in BluePrint then realized it should probably have some first-class support in Beam. Shows an indicator if an input is dirty/has not been saved yet.

image

In the future I suppose we could add this to the other BoundXFields.

### DIFF
--- a/src/forms/BoundSelectField.test.tsx
+++ b/src/forms/BoundSelectField.test.tsx
@@ -62,6 +62,30 @@ describe("BoundSelectField", () => {
     // Then formState has the latest value when onBlur is called
     expect(autoSave).toBeCalledWith("s:2");
   });
+
+  it("shows dirty indicator", async () => {
+    const author = createObjectState(formConfig, { favoriteSport: "s:1" });
+    const r = await render(<BoundSelectField field={author.favoriteSport} options={sports} isDirtyIndicator />);
+    r.favoriteSport().focus();
+    click(r.getByRole("option", { name: "Soccer" }));
+    expect(r.favoriteSport_isDirtyIndicator()).toBeInTheDocument();
+  });
+
+  it("ignores isDirtyIndicator if fieldDecoration is provided", async () => {
+    const author = createObjectState(formConfig, { favoriteSport: "s:1" });
+    const r = await render(
+      <BoundSelectField
+        field={author.favoriteSport}
+        options={sports}
+        isDirtyIndicator
+        fieldDecoration={() => <div data-testid="lookforme" />}
+      />,
+    );
+    r.favoriteSport().focus();
+    click(r.getByRole("option", { name: "Soccer" }));
+    expect(r.lookforme()).toBeInTheDocument();
+    expect(r.queryByTestId("favoriteSport_isDirtyIndicator")).not.toBeInTheDocument();
+  });
 });
 
 const formConfig: ObjectConfig<AuthorInput> = {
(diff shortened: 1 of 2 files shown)


src/forms/BoundSelectField.test.tsx
```diff
@@ -62,6 +62,30 @@ describe("BoundSelectField", () => {
...
+    const r = await render(
+      <BoundSelectField
+        field={author.favoriteSport}
+        options={sports}
+        isDirtyIndicator
+        fieldDecoration={() => <div data-testid="lookforme" />}
+      />,
+    );
+    r.favoriteSport().focus();
+    click(r.getByRole("option", { name: "Soccer" }));
```
TylerR909, line 85: I could not get `select(r.favoriteSport, "Soccer")` to work and idk why. Gave some "length of 0" error.

stephenh (commented):
I genuinely appreciate the initiative, but our process is that these changes should go through/come from UX.

I.e. reach out to Radhi and confirm:

a) Do we want to show dirty state on this page (my off-the-cuff "seems great, right?" doesn't really count :-)),
b) How does UX want to show the dirty state?

And then work from their figma.

Also/fwiw, my assumption is that UX would consider this pattern holistically as well, i.e. this PR touches this one bound select field, but we probably need to consider how the pattern works for the other common form fields like text boxes, ...those other fields, etc...

TylerR909:
Totally fair. This entire ticket turned into Tyler Design Hour so yeah I should probably start tossing ideas back to Radhi for verification.

src/forms/BoundSelectField.test.tsx
bdow: `select` helper expects the element to be a `<select>`, which is not the case here.

src/forms/BoundSelectField.tsx
```diff
@@ -5,12 +5,15 @@ import { HasIdAndName, Optional } from "src/types";
 import { maybeCall } from "src/utils";
 import { defaultLabel } from "src/utils/defaultLabel";
 import { useTestIds } from "src/utils/useTestIds";
+import { Css } from "..";
```
bdow, line 8: I'm a bit worried about this import. I think i've seen this before and it blew up `internal-frontend`. Can we make this an absolute path?

src/forms/BoundSelectField.tsx
```diff
@@ -67,6 +72,12 @@ export function BoundSelectField<T extends object, V extends Value>(
             field.focus();
             maybeCall(onFocus);
           }}
+          fieldDecoration={
+            fieldDecoration ??
+            (isDirtyIndicator && field.dirty
+              ? () => <div css={Css.wPx(8).hPx(8).br100.bgYellow600.$} {...testId.isDirtyIndicator} />
+              : undefined)
```
bdow, line 79: **request:** Can we have a storybook example for this?

src/forms/BoundSelectField.tsx
TylerR909: Sure but I may end up closing this PR altogether anyway. Got a little too ambitious with this one.

src/forms/BoundSelectField.tsx
TylerR909: Maybe now that you've added `play`

src/forms/BoundSelectField.tsx
bdow: Okay no problem. I was going to ask if this was a new pattern introduced somewhere. But now just seeing the messages between you and Stephen.

PR #606, opened 2022-07-25
feat: added `Small` width prop for SuperDrawer

0ces:
With this PR a `width` option will be added to the `openInDrawer` from the `useSuperDrawer` hook, when this option is set to `"small"` it will set the SuperDrawer width to `560px`, if this prop is not passed the width will default to `1040px`.

# Usage
```javascript
// Example from `SuperDrawer.stories.tsx`
openInDrawer({
  title: "Drawer Title",
  content: <TestDrawerContent book={Books[0]} />,
  width: "small"
});
```

### DIFF
--- a/src/components/SuperDrawer/SuperDrawer.tsx
+++ b/src/components/SuperDrawer/SuperDrawer.tsx
@@ -7,11 +7,7 @@ import { ButtonGroup, IconButton, OpenInDrawerOpts, useSuperDrawer } from "src/c
 import { useBeamContext } from "src/components/BeamContext";
 import { Css, px } from "src/Css";
 import { useTestIds } from "src/utils";
-
-export enum SuperDrawerWidth {
-  small = 560,
-  normal = 1040
-};
+import { SuperDrawerWidth } from "./utils";
 
 /**
  * Global drawer component.
@@ -70,7 +66,7 @@ export function SuperDrawer(): ReactPortal | null {
   const isDetail = currentContent !== firstContent;
   const title = content === undefined ? '' : currentContent.title || firstContent.title;
 
-  const width = firstContent && firstContent.width === "small" ? SuperDrawerWidth.small : SuperDrawerWidth.normal;
+  const { width = SuperDrawerWidth.Normal } = firstContent ?? {}
 
   return createPortal(
     <AnimatePresence>
--- a/src/components/SuperDrawer/SuperDrawer.tsx
+++ b/src/components/SuperDrawer/SuperDrawer.tsx
@@ -8,6 +8,11 @@ import { useBeamContext } from "src/components/BeamContext";
 import { Css, px } from "src/Css";
 import { useTestIds } from "src/utils";
 
+export enum SuperDrawerWidth {
+  small = 560,
+  normal = 1040
+};
+
 /**
  * Global drawer component.
  *
@@ -65,7 +70,7 @@ export function SuperDrawer(): ReactPortal | null {
   const isDetail = currentContent !== firstContent;
   const title = content === undefined ? '' : currentContent.title || firstContent.title;
 
-  const width = firstContent && firstContent.width === "small" ? 560 : 1040;
+  const width = firstContent && firstContent.width === "small" ? SuperDrawerWidth.small : SuperDrawerWidth.normal;
 
   return createPortal(
     <AnimatePresence>
--- a/src/components/SuperDrawer/SuperDrawer.tsx
+++ b/src/components/SuperDrawer/SuperDrawer.tsx
@@ -65,7 +65,7 @@ export function SuperDrawer(): ReactPortal | null {
   const isDetail = currentContent !== firstContent;
   const title = content === undefined ? '' : currentContent.title || firstContent.title;
 
-  const width = !isDetail && firstContent && firstContent.width === "small" ? 560 : 1040;
+  const width = firstContent && firstContent.width === "small" ? 560 : 1040;
 
   return createPortal(
     <AnimatePresence>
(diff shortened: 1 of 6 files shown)


src/components/SuperDrawer/SuperDrawer.tsx
```diff
@@ -65,6 +65,8 @@ export function SuperDrawer(): ReactPortal | null {
   const isDetail = currentContent !== firstContent;
   const title = content === undefined ? '' : currentContent.title || firstContent.title;
 
+  const width = !isDetail && firstContent && firstContent.width === "small" ? 560 : 1040;
```
bdow, line 68: **request:** This makes me wonder what would happen if you opened a "Detail" page after the opening a small Drawer. Seems like if opening a "Detail" page (a sub page within the Drawer) we could end up changing the width from 560 to 1040. I think we'd be good to omit the `!isDetail` and _always_ determine the width based on the `firstContent`. Then a user doesn't have to also remember to set `width: "small"` when opening sub pages and it can just be based on the initial drawer content's size.

0ces: About that, there are other `1040` hard-coded values in the `SuperDrawerContent` component
imagen
Should we change those too?, I'm not familiar with this, but from what I can see from the Storybook test it is ok to leave those hard-coded

bdow: Hey, sorry. Was in meetings. I believe those values are just for animations. It is telling it to grow from `0` to `1040` in 3ms. If we don't update it, then it'll still work fine, but the animation will really finish growing from 0 to 560 in ~1.5ms. It is not a big deal, but probably would be better for consistent animation timings if we updated these too. Thanks for catching!

src/components/SuperDrawer/SuperDrawer.tsx
```diff
@@ -8,6 +8,11 @@ import { useBeamContext } from "src/components/BeamContext";
 import { Css, px } from "src/Css";
 import { useTestIds } from "src/utils";
 
+export enum SuperDrawerWidth {
+  small = 560,
+  normal = 1040
+};
```
bdow, line 14: **request:** Can we create a `/components/SuperDrawer/utils.ts` that has these width definitions in it? I'd rather `SuperDrawerContent` not import from `SuperDrawer`. It is probably fine, but I typically prefer to move these sort of utility constants to a non-component related file. Doing this will help prevent circular dependency issues (not a problem right now as `SuperDrawer` does not import from `SuperDrawerContent`, but more of a general preference)

I'm also wondering if this makes sense as an `enum`? I'm not against it, but I usually see these sort of values as a regular object that maybe uses an `enum` as the "key". Something like:
```
const superDrawerWidthToPixels: Record<SuperDrawerWidth, number> = {
  [SuperDrawerWidth.Small]: 560,
  [SuperDrawerWidth.Normal]: 1040,
};
```
And if we are introducing an enum for these values, then I think it'd also be worth using that `enum` for the props.
```
export interface OpenInDrawerOpts {
  ...
  width?: SuperDrawerWidth;
}
```

(bear with me as I think "out loud" here, please...) So maybe we keep your SuperDrawerWidth enum as is, with the "name = pixels", but change it from "small" and "normal" to "Small" and "Normal" to just stick our conventions of `enum` naming.

Then when a user calls this they'd do something along the lines of:
```
openInDrawer({ width: SuperDrawerWidth.Small }); // a bit more typing than just doing "small", but will immediately translate to the pixel value
```
And in the SuperDrawer we can do something like:
```
const { width = SuperDrawerWidth.Normal, ... } = firstContent ?? {}
```
Thoughts?

0ces: I prefer the `enum` option over the `Record` option, just for code simplicity.

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -20,6 +21,8 @@ export interface OpenInDrawerOpts {
   /** Adds a callback that is called _after_ close has definitely happened. */
   onClose?: () => void;
   content: ReactNode;
+  /** Adds the hability to change the SuperDrawer width based on some pre-defined values */
```
bdow, line 24: **nit:** Typo on `hability` => `ability`

src/components/SuperDrawer/SuperDrawerContent.tsx
```diff
@@ -29,6 +30,9 @@ export const SuperDrawerContent = ({ children, actions }: SuperDrawerContentProp
 
   // Determine if the current element is a new content element or an detail element
   const { kind } = contentStack.current[contentStack.current.length - 1] ?? {};
+  const firstContent = contentStack.current[0]?.opts as OpenInDrawerOpts;
+
+  const width = (firstContent && firstContent.width) || SuperDrawerWidth.Normal;
```
bdow, line 35: **nit (non-blocking):** I don't have a problem with this, but sometimes seeing a mix of `&&` and `||` operators makes reading more difficult in my opinion. I probably would've done something more like: 
```
const { width = SuperDrawerWidth.Normal } = firstContent ?? {};
```
Again, non-blocking, but just wanted to provide an alternative way to define the `width`.

src/components/SuperDrawer/SuperDrawer.tsx
```diff
@@ -65,6 +66,8 @@ export function SuperDrawer(): ReactPortal | null {
   const isDetail = currentContent !== firstContent;
   const title = content === undefined ? '' : currentContent.title || firstContent.title;
 
+  const { width = SuperDrawerWidth.Normal } = firstContent ?? {}
```
bdow, line 69: Ahh, I see you did it this way here! Nice.

bdow (approved):
Awesome! Thanks for taking the time to address and implement the review comments. I also really appreciate you taking the initiative to build this into Beam. I think a few other teams will be benefiting from this very soon! 🚀

src/components/SuperDrawer/SuperDrawerContent.tsx
0ces: Forgot to change, thanks

PR #608, opened 2022-07-26
feat: add type to chip

crshannon:
[#19307](https://app.shortcut.com/homebound-team/story/19307/chip-configuration)

Added `type` prop to  `Chip` component with styled values of `caution, warning, success`

Screen Shot 2022-07-26 at 11 23 41 AM

### DIFF
--- a/src/components/Chip.test.tsx
+++ b/src/components/Chip.test.tsx
@@ -1,11 +1,17 @@
 import { Chip } from "src/components/Chip";
 import { render } from "src/utils/rtl";
+import {Css} from "../Css";
 
 describe("Chip", () => {
   it("renders", async () => {
     const r = await render(<Chip text="Chip text content" />);
     expect(r.chip().textContent).toBe("Chip text content");
     expect(r.chip()).toHaveAttribute("title", "Chip text content");
+  });
+
+  it("can set type to change background color", async () => {
+    const r = await render(<Chip text="Chip" type="caution" />);
+    expect(r.chip()).toHaveStyle(`background: ${Css.bgYellow200.$}`);
   });
 
   it("can set custom testids", async () => {
--- a/src/components/Chip.test.tsx
+++ b/src/components/Chip.test.tsx
@@ -1,6 +1,6 @@
-import { Chip } from "src/components/Chip";
+import { Chip, ChipTypes } from "src/components/Chip";
 import { render } from "src/utils/rtl";
-import {Css} from "../Css";
+import { Css } from "../Css";
 
 describe("Chip", () => {
   it("renders", async () => {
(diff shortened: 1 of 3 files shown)


stephenh:
@crshannon b/c of how github's "squash and merge" works, you'll need to "git amend" the commit and get "feat:" directly into the commit message.

(The reason is that, for PRs with more than 1 commit, github "squash and merge" creates a synthetic squashed commit and so uses the PR title as the commit message title; but for only-1-commit PRs, it doesn't create that extra commit, and so then the commit message as-is is what semantic-release needs to have a "feat:" tag in.)

src/components/Chip.test.tsx
```diff
@@ -1,5 +1,6 @@
 import { Chip } from "src/components/Chip";
 import { render } from "src/utils/rtl";
+import {Css} from "../Css";
```
stephenh, line 3: Nit: run prettier?

src/components/Chip.tsx
```diff
@@ -27,3 +33,17 @@ export function Chip<X extends Only<Xss<Margin>, X>>(props: ChipProps<X>) {
     </span>
   );
 }
+
+function getStyles(type?: TagType) {
```
stephenh, line 37: Thought: This is just fine, I write `switch`es like this a lot; fwiw another way we also do sometimes is:

```
const typeStyles: Record<TagType, Properties> = {
  [TagType.Caution]: Css.byYellow200.$,
  ...
}

// and in
...typeStyles[type ?? TagType.Neutral];
```

It might be a little bit shorter here.

Granted, both `switch` + `Record` idioms provide exhaustive "if you add a new type, you need to add a new map/case" checking from the compiler, so it's up to you, just mentioning it.

stephenh (approved):

src/components/Chip.tsx
crshannon: Just updated, I prefer to use a `Record` here but was following Tag which used a switch and didn't want to throw things off so thanks for the info.

src/components/Chip.tsx
```diff
@@ -1,23 +1,36 @@
 import React from "react";
 import { usePresentationContext } from "src/components/PresentationContext";
-import { Css, Margin, Only, Xss } from "src/Css";
+import { Css, Margin, Only, Properties, Xss } from "src/Css";
 import { useTestIds } from "src/utils/useTestIds";
 
+type ChipType = "caution" | "warning" | "success" | "neutral";
+
+// exporting for using in type prop as constant - this could be moved and become a global list for colors
+export const ChipTypes: Record<ChipType, ChipType> = {
```
crshannon, line 9: @stephenh - I exposed the `ChipTypes` Record so we can use it for consistency in our implementation values, I noticed people are manually typing in 'success` etc for the `Tag` component in blueprint.

PR #611, opened 2022-07-29
feat: [sc-19337]: Create dateRangeFilter component; update DateFieldB…

apattersonATX-HB:
.

### DIFF
--- a/src/components/Filters/DateRangeFilter.test.tsx
+++ b/src/components/Filters/DateRangeFilter.test.tsx
@@ -1,0 +1,2 @@
+// TODO: Add real tests
+export {}
--- a/src/components/Filters/DateRangeFilter.test.tsx
+++ b/src/components/Filters/DateRangeFilter.test.tsx
@@ -6,7 +6,7 @@ import { click, render, type } from "src/utils/rtl";
 
 describe("DateRangeFilter", () => {
   it("shows date placeholder text when not given an initial value", async () => {
-    const r = await render(<TestFilters defs={{ date: taskCompleteFilter}} />)
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
 
     expect(r.filter_taskCompleted_dateField())
       .toHaveProperty("placeholder", "This is some placeholder text");
@@ -16,7 +16,7 @@ describe("DateRangeFilter", () => {
   });
 
   it("shows clear btn when a value is present", async () => {
-    const r = await render(<TestFilters defs={{ date: taskCompleteFilter}} />)
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
     const dateField = r.filter_taskCompleted_dateField();
 
     expect(r.queryByTestId("xCircle")).toBeNull();
@@ -30,7 +30,7 @@ describe("DateRangeFilter", () => {
   });
 
   it("clears text when clear btn is pressed", async () => {
-    const r = await render(<TestFilters defs={{ date: taskCompleteFilter}} />)
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
     const dateField = r.filter_taskCompleted_dateField();
     fireEvent.focus(dateField);
 
@@ -45,7 +45,7 @@ describe("DateRangeFilter", () => {
   });
 
   it("does not show clear btn when datePicker is open", async () => {
-    const r = await render(<TestFilters defs={{ date: taskCompleteFilter}} />)
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
     const dateField = r.filter_taskCompleted_dateField();
     fireEvent.focus(dateField);
 
--- a/src/components/Filters/DateRangeFilter.test.tsx
+++ b/src/components/Filters/DateRangeFilter.test.tsx
@@ -6,58 +6,46 @@ import { click, render, type } from "src/utils/rtl";
 
 describe("DateRangeFilter", () => {
   it("shows date placeholder text when not given an initial value", async () => {
-    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter }} />);
 
-    expect(r.filter_taskCompleted_dateField())
-      .toHaveProperty("placeholder", "This is some placeholder text");
+    expect(r.filter_taskCompleted_dateField()).toHaveProperty("placeholder", "This is some placeholder text");
 
-    expect(r.filter_taskCompleted_dateField())
-      .toHaveValue("");
+    expect(r.filter_taskCompleted_dateField()).toHaveValue("");
   });
 
   it("shows clear btn when a value is present", async () => {
-    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter }} />);
     const dateField = r.filter_taskCompleted_dateField();
 
     expect(r.queryByTestId("xCircle")).toBeNull();
-    fireEvent.focus(dateField);
 
-    type(dateField, "01/02/20 - 01/19/20")
-
-    fireEvent.blur(dateField);
+    type(dateField, "01/02/20 - 01/19/20");
 
     expect(r.xCircle()).toBeTruthy();
   });
 
   it("clears text when clear btn is pressed", async () => {
-    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter }} />);
     const dateField = r.filter_taskCompleted_dateField();
-    fireEvent.focus(dateField);
 
-    type(dateField, "01/02/20 - 01/19/20")
-
-    fireEvent.blur(dateField);
+    type(dateField, "01/02/20 - 01/19/20");
 
     click(r.xCircle());
 
-    expect(r.filter_taskCompleted_dateField())
-      .toHaveValue("");
+    expect(r.filter_taskCompleted_dateField()).toHaveValue("");
   });
 
   it("does not show clear btn when datePicker is open", async () => {
-    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter }} />);
     const dateField = r.filter_taskCompleted_dateField();
-    fireEvent.focus(dateField);
 
-    type(dateField, "01/02/20 - 01/19/20")
-
-    fireEvent.blur(dateField);
+    type(dateField, "01/02/20 - 01/19/20");
 
     expect(r.xCircle()).toBeTruthy();
 
     fireEvent.focus(dateField);
 
-    expect(r.queryByTestId("xCircle")).toBeNull()
+    expect(r.queryByTestId("xCircle")).toBeNull();
   });
 });
 
--- a/src/components/Filters/DateRangeFilter.test.tsx
+++ b/src/components/Filters/DateRangeFilter.test.tsx
@@ -1,2 +1,73 @@
-// TODO: Add real tests
-export {}
+import { fireEvent } from "@testing-library/react";
+import { useState } from "react";
+import { FilterDefs, Filters } from "src/components/Filters";
+import { ProjectFilter, taskCompleteFilter } from "src/components/Filters/testDomain";
+import { click, render, type } from "src/utils/rtl";
+
+describe("DateRangeFilter", () => {
+  it("shows date placeholder text when not given an initial value", async () => {
+    const r = await render(<TestFilters defs={{ date: taskCompleteFilter}} />)
+
+    expect(r.filter_taskCompleted_dateField())
+      .toHaveProperty("placeholder", "This is some placeholder text");
+
+    expect(r.filter_taskCompleted_dateField())
+      .toHaveValue("");
+  });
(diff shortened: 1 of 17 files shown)


apattersonATX-HB:
Test

src/components/internal/DatePicker/Header.tsx
```diff
@@ -18,3 +18,24 @@ export function Header(props: CaptionProps) {
...
+        <IconButton color={Palette.Gray700} icon="chevronRight" onClick={() => goToMonth(addMonths(displayMonth, 1))} />
+      </div>
+      <div css={Css.df.fdr.jcsb.$ }>
+        <IconButton color={Palette.Gray700} icon="chevronLeft" onClick={() => goToMonth(addYears(displayMonth, -1))} />
+        <h1 css={Css.base.$}>{format(displayMonth, "yyyy")}</h1>
+        <IconButton color={Palette.Gray700} icon="chevronRight" onClick={() => goToMonth(addYears(displayMonth, 1))} />
+      </div>
+    </div>
+  );
+}
```
apattersonATX-HB, line 41: When using dateRangeFilter we have to support undefined. Created this second header to make date selections easier on clear.

Screen Shot 2022-08-08 at 11 50 48 AM

src/inputs/DateFields/DateFieldBase.tsx
```diff
@@ -86,9 +88,11 @@ export function DateFieldBase(props: DateRangeFieldBaseProps | DateSingleFieldBa
     onEnter,
     defaultOpen,
     mode,
+    isRangeFilterField = false,
     ...others
   } = props;
 
+  const isRangeMode = mode === "range";
```
apattersonATX-HB, line 95: Keeping things DRY.

src/inputs/DateFields/DateFieldBase.tsx
```diff
@@ -276,11 +306,15 @@ export function DateFieldBase(props: DateRangeFieldBaseProps | DateSingleFieldBa
...
             setInputValue(v);
             // If changing the value directly (vs using the DatePicker), then we always use the short format
-            const parsed = mode === "range" ? parseDateRange(v, dateFormats.short) : parseDate(v, dateFormats.short);
+            const parsed = isRangeMode ? parseDateRange(v, dateFormats.short) : parseDate(v, dateFormats.short);
             onChange(parsed);
           }
+          // User has deleted all text in field
+          else if (v === undefined) {
+            setInputValue("")
+          }
```
apattersonATX-HB, line 315: EdgeCase bug fix
Currently DateFieldBase actually didn't allow users to clear out text completely so technically if a user were to have a date from months `jan-sep` they couldnt manually change it to anything `oct-sept` since the first `0` char couldn't be removed.

src/inputs/DateFields/DateFieldBase.tsx
```diff
@@ -243,7 +251,23 @@ export function DateFieldBase(props: DateRangeFieldBaseProps | DateSingleFieldBa
...
   // How do other applications handle this defined sizing? Appears they use hard coded widths depending on format, which is similar here (using `size` instead of css `width`).
   // But would also need to allow for the input to be `fullWidth`, which is basically also what we're accomplishing here... so maybe fine?
-  const inputSize = mode !== "range" ? (format === "short" ? 8 : format === "medium" ? 10 : undefined) : undefined;
+  const inputSize = !isRangeMode ? (format === "short" ? 8 : format === "medium" ? 10 : undefined) : undefined;
+
+  // Support input range filter field w/ a clear btn that will appear when overlay is closed and input is not focused
+  const clearButton = (
+    <>
+      {(inputValue !== "" && !state.isOpen) && (
+        <IconButton
```
bdow, line 260: **question(s)** Any reason to not utilize the `clearable` property on `TextFieldBase`? And, is there a specific reason why we do not want to show the "clear" button when the overlay is open. Doesn't seems like a bad thing to me to have that.

src/components/Filters/DateRangeFilter.test.tsx
```diff
@@ -0,0 +1,73 @@
+import { fireEvent } from "@testing-library/react";
+import { useState } from "react";
+import { FilterDefs, Filters } from "src/components/Filters";
+import { ProjectFilter, taskCompleteFilter } from "src/components/Filters/testDomain";
+import { click, render, type } from "src/utils/rtl";
+
+describe("DateRangeFilter", () => {
+  it("shows date placeholder text when not given an initial value", async () => {
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
```
stephenh, line 9: Nit: formatting before the `}`

src/components/Filters/DateRangeFilter.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  it("shows clear btn when a value is present", async () => {
+    const r = await render(<TestFilters defs={{ dateRange: taskCompleteFilter}} />)
+    const dateField = r.filter_taskCompleted_dateField();
+
+    expect(r.queryByTestId("xCircle")).toBeNull();
+    fireEvent.focus(dateField);
+
+    type(dateField, "01/02/20 - 01/19/20")
+
+    fireEvent.blur(dateField);
```
stephenh, line 27: Thought: I think our `type` utility already blurs? If you remove this, does the test still pass?

src/components/Filters/DateRangeFilter.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    click(r.xCircle());
+
+    expect(r.filter_taskCompleted_dateField())
+      .toHaveValue("");
```
stephenh, line 44: Thought: I am a little surprised that this line wrapped; maybe run prettier and see if it unwraps?

src/inputs/DateFields/DateFieldBase.tsx
bdow: Nice! When I was doing this, I had so many issues trying to store local variables and keep TS happy. I ended up just grabbing everything off `props`, but glad we can at least store this one.

src/components/internal/DatePicker/DateRangePicker.tsx
```diff
@@ -13,18 +13,19 @@ export interface DateRangePickerProps {
   onSelect: (range: DateRange | undefined) => void;
   disabledDays?: Matcher | Matcher[];
   dottedDays?: Matcher[];
+  useYearPicker?: boolean;
```
bdow, line 16: This makes me wonder if we should _always_ have this available? @jeffreyqua, what do you think about showing the Year navigation all of the time?

src/forms/BoundDateRangeField.tsx
```diff
@@ -32,8 +32,11 @@ export function BoundDateRangeField(props: BoundDateRangeFieldProps) {
           label={label}
           value={field.value || undefined}
           onChange={(value) => {
-            onChange(value);
-            field.maybeAutoSave();
+            // DateRangeField will return an undefined if part of filter, but we're not doing that here yet
```
bdow, line 35: **question:** Any reason not to do that here yet?

src/inputs/DateFields/DateFieldBase.tsx
```diff
@@ -223,18 +228,21 @@ export function DateFieldBase(props: DateRangeFieldBaseProps | DateSingleFieldBa
...
 
-        if (props.mode === "single" && !isDateRange(d)) {
+        if (!isRangeMode && !isDateRange(d)) {
           props.onChange(d);
           return;
         }
+      } else if (isRangeMode && isRangeFilterField) {
+        props.onChange(undefined);
+        return;
       }
```
bdow, line 243: **question/thought:** I'm thinking we might just want to allow DateRangeField to always support `undefined`. Our props have been updated to support `undefined`, but the implementation is avoiding it here unless it is the filter field. I think we might be creating too many code paths by doing it this way, and instead I would prefer if we allow any DateRangeField to return `undefined` (as the props are going to assume that it can happen anyways)

src/inputs/DateFields/DateFieldBase.tsx
apattersonATX-HB: Our `TextFieldBase` will only show the clear `X` if the field is focused where as we wanted it shown any time the field is filled. 
Only showing it when the datepicker is closed because I wasn't able to rework the overlayTrigger and `useEffect` on ln 209 to clear the field in **ONE** click. It was always requiring **TWO** clicks. Having it removed when closed does keep the user focused on the datePicker since it can be cleared by clicking the same date twice.

src/components/Filters/DateRangeFilter.test.tsx
apattersonATX-HB: Thank you 💪

src/components/Filters/DateRangeFilter.test.tsx
apattersonATX-HB: It does! Removed the `.blur` & `.focus` events

src/components/Filters/DateRangeFilter.test.tsx
apattersonATX-HB: Prettier does unwrap it, updated my settings to actually run prettier more often.

src/forms/BoundDateRangeField.tsx
apattersonATX-HB: None anymore, if we're going to allow `DateRangeField` to support undefined in general.

src/inputs/DateFields/DateFieldBase.tsx
apattersonATX-HB: Makes sense, i'll get that updated and tested. I can see it causing some issues in our filter definitions but i should be able to work it out.

src/components/Filters/DateRangeFilter.tsx
```diff
@@ -0,0 +1,59 @@
...
+  defaultValue?: DV;
+  placeholderText?: string;
+  disabledDays?: Matcher | Matcher[];
+  // For storybook to support showing dateRange and date filters in same Filter component
+  testFieldLabel?: string;
+};
+
+// Using op to align with what is expected from DateFilter at this time.
+// In the future we should remove that coupling and rely on user to map returned value
+export type DateRangeFilterValue<V extends Value> = { op: V; value: DateRange | undefined };
```
stephenh, line 22: Thought: Hm, kind of wondering...you mention we have `op` b/c the backend `DateFilter2` has it, which is true, but the backend type looks like:

```
export interface DateFilter2 {
  op: DateOperation;
  value: Date;
  value2?: Date | null | undefined;
}
```

So it also doesn't use `value: DateRange`, but instead uses two `value` + `value2` `Date`s...

This makes me think the page using this filter is probably going to have to manually map this value anyway, and so maybe we should just skip taking `op` all together (since the rationale of 'matching the backend' isn't 100% right anyway? Afaiu?)

src/components/Filters/Filters.stories.tsx
```diff
@@ -116,6 +117,14 @@ function TestFilterPage({ vertical }: { vertical?: boolean }) {
       defaultValue: { op: "BEFORE", value: jan1 },
     });
 
+    const dateRange = dateRangeFilter({
+      label: "Completion Estimate Range",
+      testFieldLabel: "Dates",
+      placeholderText: "All",
+      // Providing a default value, otherwise the default date in the DateField will be today's date, which will cause storybook diffs every day.
```
stephenh, line 124: Praise: nice in-source comment!

src/components/internal/DatePicker/Header.tsx
```diff
@@ -18,3 +18,24 @@ export function Header(props: CaptionProps) {
...
+    <div css={Css.df.jcsb.aic.mlPx(12).mrPx(12).hPx(32).$}>
+      <div css={Css.df.fdr.jcsb.$}>
+        <IconButton color={Palette.Gray700} icon="chevronLeft" onClick={() => goToMonth(addMonths(displayMonth, -1))} />
+        <h1 css={Css.base.$}>{format(displayMonth, "MMM")}</h1>
+        <IconButton color={Palette.Gray700} icon="chevronRight" onClick={() => goToMonth(addMonths(displayMonth, 1))} />
+      </div>
+      <div css={Css.df.fdr.jcsb.$}>
+        <IconButton color={Palette.Gray700} icon="chevronLeft" onClick={() => goToMonth(addYears(displayMonth, -1))} />
+        <h1 css={Css.base.$}>{format(displayMonth, "yyyy")}</h1>
+        <IconButton color={Palette.Gray700} icon="chevronRight" onClick={() => goToMonth(addYears(displayMonth, 1))} />
```
stephenh, line 37: Neat

src/components/Filters/DateRangeFilter.tsx
apattersonATX-HB: I had gone back forth on that as well. When i ran this branch in blueprint, our GQL types in `JobLogsPage` really hated leaving out `op` in the filterDef and we were already mapping our usages of `dateFilter` in the same way with `value1` so i figured no harm in keeping it around _for now._ Removing it will just require adding a little more logic to make TS happy when the filter is used. 

If we remove `op` i'd like to make sure we track this tech debt somewhere to see about adding a new GQL resolver or util for all of our dates and cut out these `} as DateFilter2` for a hack day or something if you agree....that'd just go into the confluence page correct?

src/components/internal/DatePicker/DateRangePicker.tsx
apattersonATX-HB: @jeffreyqua ☝️

stephenh (approved):
Hey @apattersonATX-HB sorry about the delay here; feel free to ping if it seems like code reviewers have gone silent :-) but I think this lgtm, at least to try it and see what happens. :-)

PR #612, opened 2022-08-05
feat: (SC-7409) Auto-collapse completed tasks

pgamboa07:
Expose isCollapse and toggleCollapse to be able to auto collapse the rows


blambillotte (approved):
Makes sense to me - tagging @bdow for a double check though

bdow (commented):
Makes sense to me. Would love to see a test that validates that these new methods work as expected.
I am also a little bit curious to see how this is going to be used. Based on the title for this PR, it sounds like a once a user completes all tasks within a group, then we're going to collapse the group. Which, to me, seems like a pretty jarring user experience. They just clicked something, then everything shifts around. 🤷

pgamboa07:
Closed because changed the expected behavior for the ticket

PR #613, opened 2022-08-08
feat: [SC-19459] Filterable Pins

Grandlouis:
[SC-19459] - Adding the new Pin type with the filter boolean property to indicate the pin should be hidden while filtering in case it doesn't match the search term, adding stories and test for this scenario

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -141,8 +141,25 @@ export function VirtualFilteringWithFilterablePin() {
       ...zeroTo(1_000).map((i) => ({
         kind: "data" as const,
         id: String(i),
-        data: { name: `ccc ${i}`, value: i },
-        pin: i === 500 ? { at: "first" as const, filter: true } : undefined,
+        data: {
+          name:
+            i === 200
+              ? `ccc ${i} - pin first, filter true`
+              : i === 300
+              ? `ccc ${i} - pin first, filter false`
+              : i === 400
+              ? `ccc ${i} - pin first no filter`
+              : `ccc ${i}`,
+          value: i,
+        },
+        pin:
+          i === 200
+            ? { at: "first" as const, filter: true }
+            : i === 300
+            ? { at: "first" as const, filter: false }
+            : i === 400
+            ? ("first" as const)
+            : undefined,
       })),
     ],
     [],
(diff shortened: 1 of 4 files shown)


src/components/Table/sortRows.ts
```diff
@@ -40,8 +40,10 @@ function sortBatch<R extends Kinded>(
       // If both rows are pinned, we don't sort within them, because by pinning the page is taking
       // explicit ownership over the order of the rows (and we also don't support "levels of pins",
       // i.e. for change events putting "just added" rows `pin: last` and the "add new" row `pin: lastest`).
-      const ap = a.pin === "first" ? -1 : a.pin === "last" ? 1 : 0;
-      const bp = b.pin === "first" ? -1 : b.pin === "last" ? 1 : 0;
+      const ap =
```
stephenh, line 43: Suggestion: Maybe add a helper method like `getPin(a)` so this could be either:

```
const ap = getPin(a.pin) === "first" ? -1 : getPin(a.pin) === "last" ? 1 : 0;
```

Or:

```
const aAt = getAt(a.pin);
const ap = aPin === "first"  ? -1 : aPin === "last" ? 1 : 0;
const bAt = getAt(b.pin);
```

src/components/Table/GridTable.stories.tsx
```diff
@@ -134,6 +134,128 @@ export function VirtualFiltering() {
...
 
+export function VirtualFilteringWithFilterablePin() {
+  const rows: GridDataRow<Row>[] = useMemo(
+    () => [
+      simpleHeader,
+      ...zeroTo(1_000).map((i) => ({
+        kind: "data" as const,
+        id: String(i),
+        data: { name: `ccc ${i}`, value: i },
+        pin: i === 500 ? { at: "first" as const, filter: true } : undefined,
```
stephenh, line 145: Afaict both this line and line 231 are using `filter: true`, and I was expecting to see a `filter: false`...

Oh, is the difference that one story is virtual and the other is not?

I dunno, each of these stories at quite a few LOC and, with the tests, I'm not sure we really need three of them, or really any of them?

Like if we _really_ needed a story for filtering, it would be much simpler to have a small story with like three rows, where one row had `filter: true`, one row had `fliter: false`, and the 3rd had no filter. Or something like that.

As-is it seems like we're copy/pasting already-pretty-complicated stories and making them even more complicated. We should try to simplify and focus on just as small as use case as possible.

That said, I think let's just skip stories for this change all together, and assume the unit tests will cover it.

src/components/Table/GridTable.test.tsx
```diff
@@ -710,6 +710,62 @@ describe("GridTable", () => {
...
+            // And the middle rows need sorted
+            { kind: "data", id: "3", data: { name: "c", value: 3 } },
+            { kind: "data", id: "4", data: { name: "d", value: 1 } },
+            // And the last rows are pinned last
+            { kind: "data", id: "5", pin: { at: "last", filter: true }, data: { name: "e", value: 20 } },
+            { kind: "data", id: "6", pin: { at: "last", filter: true }, data: { name: "f", value: 21 } },
+          ]}
+        />,
+      );
+      // Then the a/b rows stayed first
```
stephenh, line 759: Suggestion: Similar to the story, I feel we copy/pasted an already-complicated / dedicated-boundary-case test...like when I read these assertions we're talking about first/middle/last rows changing orders...but none of that is really related to the change in code you just made.

I'd be tempted to remove this 2nd test, and make the 1st test as simple as possible by only having 3 rows, one with `filter: true` (that does not match the filter so is not shown) one with `filter: false` (that does not match the filter but _is_ shown), and one w/o any pin/filter set (that does not match so is not shown).

src/components/Table/GridTable.tsx
```diff
@@ -1000,15 +1000,22 @@ export type GridDataRow<R extends Kinded> = {
   id: string;
   /** A list of parent/grand-parent ids for collapsing parent/child rows. */
   children?: GridDataRow<R>[];
-  /** Whether to pin this sort to the first/last of its parent's children. */
-  pin?: "first" | "last";
+  /** Whether to pin this sort to the first/last of its parent's children. Providing Pin `filter` property will allow pinned rows to be hidden while filtering
```
stephenh, line 1003: Suggestion: Maybe format this comment b/c it's really long:

```suggestion
  /**
    * Whether to pin this sort to the first/last of its parent's children.
    *
    * By default, pinned rows are always shown/not filtered out, however providing
    * the pin `filter: true` property will allow pinned rows to be hidden
    * while filtering.
    */
```

src/components/Table/GridTable.tsx
```diff
@@ -1503,7 +1510,7 @@ function filterRows<R extends Kinded>(
       return acc.concat([[row, row.children?.reduce(acceptAll, []) ?? []]]);
     } else {
       const matchedChildren = row.children?.reduce(filterFn, []) ?? [];
-      if (matchedChildren.length > 0 || row.pin) {
+      if (matchedChildren.length > 0 || typeof row.pin === "string" || row.pin?.filter === false) {
```
stephenh, line 1513: Suggestion: I think `row.pin?.filter !== true` would be slightly better b/c if the user doesn't set `pin.filter`, then we still want to treat it like `filter: false`.

src/components/Table/GridTable.stories.tsx
Grandlouis: okay, i think we can leave one story then, with 3 types of pins, at least as an example for people to be aware of the differences

Grandlouis:
@stephenh feedback done, please review it again 🤓

src/components/Table/GridTable.test.tsx
```diff
@@ -710,6 +710,32 @@ describe("GridTable", () => {
...
+          filter={"d"}
+          columns={[nameColumn, valueColumn]}
+          sorting={{ on: "client", initial: [1, "ASC"] }}
+          rows={[
+            simpleHeader,
+            // pin that is filterable
+            { kind: "data", id: "1", pin: { at: "first", filter: true }, data: { name: "a", value: 11 } },
+            // pin that is not filterable
+            { kind: "data", id: "2", pin: { at: "first", filter: false }, data: { name: "b", value: 10 } },
+            // traditional pin that by default is not filterable
```
stephenh, line 727: Cool, great in-source comments!

src/components/Table/GridTable.test.tsx
```diff
@@ -710,6 +710,32 @@ describe("GridTable", () => {
...
       ).rejects.toThrow("Column 0 passed an unsortable value, use GridCellContent or clientSideSort=false");
     });
+
+    it("can pin rows and filter - filterable pin", async () => {
+      // Given the table with 3 rows with pins
+      const r = await render(
+        <GridTable<Row>
+          filter={"d"}
+          columns={[nameColumn, valueColumn]}
+          sorting={{ on: "client", initial: [1, "ASC"] }}
```
stephenh, line 720: Thought: given we're sorting, I think technically the "name: a" row with a `value: 11` could actually be the 3rd row right now and our assertions below would not catch it?

Kinda wonder, maybe just leaving off the `sorting` prop all together? We don't really care how/if the rows are sorted for this test case, we just want to know if they're shown/not shown.

That would simplify the test + remove this "...maybe the 1st row is just sorted last?" ambiguity.

src/components/Table/GridTable.stories.tsx
```diff
@@ -134,6 +134,66 @@ export function VirtualFiltering() {
...
+export function VirtualFilteringWithFilterablePin() {
+  const rows: GridDataRow<Row>[] = useMemo(
+    () => [
+      simpleHeader,
+      ...zeroTo(1_000).map((i) => ({
+        kind: "data" as const,
+        id: String(i),
+        data: {
+          name:
+            i === 200
```
stephenh, line 146: Thought: I know this is super picky, but math like this is something we try to avoid (see the "Generally Avoid Math in Tests" section https://homebound.atlassian.net/wiki/spaces/ENG/pages/2347171841/Testing+Style+Guide)

The story you're copy/pasting had 1,000 rows precisely b/c it wanted to test how "really large tables" behavior/visualize/interact; for what you're doing, we could/should just ~2-3 rows, make it as dumb/simple as possible, so that when we open up the story in ~6-9 months we know "okay, okay, I only have to think about these three rows" and not the other ~997 we're making that are just going to be a distraction.

stephenh (approved):

bdow (approved):
Awesome! Great addition

PR #618, opened 2022-08-10
feat: add outline-flag icon

0ces:
This icon contains hard-coded attributes since our current Icon/IconButton don't support any stroke property for changing it (@bdow should I implement this instead of hard-coding those values?).

imagen

### DIFF
--- a/src/components/Icon.tsx
+++ b/src/components/Icon.tsx
@@ -323,6 +323,12 @@ export const Icons = {
   flag: (
     <path d="M19 4H6V2H4V20H3V22H4H6H7V20H6V15H19C19.553 15 20 14.552 20 14V5C20 4.448 19.553 4 19 4Z" />
   ),
+  outlineFlag: (
+    <>
+      <path xmlns="http://www.w3.org/2000/svg" d="M6.38464 14H20.3846V5H6.38464" color="transparent" stroke="black" stroke-width="2" stroke-linejoin="round"/>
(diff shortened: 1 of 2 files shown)


src/components/Icon.tsx
```diff
@@ -323,6 +323,12 @@ export const Icons = {
   flag: (
     <path d="M19 4H6V2H4V20H3V22H4H6H7V20H6V15H19C19.553 15 20 14.552 20 14V5C20 4.448 19.553 4 19 4Z" />
   ),
+  outlineFlag: (
+    <>
+      <path xmlns="http://www.w3.org/2000/svg" d="M6.38464 14H20.3846V5H6.38464" color="transparent" stroke="black" stroke-width="2" stroke-linejoin="round"/>
+      <path xmlns="http://www.w3.org/2000/svg" d="M6.38464 2V11V21M8.38464 21H4.38464" stroke="black" stroke-width="2" stroke-linejoin="round"/>
```
bdow, line 329: Hey @0ces, Thanks for calling out the "stroke" styles. I do not believe we want to start introducing "stroke" for our SVGs. This would be the only SVG that uses stroke, and having to know when to use `stroke` to define the colors vs `fill` would require developers to inspect the SVG code for these icons. It would be preferable to have stick with conventions and have these lines drawn and colored using `fill`. Would you mind asking the designer to update this icon to use fill instead of stroke to stick with conventions? If you receive any push back, I'd be happy to jump in a conversation to explain why introducing stroke could introduce some problems.

Also, if possible, we want to only define the `d` attribute on the `<path />` elements. `xmlns` is unnecessary, as the enclosing `<svg />` element already defines that namespace.

bdow (approved):

PR #620, opened 2022-08-15
feat: Add LineTab style option

blambillotte:


src/components/Tabs.tsx
```diff
@@ -264,15 +287,41 @@ function TabImpl<V extends string>(props: TabImplProps<V>) {
...
+    case "lineTab":
+      // Decrease the bottom padding by the same amount of the new border width to prevent text layout shift
+      const borderBottomWidthPx = 4;
+      const borderBottomStyles = Css.bb
+        .add("borderBottomWidth", `${borderBottomWidthPx}px`)
+        .pbPx(verticalPaddingPx - borderBottomWidthPx).$;
+
+      return {
+        baseStyles: sharedBaseStyles.sm.$,
+        activeStyles: Css.add(borderBottomStyles).bLightBlue700.smEm.gray900.$,
```
blambillotte, line 307: concatenating styles with `.add(foo)` seemed nice/cute, but maybe not conventional? LMK if object spreads are preferred

src/components/Tabs.tsx
```diff
@@ -34,6 +34,8 @@ export interface Tab<V extends string = string> {
 
 type TabsContentXss = Xss<Margin>;
 
+export type TabType = "default" | "lineTab";
```
stephenh, line 37: Thought: Maybe include an in-source comment about what `lineTab` is? Afaict it's just a thicker line on the bottom?

Also naive question, but if someone on UX is asking for this b/c "they like this tab style better", should we just change all of our tabs to be this new style? Like why we keep the old one we don't like anymore?

(Disclaimer these questions to UX usually go like 50/50, but :crossed_fingers: they're fine with "ah sure, new tabs everywhere!")

If they do want to keep both styles, I'd really like (cc @bdow and @jeffreyqua ) to work into our component library documentation the rationale for "when you use default tabs" vs. "when to use line tabs", as I'd like to think there is some sort of rule-of-thumb driving this change more than just "I think line tabs look better".

src/components/Tabs.tsx
stephenh: Seems fine!

stephenh (approved):

src/components/Tabs.tsx
bdow: I agree with this comment. I believe the only reason why we are keeping both around for now is because we aren't sure how the new style will look on existing pages. @jeffreyqua can you confirm the reasoning for needing two styles?

src/components/Tabs.tsx
```diff
@@ -173,11 +178,27 @@ export function Tabs<V extends string>(props: TabsProps<V, {}> | RouteTabsProps<
...
   }
 
+  function lineTabBorderStyles() {
+    if (!includeLineTabBorder) return {};
+
+    if (includeLineTabBorder && type !== "lineTab") {
+      throw new Error("Prop `includeLineTabBorder` is currently only compatible with `type` of `lineTab`");
+    }
+
+    return { ...Css.bb.bGray200.$ };
```
bdow, line 188: **nit:** could just return `Css.bb.bGray200.$`, no need for the spread.

src/components/Tabs.tsx
blambillotte: Okay just walked through it with Jeff Qua after reviewing the current uses of Tabs and we're good to make this change a style replacement. Just pushed up those changes for a new review

stephenh: That's amazing @blambillotte (and @jeffreyqua !), thanks for taking the extra time to do this!

stephenh (approved):

PR #622, opened 2022-08-17
feat: reveal on row hover

asanchezHB:
[SC-19741](https://app.shortcut.com/homebound-team/story/19741/beam-add-revealonhover-to-beam-tables)
# Description
 - Adds ability to reveal a cell on row hover

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -1289,3 +1289,29 @@ export function SelectableRows() {
     </>
   );
 }
+
+export function RevealOnRowHover() {
+  const nameColumn: GridColumn<Row> = {
+    header: "Name",
+    data: ({ name }) => ({ content: name }),
+    mw: "160px",
+  };
+
+  const revealColumn: GridColumn<Row> = {
+    header: "Reveal",
+    data: ({ value }) => ({ content: <span>{value}</span>, revealOnRowHover: true }),
+    mw: "160px",
+  };
+
+  return (
+    <>
+      <GridTable
+        columns={[nameColumn, revealColumn]}
+        rows={[
+          simpleHeader,
+          { kind: "data", id: "1", data: { name: "Hover over me to reveal the number!", value: 7 } },
+        ]}
+      />
+    </>
+  );
+}
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -1294,13 +1294,11 @@ export function RevealOnRowHover() {
   const nameColumn: GridColumn<Row> = {
     header: "Name",
     data: ({ name }) => ({ content: name }),
-    mw: "160px",
   };
 
   const revealColumn: GridColumn<Row> = {
     header: "Reveal",
     data: ({ value }) => ({ content: <span>{value}</span>, revealOnRowHover: true }),
-    mw: "160px",
   };
 
   return (
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -2293,7 +2293,7 @@ describe("GridTable", () => {
     expect(cellAnd(r, 2, 0, "select")).toBeChecked();
 
     // And the api can fetch them
-    expect(api.current!.getSelectedRowIds()).toEqual(["2", "1"]);;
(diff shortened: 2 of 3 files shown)


src/components/Table/GridTable.stories.tsx
```diff
@@ -1289,3 +1289,29 @@ export function SelectableRows() {
     </>
   );
 }
+
+export function RevealOnRowHover() {
+  const nameColumn: GridColumn<Row> = {
+    header: "Name",
+    data: ({ name }) => ({ content: name }),
+    mw: "160px",
```
bdow, line 1297: **nit:** The `mw` properties are probably unnecessary in this storybook file.

src/components/Table/GridTable.test.tsx
```diff
@@ -2293,7 +2293,7 @@ describe("GridTable", () => {
     expect(cellAnd(r, 2, 0, "select")).toBeChecked();
 
     // And the api can fetch them
-    expect(api.current!.getSelectedRowIds()).toEqual(["2", "1"]);;
+    expect(api.current!.getSelectedRowIds()).toEqual(["2", "1"]);
```
asanchezHB, line 2296: Extra `;` removed

src/components/Table/GridTable.stories.tsx
```diff
@@ -1289,3 +1289,29 @@ export function SelectableRows() {
...
+  };
+
+  const revealColumn: GridColumn<Row> = {
+    header: "Reveal",
+    data: ({ value }) => ({ content: <span>{value}</span>, revealOnRowHover: true }),
```
bdow, line 1301: **question:** I am curious to know if this works when the `content` does not return an element. For instance, if this works with:
```
data: ({ value }) => ({ content: value, revealOnRowHover: true })
```
I'm thinking it might not based on the CSS selector. We may need to tell the user as a JSDoc above the `revealOnRowHover` prop that the content of the cell needs to be an element in order to work.

src/components/Table/GridTable.tsx
```diff
@@ -1093,6 +1099,10 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
     // Maybe add the sticky header styles
     ...((isHeader || isTotals) && stickyHeader ? Css.sticky.topPx(stickyOffset).z2.$ : undefined),
     ...getNestedCardStyles(row, openCardStyles, style, isActive),
+    ...{
+      "> .revealOnRowHover > *": Css.invisible.$,
+      ":hover > .revealOnRowHover > *": Css.visible.$,
```
bdow, line 1104: **suggestion:** Can these use the `revealOnRowHoverClass` const?
```
` > .${revealOnRowHoverClass} > *`: Css.invisible.$,
...
```

asanchezHB: ah I tried using that initially but kept getting a type error. Seems like it doesn't like using a template string as key?

src/components/Table/GridTable.stories.tsx
asanchezHB: let me give it a try!

src/components/Table/GridTable.tsx
bdow: Maybe try:
```
[` > .${revealOnRowHoverClass} > *`]: Css.invisible.$,
...
```

asanchezHB: ah that works, interesting that it needs to be wrapped

src/components/Table/GridTable.stories.tsx
asanchezHB: hmm I got the same results as we saw during vc with the different tests we tried

bdow: Which was what - that the cell's content value **isn't** hidden? I am guessing that would be the case because our CSS selector is not hiding the cell, but any direct children within the cell. I was curious if a direct value, not wrapped in an element, would hide or not. But sounds like it doesn't hide, which is kind of expected as there are no elements nested in the cell, just `textContent`.

bdow: Based on that, I'd suggest updated the JS Doc for the `revealOnRowHover` to specify that it requires the content of the cell to be wrapped in an element to be hidden.

stephenh (approved):

PR #624, opened 2022-08-23
feat: [SC-20356]  forceSameBrowserTab to make safari open link in same window

Grandlouis:
Adding forceSameBrowserTab to MenuItemBase to allow safari to open the link provided in the same browser window.

Explanation:
`window.open(url, "_blank", "noopener,noreferrer")` works as expected on all main browser, but with safari there is a particular behavior.

in Safari -> Preferences -> Tabs -> Open pages in tabs instead of windows is set to `Automatic`. Using the `features` param in `windown.open` makes the browser to open a new tab.

Adding this param makes the button to use the `window.open(url, target)` syntax, that gets by default a new tab behavior, and forces the `opener` to null to prevent exposin the origin window tab. As a caveat, this does exposes the referrer.

This behavior is also overridable if safari option is set to `Never` or `Always` so is not that `forced` really.

Please review it, i'm not 100% sure is required, due to is only an issue on safari and is option controled.

### DIFF
--- a/src/components/ButtonMenu.tsx
+++ b/src/components/ButtonMenu.tsx
@@ -47,8 +47,6 @@ type MenuItemBase = {
   onClick: string | VoidFunction;
   disabled?: boolean;
   destructive?: boolean;
-  /** If using a string as the onClick value, and setting this to true will force the browser to open the link in the same browser window as a new tab */
-  forceSameBrowserTab?: boolean;
 };
 
 export type IconMenuItemType = MenuItemBase & {
--- a/src/components/ButtonMenu.tsx
+++ b/src/components/ButtonMenu.tsx
@@ -47,6 +47,8 @@ type MenuItemBase = {
   onClick: string | VoidFunction;
   disabled?: boolean;
   destructive?: boolean;
+  /** If using a string as the onClick value, and setting this to true will force the browser to open the link in the same browser window as a new tab */
+  forceSameBrowserTab?: boolean;
 };
 
 export type IconMenuItemType = MenuItemBase & {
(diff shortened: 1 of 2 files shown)


src/components/ButtonMenu.tsx
```diff
@@ -47,6 +47,8 @@ type MenuItemBase = {
   onClick: string | VoidFunction;
   disabled?: boolean;
   destructive?: boolean;
+  /** If using a string as the onClick value, and setting this to true will force the browser to open the link in the same browser window as a new tab */
+  forceSameBrowserTab?: boolean;
```
stephenh, line 51: Ah okay, I think I misunderstood. Fwiw `forceSameBrowserTab` made me think we wanted literally the same tab, i.e. `_self`, behavior.

But we're not really concerned with `_self` vs. `_blank`, we always want `_blank`, and we always want it to be "new tab _in the same window".

And the issue was that Safari seeing `noopener.noreferrer` meant it was "new tab but _new window_"? 

If my new understanding is correct, it seems like the "fixed for Safari" behavior is probably want everyone wants, and we should add flag to request the behavior.

I think personally I'd be fine with something like changing line 37 to be:

```
// We want to do `window.open(url, "_blank", "noopener,noreferrer")` but that Safari treats
// that as "open in new window", see ...url...
//
// So we do this instead, and at least null out the opener
// as a way to manually mimic the `"noopener"` flag.
window.open(onClick, "_blank") as Window).opener = null;
```

Grandlouis: @stephenh changes updated ✅

stephenh (approved):
Cool, thanks!

PR #626, opened 2022-08-24
feat: Support HTML attributes, such as data-testid, on GridCellContent

bdow:


src/components/Table/GridTable.test.tsx
```diff
@@ -2306,6 +2306,72 @@ describe("GridTable", () => {
...
+  it("can set test ids on cell elements", async () => {
+    // Given different row types for each render function
+    type HeaderRender = { kind: "header"; id: string; data: {} };
+    type DefaultRender = { kind: "default"; id: string; data: { name: string; value: number } };
+    type RowLinkRender = { kind: "link"; id: string; data: { name: string; value: number } };
+    type RowClickRender = { kind: "click"; id: string; data: { name: string; value: number } };
+    type RowRenderFn = HeaderRender | DefaultRender | RowLinkRender | RowClickRender;
+    // And columns with `data-testid` defined
+    const nameColumn: GridColumn<RowRenderFn> = {
+      header: () => ({ content: "Name", "data-testid": "headerCell_name" }),
```
stephenh, line 2319: @bdow if this is primarily to add `data-testid`s, should we just set our own data-testids using the column name?

...oh, we don't have column names, we only have a list of columns...

:thinking:

bdow: We may also want a different data-testid value depending on the row type. For instance `projectItem` vs `division`.

bdow: But that kind of leads into another question regarding the Edit Columns request. How do we know know what the `string` value is for that column. I think we might want to add in a `name` property for that column, as the header's content could return JSX and we wouldn't be able to pull out the proper name of the column to show in that Edit Columns modal.

stephenh: It has a huge API change, but we could make it be like:

```
const columns = {
  name: {
    header: ..., link: ..., click: ...
  },
  value: {
    header: ..., 
  }
};
```

So basically `GridTable.columns` changes from a `GridColumn[]` to a `Record<string, GridColumn>`...

stephenh: Granted, a small change yeah is to add `GridColumn.name` as an optional? property that looks more like:

```
const valueColumn: GridColumn<...> = {
  name: "value",
  header: ...,
  link: ...,
}
```

bdow: Right... hopefully there is no collision on user's already have a `kind: "name"` row, which I doubt.

stephenh: (Fwiw just made an edit to the comment ^ about `[columnA, `columnB]` --> `{ columnA, columnB }` being a potentially cute way to handle this...)

bdow: Sure that's not so bad. Though, I am a bit hesitant to pull in such a big breaking change.

stephenh: > Though, I am a bit hesitant to pull in such a big breaking change.

Yeah... Hack day is on Friday though! :-D

PR #628, opened 2022-08-26
feat: Bundle with esbuild for better vite support.

bdow:


esbuild.js
```diff
@@ -0,0 +1,16 @@
...
+
+// See the Bundling section of the readme for rationale/thoughts.
+
+esbuild.build({
+  entryPoints: ["src/index.ts"],
+  sourcemap: true,
+  outfile: "dist/index.mjs",
+  bundle: true,
+  format: "esm",
+  plugins: [nodeExternalsPlugin()],
```
stephenh, line 12: @bdow honestly I can't remember if this is really required, or if I was just mashing buttons; can you tell if we need it or not?

bdow: Appears to build fine without it. Will push an update.

stephenh (approved):
Love it!

PR #631, opened 2022-08-26
fix: Run esbuild script during release

bdow:


release.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     [
       "@semantic-release/exec",
       {
-        prepareCmd: "yarn build",
+        prepareCmd: "yarn build && node ./esbuild.js",
```
stephenh, line 9: We could probably just skip `yarn build`, although maybe having it do another type check doesn't hurt; in theory this should only be ran if the `build` job passes, but doesn't hurt to have it again.

stephenh (approved):

PR #636, opened 2022-08-30
feat: [SC-19452] Chip custom title support

Grandlouis:
[SC-19452]
Based on the requirement to have the chip title to display detailed information different than the chip content, we are adding support for a object having text and title properties, adding also unit tests and stories to cover this usage
imagen


chr1sjf0x (approved):
TY @Grandlouis 🍣

PR #641, opened 2022-09-06
feat: Sc-19444 active cell

JonnCharpentier:
Feature request from [ticket](https://app.shortcut.com/homebound-team/story/19444/add-cell-highlighting-to-bulk-schedule-phases)

Add cell highlighting 

This branch is based off https://github.com/homebound-team/beam/pull/626
image

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -93,6 +93,39 @@ export const Hovering = newStory(
   { decorators: [withRouter()] },
 );
 
+export const ActiveCell = newStory(
+    () => {
+        const nameColumn: GridColumn<Row> = { name: "name", header: "Name", data: ({ name }) => name };
+        const valueColumn: GridColumn<Row> = { name: "value", header: "Value", data: ({ value }) => value };
+        const actionColumn: GridColumn<Row> = { name: "actions", header: "Action", data: () => <div>Actions</div> };
+        const rowStyles: GridRowStyles<Row> = useMemo(
+            () => ({
+                data: {
+                    onClick: (row, api) => {
+                        api.setActiveRowId(`data_${row.id}`);
+                    },
+                },
+            }),
+            [],
+        );
+        return (
+            <GridTable<Row>
+                columns={[nameColumn, valueColumn, actionColumn]}
+                activeCellId={`data_1_value`}
+                style={{ cellHighlight: true }}
+                rowStyles={rowStyles}
+                rows={[
+                    simpleHeader,
+                    { kind: "data", id: "1", data: { name: "c", value: 1 } },
+                    { kind: "data", id: "2", data: { name: "b", value: 2 } },
+                    { kind: "data", id: "3", data: { name: "a", value: 3 } },
+                ]}
+            />
+        );
+    },
+    { decorators: [withRouter()] },
+);
+
 export function VirtualFiltering() {
   const rows: GridDataRow<Row>[] = useMemo(
     () => [
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -2306,6 +2306,72 @@ describe("GridTable", () => {
     // And they can no longer be fetched by the api
     expect(api.current!.getSelectedRowIds()).toEqual([]);
   });
+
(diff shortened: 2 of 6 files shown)


src/components/Table/GridTable.stories.tsx
```diff
@@ -93,6 +93,39 @@ export const Hovering = newStory(
   { decorators: [withRouter()] },
 );
 
+export const ActiveCell = newStory(
+    () => {
+        const nameColumn: GridColumn<Row> = { name: "name", header: "Name", data: ({ name }) => name };
```
stephenh, line 98: Nit: maybe run prettier?

src/components/Table/GridTable.test.tsx
```diff
@@ -2306,6 +2306,72 @@ describe("GridTable", () => {
...
+  it("can set test ids on cell elements", async () => {
+    // Given different row types for each render function
+    type HeaderRender = { kind: "header"; id: string; data: {} };
+    type DefaultRender = { kind: "default"; id: string; data: { name: string; value: number } };
+    type RowLinkRender = { kind: "link"; id: string; data: { name: string; value: number } };
+    type RowClickRender = { kind: "click"; id: string; data: { name: string; value: number } };
+    type RowRenderFn = HeaderRender | DefaultRender | RowLinkRender | RowClickRender;
+    // And columns with `data-testid` defined
+    const nameColumn: GridColumn<RowRenderFn> = {
+      header: () => ({ content: "Name", "data-testid": "headerCell_name" }),
```
stephenh, line 2319: Thought: I'm pretty worried about the boilerplate this will create, insofar as _every_ page in blueprint is going to start typing out `data-testid: <column><row>` over and over again.

I kind of wonder if we actually need the row kind (header vs. default vs. link. vs. click)...

Like if we added `GridColumn.name`, could our cell ids be just `name`, so like in a table of 3 rows (2 cells each) it would:

* `data-testid=name` bob + `data-testid=value` 1
* `data-testid=name` fred + `data-testid=value` 2
* `data-testid=name` greg + `data-testid=value` 3

And then use "get me name_1" / "get me name_2", like:

```
expect(r.name_1).toBe("bob");
expect(r.name_2).toBe("fred");
```

Like do we really need to know that row 1 is kind=header, and row 2 is kind=default?

It would also really cut down the boilerplate if, by just setting like `name`:

```
const actionColumn: GridColumn<Row> = {
  name: "actions",
  header: "Action",
  data: () => <div>Actions</div>
};
```

That's enough to set the `data-testid` for the cell.

stephenh: I.e. I want something where this `data-testid` goes away:

```
default: ({ name }) => ({ content: name, "data-testid": "defaultCell_name" }),
```

From being something that is "typed by hand", and instead the system "just knows" (even if it means tweaking the id to not literally be `defaultCell_name` but just like `name` or `nameCell`.

JonnCharpentier: I agree, but looks like that change is part of this pr: https://github.com/homebound-team/beam/pull/626

stephenh: @bdow / @JonnCharpentier was the plan to land #626 first, or just have this PR land it's change + the changes Jonn made on top of it?

Jonn, I don't see the `GridColumn.name` change in that other PR? Is that something you added on top? (Which is great if so, just clarifying)

bdow: No plan to land #626 ahead of time. I still wasn't sure about supporting the `data-testid` as a prop like this, per our conversation that PR. I much prefer the idea of introducing a `GridColumn.name` property, but just haven't had time to come back around to that.

stephenh: Cool, @JonnCharpentier do you mind just making #626 fundamentally part of this PR, and updating it based on ^ suggestions? I.e. remove needing to type out the `data-testid`

JonnCharpentier: Ok I'll  work on that

JonnCharpentier: @stephenh and @bdow  here is a new PR with just my changes: https://github.com/homebound-team/beam/pull/643

PR #642, opened 2022-09-07
feat: update MultiSelectField.mock to support disabled and helperText

0ces:


bdow (approved):
Awesome. Thanks for updating. Would you be able to make the same changes to the `SelectField.mock.tsx` file as well?

bdow (approved):
Perfect! Thank you for updating these.

PR #643, opened 2022-09-07
feat: SC-19444 cell highlighting

JonnCharpentier:
Feature request from [ticket](https://app.shortcut.com/homebound-team/story/19444/add-cell-highlighting-to-bulk-schedule-phases)

Add cell highlighting 

This branch is based off https://github.com/homebound-team/beam/pull/626
image

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -95,7 +95,7 @@ export const Hovering = newStory(
 
 export const ActiveCell = newStory(
   () => {
-    const nameColumn: GridColumn<Row> = {  header: "Name", data: ({ name }) => name };
+    const nameColumn: GridColumn<Row> = { name:"name", header: "Name", data: ({ name }) => name };
     const valueColumn: GridColumn<Row> = { name: "value", header: "Value", data: ({ value }) => value };
     const actionColumn: GridColumn<Row> = { name: "actions", header: "Action", data: () => <div>Actions</div> };
     const rowStyles: GridRowStyles<Row> = useMemo(
@@ -111,7 +111,7 @@ export const ActiveCell = newStory(
     return (
       <GridTable<Row>
         columns={[nameColumn, valueColumn, actionColumn]}
-        activeCellId={`data_1_value`}
+        activeCellId={"data_1_value"}
         style={{ cellHighlight: true }}
         rowStyles={rowStyles}
         rows={[
@@ -1135,7 +1135,7 @@ export function StickyColumnsAndHeaderVirtualized() {
 export function ActiveRow() {
   const nameColumn: GridColumn<Row> = { header: "Name", data: ({ name }) => name, w: "200px" };
   const valueColumn: GridColumn<Row> = { header: "Value", data: ({ value }) => value, w: "200px" };
-  const actionColumn: GridColumn<Row> = { header: "Actions", data: "Actions", w: "200px" };
+  const actionColumn: GridColumn<Row> = { header: "Actions", data: "Actions" , w:"200px" };
   const rowStyles: GridRowStyles<Row> = useMemo(
     () => ({
       data: {
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -94,36 +94,36 @@ export const Hovering = newStory(
 );
 
 export const ActiveCell = newStory(
-    () => {
-        const nameColumn: GridColumn<Row> = { name: "name", header: "Name", data: ({ name }) => name };
-        const valueColumn: GridColumn<Row> = { name: "value", header: "Value", data: ({ value }) => value };
-        const actionColumn: GridColumn<Row> = { name: "actions", header: "Action", data: () => <div>Actions</div> };
-        const rowStyles: GridRowStyles<Row> = useMemo(
-            () => ({
-                data: {
-                    onClick: (row, api) => {
-                        api.setActiveRowId(`data_${row.id}`);
-                    },
-                },
-            }),
-            [],
-        );
-        return (
-            <GridTable<Row>
-                columns={[nameColumn, valueColumn, actionColumn]}
-                activeCellId={`data_1_value`}
-                style={{ cellHighlight: true }}
-                rowStyles={rowStyles}
-                rows={[
-                    simpleHeader,
-                    { kind: "data", id: "1", data: { name: "c", value: 1 } },
-                    { kind: "data", id: "2", data: { name: "b", value: 2 } },
-                    { kind: "data", id: "3", data: { name: "a", value: 3 } },
-                ]}
-            />
-        );
-    },
-    { decorators: [withRouter()] },
+  () => {
+    const nameColumn: GridColumn<Row> = {  header: "Name", data: ({ name }) => name };
+    const valueColumn: GridColumn<Row> = { name: "value", header: "Value", data: ({ value }) => value };
+    const actionColumn: GridColumn<Row> = { name: "actions", header: "Action", data: () => <div>Actions</div> };
+    const rowStyles: GridRowStyles<Row> = useMemo(
+      () => ({
+        data: {
+          onClick: (row, api) => {
+            api.setActiveRowId(`data_${row.id}`);
+          },
+        },
+      }),
+      [],
+    );
+    return (
+      <GridTable<Row>
+        columns={[nameColumn, valueColumn, actionColumn]}
+        activeCellId={`data_1_value`}
+        style={{ cellHighlight: true }}
+        rowStyles={rowStyles}
+        rows={[
+          simpleHeader,
+          { kind: "data", id: "1", data: { name: "c", value: 1 } },
+          { kind: "data", id: "2", data: { name: "b", value: 2 } },
+          { kind: "data", id: "3", data: { name: "a", value: 3 } },
+        ]}
+      />
+    );
+  },
+  { decorators: [withRouter()] },
 );
 
 export function VirtualFiltering() {
@@ -470,7 +470,6 @@ export function ObservableRows() {
       <IconButton onClick={() => (o.a = o.a + 1)} icon="plus" />
       <GridTable
         columns={[nameColumn]}
-        observeRows={true}
         rows={[
           simpleHeader,
           { kind: "data", id: "1", data: { name: "a", value: 1 } },
@@ -1157,7 +1156,7 @@ export function ActiveRow() {
     [],
   );
   const columns = useMemo(() => [nameColumn, valueColumn, actionColumn], []);
-  return <GridTable columns={columns} activeRowId="data_2" rowStyles={rowStyles} rows={rows} observeRows />;
+  return <GridTable columns={columns} activeRowId="data_2" rowStyles={rowStyles} rows={rows} />;
 }
 
 export function ActiveRowNestedCard() {
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -93,6 +93,39 @@ export const Hovering = newStory(
   { decorators: [withRouter()] },
 );
 
+export const ActiveCell = newStory(
+    () => {
+        const nameColumn: GridColumn<Row> = { name: "name", header: "Name", data: ({ name }) => name };
+        const valueColumn: GridColumn<Row> = { name: "value", header: "Value", data: ({ value }) => value };
+        const actionColumn: GridColumn<Row> = { name: "actions", header: "Action", data: () => <div>Actions</div> };
+        const rowStyles: GridRowStyles<Row> = useMemo(
+            () => ({
+                data: {
+                    onClick: (row, api) => {
+                        api.setActiveRowId(`data_${row.id}`);
+                    },
+                },
+            }),
+            [],
+        );
+        return (
+            <GridTable<Row>
+                columns={[nameColumn, valueColumn, actionColumn]}
+                activeCellId={`data_1_value`}
+                style={{ cellHighlight: true }}
+                rowStyles={rowStyles}
+                rows={[
+                    simpleHeader,
+                    { kind: "data", id: "1", data: { name: "c", value: 1 } },
+                    { kind: "data", id: "2", data: { name: "b", value: 2 } },
+                    { kind: "data", id: "3", data: { name: "a", value: 3 } },
+                ]}
+            />
+        );
+    },
+    { decorators: [withRouter()] },
+);
+
 export function VirtualFiltering() {
   const rows: GridDataRow<Row>[] = useMemo(
     () => [
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -2075,8 +2075,7 @@ describe("GridTable", () => {
     expect(row(r, 2).getAttribute("data-render")).toEqual("1");
   });
 
-  // Skip since we are always using Observable rows
-  it.skip("memoizes rows based on the data attribute", async () => {
+  it("memoizes rows based on the data attribute", async () => {
     const [header, row1, row2] = rows;
     const columns = [nameColumn];
     // Given a table is initially rendered with 2 rows
@@ -2109,19 +2108,6 @@ describe("GridTable", () => {
 
     // Then the first row/cell has the 'active' background color
     expect(cell(r, 1, 1)).toHaveStyleRule("background-color", Palette.LightBlue50);
-  });
-
-  it("reacts to setting activeCellId", async () => {
-    // Given a table initially rendered with an active cell id
-    const r = await render(
-      <GridTable columns={columns} rows={rows} style={{ cellHighlight: true }} activeCellId={"data_1_name"} />,
-    );
-    // Then expect the cell to have highlight style
-    expect(cell(r, 1, 0)).toHaveStyleRule("border-width", "1px");
-    expect(cell(r, 1, 0)).toHaveStyleRule("border-color", Palette.LightBlue900);
-    // Then expect other cells not to have highlight style
-    expect(cell(r, 1, 1)).not.toHaveStyleRule("border-width", "1px");
-    expect(cell(r, 1, 1)).not.toHaveStyleRule("border-color", Palette.LightBlue900);
   });
 
   it("can render with rows with initCollapsed defined", async () => {
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -2075,7 +2075,8 @@ describe("GridTable", () => {
     expect(row(r, 2).getAttribute("data-render")).toEqual("1");
   });
 
-  it("memoizes rows based on the data attribute", async () => {
+  // Skip since we are always using Observable rows
+  it.skip("memoizes rows based on the data attribute", async () => {
     const [header, row1, row2] = rows;
     const columns = [nameColumn];
     // Given a table is initially rendered with 2 rows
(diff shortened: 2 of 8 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -1108,129 +1126,146 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
 
   let firstContentColumnStylesApplied = false;
 
-  const rowNode = (
-    <Row css={rowCss} {...others} data-gridrow {...getCount(row.id)}>
-      {columns.map((column, columnIndex) => {
-        const { wrapAction = true, isAction = false } = column;
+  const rowNode = useComputed(
```
stephenh, line 1129: Why the `useComputed`? I suppose it's b/c we're accessing `rowState.activeCellId`...

In theory `GridTable` will also do this if `GridTable.observeRows` is set (i.e. wrap `GridRow` in an `ObservableGridRow`)...

Maybe we should change:

```
      const RowComponent = observeRows ? ObservedGridRow : MemoizedGridRow;
```

To be `observeRows || cellHighlight ? ... :`

Or we could just _always_ use `ObservedGridRow`, it probably wouldn't hurt anything, and we could remove the `observeRows` prop all together (b/c it's always set true now); maybe just do that?

src/components/Table/GridTable.tsx
```diff
@@ -245,6 +247,11 @@ export interface GridTableProps<R extends Kinded, S, X> {
    * Example "data_123"
    */
   activeRowId?: string;
+  /**
+   * Defines which cell in the table should be provided with an "active" styling.
+   * Expected format is `${row.kind}_${row.id}_${column.name}`.
+   */
+  activeCellId?: string;
```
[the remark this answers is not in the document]
stephenh: (I suppose if page really wanted to set it, they could use the `GridTableApi`.)

src/components/Table/GridTable.tsx
JonnCharpentier: by doing this, removing the observeRows prop, are we introducing a breaking change on internal-frontend?

JonnCharpentier: @bdow what do you think?

src/components/Table/GridTable.tsx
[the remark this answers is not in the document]
JonnCharpentier: yeah, in this case the active cell is not only highlighted, but also open up a side pane that display task details

src/components/Table/GridTable.tsx
bdow: I kind of like the idea of just _always_ doing `observeRows`. Maybe for now keep the prop, but default it to be `true`. This should not be a breaking change on BP.

bdow: I'm also good with removing the prop too... Would just mean we need to clean up places that set it, which would be an easy fix.

src/components/Table/GridTable.tsx
```diff
@@ -93,6 +93,8 @@ export interface GridStyle {
   levels?: Record<number, { cellCss?: Properties; firstContentColumn?: Properties }>;
   /** Allows for customization of the background color used to denote an "active" row */
   activeBgColor?: Palette;
+  /** Enables cells Highlight and hover */
+  cellHighlight?: boolean;
```
bdow, line 97: **request:** I'm thinking I would prefer this to be an option on the `GridStyleDef` in `styles.tsx`, and not here. If that is set, then we can add in hover styles for the `cellCss` defined in the `memoizedTableStyles` function. 
The reasoning for this request is because we want to move away from people using `GridStyle` and instead choose a `GridStyleDef`. This would ensure better consistency between all of our tables.

src/components/Table/GridTable.tsx
```diff
@@ -1108,129 +1126,146 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
...
+            // Add any cell specific style overrides
+            ...(isGridCellContent(maybeContent) && maybeContent.typeScale ? Css[maybeContent.typeScale].$ : {}),
+            // And any cell specific css
+            ...(isGridCellContent(maybeContent) && maybeContent.css ? maybeContent.css : {}),
+            // Apply cell highlight styles to active cell and hover
+            ...(applyCellHighlight
+              ? isCellActive
+                ? Css.bw1.bsSolid.bLightBlue900.br4.$
+                : { ":hover": Css.bgGray100.$ }
+              : {}),
```
bdow, line 1241: When we add the `hover` styles to GridStylesDef, then we can remove it here.

src/components/Table/GridTable.tsx
```diff
@@ -1108,129 +1126,146 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
...
+              ? Css.bgColor(style.activeBgColor ?? Palette.LightBlue50).$
+              : {}),
+            // Add any cell specific style overrides
+            ...(isGridCellContent(maybeContent) && maybeContent.typeScale ? Css[maybeContent.typeScale].$ : {}),
+            // And any cell specific css
+            ...(isGridCellContent(maybeContent) && maybeContent.css ? maybeContent.css : {}),
+            // Apply cell highlight styles to active cell and hover
+            ...(applyCellHighlight
+              ? isCellActive
+                ? Css.bw1.bsSolid.bLightBlue900.br4.$
```
bdow, line 1239: **question:** Does this cause all the cells to shift when we add in this border? This could impact how we define our column widths and potentially cause horizontal scrollbars when a table is already the full width of the page 🤔. It might be fine since we are using `box-sizing: border-box`, but it might cause the content within the cell to shift around. We might want to add a transparent border to every cell to prevent any shifting.

src/components/Table/GridTable.tsx
```diff
@@ -1413,14 +1431,18 @@ const rowLinkRenderFn: (as: RenderAs) => RenderCellFn<any> =
 
 /** Renders a cell that will fire the RowStyle.onClick. */
 const rowClickRenderFn: (as: RenderAs, api: GridTableApi<any>) => RenderCellFn<any> =
-  (as: RenderAs, api: GridTableApi<any>) => (key, css, content, row, rowStyle, classNames: string | undefined) => {
+  (as: RenderAs, api: GridTableApi<any>) =>
+  (key, css, content, row, rowStyle, classNames: string | undefined, onClick) => {
     const Row = as === "table" ? "tr" : "div";
     return (
       <Row
```
bdow, line 1438: **request:** This isn't your fault, but it really confused me when reading this. We **aren't** rendering a `Row` here, but instead a `Cell`. This naming and the use of a `tr` tag is incorrect in this situation. Would you mind fixing this and updating it to:
```
const Cell = as === "table" ? "td" : "div";
return (
  <Cell>...</Cell>
);
```
Please and thanks!

src/components/Table/GridTable.tsx
```diff
@@ -341,10 +349,9 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
       // We only pass sortState to header rows, b/c non-headers rows shouldn't have to re-render on sorting
       // changes, and so by not passing the sortProps, it means the data rows' React.memo will still cache them.
       const sortProps = row.kind === "header" ? { sortOn, sortState, setSortKey } : { sortOn };
-      const RowComponent = observeRows ? ObservedGridRow : MemoizedGridRow;
```
bdow, line 344: **question:** Did we get rid of `MemoizedGridRow`? I don't think its being used still is it?

src/components/Table/GridTable.test.tsx
```diff
@@ -2075,7 +2075,8 @@ describe("GridTable", () => {
     expect(row(r, 2).getAttribute("data-render")).toEqual("1");
   });
 
-  it("memoizes rows based on the data attribute", async () => {
+  // Skip since we are always using Observable rows
+  it.skip("memoizes rows based on the data attribute", async () => {
```
bdow, line 2079: **question:** Does this fail now? I'm assuming so since you skipped it. 
If so, @stephenh do you think this is a bad choice to remove the `MemoizedGridRow` if `ObservedGridRow` will cause rerenders like this?

src/components/Table/GridTable.tsx
```diff
@@ -1208,6 +1221,8 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
           ...(isGridCellContent(maybeContent) && maybeContent.typeScale ? Css[maybeContent.typeScale].$ : {}),
           // And any cell specific css
           ...(isGridCellContent(maybeContent) && maybeContent.css ? maybeContent.css : {}),
+          // Apply cell highlight styles to active cell and hover
+          ...Css.bw1.bsSolid.br4.if(applyCellHighlight && isCellActive).bLightBlue900.else.bTransparent.$,
```
bdow, line 1225: **nit:** You do not need the `else` block here. You could _always_ set `bTransparent` and only override it with the `if` statement.

src/components/Table/GridTable.test.tsx
```diff
@@ -2110,6 +2111,19 @@ describe("GridTable", () => {
...
+    // Given a table initially rendered with an active cell id
+    const r = await render(
+      <GridTable columns={columns} rows={rows} style={{ cellHighlight: true }} activeCellId={"data_1_name"} />,
+    );
+    // Then expect the cell to have highlight style
+    expect(cell(r, 1, 0)).toHaveStyleRule("border-width", "1px");
+    expect(cell(r, 1, 0)).toHaveStyleRule("border-color", Palette.LightBlue900);
+    // Then expect other cells not to have highlight style
+    expect(cell(r, 1, 1)).not.toHaveStyleRule("border-width", "1px");
+    expect(cell(r, 1, 1)).not.toHaveStyleRule("border-color", Palette.LightBlue900);
```
bdow, line 2124: **question:** Shouldn't the other cells have the border style, but with a transparent color?

src/components/Table/GridTable.test.tsx
```diff
@@ -2110,6 +2111,19 @@ describe("GridTable", () => {
...
+  });
```
bdow, line 2125: I really like the test for the style rules. Can we write tests to ensure (1) The active cell is updated if the prop value changes, and (2) the active cell can be changed via the GridTableApi?

src/components/Table/GridTable.stories.tsx
```diff
@@ -93,6 +93,39 @@ export const Hovering = newStory(
...
+    const nameColumn: GridColumn<Row> = {  header: "Name", data: ({ name }) => name };
+    const valueColumn: GridColumn<Row> = { name: "value", header: "Value", data: ({ value }) => value };
+    const actionColumn: GridColumn<Row> = { name: "actions", header: "Action", data: () => <div>Actions</div> };
+    const rowStyles: GridRowStyles<Row> = useMemo(
+      () => ({
+        data: {
+          onClick: (row, api) => {
+            api.setActiveRowId(`data_${row.id}`);
+          },
+        },
```
bdow, line 107: **question:** Is this necessary? Won't the `cellOnClick ` in the GridRow do this for us?

src/components/Table/GridTable.tsx
stephenh: Ah damn; @bdow / @JonnCharpentier it looks like using `ObservedGridRow` means that (even before Jonn's PR) we would just never use `MemoizedGridRow`.

I think this was probably a bug/mistake in the old code, b/c the idea was less of "choose either observed or memoized" (which is what it was doing) but the intention was "observed would be layered on top of memoized".

@JonnCharpentier I wonder if we can try:

```
const MemoizedGridRow = React.memo(observer(GridRow), (one, two) => {
  const { row: row1, ...others1 } = one;
  const { row: row2, ...others2 } = two;
  return shallowEqual(row1, row2) && shallowEqual(others1, others2);
}) as typeof GridRow;
```

And then make line 354 `<MemoizedGridRow ...`.

I believe this will achieve both goals, of memoizing the component based on the `row` prop, but also wrapping it in an observer (now using the out-of-the-box `observer` function instead of the previous hand-written `ObservedGridRow` which was just doing the same thing).

src/components/Table/GridTable.stories.tsx
JonnCharpentier: on this story I'm doing both active cell and active Row this onClick is just for setting active row

stephenh: :thinking: should we have this line do both:

```
const cellOnClick = applyCellHighlight ? () => api.setActiveCellId(cellId) : undefined;
```

Set the active cell id as well as the active row id?

My initial thought is that this seems great, b/c it starts to move control of "what is the active row?" into GridTable, instead of being something each page needs to know about / reimplement.

JonnCharpentier: Y es, we can do that

JonnCharpentier: @stephenh what if we only want to highlight a cell, but don't want to use active row styles? I believe this is the case on my ticket

bdow: Sorry, I misread that code, I thought you were setting the active cell in that `onClick` callback. I'm good with them being treated separately.

stephenh: @JonnCharpentier ah I see, merely setting the active row id turns on active row styles?

I wonder if we need some sort of `rowHighlight`, similar to `cellHighlight`, that tables need to set to explicitly turn on row highlighting behavior...

We can probably deal with that later if you'd like to leave the current `cellOnClick` as-is.

src/components/Table/GridTable.tsx
```diff
@@ -1068,7 +1077,8 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
   } = props;
 
   const { rowState } = useContext(RowStateContext);
-  const isActive = useComputed(() => rowState.activeRowId === `${row.kind}_${row.id}`, [row, rowState]);
+  const rowId = `${row.kind}_${row.id}`;
+  const isActive = useComputed(() => rowState.activeRowId === rowId, [row, rowState]);
```
bdow, line 1081: **nit:** `rowId` should now be part of the dependency array.

bdow (commented):
Looks like a lot of storybook diff issues with how the borders are applied now. Appears as though it is an issue with how the row borders are applied, I think via `betweenRowsCss` in the `export const defaultStyle: GridStyle`. I _think_ they way these styles are defined is done on each cell element, vs the row element. We initially did this due to the row element being `display: contents` and couldn't have styles applied to it. This is no longer the case, so we should be able to apply the `betweenRowCss` on the row elements themselves.

src/components/Table/GridTable.tsx
JonnCharpentier: question, does'nt  putting rowId on that dependency array will make it fire useComputed every render? since is a variable that get re-defined on every render, also, if we already have row object as a dependency, then if the row changes we will get the new rowId, and rowId can not change unless row changed

JonnCharpentier:
> is is no longer the case, so we should be able to apply the `betweenRowCss` on the row elements themselves.

I don't understand what the request is here

bdow:
> > is is no longer the case, so we should be able to apply the `betweenRowCss` on the row elements themselves.
> 
> I don't understand what the request is here

There are a bunch of GridTable stories that look very off due to borders being applied around the _whole_ cell, except the cells in the first row ([current](https://main--60466f7d43c37c0021788867.chromatic.com/?path=/story/components-gridtable--client-side-sorting) vs [this PR](https://60466f7d43c37c0021788867-rgothkhvbb.chromatic.com/?path=/story/components-gridtable--client-side-sorting)). This is a result of the new cell border styles in this PR and how the `betweenRowCss` is applied are not compatible. I am thinking the `betweenRowCss` is being applied to each cell element, and causing this.

bdow (commented):
Looks like there are still some issues with the borders showing up when they shouldn't in the Storybook diffs: 
https://www.chromatic.com/test?appId=60466f7d43c37c0021788867&id=6320b324f201b76a50f5dd9f

bdow:
@JonnCharpentier Another thought I had, which I _think_ I mentioned in our meeting, is what if we used a `box-shadow` for the active cell styling? Then we wouldn't need to set any border styles, transparent or with a color. That might help solve some of this.
```
box-shadow: inset 0 0 0 2px Palette.LightBlue900;
```

src/components/Table/GridTable.tsx
```diff
@@ -540,7 +546,7 @@ function renderDiv<R extends Kinded>(
           the + operator as an offset.
           Inspired by: https://stackoverflow.com/a/25005740/2551333
         */
-        ...(style.betweenRowsCss ? Css.addIn(`& > div:nth-of-type(n+3) > *`, style.betweenRowsCss).$ : {}),
+        ...(style.betweenRowsCss ? Css.addIn(`& > div:nth-of-type(n+${headerRows.length + totalsRows.length + 2})`, style.betweenRowsCss).$ : {}),
```
bdow, line 549: Let's add ` > *` back in to the selector, but leave the rest as is.

src/components/Table/GridTable.tsx
bdow: Since `rowId` is a string, it will be properly evaluated as `"rowId" === "rowId"`. There should be no difference, and would not cause the `useComputed` to fire again.

src/components/Table/GridTable.tsx
```diff
@@ -1208,6 +1222,8 @@ function GridRow<R extends Kinded, S>(props: GridRowProps<R, S>): ReactElement {
           ...(isGridCellContent(maybeContent) && maybeContent.typeScale ? Css[maybeContent.typeScale].$ : {}),
           // And any cell specific css
           ...(isGridCellContent(maybeContent) && maybeContent.css ? maybeContent.css : {}),
+          // Apply cell highlight styles to active cell and hover
+          ...Css.if(applyCellHighlight && isCellActive).boxShadow(`inset 0 0 0 2px ${Palette.LightBlue700}`).$,
```
bdow, line 1226: Maybe need the `br4` here?

src/components/Table/GridTable.tsx
```diff
@@ -1388,11 +1406,11 @@ const headerRenderFn: (
 
 /** Renders a cell element when a row link is in play. */
 const rowLinkRenderFn: (as: RenderAs) => RenderCellFn<any> =
-  (as: RenderAs) => (key, css, content, row, rowStyle, classNames: string | undefined) => {
+  (as: RenderAs) => (key, css, content, row, rowStyle, classNames: string | undefined, onClick) => {
     const to = rowStyle!.rowLink!(row);
     if (as === "table") {
       return (
-        <td key={key} css={{ ...css, ...tableRowStyles(as) }} className={classNames}>
+        <td key={key} css={{ ...css, ...tableRowStyles(as) }} className={classNames} onClick={onClick}>
```
bdow, line 1413: Let's remove the `onClick` from the `rowLinkRenderFn`.

src/components/Table/styles.tsx
```diff
@@ -70,9 +72,11 @@ function memoizedTableStyles() {
         firstRowMessageCss: Css.tc.py3.$,
         headerCellCss: Css.gray700.xsEm.bgGray200.aic.nowrap.pxPx(12).hPx(40).$,
         totalsCellCss: Css.bgWhite.gray700.smEm.hPx(52).pt0.pbPx(12).boxShadow("none").$,
+        betweenRowsCss: Css.bt.bGray200.$,
```
bdow, line 75: Remove this.

src/components/Table/styles.tsx
```diff
@@ -70,9 +72,11 @@ function memoizedTableStyles() {
...
         cellCss: {
-          ...Css.gray900.xs.bgWhite.aic.pxPx(12).boxShadow(`inset 0 -1px 0 ${Palette.Gray200}`).$,
+          ...Css.gray900.xs.bgWhite.aic.pxPx(12).$,
```
bdow, line 77: Add box-shadow back

bdow (approved):
As far as I can tell, all the snapshots look right. I'll write this up somewhere, but I tend to find it better to not approve storybook diffs until after the code has been approved. That way you're usually comparing against `main`.

JonnCharpentier:
> As far as I can tell, all the snapshots look right. I'll write this up somewhere, but I tend to find it better to not approve storybook diffs until after the code has been approved. That way you're usually comparing against `main`.

I didn't know that's how it works,  Thanks! I'll do a beam version update

PR #647, opened 2022-09-13
feat: Accordion component

roberthdg:


src/components/Accordion.tsx
```diff
@@ -0,0 +1,44 @@
+import { Css, IconButton, useTestIds } from "src/index";
+import { ReactNode, useState } from "react";
+
+export interface AccordionProps {
+  title: string;
+  children: ReactNode;
+  disabled?: boolean;
+  defaultExpanded?: boolean;
+  bottomSection?: boolean;
```
bdow, line 9: **thought:** This `bottomSection` property makes me think we can maybe do this smarter. I don't like the idea of a user having to define that this Accordion is the last Accordion in my list. Maybe instead what we can do is create another component in this PR, for instance `<AccordionList accordions=[[] as AccordionProps] />`. Then this component can do the proper styling for a list of Accordions by defining the `bottomSection`

src/components/Accordion.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function Accordion(props: AccordionProps) {
+  const { title, children, disabled = false, defaultExpanded = false, bottomSection = false} = props;
+  const [expanded, setExpanded] = useState(defaultExpanded);
+  const [focused, setFocused] = useState(false);
+  const testIds = useTestIds({}, "accordion");
```
bdow, line 16: **suggestion:** Use `props` for the first parameter in `useTestIds`. This would allow a user to specify their own test id for the component.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,44 @@
...
+        </div>
+        <div css={Css.ml("auto").my("auto").pr1.$}>
+          <IconButton
+            {...testIds.button}
+            disabled={disabled}
+            icon={expanded ? "chevronUp" : "chevronDown"}
+            onClick={() => setExpanded(!expanded)}
+          />
+        </div>
+      </div>
```
bdow, line 38: **suggestion:** Per my note in Figma, I am thinking that this surrounding element becomes a `<button />` which would allow the user to click anywhere in the title to expand/collapse the content. We'll wait for Radhi's reply, but I'm guessing that is the expected behavior.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <div {...testIds.details} css={Css.px2.pb2.pt1.if(!expanded).dn.$}>
```
bdow, line 39: **nit (non-blocking):** Not necessary, but could be fun to see how we can use a transition or animation to have this show. Depends on how quick you want to get this done. And that'd be another thing we'd need to check with design about - we don't have any definitions around animations or transitions for Beam. So maybe not!?

src/components/Accordion.tsx
```diff
@@ -0,0 +1,44 @@
+import { Css, IconButton, useTestIds } from "src/index";
+import { ReactNode, useState } from "react";
+
+export interface AccordionProps {
+  title: string;
```
bdow, line 5: **suggestion:** Doing something like `title: string` has been where we started, but end up with `title: ReactNode`, because eventually someone wants to put in more than just text. Maybe another Icon, or Tag... So I would suggest allow this to be a `ReactNode` to start out with.

src/components/Accordion.tsx
bdow: Another cool thing with an AccordionList is that we could provide a prop that can allow the list to only allow one Accordion open at a time vs multiple. Could be neat.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,44 @@
...
+  children: ReactNode;
+  disabled?: boolean;
+  defaultExpanded?: boolean;
+  bottomSection?: boolean;
+}
+
+export function Accordion(props: AccordionProps) {
+  const { title, children, disabled = false, defaultExpanded = false, bottomSection = false} = props;
+  const [expanded, setExpanded] = useState(defaultExpanded);
+  const [focused, setFocused] = useState(false);
```
bdow, line 15: **suggestion:** I don't believe we want the focus state to always show when focused. It should be a keyboard sort of focus (like showing an outline when you hit tab vs showing when you click). For that, we use a helper called `useFocusRing`. You should be able to find example uses of it in the application. It is pretty straight forward, but let me know if you want any more details.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <div
+      {...testIds}
+      onBlur={() => setFocused(false)}
+      onFocus={() => setFocused(true)}
+      // Remove default border when focused
+      css={Css.if(!focused).bt.bGray400.if(bottomSection && !(!expanded && focused)).bb.$}
+    >
+      <div {...testIds.summary} css={Css.df.if(focused).ba.bLightBlue600.bw2.addIn(":hover", Css.bgGray100.$).$}>
```
bdow, line 26: **suggestion:** Use a `box-shadow` to provide the focus outline instead of a border. This will stop the content from shifting when you introduce the focus styles.

bdow: (probably something like: `box-shadow: inset 0 0 0 2px Palette.LightBlue600`.

bdow (commented):
Great start. A few suggestions, and I think the inclusion of a AccordionList component would be beneficial to auto-add that styling.

Also, another inclusion I'd like to see is including `aria-attributes` for better accessibility. You could do something along the lines of:
```
import { useId } from "@react-aria/utils";

export function Accordion() {
  const id = useId();
  const [isExpanded, setExpanded] = useState(defaultExpanded);
  
  return (
    <>
      <button aria-controls={id} aria-expanded={isExpanded}>{title}</button>
      <div id={id} aria-hidden={!isExpanded}>{children}</div>
    </>
```

src/components/Accordion.tsx
bdow: This also makes me wonder, if you have only one Accordion, then how should it look? Should a single accordion include the bottom border? I'll ping the Figma again.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+          </span>
+        </div>
+      </button>
+      <div
+        {...testIds.details}
+        id={id} 
+        aria-hidden={!expanded}
+        css={{
+          // Use max-height for grow/shrink animation
+          ...Css.overflowHidden.maxhPx(1000)
```
roberthdg, line 105: For some reason, the transition animation doesn't get triggered with absolute values, I have to set it as `maxhPx(...)`

src/components/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+    if (!allowMultipleExpanded && index !== currentSelectedIndex) {
+      setExpanded(false)
+    }
+  }, [allowMultipleExpanded, currentSelectedIndex, index]);
+
+  return (
+    <div
+      {...testIds}
+      css={
+       {...Css.mbPx(4)
```
roberthdg, line 64: I had to add this margin because on the `AccordionList`, when one accordion was highlighed, it created an overlap with the next one

image

src/components/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+        }}
+      >
+        <div {...testIds.title} css={Css.p2.baseEm.if(disabled).gray500.$}>
+          {title}
+        </div>
+        <div css={Css.ml("auto").my("auto").prPx(14).$}>
+          <span
+            css={{
+              ...Css.mwPx(24).mhPx(24).pyPx(9).db.tc.cursorPointer.$,
+              ...{
```
bdow, line 89: **suggestion:** Can we remove this extra spread?
```
css={{
  ...Css.$,
  transition: "transform ...",
}}
```

src/components/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+        css={Css.w100.df.addIn(":focus", { outline: "none" }).addIn(":hover", Css.bgGray100.$).if(isFocusVisible).bshFocus.$}
+        onClick={() => {
+          setExpanded(!expanded)
+          if (!allowMultipleExpanded && setCurrentSelectedIndex) setCurrentSelectedIndex(index)
+        }}
+      >
+        <div {...testIds.title} css={Css.p2.baseEm.if(disabled).gray500.$}>
+          {title}
+        </div>
+        <div css={Css.ml("auto").my("auto").prPx(14).$}>
```
bdow, line 85: **suggestion:** Can we use flex styles on the button to position the content within?
```
<button css={Css.df.jcsb.gap2.if(disabled).gray500.$}>
  <div>{title}</div>
  <span><Icon /></span>
</button>
```
I'm thinking some of the existing css you have might not be completely necessary. Would be interesting to see how we can achieve the desired styles with less CSS.

src/components/Accordion.tsx
bdow: `height` animations are difficult as you cannot animate from `height: 0` to `height: auto`. Doing a max-height transition has been a workaround, but it isn't great. You end up transition from `max-height: 0` to `max-height: 1000` in `.25s`. This results in the transition appearing faster or slower depending on how tall the content really is. 
If you really want to animate the height, then you'd need to measure the height of the content using JS, then set the height to the measured value. (And typically I've see where upon animation complete, people will set `height: auto`;)

src/components/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+}
+
+interface AccordionListProps {
+  accordions: AccordionProps[];
+  /** Allows multiple accordions to be expanded simultaneously (enabled by default) */
+  allowMultipleExpanded?: boolean;
+  size?: AccordionSize;
+}
+
+export function AccordionList({ accordions, size, allowMultipleExpanded = true }: AccordionListProps) {
```
bdow, line 125: **suggestion:** Can this be its own file?

src/components/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
+import { Css, Icon, useTestIds, Palette } from "src/index";
+import { Dispatch, ReactNode, SetStateAction, useEffect, useState } from "react";
+import { useFocusRing } from "react-aria";
+import { useId } from "@react-aria/utils";
+
+export interface AccordionProps {
+  label?: string;
```
bdow, line 7: **question:** Is `label` necessary? I see we're only using it to provide a default testId. I would prefer we just remove this and set the testids hook as `useTestIds(props, "accordion");`

src/components/Accordion.tsx
bdow: **suggestion:** I would suggest we use an `inset` box-shadow for the focus state. That'd remove the issue of other accordions overlapping the focus styles.

src/components/AccordionList.tsx
```diff
@@ -0,0 +1,31 @@
+import { useState } from "react";
+import { useTestIds } from "..";
+import { Accordion, AccordionProps, AccordionSize } from "./Accordion";
+
+interface AccordionListProps {
+    accordions: AccordionProps[];
+    /** Allows multiple accordions to be expanded simultaneously (enabled by default) */
+    allowMultipleExpanded?: boolean;
+    size?: AccordionSize;
+}
```
bdow, line 10: **suggestion** Appears as though we have 4spaces for our tabs, where we should only have 2. Can you run `yarn format` on your workspace?

src/components/AccordionList.tsx
```diff
@@ -0,0 +1,31 @@
...
+        {accordions.map((accordionProps, index, arr) => (
+            <Accordion
+                {...accordionProps}
+                {...tid[index]}
+                key={index}
+                size={size}
+                bottomBorder={index === arr.length - 1}
+                allowMultipleExpanded={allowMultipleExpanded}
+                index={index}
+                currentSelectedIndex={currentSelectedIndex}
```
bdow, line 26: **question:** I wondering if we can hide some of the `allowMultipleExpanded`, `index`, `currentSelectedIndex` logic from the single Accordion and have the list handle this?

src/components/AccordionList.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export function AccordionList({ accordions, size, allowMultipleExpanded = true }: AccordionListProps) {
+    const [currentSelectedIndex, setCurrentSelectedIndex] = useState<number>();
+    const tid = useTestIds({}, "accordionList");
+
+    return (<>
+        {accordions.map((accordionProps, index, arr) => (
+            <Accordion
+                {...accordionProps}
+                {...tid[index]}
```
bdow, line 20: **suggestion** I would prefer we just do `{...tid}` without the index. In our tests we can always get the index if needed either by `r.queryAllByTestId("accordionList_title")[0]`, or our own helper `r.accordionList_title_0`.

src/components/AccordionList.tsx
bdow: Specifically, I'm hoping we can make Accordion component unaware of other accordions around it, and instead have the AccordionList be responsible for making sure the `allowMultipleExpanded` property is respected.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,100 @@
...
+          <Icon icon="chevronDown" />
+        </span>
+      </button>
+      <div
+        {...testIds.details}
+        id={id}
+        aria-hidden={!expanded}
+        css={{
+          // Use max-height for grow/shrink animation
+          ...Css.overflowHidden.maxhPx(1000).add("transition", "max-height 0.25s ease-in-out").$,
```
bdow, line 83: I really think if we want to animate the height then we should calculate it via JS. Otherwise we are introducing some weird delays in the `allowMultipleExapanded === false` case.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const [expanded, setExpanded] = useState(defaultExpanded);
+  const { isFocusVisible, focusProps } = useFocusRing();
+
+  useEffect(() => { setExpanded(defaultExpanded); }, [defaultExpanded])
+
+  return (
+    <div
+      {...testIds.container}
+      css={{
+        ...Css.if(topBorder).bt.bGray400.if(bottomBorder && !(!expanded && isFocusVisible)).bb.$,
```
bdow, line 47: **suggestion:** Maybe we should always set `bGray400`. Right now if `topBorder === false`, but `bottomBorder === true`, then you won't get the border because aren't setting `bGray400`. 
Also, I don't think we need the `&& !(!expanded && isFocusVisible)` any longer do we after we updated the focus state to use box-shadows.

src/components/Accordion.tsx
roberthdg: @bdow I added this condition `max-height ${expanded? "300ms" : "0"} ease-in-out` and it solves the delay issues (although it doesn't look as smooth when closing, but at least it keeps it just CSS)

roberthdg: I could also add to calculate this only if `index` is set (meaning it comes from an `AccordionList`), leaving the single accordion with smooth animation while closing

src/components/Accordion.tsx
```diff
@@ -0,0 +1,100 @@
...
+        </span>
+      </button>
+      <div
+        {...testIds.details}
+        id={id}
+        aria-hidden={!expanded}
+        css={{
+          // Use max-height for grow/shrink animation
+          ...Css.overflowHidden.maxhPx(1000).add("transition", `max-height ${expanded? "300ms" : "0"} ease-in-out`).$,
+          ...(!expanded || disabled ? Css.maxh0.$ : {}),
```
bdow, line 84: Sorry to make you go over this open and closing animation again. At this point I don't think getting the animation 100% correct is necessary. We can always follow up to address this later. As long as design is happy with the result of the animation, then I am okay with moving forward however you have it now.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,100 @@
...
+        onClick={() => {
+          setExpanded(!expanded);
+          if (setExpandedIndex) setExpandedIndex(index);
+        }}
+      >
+        <span>{title}</span>
+        <span
+          css={{
+            transition: "transform 250ms linear",
+            transform: expanded ? "rotate(180deg)" : "rotate(0deg)",
```
bdow, line 71: **thought (non-blocking):** This is not a request at all, more-so for my own musing. If we could come up with a list of support transitions or transforms and some how bake them into Truss, that'd be pretty cool. Maybe something like:
```
Css.transition("transform").fast.rotate0.if(expanded).rotate180.$
```
"fast" could represent a predetermined duration, and we could also include ease-in or linear somehow? This will be a bigger convo with design to figure out some animation patterns.

src/components/AccordionList.tsx
```diff
@@ -0,0 +1,32 @@
...
+  size?: AccordionSize;
+}
+
+export function AccordionList({ accordions, size, allowMultipleExpanded = true }: AccordionListProps) {
+  const [expandedIndex, setExpandedIndex] = useState<number>();
+  const tid = useTestIds({}, "accordionList");
+
+  return (
+    <>
+      {accordions.map((accordionProps, index, arr) => (
```
bdow, line 18: **question:** I have never used the third parameter in the `map` function. I'm curious if there is a benefit to using the third param (`arr`) vs referencing `accordions` when you check `bottomBorder={index === arr.length - 1}`. I'm guessing you'd typically use this if the map function was in a separate scope and didn't have access to the variable you're `.map`-ing over. 
Do you happen to know if there is a benefit to using one over the other if the `map`'d variable is in scope?

src/components/AccordionList.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface AccordionListProps {
+  accordions: AccordionProps[];
+  /** Allows multiple accordions to be expanded simultaneously (enabled by default) */
+  allowMultipleExpanded?: boolean;
+  size?: AccordionSize;
+}
+
+export function AccordionList({ accordions, size, allowMultipleExpanded = true }: AccordionListProps) {
+  const [expandedIndex, setExpandedIndex] = useState<number>();
+  const tid = useTestIds({}, "accordionList");
```
bdow, line 14: **nit:** Could you update to pass `props` into `useTestids` (`useTestIds(props, "accordionList")`). Typically in Beam where we are making generic components, it makes sense to use the component's `props` as the first param of `useTestIds`. This allows the caller to define their own custom testid for the component if they wish. 
In `internal-frontend` we don't do that as much because our components aren't as generic and it doesn't make sense to allow for overriding the default test id.

src/components/Accordion.tsx
```diff
@@ -0,0 +1,100 @@
...
+          ...(!expanded || disabled ? Css.maxh0.$ : {}),
+        }}
+      >
+        <div css={Css.px2.pb2.pt1.$}>{children}</div>
+      </div>
+    </div>
+  );
+}
+
+export type AccordionSize = "xs" | "s" | "m" | "l";
```
bdow, line 93: **suggestion:** Let's align with existing size names that are use else where. Specifically, `"xs" | "sm" | "md" | "lg"`

src/components/AccordionList.tsx
roberthdg: That's a feedback I got from Tyler a while back, I guess it makes it more succinct that you're re-using that particular array you're mapping. I'm not very opinionated on the matter, probably comes down to personal preferences (and some very specific edge cases)

src/components/Accordion.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    expect(r.accordion_title()).not.toHaveAttribute("disabled");
+  });
+
+  it("can expand by default", async () => {
+    // Given an accordion component with defaultExpanded set
+    // When rendered
+    const r = await render(
+      <Accordion title="Test title" defaultExpanded>
+        {" "}
+        Test description{" "}
```
bdow, line 26: **nit (non-blocking):** I'm guessing the formatter did this, but those `{" "}` are unnecessary for the test.

src/components/Accordion.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    expect(r.accordion_container()).toHaveStyleRule("border-bottom-style", "solid");
+  });
+
+  it("can hide the top border", async () => {
+    // Given an accordion component with topSection set as false
+    // When rendered
+    const r = await render(
+      <Accordion title="Test title" topBorder={false}>
+        {" "}
+        Test description{" "}
```
bdow, line 57: **nit: (non-blocking)** Same here regarding `{" "}`

src/components/Accordion.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    // When rendered
+    const r = await render(
+      <Accordion title="Test title" disabled>
+        {" "}
+        Test description{" "}
+      </Accordion>,
+    );
+
+    // Then the dropdown button is disabled
+    expect(r.accordion_title()).toHaveAttribute("disabled");
```
bdow, line 76: **suggestion:** Maybe validate `expect(r.accordion_details).not.toBeVisible()` as well.

src/components/Accordion.tsx
roberthdg: I think it comes pretty close to what Radhi asked for on the Figma, I'll ping her on the ticket with the SB link so she gives it a look. I also added the `expanded || !index ? "250ms" : "0"` so the single accordion animation will be complete, and when it's used on the `AccordionList` it's not as noticeable.

src/components/Accordion.test.tsx
bdow, line 76: **suggestion:** Use `toBeDisabled()` instead of `toHaveAttribute("disabled")` as that could potentially pass if for some reason the element had `disabled="false"`.

bdow (approved):
Just a couple test suggestions, otherwise good to go. Thanks for contributing to Beam!

PR #648, opened 2022-09-14
feat: create `ToggleButton`

0ces:
imagen


src/inputs/ToggleButton.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+describe("ToggleButton", () => {
+  it("renders", async () => {
+    // Given a ToggleButton
+    const r = await render(
+      <ToggleButton
+        label={"Toggle button"}
+        onClick={() => {}}
+      />
+    );
```
bdow, line 12: **nit:** Can this be done in a single line? I personally find the code easier to scan and read through when we try to limit the number of lines of code.
```
const r = await render(<ToggleButton label={"Toggle button"} onClick={() => {}} />);
```

0ces: Sure, I'll change them :)

src/inputs/ToggleButton.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    //Given a non selected ToggleButton
+    await r.rerender(
+      <ToggleButton
+        label={"Toggle button"}
+        onClick={() => {}}
+      />
+    );
+    // Then it isn't checked
+    expect(r.toggleButton_value()).not.toBeChecked();
```
bdow, line 37: **suggestion:** What do you think about removing this extra render, and instead validating `not.toBeChecked` in the `it("renders")` test?

src/inputs/ToggleButton.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  });
+  it("fires onClick", async () => {
```
bdow, line 39: **nit:** Please add a new line between each `it` method.

src/inputs/ToggleButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+        css={{
+          ...Css.relative.cursorPointer.df.w("max-content").smEm.selectNone.$,
+          ...(isDisabled && Css.cursorNotAllowed.gray400.$),
+        }}
+        aria-label={label}
+      >
+        <button
+           aria-hidden="true"
+          {...buttonAttrs}
+        >
```
bdow, line 91: **thought** I am thinking we don't really want a button element here. As it is right now, we have nested interactions. When a user clicks on the `<button />` element, then we fire the `onPress`, then that click bubbles up to the `<label/>` which toggle the state of the `<input />`. 
I would suggest we get rid of the button and `useButton`. The label should be sufficient for our click interaction.

chr1sjf0x (commented):
TY @0ces

src/inputs/ToggleButton.tsx
0ces: Yeah, I tried this at first, but the problem is that `useSwitch` doesn't support a `onPress` so we won't have the `isPressed` state, which is minor thing but I was just following the figma design

0ces: Also, the async behavior doesn't seem to work properly when applying the same logic on the `onChange`

bdow: Hmm.. I think we can still get around this. Let me pull this down and take a look as well.

0ces: @bdow pushed!

src/components/Icon.tsx
```diff
@@ -344,6 +344,9 @@ export const Icons = {
   house: (
     <path d="M5.00299 22H9.00299H15.003H19.003C20.106 22 21.003 21.103 21.003 20V11C21.003 10.735 20.898 10.48 20.71 10.293L12.71 2.29301C12.319 1.90201 11.687 1.90201 11.296 2.29301L3.29599 10.293C3.10799 10.48 3.00299 10.735 3.00299 11V20C3.00299 21.103 3.89999 22 5.00299 22ZM10.003 20V15H14.003V20H10.003ZM5.00299 11.414L12.003 4.41401L19.003 11.414L19.004 20H16.003V15C16.003 13.897 15.106 13 14.003 13H10.003C8.89999 13 8.00299 13.897 8.00299 15V20H5.00299V11.414Z" />
   ),
+  bell: (
+    <path d="M16 11.586V8C16 4.783 13.815 2.073 10.855 1.258C10.562 0.52 9.846 0 9 0C8.154 0 7.438 0.52 7.145 1.258C4.185 2.074 2 4.783 2 8V11.586L0.293 13.293C0.105 13.48 0 13.734 0 14V16C0 16.553 0.447 17 1 17H17C17.553 17 18 16.553 18 16V14C18 13.734 17.895 13.48 17.707 13.293L16 11.586ZM16 15H2V14.414L3.707 12.707C3.895 12.52 4 12.266 4 12V8C4 5.243 6.243 3 9 3C11.757 3 14 5.243 14 8V12C14 12.266 14.105 12.52 14.293 12.707L16 14.414V15ZM9 20C10.311 20 11.407 19.166 11.818 18H6.182C6.593 19.166 7.689 20 9 20Z" />
```
bdow, line 348: This icon doesn't appear to have been copied to properly sit within a 24x24 bounding box. See the result here:
Screen Shot 2022-09-14 at 3 07 18 PM
This causes the button to not align as expected against text.

0ces: Huh, you're right, I'll check it

src/inputs/ToggleButton.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    click(r.toggleButton, { allowAsync: true });
+    expect(r.toggleButton_value()).toBeDisabled();
+    await wait();
+    expect(r.toggleButton_value()).not.toBeDisabled();
+    expect(onError).toBeCalledWith("Promise error");
+  });
+
+  it("can define a custom testid", async () => {
+    const r = await render(<ToggleButton label={"Custom test id"} onChange={() => {}} />);
+    expect(r.customTestId()).toBeInTheDocument();
```
bdow, line 68: **suggestion:** The "custom testid" test should validate that the caller can specify a `data-testid` property on the component. This is rather than having it defined by the `label`, which is the passed to `useTestIds` as the default value.
Instead the test should look something like:
```
const r = await render(<ToggleButton label="Not the test id" data-testid="customTestId" ... />);
expect(r.customTestId).toBeInTheDocument();

Comment (0ces) [reply src/inputs/ToggleButton.test.tsx:68]: You're right, I thought that but didn't implement it at the end, I'll do it right away

File: src/inputs/ToggleButton.tsx [hunk 7]
```diff
@@ -0,0 +1,82 @@
...
+  /** Input label */
+  label: string;
+  selected?: boolean;
+  onChange: ((selected: boolean) => void) | ((active: boolean) => Promise<void>);
+  autoFocus?: boolean;
+  icon?: IconKey;
+  /** Whether the field is disabled. If a ReactNode, it's treated as a "disabled reason" that's shown in a tooltip. */
+  disabled?: boolean | ReactNode;
+  tooltip?: ReactNode;
+  /** Handler when the interactive element state changes. */
```
bdow, line 18: **nit:** Delete if not needed

src/inputs/ToggleButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { useToggleState } from "react-stately";
+import { Icon, IconKey, maybeTooltip, resolveTooltip } from "src/components";
+import { Css } from "src/Css";
+import { isPromise, useTestIds } from "src/utils";
+
+export interface ToggleButtonProps {
+  /** Input label */
+  label: string;
+  selected?: boolean;
+  onChange: ((selected: boolean) => void) | ((active: boolean) => Promise<void>);
```
bdow, line 12: **suggestion:**  Can we use the same variable name for the two callbacks? Right now we have `selected` or `active`. I think `selected` makes sense.

src/inputs/ToggleButton.test.tsx
bdow: No rush, it can wait until tomorrow.

bdow (approved):

PR #650, opened 2022-09-16
feat: Remove 'nestedCard' styles from GridTable

bdow:
The nested card styling for GridTable was a prototype style that we have no moved away from after gathering feedback. The consensus of the feedback was that the card styles adding more spacing than necessary to the rows of data causing fewer rows of the table to be visible within the user's viewport.


stephenh (approved):
:dagger: :dagger: :dagger: 

:-D

PR #651, opened 2022-09-16
fix: Respect initCollapsed when rows are lazily rendered

bdow:


chr1sjf0x (approved):

src/components/Table/GridTable.test.tsx
```diff
@@ -2226,6 +2226,40 @@ describe("GridTable", () => {
...
   });
 
+  it("can lazily initialize table with collapsed rows", async () => {
+    // Given a table that can update its set of rows
+    function TestComponent() {
+      const [rows, setRows] = useState<GridDataRow<NestedRow>[]>(staticRows);
+      return (
+        <>
+          <button onClick={() => setRows(initRows)} data-testid="initRows" />
+          <GridTable columns={nestedColumns} rows={rows} fallbackMessage="Loading..." />
```
stephenh, line 2236: Thought: We're not actually using `persistCollapse` here? The code changes seemed to be focus on "if `persistCollapse`", so I expected it in the test.

src/components/Table/RowState.ts
```diff
@@ -80,23 +80,26 @@ export class RowState {
...
-    // If the set of rows are different, and this is the first time adding "data" rows (non "totals" or "header" rows), then collapsedRows may need to be updated.
-    if (rows !== this.rows && this.rows.some((r) => r.kind !== "totals" && r.kind !== "header")) {
+    // If the set of rows are different
+    if (rows !== this.rows) {
       const currentCollapsedIds = this.collapsedIds;
       // Create a list of the (maybe) new rows that should be initially collapsed
       const maybeNewCollapsedRows = flattenRows(rows).filter((r) => r.initCollapsed);
+      // Check against local storage for collapsed state only if this is the first render of "data" (non-header or totals) rows.
+      const checkLocalStorage =
+        this.persistCollapse && !this.rows.some((r) => r.kind !== "totals" && r.kind !== "header");
```
stephenh, line 102: Thought: I was surprised that `persistCollapse` has not already overwritten local storage; i.e. `RowState.loadCollapse` calls `sessionStorage.setItem`, but looks like that's only if there are non-zero collapsed rows, and the header/totals wouldn't be collapsed...

src/components/Table/GridTable.test.tsx
bdow: The bug this fixed was happening when we lazily load rows regardless of `persistCollapse` being set. The `setRows` function was only checking for `initCollapsed` if it was **not** the first time loading "data" rows.
Previous if statement:
```
if (rows !== this.rows && this.rows.some((r) => r.kind !== "totals" && r.kind !== "header")) {
```

src/components/Table/RowState.ts
```diff
@@ -80,23 +80,26 @@ export class RowState {
...
-    })
+    });
 
     this.selectedRows.merge(map);
   }
 
   // Updates the list of rows and regenerates the collapsedRows property if needed.
   setRows(rows: GridDataRow<any>[]): void {
-    // If the set of rows are different, and this is the first time adding "data" rows (non "totals" or "header" rows), then collapsedRows may need to be updated.
-    if (rows !== this.rows && this.rows.some((r) => r.kind !== "totals" && r.kind !== "header")) {
```
stephenh, line 96: Fwiw @bdow I was able to get the new test to pass by just changing this line to:

```
    if (rows !== this.rows && rows.some((r) => r.kind !== "totals" && r.kind !== "header")) {
```

I.e. checking the _incoming_ rows for "we have a real row" now.

Previously both `setRows` calls from the test were not entering this block b/c

* First call, this.rows is empty
* Second call, this.rows is the original static total+header

...I guess your new line 102 does add a `!` that was not here before:

```
    if (rows !== this.rows && !this.rows.some((r) => r.kind !== "totals" && r.kind !== "header")) {
```

So perhaps that is the core fix? If so, that makes sense; given the lack of `persistCollapse` in the test and just not 100% following line 116, I wonder if the `checkLocalStorage` aspect of this code either wasn't necessary and/or isn't being covered...

src/components/Table/GridTable.test.tsx
stephenh: Ah okay; I see we've got a separate test for the persistCollapse test; cool, looks good then. Thanks!

PR #653, opened 2022-09-20
fix: Export Accordion component

roberthdg:


src/components/index.ts
```diff
@@ -31,3 +31,5 @@ export { TabActions, TabContent, Tabs, TabsWithContent } from "./Tabs";
 export type { RouteTab, RouteTabWithContent, Tab, TabWithContent } from "./Tabs";
 export * from "./Tag";
 export * from "./Tooltip";
+export * from "./Accordion";
```
roberthdg, line 34: If I put this at the top of the file (in alpbabetical order), it throws an export error, probably because it's using `Icon` that hasn't been exported yet

stephenh (approved):

src/components/index.ts
bdow: Ah, maybe. You could always import the Icon from the Icon file directly vs from `src/index`.

PR #656, opened 2022-09-22
fix: Add 'cellHighlight' to grid styles cache key

bdow:


src/components/Table/styles.tsx
```diff
@@ -54,8 +54,8 @@ export interface GridStyleDef {
 function memoizedTableStyles() {
   const cache: Record<string, GridStyle> = {};
   return (props?: GridStyleDef) => {
-    const { inlineEditing = false, grouped = false, rowHeight = "flexible", cellHighlight } = props || {};
-    const key = `${inlineEditing}|${grouped}|${rowHeight}`;
+    const { inlineEditing = false, grouped = false, rowHeight = "flexible", cellHighlight = false } = props || {};
+    const key = `${inlineEditing}|${grouped}|${rowHeight}|${cellHighlight}`;
```
stephenh, line 58: @bdow wdyt about a `JSON.stringify(props)` so that new props going forward are handled for free/

chr1sjf0x (approved):
TY 🍣

src/components/Table/styles.tsx
bdow: I like the idea. I wonder how that'd look... Will try.

bdow: Would be worried props may not always be returned in the same order, so could be tricky. Might be able to do something like `Object.keys`, sort, get values, then stringify.

bdow: Pushed up a change, let me know what you think! Thanks for the suggestion.

src/components/Table/styles.tsx
```diff
@@ -53,9 +54,12 @@ export interface GridStyleDef {
 
 function memoizedTableStyles() {
   const cache: Record<string, GridStyle> = {};
-  return (props?: GridStyleDef) => {
-    const { inlineEditing = false, grouped = false, rowHeight = "flexible", cellHighlight } = props || {};
-    const key = `${inlineEditing}|${grouped}|${rowHeight}`;
+  return (props: GridStyleDef = {}) => {
+    const { inlineEditing = false, grouped = false, rowHeight = "flexible", cellHighlight = false } = props;
```
stephenh, line 58: Nit: could drop this line now? Good call on the sort!

stephenh (approved):

src/components/Table/styles.tsx
bdow: I don't think we can drop this line, we still use the destructured props to apply the styles.

stephenh: Oh right :-) Nm!

PR #658, opened 2022-09-23
chore: add eslint config

roberthdg:
Get Beam onto the shared eslint/prettier/vscode setup


bdow:
Thanks! I would suggest updating the prefix to "chore" rather than "fix" as we don't really need to publish a new version of Beam for this.

bdow (approved):

stephenh (approved):
Looks great!

Fwiw I'd also like to get the eslint hooks usage rule (`react-hooks/exhaustive-deps`) turned on for both Beam + internal-frontend...I don't think we disabled it on purpose, it just happened by accident when we ejected from CRA/something/...

@roberthdg if you're looking for other eslint-related things to do... :-D

PR #659, opened 2022-09-23
chore(storybook): Update storybook to use controls and update theme

kbesingeryeh:
* Updated Storybook mdx files to use `Prop` component to display controls
* Added a Storybook theme to be able to update colors/fonts/typography

### DIFF
--- a/.storybook/beamTheme.tsx
+++ b/.storybook/beamTheme.tsx
@@ -1,0 +1,37 @@
+import { create } from "@storybook/theming";
+
+export default create({
+  base: "light",
+
+  colorPrimary: "#0369A1",
+  colorSecondary: "#646464",
+
+  // UI
+  appBg: "white",
+  appContentBg: "#F7F5F5",
+  appBorderRadius: 4,
+
+  // Typography
+  fontBase: '"Inter", sans-serif',
+  fontCode: "monospace",
+
+  // Text colors
+  textColor: "black",
+  textInverseColor: "rgba(255,255,255,0.9)",
+
+  // Toolbar default and active colors
+  barTextColor: "white",
+  barSelectedColor: "black",
+  barBg: "#0369A1",
+
+  // Form colors
+  inputBg: "white",
+  inputBorder: "#646464",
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -7626,11 +7626,6 @@ ast-types@^0.14.2:
   dependencies:
     tslib "^2.0.1"
 
-astral-regex@^2.0.0:
-  version "2.0.0"
-  resolved "https://registry.npmjs.org/astral-regex/-/astral-regex-2.0.0.tgz"
-  integrity sha512-Z7tMw1ytTXt5jqMcOP+OQteU1VuNK9Y02uuJtKQ1Sv69jXQKKg5cibLwGJow8yzZP+eAc18EmLGPal0bp36rvQ==
-
```
bdow, line 7633: This is surprising... but fine with me. Maybe it was a transitive dependency that got removed... 🤷

beamLogo.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
```
bdow, line 1: **question:** Did we need to import React? I sometimes see this is a warning, but never an error. Typically it can be ignored.

.storybook/beamTheme.tsx
```diff
@@ -0,0 +1,37 @@
...
+  inputBg: "white",
+  inputBorder: "#646464",
+  inputTextColor: "black",
+  inputBorderRadius: 4,
+
+  brandTitle: "Beam Storybook",
+  brandUrl: "https://homebound.com",
+  brandImage: "https://upload.wikimedia.org/wikipedia/commons/thumb/6/6a/JavaScript-logo.png/600px-JavaScript-logo.png",
+  brandTarget: "_self",
+});
```
bdow, line 37: Thanks for starting this. I think it'll be fun to style Storybook a bit.

bdow (approved):

beamLogo.tsx
kbesingeryeh: no I don't think so!

PR #660, opened 2022-09-26
chore: Hack Day Storybook Reorganization

bdow:


src/components/Accordion.stories.tsx
```diff
@@ -4,7 +4,7 @@ import { Accordion } from "./Accordion";
 
 export default {
   component: Accordion,
-  title: "Components/Accordion",
+  title: "Workspace/Components/Accordion",
```
stephenh, line 7: Fwiw eventually I'd love to see these `title`s inferred from file system location:

https://storybook.js.org/docs/react/configure/sidebar-and-urls#csf-30-auto-titles

bdow: Are you suggesting we organize our folder structure to fit how we want it to show up in Storybook? Where we'd end up having `src/catalog`, `src/workspace` and `src/design-qa` folders.
That would break with our current file structure of keeping the component and its respective files next to it.
Maybe we end up with something like:
```
/src/workspace/components/Button.stories.tsx
/src/catalog/components/Button.stories.tsx
/src/design-qa/components/Button.stories.tsx
/src/impl/components/Button.tsx
/src/impl/components/Button.test.tsx
````

stephenh: Yeah, I guess... What was the idea going to be, in terms of file organization?

Like I see `Accordian.stories.tsx` looks like a `Workspace/` and `Accordian.stories.mdx` is a `Catalog/` ... so just using the `tsx` vs. `mdx` file extension is the convention for workspace vs. catalog? What about design catalog?

...just reading the design-qa overview, if it's based on Figma boards, I wouldn't be surprised if it's structure is just going to fundamentally be different, and so having its own `design-qa` folder makes sense.

bdow: Initial idea was that workspace stories would remain as they are, Catalog stories would be in `mdx` format, and the Design QA stories would be named as such...
**Workspace**: `Button.stories.tsx`
**Catalog**: `Button.stories.mdx`
**Design QA**: `Button.qa.stories.tsx` (not the `.qa` in the name)

I'm good with thinking back through the file structure.

stephenh: Ah yeah, I mean, I do like all of the Button file+test+stories living next to each other...

It seems like that convention does assume all "pages" for each set of docs (workspace/catalog/design) are always `<some component name>.qa|tsx|mdx`... 

I like that convention of `<some component name>...` makes a lot of sense for the workspace stories ... and probably the catalog?, but I'd kinda assume the design qa stories would drift quite a bit to be more like `<some board name>`? Not sure...

bdow: Sure. I could see that drift. My only thing is that I wouldn't want to block moving in the right direction by not having it all figured out yet. (which I think the storybook reorg is the right direction and wouldn't impact how we change our file structure in the future)

stephenh (approved):

PR #663, opened 2022-09-29
feat: Support backgroundColor on Tabs and ScrollableContent

bdow:


src/components/Layout/ScrollableParent.tsx
```diff
@@ -64,13 +64,13 @@ export function ScrollableParent(props: PropsWithChildren<ScrollableParentContex
...
             ...Css.pl(context.pl).pr(context.pr).$,
-            ...(!hasScrollableContent ? { ...Css.overflowAuto.$, ...scrollContainerBottomPadding } : undefined),
+            ...(!hasScrollableContent ? { ...Css.overflowAuto.h100.$, ...scrollContainerBottomPadding } : undefined),
           }}
         >
           {children}
         </div>
         {/* Set fg1 to take up the remaining space in the viewport.*/}
-        <div css={Css.fg1.overflowAuto.pl(context.pl).pr(context.pr).$} ref={scrollableRef} />
+        <div css={Css.fg1.overflowAuto.$} ref={scrollableRef} />
```
bdow, line 73: The padding has been moved into ScrollableContent

src/components/Layout/ScrollableContent.tsx
```diff
@@ -28,18 +28,15 @@ export function ScrollableContent(props: ScrollableContentProps): ReactPortal |
...
-    ),
+    <div
+      css={{
+        ...Css.pr(pr).pl(pl).if(virtualized).pr0.h100.$,
+        ...(bgColor && Css.bgColor(bgColor).$),
+        ...(!omitBottomPadding && !virtualized && scrollContainerBottomPadding),
+      }}
+    >
+      {children}
+    </div>,
```
bdow, line 39: Moving the `paddingLeft` and `paddingRight` definitions here allows us to remove the need of the `<FullBleed />` component when the `virtualized` prop is `true`.  This helped simplify the return logic as well.

stephenh (approved):

src/components/Layout/Layout.stories.tsx
```diff
@@ -110,6 +110,29 @@ export function ScrollableParentFallback() {
...
+  ];
+  return (
+    <TestProjectLayout>
+      <TestHeader title="Change Event - Mud Room" />
+      <div css={Css.pt1.$}>
+        <TabsWithContent
+          selected={selectedTab}
+          tabs={tabs}
+          onChange={(t) => setSelectedTab(t)}
+          contentXss={Css.bgColor(contentBgColor).$}
```
bdow, line 129: Unfortunately to really get this to work as we'd want with tabs, we need to define the same background color on the `<Tab contentXss />` and defined on `<ScrollableContent bgColor />`. Can get into more details if needed, but basically it's due to the Tab's content being moved into the ScrollableContent portal.

PR #664, opened 2022-09-29
fix: Always apply 'h100' on ScrollableContent wrapper

bdow:
This fixes issues where users may have a Virtualized component within the ScrollableContent, but forgot to set the ScrollableContent.virtualized prop


Grandlouis (approved):
🍕

PR #665, opened 2022-09-29
feat: (SC-18433) Edit columns button component

pgamboa07:
[SC-18433](https://app.shortcut.com/homebound-team/story/18433/variance-baseline-edit-columns-in-schedule)

Screen Shot 2022-09-29 at 15 33 01


src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export function useColumns<R extends Kinded, S = {}>(tableColumns: GridColumn<R, S>[]):
+[Columns<R, S>, Dispatch<SetStateAction<Columns<R, S>>>] {
+
+  const initColumns: Columns<R, S> = useMemo(() => {
+    return {
+      allColumns: tableColumns,
+      visibleColumns: tableColumns.filter((column) => ((column.canHide && column.visible) || !column.canHide))
+    }
+  },[tableColumns]);
```
bdow, line 87: **nit:** I think there are some formatting issues in this. Can you try running `yarn format`?

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+    </OverlayTrigger>
+  );
+}
+
+export interface Columns<R extends Kinded, S = {}> {
+  allColumns: GridColumn<R, S>[],
+  visibleColumns: GridColumn<R, S>[]
+}
+
+export function useColumns<R extends Kinded, S = {}>(tableColumns: GridColumn<R, S>[]):
```
bdow: **suggestion:** Can we create a different file for the hook?

src/components/Table/GridTable.tsx
```diff
@@ -827,6 +827,10 @@ export type GridColumn<R extends Kinded, S = {}> = {
   isAction?: true;
   /** Column name that will be used to generate an unique identifier for every row cell */
   name?: string;
+  /** Flag that will allow to know which columns are hide-able */
+  canHide?: boolean;
+  /** Flag that will allow to know which columns are hide-able */
```
bdow, line 832: **nit:** Looks like a copy and paste of the JS doc above.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    }}
+                  />
+                </div>
+              )
+            );
+          })}
+        </div>
+        <div css={Css.smMd.$}>
+          <Button size="sm" variant={"text"} label={buttonText || "Clear selections"} onClick={clearSelections} />
+        </div>
```
bdow, line 74: **request:** This should be a tertiary button per Figma mock. Using the tertiary button will remove the need for the wrapping element as well.
See Figma:
Screen Shot 2022-09-30 at 8 29 36 PM

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+      visibleColumns: columns.filter((column) => (column.canHide && column.visible) || !column.canHide),
+    }));
+  }
+
+  return (
+    <OverlayTrigger {...props} menuTriggerProps={menuTriggerProps} state={state} buttonRef={buttonRef} {...tid}>
+      <div
+        css={{
+          ...Css.df.fdc.bgWhite.py3.px2.maxw(px(380)).$,
+          "&:hover": Css.bshHover.$,
```
bdow, line 45: **nit/question:** Any reason we need this to be `display: flex`? As far as I can tell we don't need this to be a flex container. I'd prefer to remove `df.fdc` as we should get the same result as you have here already.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+      ...prevState,
+      visibleColumns: columns.filter((column) => (column.canHide && column.visible) || !column.canHide),
+    }));
+  }
+
+  return (
+    <OverlayTrigger {...props} menuTriggerProps={menuTriggerProps} state={state} buttonRef={buttonRef} {...tid}>
+      <div
+        css={{
+          ...Css.df.fdc.bgWhite.py3.px2.maxw(px(380)).$,
```
bdow, line 44: **suggestion:** I noticed in the mocks that the `py` is 40px top and bottom, so maybe we do `py5` instead of `py3` here. Also, the horizontal padding appears to be 24px in the mocks, so that should be `px3`.

bdow, line 44: **nit/suggestion:** Most of our helpers have a `Px` suffix so you do not need to wrap with `maxw(px(380))`, instead you can do `maxwPx(380)`. Small change 🤷

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+          "&:hover": Css.bshHover.$,
+        }}
+      >
+        <div css={Css.gray500.sm.ttu.$}>{title || "Select columns to show"}</div>
```
bdow, line 48: **suggestion:** Based on the mocks, this text should be `xsSb` - 12px in height and font-weight of 600. I've ping Andrea on this mock to see if she can have this better defined in the Inspect panel so the design to eng handoff is a bit smoother.

bdow: She also used a color that doesn't exist in our palette, so I'm good with the one you chose.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <div css={Css.df.add({ flexWrap: "wrap" }).gap2.py2.$}>
+          {columns.map((column, i) => {
+            return (
+              column.canHide && (
+                <div css={Css.add({ flex: "40%" }).ttc.$} key={i}>
```
bdow, line 53: **question/suggestion:** Do we really need this to be display flex? I feel like we just want two columns of checkboxes, with results in us needing to set the flex-basis for every element to 40%. I would suggest using CSS Grid for this, I think we'll have better luck. Something like:

```suggestion
        <div css={Css.dg.gtc("auto auto").gap2.py2.$}>
          {columns.map((column, i) => {
            return (
              column.canHide && (
                <div key={i}>
```

bdow, line 53: **suggestion:** I would suggest not doing a `text-transform: capitalize` here. I think it might assume too much.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+                      column.visible = value;
+                      setColumns((prevState) => ({
+                        ...prevState,
+                        visibleColumns: columns.filter(
+                          (column) => (column.canHide && column.visible) || !column.canHide,
+                        ),
+                      }));
+                    }}
+                  />
+                </div>
```
bdow, line 67: **suggestion/thought:** I think we might be better off using a `<CheckboxGroup />` component here. That'd result in fewer `onChange` callbacks to define (one per the group vs one per column).

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+                <div css={Css.add({ flex: "40%" }).ttc.$} key={i}>
+                  <Checkbox
+                    label={column.name || ""}
+                    selected={!!column.visible}
+                    onChange={(value) => {
+                      column.visible = value;
+                      setColumns((prevState) => ({
+                        ...prevState,
+                        visibleColumns: columns.filter(
+                          (column) => (column.canHide && column.visible) || !column.canHide,
```
bdow, line 62: **suggestion:** What do you think if we change the filter logic to:
```
(column) => column.canHide ? column.visible : true,
```
So if the column can be hidden, then we check the `visible` property. Otherwise, it can't be hidden so always return true to make it visible.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+                        ),
+                      }));
+                    }}
```
bdow, line 65: **thought:** This logic will change up a bit once we move towards using the `<CheckboxGroup />`, but I'm thinking we don't need to set `column.visible = value` here. It feels a little out of place setting a property that is not reactive, then following up immediately with a `setState` call. I'm wondering if we can do all of our logic within the `setColumns`.. 🤔 . We might have to when we switch to the `<CheckboxGroup />`

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+        }}
+      >
+        <div css={Css.gray500.sm.ttu.$}>{title || "Select columns to show"}</div>
+        <div css={Css.df.add({ flexWrap: "wrap" }).gap2.py2.$}>
+          {columns.map((column, i) => {
+            return (
+              column.canHide && (
+                <div css={Css.add({ flex: "40%" }).ttc.$} key={i}>
+                  <Checkbox
+                    label={column.name || ""}
```
bdow, line 55: **thought:** I wonder if we should do a `console.warn` if the column does not have a `name` property. Also, rather than showing a blank `label`, I think we should filter out columns without a name.
So maybe ahead of this loop we do something along the lines of:
```
const editableColumns = useMemo(() => columns.filter((c) => {
  if (!c.canHide) return false;
  if (!c.name || c.name.length === 0) { 
    console.warn("Column is missing 'name' property required by the Edit Columns button", column);
    return false;
  }
  return true;
}), [columns])
```
(pseudo code, so please don't expect it to immediately work 😄)

src/components/Table/useColumns.ts
```diff
@@ -0,0 +1,21 @@
+import { Dispatch, SetStateAction, useMemo, useState } from "react";
+import { GridColumn, Kinded } from "./GridTable";
+
+export interface Columns<R extends Kinded, S = {}> {
```
bdow, line 4: **nit/suggestion:** My gut tells me this is more of a `type` than an `interface`. Typically we use `interface`s as the definition of what props are required for "interfacing" with a component. Where a type is more the structure of an object. Can we update this to be a `type` instead of an interface, as it is the return type of the memo'd data.

bdow, line 4: **nit/suggestion:** I feel like `Columns` is too generic, but I can't think of a better name. Maybe something like: `GridColumnState`?

src/components/Table/useColumns.ts
```diff
@@ -0,0 +1,21 @@
...
+  allColumns: GridColumn<R, S>[];
+  visibleColumns: GridColumn<R, S>[];
+}
+
+export function useColumns<R extends Kinded, S = {}>(
+  tableColumns: GridColumn<R, S>[],
+): [Columns<R, S>, Dispatch<SetStateAction<Columns<R, S>>>] {
+  const initColumns: Columns<R, S> = useMemo(() => {
+    return {
+      allColumns: tableColumns,
```
bdow, line 14: **question:** Do we need `allColumns`? What is that providing us?

src/components/Table/useColumns.ts
```diff
@@ -0,0 +1,21 @@
...
+      visibleColumns: tableColumns.filter((column) => (column.canHide && column.visible) || !column.canHide),
```
bdow, line 15: **suggestion:** Same we earlier, I think we can make this a bit easier to read via:
```suggestion
      visibleColumns: tableColumns.filter((column) => column.canHide ? column.visible : true,
```

src/components/Table/useColumns.ts
bdow, line 4: Can we move this `type` to the bottom of the file. Otherwise people may assume that it is the interface for the hook, which it is not. This is per our component styling guide: https://homebound.atlassian.net/wiki/spaces/ENG/pages/339574863/Component+Style+Guide#Component-Structure

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+                      }));
+                    }}
+                  />
+                </div>
+              )
+            );
+          })}
+        </div>
+        <div css={Css.smMd.$}>
+          <Button size="sm" variant={"text"} label={buttonText || "Clear selections"} onClick={clearSelections} />
```
bdow, line 73: **nit/suggestion:** Can we remove the `buttonText` prop? I don't see a reason why someone would need to override this, and I don't see an ask for this in the associated ticket.

bdow, line 73: **nit:** You can remove `size="sm"`, that is the default size.

src/components/Table/EditColumnsButton.tsx
bdow, line 53: Not sure if we'll need the `key` any more once we use the `<CheckboxGroup />`, but I would push to not use an `index` value as much as possible. We could instead use `column.name` in this instance (as long as we removed columns that don't have names prior to this). But again, might not be necessary to worry about if when moving to `<CheckboxGroup />`

bdow: fun fact! `<CheckboxGroup />` already uses `display: grid`. However, we not allow for defining `grid-template-columns`. I think that'd be an easy addition though, something like:
```
interface CheckboxGroupProps {
  ...
  columns?: number;
}

function CheckboxGroup(props) {
  const { columns = 1, ... } = props;
  
  return <div css={Css.dg.gtc(`repeat(${columns}, auto)`).$}>....
```

src/components/Table/useColumns.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  const nameColumn: GridColumn<Row> = { name: "name", header: "Name", data: ({ name }) => name };
+  const valueColumn: GridColumn<Row> = {
+    name: "value",
+    header: "Value",
+    canHide: true,
+    visible: false,
+    data: ({ value }) => value,
+  };
+
+  it("Should return all Columns", async () => {
```
bdow, line 18: **nit:** Please update the test names from `it("Should ...")` to `it("should ...")`. We typically do not capitalize the first word in the test name.

src/components/Table/EditColumnsButton.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+        trigger={{ label: "Columns" }}
+        columns={[nameColumn, valueColumn, actionColumn]}
+        setColumns={setColumns}
+        defaultOpen={true}
+      />,
+    );
+    // When click clearSelections button
+    click(r.clearSelections);
+    // Expect for setColumns to be called
+    expect(setColumns).toHaveBeenCalled();
```
bdow, line 79: **nit:** The logic for clearing the selections lives within the EditColumnsButton. In that sense, I don't think validating that `setColumns` was called is enough, but instead we should ensure `setColumns` was called and none of the columns are visible any longer.

src/components/Table/EditColumnsButton.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+        trigger={{ label: "Columns" }}
+        columns={[nameColumn, valueColumn, actionColumn]}
+        setColumns={setColumns}
+        defaultOpen={true}
+      />,
+    );
+    // When click on an option
+    click(r.value);
+    // Expect for setColumns to be called
+    expect(setColumns).toHaveBeenCalled();
```
bdow, line 62: **suggestion:** Same thing here about checking that `setColumns` was called. I don't think this is enough of a test as we need to make sure that setColumns was called and the column state was properly updated.

src/components/Table/EditColumnsButton.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+const valueColumn: GridColumn<Row> = { name: "value", header: "Value", canHide: true, data: ({ value }) => value };
+const actionColumn: GridColumn<Row> = {
+  name: "actions",
+  header: "Action",
+  canHide: true,
+  data: () => <div>Actions</div>,
+};
+
+describe("EditColumnsButton", () => {
+  it("Should render editColumnsButton", async () => {
```
bdow, line 19: **suggestion:** Same request about changing "Should" to "should".  (also, I think it'd make sense to uppercase the "E" in "editColumnsButton")

src/components/Table/EditColumnsButton.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    // Given an edit columns button
+    const r = await render(
+      <EditColumnsButton
+        trigger={{ label: "Columns" }}
+        columns={[nameColumn, valueColumn, actionColumn]}
+        setColumns={noop}
+      />,
+    );
+    // Expect to render the button with the expected label
+    expect(r.columns().textContent).toBe("Columns");
```
bdow, line 29: **nit/request:** Please don't use `Expect` in the comments, as it is redundant with the following line of `expect(...)`. Typically we use `Given / When /Then` comments, so this would be something like:
```
// Then the button renders with the correct label
```
(There are more `// Expect ...` comments like this in the rest of the tests. Could you update those too, please?)

src/components/Table/EditColumnsButton.test.tsx
```diff
@@ -0,0 +1,81 @@
+import { noop } from "src/utils";
+import { click, render } from "src/utils/rtl";
+import { EditColumnsButton } from "./EditColumnsButton";
+import { GridColumn } from "./GridTable";
+import { SimpleHeaderAndData } from "./simpleHelpers";
+
+type Data = { name: string | undefined; value: number | undefined };
+type Row = SimpleHeaderAndData<Data>;
+const nameColumn: GridColumn<Row> = { name: "name", header: "Name", data: ({ name }) => name };
+const valueColumn: GridColumn<Row> = { name: "value", header: "Value", canHide: true, data: ({ value }) => value };
```
bdow, line 10: **nit:** I see why you're using `ttc` now. I think we'd be okay if you just did: `name: "Value", header: "Value"`. Rather than assuming the capitalization.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+    isTextButton(trigger) ? trigger.label : isIconButton(trigger) ? trigger.icon : trigger.name,
+  );
+
+  function clearSelections() {
+    columns.map((column) => (column.visible = false));
+    setColumns((prevState) => ({
+      ...prevState,
+      visibleColumns: columns.filter((column) => (column.canHide && column.visible) || !column.canHide),
+    }));
+  }
```
bdow, line 38: **suggestion:** Let's memoize this function and I think we can simplify it a bit more. For instance, I don't think we need to do the `columns.map((c) => column.visible = false`. That may incorrectly set `visible: false` on columns that can't be hidden. It'd be easy to accidentally introduce a bug that way. Instead, let's just do the setColumns call to be something like:
```
// When clearing all selections, all hide-able columns should be filtered out of `visibleColumns`.
setColumns((prevState) => ({ ...prevState, visibleColumns: columns.filter((c) => !column.canHide) }));
```

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  return (
+    <OverlayTrigger {...props} menuTriggerProps={menuTriggerProps} state={state} buttonRef={buttonRef} {...tid}>
+      <div
+        css={{
+          ...Css.bgWhite.py5.px3.maxwPx(380).dg.gap1.$,
+          "&:hover": Css.bshHover.$,
+        }}
+      >
+        <CheckboxGroup
```
pgamboa07, line 78: @bdow , is there a way to set `xsSb` to the `CheckboxGroup` label?

bdow: Not at the moment. I would suggest maybe adding in a `hideLabel?: boolean` property. 
```
<CheckboxGroup label="Columns" hideLabel />
```
Then you can put your own "label" with the desired styles. But you'd still define the label for accessibility purposes, it just wouldn't show.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+      if (column.canHide && column.visible) return column.name;
+    })
+    .filter((column) => !!column) as string[];
+
+  const options: CheckboxGroupItemOption[] = editableColumns.map((column) => ({
+    label: column.name!,
+    value: column.name!,
+  }));
+
+  const [selectedValues, setSelectedValues] = useState<string[]>(selectedColumns ?? []);
```
bdow, line 55: **question:** Can we accomplish this logic with introducing a second state variable for the selected columns?

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const selectedColumns: string[] = editableColumns
+    .map((column) => {
+      if (column.canHide && column.visible) return column.name;
+    })
+    .filter((column) => !!column) as string[];
+
+  const options: CheckboxGroupItemOption[] = editableColumns.map((column) => ({
+    label: column.name!,
+    value: column.name!,
+  }));
```
bdow, line 53: **suggestion:** Can we memoize the options? Maybe we when do the `useMemo` above to get `editableColumns`, we can also do this map and return (1) The selected columns and (2) the options. We dont' use `editableColumns` anywhere else directly, so the current memoization logic never really helps us. We just use it to build out `selectedColumns` and `options`, which will now appear as different lists every time we re-render, causing the whole thing to rerender. Granted, its not a big deal with this small of a component, but we want to make sure we memoize in the right place. Sorry if I led you in the wrong direction. This has changed quite a bit, so some of my previous comments may no longer be applicable. Still, I do think we want to memoize `options` and `selectedColumns`. I'd suggest something like:
```
const { options, selectedColumns } = useMemo(() => {
  ...
}, [columns]);
```

src/components/Table/EditColumnsButton.tsx
pgamboa07: The selected values state handles the selected columns, what is the reason to use another state for it? I'm not clear

pgamboa07: The `columns` for `EditColumnsButton` are all columns for the table, is not the same as `useColumns` that handles `columns.visibleColumns`  for visible columns in the table, while setSelectedValues should handle all columns with canHide no matter if it is visible or not in the table

src/inputs/CheckboxGroup.tsx
```diff
@@ -31,19 +31,23 @@ export interface CheckboxGroupProps {
...
   const state = useCheckboxGroupState({ ...props, value: values });
   const { groupProps, labelProps } = useCheckboxGroup(props, state);
   const tid = useTestIds(props);
 
   return (
     <div {...groupProps} onBlur={onBlur} onFocus={onFocus} {...tid}>
-      <Label label={label} {...labelProps} {...tid.label} />
-      <div css={Css.dg.gap2.$}>
+     {!hideLabel && <Label label={label} {...labelProps} {...tid.label} />}
+      <div css={Css.dg.gtc(`repeat(${columns}, auto)`).gap2.$}>
```
bdow, line 50: **praise:** Thanks!

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+    <OverlayTrigger {...props} menuTriggerProps={menuTriggerProps} state={state} buttonRef={buttonRef} {...tid}>
+      <div
+        css={{
+          ...Css.bgWhite.py5.px3.maxwPx(380).$,
+          "&:hover": Css.bshHover.$,
+        }}
+      >
+        <div css={Css.gray500.xsSb.mb1.$}>{title || "Select columns to show"}</div>
```
bdow, line 77: **question:** Based on the mocks, it looks like this should be `text-transform: uppercase`. Should we add that here?

src/components/Table/useColumns.ts
```diff
@@ -0,0 +1,19 @@
...
+    };
+  }, [tableColumns]);
+
+  const [columns, setColumns] = useState<GridColumnState<R, S>>(initColumns);
+  return [columns, setColumns];
+}
+
+export type GridColumnState<R extends Kinded, S = {}> = {
+  visibleColumns: GridColumn<R, S>[];
+}
```
bdow, line 19: **request:** Can we run the formatting again on this?

src/components/Table/useColumns.ts
```diff
@@ -0,0 +1,19 @@
...
+import { GridColumn, Kinded } from "./GridTable";
+
+export function useColumns<R extends Kinded, S = {}>(
+  tableColumns: GridColumn<R, S>[],
+): [GridColumnState<R, S>, Dispatch<SetStateAction<GridColumnState<R, S>>>] {
+  const initColumns: GridColumnState<R, S> = useMemo(() => {
+    return {
+      visibleColumns: tableColumns.filter((column) => column.canHide ? column.visible : true),
+    };
+  }, [tableColumns]);
```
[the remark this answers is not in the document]
pgamboa07: Updated to have `columns` and `setColumns` as a useState with the visible columns for the table,  `setColumns` is called on `EditColumnsButton` to update the visible columns according with user selection in the menu, updating the columns of the table

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+    }));
+
+    return { selectedColumns, options };
+  }, [columns]);
+
+  const [selectedValues, setSelectedValues] = useState<string[]>(selectedColumns ?? []);
+
+  const clearSelections = useCallback(() => {
+    // When clearing all selections, all hide-able columns should be filtered out of `visibleColumns`.
+    setColumns(columns.filter((column) => !column.canHide));
```
bdow, line 58: **suggestion:** We should remove this `setColumns` call, as it we're already going to call `setColumns` once we call `setSelectedValues`, as that triggers the `useEffect` and makes another `setColumns` call. Doing this call twice as we have it now will cause unnecessary rerenders.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+        return false;
+      }
+      return true;
+    });
+
+    const selectedColumns = editableColumns
+      .map((column) => {
+        if (column.canHide && column.visible) return column.name;
+      })
+      .filter((column) => !!column) as string[];
```
bdow, line 44: **question:** Do we need this filter? I believe `column.name` will always be present by this point due to the check above for `editableColumns`.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    const options: CheckboxGroupItemOption[] = editableColumns.map((column) => ({
+      label: column.name!,
+      value: column.name!,
+    }));
+
+    return { selectedColumns, options };
```
bdow, line 51: **thought (non-blocking):** Personally, I like to see how few times I need to loop in order to get the values I need. In this instance you could use something like a `.reduce` method, which could achieve the `console.warn` and properly group the `selectedColumns`, and `options` values. I don't think this is a problem the way you have it, its a very small component, which will typically take a very small data set, so doing this loop a few times isn't going to hurt us.
Just because I like to explore these sort of things (they remind me of interview questions), then I'm going to take a stab at what that might look like:
```
const { options, selectedColumns } = useMemo(() => {
  return columns.reduce((acc, col) => {
    // Only include options that can be hidden and have the `name` property defined.
    if (!column.canHide) return acc;
    if (!column.name || column.name.length === 0) {
      console.warn("Column is missing 'name' property required by the Edit Columns button", column);
      return acc;
    }
    
    // If currently visible, then add to selectedColumns
    if (column.canHide && column.visible) {
      acc.selectedColumns.push(column);
    }
    
    // Add current column as an option
    return { ...acc, options: acc.options.concat({ label: column.name!, value: column.name! }) };
  }, { options: [] as CheckboxGroupItemOption[], selectedColumns as string[] }); 
```

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,89 @@
...
+  useEffect(() => {
+    setColumns(columns.filter((column) => (column.canHide ? selectedValues.includes(column.name!) : true)));
+  }, [selectedValues]);
+
+  return (
+    <OverlayTrigger {...props} menuTriggerProps={menuTriggerProps} state={state} buttonRef={buttonRef} {...tid}>
+      <div
+        css={{
+          ...Css.bgWhite.py5.px3.maxwPx(380).$,
+          "&:hover": Css.bshHover.$,
```
bdow, line 71: **suggestion:** Can we add the `bshBasic` by default? I think it needs some sort of elevation treatment, otherwise it blends in too much to the rest of the page, imo.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+  useEffect(() => {
+    setColumns(columns.filter((column) => (column.canHide ? selectedValues.includes(column.name!) : true)));
+  }, [selectedValues]);
+
+  return (
+    <OverlayTrigger {...props} menuTriggerProps={menuTriggerProps} state={state} buttonRef={buttonRef} {...tid}>
+      <div
+        css={{
+          ...Css.bgWhite.py5.px3.maxwPx(380).$,
+          "&:hover": Css.bshBasic.$,
```
bdow, line 67: Sorry if I wasn't more explicit. My request here was that we always apply `bshBasic`, not to replace the hover style.
```suggestion
          ...Css.bgWhite.py5.px3.maxwPx(380).bshBasic.$,
          "&:hover": Css.bshHover.$,
```

src/components/Table/EditColumnsButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+      },
+      { options: [] as CheckboxGroupItemOption[], selectedColumns: [] as string[] },
+    );
+  }, [columns]);
+
+  const [selectedValues, setSelectedValues] = useState<string[]>(selectedColumns);
+
+  const clearSelections = useCallback(() => {
+    setSelectedValues([]);
+  }, [columns]);
```
bdow, line 56: **nit (non-blocking):** We should try to recognize when a function would benefit or not benefit from being memoized. There is so little logic in this callback now that I don't think it needs to be memoized, and perhaps doesn't even need a defined function as it is only used in one place. We could just do this inline for the "Clear Selections" button. 
If we do decide to keep `useCallback`, there is no reason to any longer have `columns` in the dependency array, so let's at least remove that.

bdow (approved):
Thanks for working through all the feedback

PR #666, opened 2022-09-30
feat: Homebound Loading Spinner

TylerR909:
Upstreams the Homebound Loading Spinner from BluePrint for anyone to use 

hb_loading_hq_2

1. `<HbSpinnerProvider />` may be used to customize app-wide quips. `HB_QUIPS_MISSION` and `HB_QUIPS_FLAVOR` are exported for some default quips you may use. You could also choose to provide custom app-level quips like "Loading the blue into BluePrint." 
    - Suggested usage is:
    ```js
    import { HbSpinnerProvider, HB_QUIPS_MISSION, HB_QUIPS_FLAVOR } from "@homebound/beam";
    
    const quips = [
      ...HB_QUIPS_MISSION,
      ...HB_QUIPS_FLAVOR,
      "Loading the blue into BluePrint...",
      "Expanding into new markets...",
    ];
    
    function App() {
      return (
        <OtherProvider>
          <HbSpinnerProvider quips={quips}>
           <MyApp />
          </HbSpinnerProvider>
        </OtherProvider>
       );
     }
3. Individual instances of the Spinner can be customized per the Props
4. The `.gif` is Base64'd directly into the component, so there is no need to manually import the gif into `public/images`

<details><summary>On the topic of GIF...</summary>

1. GIF is a Soft G
3. `i` and `f` are frontal vowels & consonants, so it doesn't make sense that the G would be a back consonant suddenly: ![](https://www.sltinfo.com/wp-content/uploads/2016/08/Table-5.-Distribution-of-front-and-back-consonants.jpg)
4. Society has rules
5. acronym pronunciation has nothing to do with the words they're based on unless you're going to start pronouncing POTUS as "Puh-thOOs" or LASER " 🐶  Lassie-r"
6. "gin" and "giraffe" are the rules, "gift" is the exception
7. [A linguist](https://www.languagejones.com/blog-1/2017/7/1/a-linguists-take-on-the-great-gif-controversy) added fuel to the fire

</details>


TylerR909:
PR #666 fuuuuuuuuuuuuuuuuuuuuuuuuuu... This is fine. I'm sure this'll be fine.

TylerR909:
uh oh... Storybook snapshots for this might get gross if they start screenshotting different frames of the gif 😐

src/components/HbLoadingSpinner.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const quip = useMemo(() => {
+    // If quips are off globally but custom quips were provided, then override global
+    const forceQuips = extraQuips.length !== 0 || noQuips === false;
+    if ((ctx.noQuips && !forceQuips) || noQuips) {
+      return "Loading...";
+    }
+
+    const allQuips = extraQuipsOnly && extraQuips.length !== 0 ? extraQuips : [...ctx.quips, ...extraQuips];
+    return allQuips[random(allQuips.length - 1)];
```
bdow, line 27: **suggestion:** Could we remove the need of `lodash` as a dependency and use `Math.random` here instead?
```suggestion
    return allQuips[Math.floor(Math.random() * allQuips.length)];
```

src/components/HbLoadingSpinner.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }, [ctx.noQuips, ctx.quips, extraQuips, extraQuipsOnly, noQuips]);
+
+  return (
+    <div data-testid="hbLoadingSpinner" css={Css.df.fdc.jcc.aic.$}>
+      <img
+        src={SpinnerGifBase64}
+        data-chromatic="ignore" // Chromatic timing would snapshot this on different frames of the 24-frame gif, so ignore it
```
bdow, line 34: Huh, does this really work?

src/components/HbLoadingSpinner.tsx
TylerR909: Vite tree-shakes it out when you import modules using the `import default from 'lodash/specific-module'` format.

src/components/HbLoadingSpinner.tsx
TylerR909: According to their docs yeah. According to real-world experience, unknown.

src/components/HbLoadingSpinner.tsx
TylerR909: ^ For which we have an ESLint rule

src/components/BeamContext.tsx
```diff
@@ -81,17 +82,19 @@ export function BeamProvider({ children, ...presentationProps }: BeamProviderPro
...
-            {/* OverlayProvider is required for Modals generated via React-Aria */}
-            <OverlayProvider>
-              {children}
-              {/*If the drawer is open, assume it will show modal content internally. */}
-              {modalRef.current && drawerContentStackRef.current.length === 0 && <Modal {...modalRef.current} />}
-            </OverlayProvider>
-            <SuperDrawer />
-          </SnackbarProvider>
-        </AutoSaveStatusProvider>
+        <HbSpinnerProvider>
```
bdow, line 85: **thought:** I'm wondering if it really makes sense to include this in the `BeamContext`. Thinking of it from "Beam is only used by Homebound", then yes, I could see how it makes sense. But since Beam is open sourced, and maybe one day folks outside of Homebound will use Beam (🤷), then I'd lean against automatically including this spinner context to the rest of the Beam contexts.

src/components/HbLoadingSpinner.tsx
```diff
@@ -0,0 +1,90 @@
...
+            // Image is white-on-black, so this flips it to black-on-white for our almost-always white backgrounds
+            filter: "invert(1)",
+          }).$
+        }
+        alt="loading"
+      />
+      {!iconOnly && (
+        <div
+          data-chromatic="ignore" // ignore spapshotting on randomized loading quips
+        >
```
bdow, line 46: **question:** Thoughts on putting `data-testid`s on this or the image tag? Then we could have more explicit tests that check for a specific element, rather than just checking that any 1 or 2 elements are present. In my mind that'd be a more robust way of testing.

src/components/HbLoadingSpinner.tsx
bdow: Yea, it is good we can tree shake out unnecessary modules, but my push for this is that we don't need an extra dependency to accomplish the same thing, especially if it the same number of lines of code.
Out of curiosity I took a look at `lodash`'s [random](https://github.com/lodash/lodash/blob/master/random.js) script. Definitely more than we need, though still quite small. Either way, if using native javascript to accomplish the same task is as many lines of code, then I'd push to use the native solution.

src/components/HbLoadingSpinner.tsx
bdow: With both of these elements being ignored in chromatic, maybe we can just ignore the stories all together?
```
export const StoryName = setStory(...);
StoryName.parameters = {
  // disables Chromatic's snapshotting on a story level
  chromatic: { disableSnapshot: true },
};
```

bdow: Or better yet:
```
export default {
  component: MyComponent,
  // Disables snapshotting for the component
  parameters: {
    chromatic: { disableSnapshot: true },
  },
  title: 'MyComponent',
};
```

bdow (commented):
The loader looks quite slick. We've had other recent directions on how we want to approach loading, including using "skeleton" loaders. Have you chatted with anyone from Design, in particular @jeffreyqua about this approach? Might be worth getting the whole org on the same page on which loading states to use.

src/components/BeamContext.tsx
TylerR909: I can pull it out.

I'd wanted it to be usable out-of-the-box, where you _could_ use it inside your app to customize it, otherwise you get it for free. But the base `createContext()` even without a provider has everything it needs. I'm also not sure if/where I should drop a README to make it easier to set up.

src/components/HbLoadingSpinner.tsx
TylerR909: The other issue to consider is Consumers. i.e. BluePrint or Underwriting snapshots shouldn't be triggering because someone decided to use this somewhere.

I can control Beam's stories easily enough but I don't want BP/UW/HOP/etc to have to jump through hoops as well. So I still need to `data-chromatic="ignore"`

TylerR909:
> The loader looks quite slick. We've had other recent directions on how we want to approach loading, including using "skeleton" loaders. Have you chatted with anyone from Design, in particular @jeffreyqua about this approach? Might be worth getting the whole org on the same page on which loading states to use.

Nope, I snuck it in alongside another feature for Hackday, and almost immediately started getting DMs about how to use it in other places, so I'd like to upstream it. It's here when Design wants it.

src/components/HbLoadingSpinner.tsx
```diff
@@ -0,0 +1,90 @@
...
+  "Loading",
+  "HOM is HOW",
+  "One Team",
+  "On a Mission",
+  "Executing Relentlessly",
+  "Building Better",
+  "In Service of Our Customers",
+  "Scaling Massively",
+  "#LoveIt",
+].map((quip) => quip.concat("..."));
```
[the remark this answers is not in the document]
TylerR909: Updated as an export. Users can choose to spread it into their quips if they want. No longer default.

src/components/HbLoadingSpinner.tsx
bdow: Smart!

src/components/HbLoadingSpinner.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { HbLoadingSpinner, HbSpinnerProvider } from "./HbLoadingSpinner";
+
+describe("HbLoadingSpinner", () => {
+  it("picks a quip", async () => {
+    // When we render normally
+    const r = await render(
+      <HbSpinnerProvider quips={["*"]}>
+        <HbLoadingSpinner />
+      </HbSpinnerProvider>,
+      { omitBeamContext: true },
```
bdow, line 11: **question:** Why do we use `omitBeamContext` in all these tests?

bdow (approved):
Cool little exploration for hackday. I will be interested to see how this gets picked up by design and used in our application.

src/components/HbLoadingSpinner.test.tsx
TylerR909: Artifact from when Beam provided the Provider.

PR #671, opened 2022-10-06
feat: Update useColumns to use sessionStorage

pgamboa07:
Update useColumns to use sessionStorage instead of state to be data persistent for the user

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@homebound/beam",
-  "version": "2.186.3",
+  "version": "2.186.6",
   "author": "Homebound",
   "license": "MIT",
   "main": "dist/index.js",
--- a/src/components/Avatar.stories.tsx
+++ b/src/components/Avatar.stories.tsx
@@ -32,6 +32,13 @@ export function Examples() {
         ))}
       </div>
 
+      <h1>Icon fallback</h1>
+      <div css={Css.df.gap2.aifs.$}>
+        {sizes.map((size) => (
+          <Avatar src={undefined} size={size} />
+        ))}
+      </div>
+
       <h1>With name</h1>
       <div css={Css.df.gap2.aifs.$}>
         {sizes.map((size) => (
--- a/src/components/Avatar.tsx
+++ b/src/components/Avatar.tsx
@@ -1,10 +1,11 @@
 import { useState } from "react";
+import { Icon } from "src/components/Icon";
 import { Css, Typography } from "src/Css";
 import { useTestIds } from "src/utils";
 
 export interface AvatarProps {
   src: string | undefined;
-  name: string;
+  name?: string;
   size?: AvatarSize;
   showName?: boolean;
 }
@@ -16,8 +17,8 @@ export function Avatar({ src, name, size = "md", showName = false, ...others }: 
   const styles = Css.br100.wPx(px).hPx(px).overflowHidden.$;
 
   const img = showFallback ? (
-    <div css={{ ...styles, ...Css[sizeToFallbackTypeScale[size]].bgLightBlue700.white.df.aic.jcc.$ }} {...tid}>
-      {nameToInitials(name)}
+    <div css={{ ...styles, ...Css[sizeToFallbackTypeScale[size]].bgGray400.gray100.df.aic.jcc.$ }} {...tid}>
+      {name ? nameToInitials(name) : <Icon icon="userCircle" inc={sizeToIconInc[size]} />}
     </div>
   ) : (
     <img
@@ -30,7 +31,7 @@ export function Avatar({ src, name, size = "md", showName = false, ...others }: 
     />
   );
 
-  return showName ? (
+  return showName && name ? (
     <div css={Css.dif.aic.gap1.if(size === "lg" || size === "xl").fdc.$}>
       {img}
       <span css={Css[sizeToTypeScale[size]].$}>{name}</span>
@@ -56,6 +57,13 @@ const sizeToFallbackTypeScale: Record<AvatarSize, Typography> = {
   xl: "xl3",
 };
 
+const sizeToIconInc = {
+  sm: 2.5,
+  md: 4,
+  lg: 5,
+  xl: 8,
+};
+
 const sizeToTypeScale: Record<AvatarSize, Typography> = {
   sm: "baseMd",
   md: "baseMd",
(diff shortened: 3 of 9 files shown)


src/components/Table/useColumns.ts
```diff
@@ -1,10 +1,12 @@
-import { Dispatch, SetStateAction, useState } from "react";
+import { useSessionStorage } from "src/hooks/useSessionStorage";
 import { GridColumn, Kinded } from "./GridTable";
 
 export function useColumns<R extends Kinded, S = {}>(
+  storageKey: string,
   tableColumns: GridColumn<R, S>[],
-): [GridColumn<R, S>[], Dispatch<SetStateAction<GridColumn<R, S>[]>>] {
```
bdow, line 6: **suggestion:** Can we make the storage key optional? (It'd need to move to the second parameter)

stephenh: I like ^ (inferring storage key) but fwiw `c.name` isn't required so we'd just need to handle that case.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -11,21 +11,25 @@ import { GridColumn, Kinded } from "./GridTable";
...
+  const { defaultOpen, disabled, columns, setColumns, trigger, title, storageKey } = props;
   const state = useMenuTriggerState({ isOpen: defaultOpen });
   const buttonRef = useRef<HTMLButtonElement>(null);
   const { menuTriggerProps } = useMenuTrigger({ isDisabled: !!disabled }, state, buttonRef);
   const tid = useTestIds(
     props,
     isTextButton(trigger) ? trigger.label : isIconButton(trigger) ? trigger.icon : trigger.name,
   );
+  
+  const storageData = storageKey ? (JSON.parse(sessionStorage.getItem(storageKey)!) as GridColumn<R, S>[]) : null;
```
stephenh, line 31: Suggestion: I don't think we usually read `sessionStorage.getItem` during render, should probably use `useSessionStorage` for this.

src/components/Table/useColumns.ts
```diff
@@ -1,10 +1,14 @@
...
 import { GridColumn, Kinded } from "./GridTable";
 
 export function useColumns<R extends Kinded, S = {}>(
   tableColumns: GridColumn<R, S>[],
-): [GridColumn<R, S>[], Dispatch<SetStateAction<GridColumn<R, S>[]>>] {
-  const [columns, setColumns] = useState<GridColumn<R, S>[]>(
+  maybeStorageKey?: string,
+): [GridColumn<R, S>[], (value: GridColumn<R, S>[]) => void] {
+  const storageKey = maybeStorageKey ?? camelCase(tableColumns.map((c) => c.name).join(""));
+  const [columns, setColumns] = useSessionStorage<GridColumn<R, S>[]>(
```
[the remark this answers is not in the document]
bdow: +1.
@pgamboa07 - I think we'll want to get the `name` property out of each visible column and use that as the sessionStorage value. Probably should do the same check we're doing in "EditColumnsButton" to warn and ignore columns that do not have a `name` property defined. We can use the same loop to define both the storage key and value.

src/components/Table/EditColumnsButton.tsx
```diff
@@ -11,21 +11,25 @@ import { GridColumn, Kinded } from "./GridTable";
...
+  const storageColumns = storageData && (storageData.map((column) => column.name) as string[]);
```
bdow, line 32: **question:** Do we need to check the session storage here? I believe if we are using the columns returned by the `useColumns` hook, then we shouldn't need to look in the session storage again.

src/components/Table/useColumns.ts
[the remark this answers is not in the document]
bdow: @pgamboa07 - let me know if you need additional guidance here, but basically you'll want `useColumns` to return (1) the list of columns based on values saved in session storage, and (2) a `setColumns` method that can update both the list of columns and the session storage.

src/components/Table/EditColumnsButton.test.tsx
```diff
@@ -47,45 +51,39 @@ describe("EditColumnsButton", () => {
...
     const r = await render(
       <EditColumnsButton
         trigger={{ label: "Columns" }}
-        columns={[nameColumn, valueColumn, actionColumn]}
+        allColumns={[nameColumn, valueColumn, actionColumn]}
+        selectedColumns={result.current[0]}
         setColumns={setColumns}
         defaultOpen={true}
       />,
     );
```
pgamboa07, line 83: Now the columnsSelected is handle by useColumns hook

src/components/Table/useColumns.ts
```diff
@@ -1,11 +1,41 @@
...
+      if (!column.name || column.name.length === 0) {
+        console.warn("Column is missing 'name' property required by the Edit Columns button", column);
+        return acc;
+      }
+      // Add hide-able columns
+      if (column.canHide) {
+        acc.hideColumns.push(column.name);
+      }
+      // Add selected columns
+      if (!column.canHide || (column.canHide && column.visible)) {
```
stephenh, line 22: @pgamboa07 / @bdow the `column.visible` flag is a little surprising to me; does the user toggle that on/off to get the column shown/not shown in response to `setColumns` called? like:

* User clicks show/hide column
* `EditColumnsButton` calls `setColumns`
* The page re-memozies its `createColumns` b/c each column has a `GridColumn.visible = columns.includes(me)`?

Is that how it currently works?

It seems like it'd be nice to avoid breaking the `createColumns` `useMemo` every time a user shows/hides a column, but I guess realistically as our table is currently rewritten we need to rerender every row anyway...

If the `EditColumnsButton` is now taking a `selectedColumns` prop, can the `column.visible` go away?

src/components/Table/EditColumnsButton.stories.tsx
```diff
@@ -51,8 +52,9 @@ export function EditColumnButtonInAction() {
         <EditColumnsButton
           trigger={{ label: "Columns" }}
           placement="right"
-          columns={tableColumns}
+          allColumns={tableColumns}
           setColumns={setColumns}
```
stephenh, line 56: Thought: should we rename `setColumns` to `setSelectedColumns` to match the new `selectedColumns` name? Afaiu that is what happens, i.e. if this component calls `setColumns` it's really changing the incoming `selectedColumns` and not the `allColumns` prop?

src/components/Table/useColumns.ts
pgamboa07: The visible flag is used to identify which hide-able columns are visible/selected by default on initial setup

bdow (approved):

PR #674, opened 2022-10-07
fix: Bug fix for Avatar when name is empty string.

bdow:
Also allow name to not be set and fallback to the user circle icon.

After updating this in Blueprint it was clear the `name` property may not always be available. Updating upstream to handle these cases better.

### DIFF
--- a/src/components/Avatar.tsx
+++ b/src/components/Avatar.tsx
@@ -1,10 +1,11 @@
 import { useState } from "react";
+import { Icon } from "src/components/Icon";
 import { Css, Typography } from "src/Css";
 import { useTestIds } from "src/utils";
 
 export interface AvatarProps {
   src: string | undefined;
-  name: string;
+  name?: string;
   size?: AvatarSize;
   showName?: boolean;
 }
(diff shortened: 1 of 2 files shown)


src/components/Avatar.tsx
```diff
@@ -1,10 +1,11 @@
...
+import { Icon } from "src/components/Icon";
 import { Css, Typography } from "src/Css";
 import { useTestIds } from "src/utils";
 
 export interface AvatarProps {
   src: string | undefined;
-  name: string;
+  name?: string;
   size?: AvatarSize;
   showName?: boolean;
```
stephenh, line 10: Maybe showName can go away b/c callers can now just not pass the name prop? Or maybe we still use it for a tooltip, even when showName is false?

stephenh (approved):

src/components/Avatar.tsx
bdow: Right, we could still use the name to generate the initials instead of the icon fallback. And, if we supported tooltips, then yes we'd do that too. Probably should add in that support.

PR #676, opened 2022-10-09
fix: Textfield 'errorInTooltip' to show the tooltip on an icon, not the whole field

bdow:
The conditional wrapping of the tooltip when a textfield is in error causes some usability issues. Specifically, when the field is in error and the user then corrects the problem, it will rerender removing the conditional Tooltip wrapper. This rerender loses focus on the input, which can be very annoying for users.

Updating this error state to be inline with what is design for [Beam](https://www.figma.com/file/aWUE4pPeUTgrYZ4vaTYZQU/%E2%9C%A8Beam-Design-System?node-id=1944%3A14)

### DIFF
--- a/src/inputs/TextFieldBase.tsx
+++ b/src/inputs/TextFieldBase.tsx
@@ -11,7 +11,7 @@ import React, {
   useState,
 } from "react";
 import { chain, mergeProps, useFocusWithin, useHover } from "react-aria";
-import { IconButton, maybeTooltip } from "src/components";
+import { Icon, IconButton, maybeTooltip } from "src/components";
 import { HelperText } from "src/components/HelperText";
 import { InlineLabel, Label } from "src/components/Label";
 import { usePresentationContext } from "src/components/PresentationContext";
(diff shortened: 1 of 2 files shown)


src/inputs/TextFieldBase.tsx
```diff
@@ -255,6 +255,11 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
                 }}
               />
             )}
+            {errorInTooltip && errorMsg && (
```
stephenh, line 258: Thought: Do you think we should reserve space for the icon, whenever `errorInTooltip` is true, so that the UI doesn't jump as the field flips between in-error/no-error/in-error?

stephenh (approved):

src/inputs/TextFieldBase.tsx
bdow: Hmm... I think we'd need to explore that a little more. We have other places where icons/adornments show up and leave, and we haven't had any reports about weird behavior yet. But would be good to check. Thanks for flagging.

PR #677, opened 2022-10-11
chore: Reorganize table components, types and util methods

bdow:
Mostly moving components, methods and types out of `GridTable.tsx` and into separate folders/files.
Renamed:
1. `MemoizedGridRow` to `Row`
2. `GridRowStyles` to `RowStyles`


src/components/Table/types.ts
```diff
@@ -0,0 +1,135 @@
+import { ReactElement, ReactNode } from "react";
+import { GridCellContent } from "src/components/Table/components";
+import { GridTableApi } from "src/components/Table/GridTableApi";
+import { Margin, Xss } from "src/Css";
+
+export type Kinded = { kind: string };
```
stephenh, line 6: Suggestion: Fwiw I'm not a huge fan of `types.ts`-only-types files; would generally prefer to have the types live closest to the code that primarily owns it, like probably `GridTableProsp` for most of these first few...

Disclaimer would be true utility types, like `DiscriminateUnion`.

src/components/Table/types.ts
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * The data for any row in the table, marked by `kind` so that each column knows how to render it.
+ *
+ * Each `kind` should contain very little presentation logic, i.e. mostly just off-the-wire data from
+ * a GraphQL query.
+ *
+ * The presentation concerns instead mainly live in each GridColumn definition, which will format/render
+ * each kind's data for that specific row+column (i.e. cell) combination.
+ */
+export type GridDataRow<R extends Kinded> = {
```
stephenh, line 107: I.e. seems like this should live in the the row impl file?

src/components/Table/utils/index.ts
```diff
@@ -0,0 +1,14 @@
+export * from "./columns";
+export { createRowLookup } from "./GridRowLookup";
```
stephenh, line 2: Thought: Wdyt about skipping the barrel export for these sub-directories like components/utils/etc, and have importers just go directly to the file?

I think the barrel file makes since for `Table/index.ts` where we're controlling our public API, but :shrug: otherwise/internally it seems like more exports to forget adding new symbols to :-)

src/components/Table/utils/styles.tsx
```diff
@@ -0,0 +1,182 @@
+import { ReactNode } from "react";
+import { PresentationContextProps, PresentationFieldProps } from "src/components/PresentationContext";
+import { DiscriminateUnion, GridColumn, GridDataRow, GridTableApi, RenderCellFn } from "src/components/Table";
+import { Kinded, RenderAs } from "src/components/Table/types";
+import { Css, Palette, Properties } from "src/Css";
+import { safeKeys } from "src/utils";
+
+/** Completely static look & feel, i.e. nothing that is based on row kinds/content. */
+export interface GridStyle {
```
stephenh, line 9: Thought: Fwiw I think I'd put this in a `GridStyle.ts` right next to `GridTable.ts`; can't quite articulate, just that it seems more important / part of the core public API and not just a utils things.

src/components/Table/utils/utils.tsx
```diff
@@ -0,0 +1,240 @@
...
+}
+
+/**
+ * Filters rows given a client-side text `filter.
+ *
+ * Ensures parent rows remain in the list if any children match the filter.
+ *
+ * We return a copy of `[Parent, [Child]]` tuples so that we don't modify the `GridDataRow.children`.
+ */
+export function filterRows<R extends Kinded>(
```
stephenh, line 103: Question: Is this really called by multiple places now? Otherwise would wonder if it just should live in whatever file is calling it, same question for the others, my initial guess is that most? of them are only used by one file so imo would be cleaner to just keep them next to their main/only caller.

stephenh (approved):

src/components/Table/types.ts
bdow: Sounds good. I'm indifferent on it. I like the push to put more related types with the implementation that uses those types, such as `GridDataRow` moves into `Row.tsx` and maybe `GridColumn` moves into `GridTable`? I wasn't sure where to put everything yet. Though, I followed a pattern of moving components to their own files + the associated interfaces. If there is a method/type used my multiple files, then put it in a "shared" area. I'm happy to keep reworking this, but some of this stuff isn't really straight forward on where it should go. At least this is a start.

src/components/Table/utils/index.ts
bdow: Sure sounds good to me.

src/components/Table/utils/utils.tsx
bdow: Nope, just in one. Moved it.

src/components/Table/types.ts
stephenh: > but some of this stuff isn't really straight forward on where it should go. At least this is a start.

Yeah, 100% it had grown organically for quite awhile now; appreciate the clean up pass!

PR #679, opened 2022-10-12
feat: sc-19881 | Compact Chip

Grandlouis:
Exposing the compact property from the Chip component to allow having chips used individually to have the xs font type

### DIFF
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -25,9 +25,8 @@ export interface ChipProps<X> {
 
 /** Kinda like a chip, but read-only, so no `onClick` or `hover`. */
 export function Chip<X extends Only<Xss<Margin>, X>>({ type = ChipTypes.neutral, ...props }: ChipProps<X>) {
-  const { text, title = text, xss = {}, compact: propsCompact } = props;
   const { fieldProps } = usePresentationContext();
-  const compact = propsCompact ?? fieldProps?.compact;
+  const { text, title = text, xss = {}, compact = fieldProps?.compact } = props;
   const tid = useTestIds(props, "chip");
 
   return (
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -45,7 +45,7 @@ export function Chip<X extends Only<Xss<Margin>, X>>({ type = ChipTypes.neutral,
 }
 
 const typeStyles: Record<ChipType, Properties> = {
-  caution: Css.bgYellow200.$,
+  caution: Css.bgYellow400.$,
   warning: Css.bgRed100.$,
   success: Css.bgGreen100.$,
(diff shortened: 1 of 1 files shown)


src/components/Chip.tsx
```diff
@@ -20,13 +20,14 @@ export interface ChipProps<X> {
...
+  compact?: boolean;
 }
 
 /** Kinda like a chip, but read-only, so no `onClick` or `hover`. */
 export function Chip<X extends Only<Xss<Margin>, X>>({ type = ChipTypes.neutral, ...props }: ChipProps<X>) {
-  const { text, title = text, xss = {} } = props;
+  const { text, title = text, xss = {}, compact: propsCompact } = props;
   const { fieldProps } = usePresentationContext();
-  const compact = fieldProps?.compact;
+  const compact = propsCompact ?? fieldProps?.compact;
```
bdow, line 30: Does this `compact` variable give us the sizes we want? If so, then `fieldProps.compact` should be set to `true` when in a table... 🤔

src/components/Chip.tsx
```diff
@@ -20,13 +20,14 @@ export interface ChipProps<X> {
   title?: string;
   xss?: X;
   type?: ChipType;
+  compact?: boolean;
 }
 
 /** Kinda like a chip, but read-only, so no `onClick` or `hover`. */
 export function Chip<X extends Only<Xss<Margin>, X>>({ type = ChipTypes.neutral, ...props }: ChipProps<X>) {
-  const { text, title = text, xss = {} } = props;
+  const { text, title = text, xss = {}, compact: propsCompact } = props;
```
bdow, line 28: Maybe we can move this code around and avoid the `compact: propsCompact` and have it default to `fieldProps.compact`. Something like:
```
const { fieldProps } = usePresentationContext();
const { text, title = text, xss = {}, compact = fieldProps?.compact } = props;
```

src/components/Chip.tsx
Grandlouis: it does give the size showed on the design 👍 the table is picking it up as compact i don't know from where, but is rendering with the `xs` size

bdow: Okay awesome. So was this change even necessary to accomplish your ticket?

Grandlouis: yes to match the design mock, i also had to add 2px padding for the text content to compensate for the chip 2px padding that shows the background color, but that was necessary to prevent the card height to change between edit and read mode

bdow (approved):

src/components/Chip.tsx
bdow: My hope is that you won't need to make any changes to the Chip component from `internal-frontend` and we can use it as is.

PR #692, opened 2022-10-20
feat: New contextual modal

kbesingeryeh:
# [SC-22988](https://app.shortcut.com/homebound-team/story/22988/new-contextual-modal-beam-component)

* Added a new `ContextualModal` that an engineer/designer can use on the front-end to render whatever content they'd like after clicking on a Button. The modal allows us to reuse the same container, but put different things inside.

Screen Shot 2022-10-20 at 1 37 18 PM

### DIFF
--- a/src/components/internal/ContextualModal.tsx
+++ b/src/components/internal/ContextualModal.tsx
@@ -13,7 +13,7 @@ export function ContextualModal(props: ContextualModalProps) {
   const tid = useTestIds(props, "popup");
   return (
     <FocusScope restoreFocus autoFocus>
-      <div css={Css.p3.df.fdc.myPx(4).gap3.bgWhite.bshModal.br4.maxh("inherit").overflowAuto.$}>
+      <div css={Css.p3.df.fdc.gap3.bgWhite.bshModal.br4.maxh("inherit").overflowAuto.$}>
         {title && (
           <div css={Css.lg.tc.$} {...tid.title}>
             {title}
--- a/src/components/internal/ContextualModal.stories.tsx
+++ b/src/components/internal/ContextualModal.stories.tsx
@@ -13,5 +13,5 @@ export function BasicContextualModal() {
 }
 
 export function BasicContextualModalWithTitle() {
-  return <ContextualModal content={"Hello"} title={"Modal title"} />;
+  return <ContextualModal content="Hello" title="Modal title" />;
 }


src/components/ButtonMenu.tsx
```diff
@@ -62,3 +85,19 @@ export type ImageMenuItemType = MenuItemBase & {
 export type MenuItem = MenuItemBase | IconMenuItemType | ImageMenuItemType;
 // This is done just to adapt to the React-Aria API for generating Sectioned lists of Menu Items.
 export type MenuSection = MenuItem & { items?: MenuItem[] };
+
+export function ContextualModal(props: ContextualModalProps) {
```
bdow, line 89: **suggestion:** Would prefer this to be its own file. Maybe in `src/components/internal/ContextualModal.tsx`.

src/components/ButtonMenu.tsx
```diff
@@ -62,3 +85,19 @@ export type ImageMenuItemType = MenuItemBase & {
...
+  const { content, title } = props;
+  return (
+    <FocusScope restoreFocus autoFocus>
+      <div css={Css.p3.df.fdc.myPx(4).gap3.bgWhite.bshModal.br4.maxh("inherit").overflowAuto.$}>
+        {title && (
+          <div css={Css.lg.tc.$} data-testid={title}>
```
bdow, line 95: **suggestion:** Could we use the `useTestIds` hook here? Something like:
```
const tid = useTestIds(props, "popup"); 

<div {...tid.title}>
```

src/components/ButtonMenu.tsx
```diff
@@ -62,3 +85,19 @@ export type ImageMenuItemType = MenuItemBase & {
...
+            {title}
+          </div>
+        )}
+        <div data-testid={content}>{content}</div>
```
bdow, line 99: Same thing regarding test ids here.

src/components/internal/OverlayTrigger.tsx
```diff
@@ -110,3 +111,8 @@ export function isIconButton(
 ): trigger is IconButtonTriggerProps {
   return trigger && typeof trigger === "object" && "icon" in trigger;
 }
+
+// adding typeguard for items in props to conditionally render inside ButtonMenu
+export function isItemsProps(props: ButtonMenuProps): props is ButtonMenuProps {
+  return "items" in props && typeof props === "object";
+}
```
bdow, line 118: **suggestion**: Can we move this util to the ButtonMenu component?

src/components/internal/OverlayTrigger.tsx
```diff
@@ -110,3 +111,8 @@ export function isIconButton(
 ): trigger is IconButtonTriggerProps {
   return trigger && typeof trigger === "object" && "icon" in trigger;
 }
+
+// adding typeguard for items in props to conditionally render inside ButtonMenu
+export function isItemsProps(props: ButtonMenuProps): props is ButtonMenuProps {
```
bdow, line 116: **suggestion:** When we do these typeguards we should be specify what `prop` could be. Something like:
```
function isItemsProps(props: ButtonMenuProps | SomeOtherProps): props is ButtonMenuProps {
```

kbesingeryeh: got it because its not only using `WithMenuItemsProps` but also uses `ButtonMenuBaseProps`

src/components/ButtonMenu.tsx
kbesingeryeh: got it. I noticed that when I created a file called `ContextualModal` Webstorm changes the file to a `text` file instead of `tsx`... if I rename the file to `ConTextualModal` it keeps the file as `tsx` but its kind of strange naming convention...

src/components/ButtonMenu.stories.tsx
```diff
@@ -93,6 +93,7 @@ export function InteractiveMenu() {
         />
       </div>
       <ButtonMenu trigger={{ label: "Menu trigger" }} items={menuItems} persistentItems={persistentItems} />
+      <ButtonMenu trigger={{ label: "Contextual Modal trigger" }} content={"Lorem Ipsum"} title={"Modal Title"} />
```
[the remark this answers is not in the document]
stephenh: cc @bdow for naming thoughts

src/components/ButtonModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+import {
+  isIconButton,
+  isTextButton,
+  OverlayTrigger,
+  OverlayTriggerProps,
+} from "src/components/internal/OverlayTrigger";
+import { useTestIds } from "src/utils";
+import { ContextualModal } from "./internal/ContextualModal";
+
+export interface ContextualModalProps
```
stephenh, line 13: Suggestion: Probably rename this to `ButtonModalProps` so that it matches the component name?

src/components/ButtonModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  extends Pick<OverlayTriggerProps, "trigger" | "placement" | "disabled" | "tooltip"> {
+  content: ReactNode;
+  title?: string;
+  // for storybook purposes
+  defaultOpen?: boolean;
```
stephenh, line 18: Suggestion: Maybe prefix this like `storybookDefaultOpen` so that engineers using the component, who may not see the in-source comment, easily know "ah yeah I don't need to worry about this".

src/components/ButtonModal.tsx
bdow, line 13: **nit:** Can you call this `ButtonModalProps`?

stephenh (commented):
Will defer to @bdow for the +1, few suggestions, but otherwise lgtm!

src/components/internal/ContextualModal.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Css } from "src/Css";
+import { useTestIds } from "src/utils";
+import { ContextualModalProps } from "../ButtonModal";
+
+export function ContextualModal(props: ContextualModalProps) {
+  const { content, title } = props;
+  const tid = useTestIds(props, "popup");
+  return (
+    <FocusScope restoreFocus autoFocus>
+      <div css={Css.p3.df.fdc.myPx(4).gap3.bgWhite.bshModal.br4.maxh("inherit").overflowAuto.$}>
```
bdow, line 11: **question:** What is the `myPx(4)` doing? Also curious about the `maxh("inherit")`.

kbesingeryeh: I was looking at the `Menu` component and trying to match certain properties, but I think I can remove the `margin-top/margin-bottom`... 

I added the `maxh("inherit")` from a comment in the `Menu` component that mentioned using it to allow us to take advantage of the height set on the overlay container, so I thought it made sense to use it here.

src/components/ButtonModal.tsx
kbesingeryeh: got it. ty!

src/components/ButtonModal.tsx
kbesingeryeh: yes, will do.

kbesingeryeh:
@aeppy @kgrant211 @jeffreyqua here is the storybook link: https://60466f7d43c37c0021788867-brlzghvjsy.chromatic.com/?path=/story/workspace-components-button-modal--button-modal-with-title

let me know if it doesn't work!

src/components/internal/ContextualModal.tsx
bdow: Sounds good.

bdow (commented):
Before to merging, please have @jeffreyqua QA the changes in the storybook built by this PR. Once he gives the 👌 , then we can merge it in.

src/components/internal/ContextualModal.tsx
```diff
@@ -0,0 +1,26 @@
...
+export function ContextualModal(props: ContextualModalProps) {
+  const { content, title } = props;
+  const tid = useTestIds(props, "popup");
+  return (
+    <FocusScope restoreFocus autoFocus>
+      <div css={Css.p3.df.fdc.gap3.bgWhite.bshModal.br4.maxh("inherit").overflowAuto.$}>
+        {title && (
+          <div css={Css.lg.tc.$} {...tid.title}>
+            {title}
+          </div>
```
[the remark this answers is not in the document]
kbesingeryeh: @jeffreyqua can you confirm if this is needed? I will add it in if so, also please let me know if this is approved. this modal is blocking 4 other stories!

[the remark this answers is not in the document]
jeffreyqua: @kbesingeryeh We can remove that. I don't think that's needed. I will update the mocks.

jeffreyqua:
This looks good to me. Let's 🍣 !

kgrant211 (approved):
Looks good to me! Thanks @kbesingeryeh

aeppy:
Thanks @kbesingeryeh good to me!

PR #694, opened 2022-10-21
feat: Sorting to be based on column.id instead of column index

bdow:
Changes include:
 - RowState -> TableState. TableState now handles all the sorting logic. useSortState hook has been removed.
 - Use column.id to define which column is being sorted. This helps as the new logic for adding/removing columns could cause unintended changes to which column is sorted when index based.
 - If a column did not set an 'id' property, then it will default to a value based on the column's index.


src/components/Table/GridTable.tsx
```diff
@@ -185,6 +184,8 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     activeCellId,
   } = props;
 
+  const columns = useMemo(() => assignDefaultColumnIds(_columns), [_columns]);
```
stephenh, line 187: Nice

src/components/Table/GridTable.tsx
```diff
@@ -195,21 +196,25 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     api.init(persistCollapse, virtuosoRef, rows);
     api.setActiveRowId(activeRowId);
     api.setActiveCellId(activeCellId);
+    api.tableState.initSortState(props.sorting, columns);
```
stephenh, line 199: Thought: We don't have it turned on for this repo (we should) but this will fail the eslint rule that `props.sorting` and `columns` are not part of the `useMemo` deps array.

So, I either led you astray by suggesting we put it in (i.e. if we really do want to call `initSortState` each time `columns` changes) _or_ (if we don't want to call `initSortState` each time `columns` changes) we could keep it here but add an in-source comment about why `props.sorting` and `columns` ar purposefully not in our `deps` array (in which case, when we eventually add that eslint rule, we'd add a `// eslint-ignore` or what not line for it)

src/components/Table/GridTable.tsx
```diff
@@ -195,21 +196,25 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
     return api;
   }, [props.api]);
 
   const style = resolveStyles(maybeStyle);
+  const { tableState } = api;
+  const { sortConfig, setSortKey } = tableState;
+  const { on: sortOn } = sortConfig ?? ({} as GridSortConfig);
```
stephenh, line 206: Thought: what does the type of `sortOn` end up being with the `as GridSortConfig`? I think the `as` means that `sortOn` could be `undefined` but it may be getting inferred as something else, like only `client | server`.

src/components/Table/components/Row.tsx
```diff
@@ -27,17 +26,17 @@ import { Css, Palette } from "src/Css";
...
   row: GridDataRow<R>;
   style: GridStyle;
   rowStyles: RowStyles<R> | undefined;
   stickyHeader: boolean;
   stickyOffset: number;
   sortOn: SortOn;
-  sortState?: SortState<S>;
-  setSortKey?: (value: S) => void;
+  sortState?: SortState;
+  setSortKey?: (value: string) => void;
```
stephenh, line 39: Suggestion: Since the `RowImpl` is getting the `tableState` directly on line 70, we could probably remove these two props, and just access them from `tableState`?

src/components/Table/utils/TableState.ts
```diff
@@ -342,3 +401,59 @@ function flattenRows(rows: GridDataRow<any>[]): GridDataRow<any>[] {
...
+    return { ...initialSortState, current: { columnId: clickedKey, direction: DESC } };
+  }
+
+  // Else, direction is already DESC, so revert to original sort value.
+  return initialSortState;
+}
+
+type ColumnSort = {
+  columnId: string | undefined;
+  direction: Direction | undefined;
```
stephenh, line 447: Suggestion: Can these `| undefined`s go away, with the idea that maybe we don't have sort in place (so `tableState.sort = {} or undefined`) but if we do have current/persistent sort in place, then that sort should always have both an id and a direction.

src/components/Table/utils/TableState.ts
```diff
@@ -342,3 +401,59 @@ function flattenRows(rows: GridDataRow<any>[]): GridDataRow<any>[] {
...
+};
+
+export type SortState = {
+  current: ColumnSort;
+  persistent?: ColumnSort;
+};
+
+export type SortOn = "client" | "server" | undefined;
+
+function isSortState(maybeSortState: SortState | {}): maybeSortState is SortState {
```
stephenh, line 457: Suggestion: This makes me think that instead of `SortState | {}` we should just be using `SortState | undefined` in the various places we're doing that, b/c then this could go away. Like for some reason I thought it'd be cute to use `{}` as an empty marker, but not sure why, seems like `undefined` is more straightforward?

src/components/Table/utils/columns.tsx
```diff
@@ -146,3 +148,10 @@ export function calcColumnSizes(
 
   return sizes;
 }
+
+/** Assign column ids if missing */
+export function assignDefaultColumnIds<T extends Kinded>(columns: GridColumn<T>[]): GridColumnWithId<T>[] {
+  return columns.map((c, idx) => (c.id ? (c as GridColumnWithId<T>) : { ...c, id: c.id ?? generateColumnId(idx) }));
```
stephenh, line 154: Suggestion: Could do

```
  return columns.map((c, idx) => ({ id: generateColumnId(idx), ...c }));
```

To avoid the `as`. Granted it always makes a copy but :shrug: this is in a useMemo so should be really cheap.

stephenh (commented):
Still reading, but looking good

src/components/Table/utils/columns.tsx
bdow: I had that before, but the spread breaks the proxy for the selectColumn and collapseColumn. I can put an in source comment for why we aren't always spreading.

src/components/Table/utils/TableState.ts
bdow: We used the empty object instead of undefined because we needed to initialize the property as something for it to be reactive

stephenh: Oh damn, you're right, I forgot about that...

Should we just make `SortState` look like:

```
export type SortState = {
  current?: ColumnSort;
  persistent?: ColumnSort;
};
```

Because then we can assign `{}` to it and not have to worry about `| {}` at misc places.

src/components/Table/GridTable.tsx
```diff
@@ -220,13 +225,13 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
   const columnSizes = useSetupColumnSizes(style, columns, resizeTarget ?? resizeRef);
 
   // Make a single copy of our current collapsed state, so we'll have a single observer.
-  const collapsedIds = useComputed(() => rowState.collapsedIds, [rowState]);
+  const collapsedIds = useComputed(() => tableState.collapsedIds, [tableState]);
+  const sortState = useComputed(() => tableState.sortState, [tableState]);
```
stephenh, line 229: Question: I'm kinda surprised, how does `sortState.current` being changed get this `sortState` to change & trigger a re-render?

Is it because we do:

```
      this.sort = newState ?? {};
```

If so, I'm just a little surprised... B/c if someone did:

```
    this.sort.current = ...new current...;
```

Then AFAIU this line would definitely not trigger/re-render, b/c the only observer we're triggering is "read the `sortState`" value.

But I guess mobx could see the `this.sort = ...new value...` as fundamentally changing the proxy identity I guess? That works out.

FWIW there is also a mobx utility method, `toJS`, that is basically a deep copy, similar to the `{ ...copy }` trick to trigger a re-render whenever a field on the object changes, b/c we've "read each field" while doing the `...`, but `toJS` will do it deeply.

I don't think we need it here, but just calling out that something like:

```
useComputed(() => toJS(tableState.sortState))
```

Would be safer if deep keys were involved.

I kind wonder if maybe `useComputed` should examine its return value, and if its a proxy, just auto-toJS it... :thinking: 

Anyway, just thoughts.

src/components/Table/GridTable.tsx
```diff
@@ -195,21 +196,25 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
...
     api.setActiveCellId(activeCellId);
+    api.tableState.initSortState(props.sorting, columns);
     return api;
   }, [props.api]);
 
   const style = resolveStyles(maybeStyle);
+  const { tableState } = api;
+  const { sortConfig, setSortKey } = tableState;
+  const { on: sortOn } = sortConfig ?? ({} as GridSortConfig);
+  const caseSensitive = sortConfig?.on === "client" ? !!sortConfig.caseSensitive : false;
```
stephenh, line 207: Thought: In general the destructure of `tablestate` and these `sortOn` / `caseSensitive` values being done directly in the render method worry me just a little bit b/c, b/c they're not wrapped in an observer, we will get no reactivity if they change.

Granted, it seems unlikely a page would actually change any of these, but, dunno, just fwiw would be nice to see these accessed in a `useComputed`. Not required if it's not affecting anything currently.

bdow: Yea, part of me did this on purpose as I don't think we care to make them reactive properties. But who knows in the future if we'll give the user an option to edit the sorting behavior later.. 🤷 So I'm good with updating.

src/components/Table/GridTable.tsx
bdow: Yup, makes sense. I will move it back out of this `useMemo` for now.

src/components/Table/GridTable.tsx
bdow: Good point. It does come back as expecting it to be `client | server`. Which isn't correct, it should be allowed to be `undefined`. Will update. Thank for catching.

src/components/Table/GridTable.tsx
bdow: Yea, thats a good point. We did get lucky here because we are _always_ updating/replacing the whole `this.sort` object and not just the inner values. You are correct that just changing `current` would not be reactive, though probably should be.
I gave a test with the `toJS` recommendation, and that works as expected when just changing `current`, or an inner value.

src/components/Table/GridTable.tsx
```diff
@@ -241,10 +253,6 @@ export function GridTable<R extends Kinded, S = {}, X extends Only<GridTableXss,
     string[],
   ] = useMemo(() => {
     function makeRowComponent(row: GridDataRow<R>, level: number): JSX.Element {
-      // We only pass sortState to header rows, b/c non-headers rows shouldn't have to re-render on sorting
-      // changes, and so by not passing the sortProps, it means the data rows' React.memo will still cache them.
-      const sortProps = row.kind === "header" ? { sortOn, sortState, setSortKey } : { sortOn };
```
stephenh, line 246: Nice!

stephenh: Maybe can remove `sortState` and `setSorKey` from this `useMemo`'s deps?

src/components/Table/components/cell.tsx
```diff
@@ -57,21 +57,16 @@ export const defaultRenderFn: (as: RenderAs) => RenderCellFn<any> =
...
-  setSortKey: Function | undefined,
-  as: RenderAs,
-) => RenderCellFn<any> =
-  (columns, column, sortState, setSortKey, as) =>
-  (key, css, content, row, rowStyle, classNames: string | undefined) => {
-    const [currentKey, direction] = sortState || [];
+export const headerRenderFn: (columns: GridColumn<any>[], column: GridColumn<any>, as: RenderAs) => RenderCellFn<any> =
+  (columns, column, as) => (key, css, content, row, rowStyle, classNames: string | undefined) => {
+    const { tableState } = useContext(RowStateContext);
+    const { current } = tableState.sortState ?? {};
```
stephenh, line 63: Question: Do we need to read this in a `useComputed` so that we re-render when it changes? Otherwise what drives the re-render of a header to flip up arrow to down arrow or what happen? 

...fwiw I think we could probably remove `GridSortContextProps` and have whoever was reading that read from `useContext(RowStateContext)` instead.

We don't have to remove `GridSortContextProps` in this PR, but I am curious about how this is re-rendering.

src/components/Table/utils/TableState.ts
```diff
@@ -90,6 +101,61 @@ export class RowState {
     this.selectedRows.merge(map);
   }
 
+  initSortState(sortConfig: GridSortConfig | undefined, columns: GridColumnWithId<any>[]) {
+    if (!this.sortConfig) {
```
stephenh, line 105: Thought: Really minor but `if (this.sortConfig) { return }` would let the rest of this method de-dent by one.

src/components/Table/utils/TableState.ts
```diff
@@ -342,3 +408,58 @@ function flattenRows(rows: GridDataRow<any>[]): GridDataRow<any>[] {
...
+};
+
+export type SortState = {
+  current?: ColumnSort;
+  persistent?: ColumnSort;
+};
+
+export type SortOn = "client" | "server" | undefined;
+
+function isSortState(maybeSortState: SortState | {}): maybeSortState is SortState {
```
stephenh, line 463: Question: Do we still need this now that `TableStart.sort` is always `SortState`?

src/components/Table/utils/columns.tsx
stephenh: Thanks for the in-source column; what aspect of proxies was breaking for this?

I didn't think we made the columns observable, so initially am not following where a proxy would come in; not surprised that there is, just curious where/what it was.

stephenh (approved):

src/components/Table/utils/columns.tsx
bdow: It was the `newMethodMissingProxy`. Basically removing wasn't able to generate the new methods any longer.

src/components/Table/utils/TableState.ts
bdow: Oh, shoot. I thought I deleted it. I think I removed its uses. Thanks for pointing out.

src/components/Table/GridTable.tsx
bdow: 🤦 - of course.

src/components/Table/components/cell.tsx
bdow: Good question. I didn't think of this. It appears the whole header row is re-rendering. It says some hook in `RowImpl` is changing.. trying to figure out which.

bdow: I can't seem to figure it out... It is really hard to track down hooks in `RowImpl` as all the hooks being triggered from the the cells, such as the `headerRenderFn` are considered a hook of the `RowImpl`. I have updated this to use a `useComputed`, but still going to try and track this down.

PR #697, opened 2022-10-21
feat: SC-11735 - New toast component - WIP

kbesingeryeh:


src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+export type ToastContextProps = {
+  setNotice: React.Dispatch<React.SetStateAction<ToastNoticeProps>>;
+};
+
+export const ToastContext = createContext<ToastContextProps>({
+  setNotice: () => {},
+});
+
+export function ToastProvider(props: PropsWithChildren<{}>) {
+  const [notice, setNotice] = useState<ToastNoticeProps>({} as ToastNoticeProps);
```
stephenh, line 14: Suggestion/question: can we avoid the `as ToastNoticeProps`? I'm worried that we're letting a `notice` that the type system _thinks_ is a `ToastNoticeProps` (but is not actually) into the runtime that later will lead to weirdness/bugs.

Maybe this should be `useState<ToastNoticeProps | undefined>(undefined)` as a better signal for "we don't have a notice right now"?

src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { createContext, PropsWithChildren, useContext, useMemo, useState } from "react";
+import { Toast } from "./Toast";
+import { ToastNoticeProps } from "./ToastNotice";
+
+export type ToastContextProps = {
+  setNotice: React.Dispatch<React.SetStateAction<ToastNoticeProps>>;
+};
+
+export const ToastContext = createContext<ToastContextProps>({
+  setNotice: () => {},
```
stephenh, line 10: Suggestion: Could make this:

```
setNotice: () => throw new Error("Missing ToastProvider")
```

Just to more explicitly "fail fast" if someone tries to `showNotice` while missing the provider.

src/components/Toast/ToastNotice.test.tsx
```diff
@@ -0,0 +1,24 @@
+import { noop } from "src/utils";
+import { click, render } from "src/utils/rtl";
+import { ToastNotice } from "./ToastNotice";
+
+describe("ToastNotice", () => {
+  it("renders the default Toast notice layout", async () => {
+    // Given a Toast Notice component
+    const r = await render(<ToastNotice type={"error"} message={"Error"} onClose={noop} />);
+    // Then expect the default properties to exist in the layout
+    expect(r.getAllByTestId("toast_type")[0]).toBeInTheDocument();
```
stephenh, line 10: Suggestion: I think `r.toast_type_0` is a "cute" syntax that would be the same thing as this line?

Although, do we expect there to be `N` `toast_type`s? We could probably just do:

```
expect(r.toast_type()).toBeInTheDocument();
```

?

src/components/Toast/ToastNotice.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export type ToastNoticeTypes = "error" | "warning" | "success" | "info" | "alert";
+export interface ToastNoticeProps {
+  type: ToastNoticeTypes;
+  message: ReactNode;
+  /** Removes the snackbar notice from the stack */
+  onClose: () => void;
+}
+
+export function ToastNotice(props: ToastNoticeProps) {
```
stephenh, line 15: Question/suggestion: Just curious, but why is `ToastNotice` extracted/separate from `Toast`?

Like could we put all of these props/JSX directly into `Toast`? And have a `ToastType` instead of a `ToastNoticeType`?

src/components/Toast/useToast.tsx
```diff
@@ -0,0 +1,18 @@
...
+  showToast: (props: ToastNoticeProps) => void;
+}
+export function useToast(): UseToastHook {
+  const { setNotice } = useToastContext();
+
+  const showToast = useCallback((props) => {
+    setNotice({
+      ...props,
+    });
+  }, []);
```
stephenh, line 15: Suggestion: Even though we "just know" that `setNotice` won't change, I think the react eslint rule will still want deps to be `[setNotice]`.

src/components/Toast/useToast.tsx
```diff
@@ -0,0 +1,18 @@
...
+export interface UseToastHook {
+  showToast: (props: ToastNoticeProps) => void;
+}
+export function useToast(): UseToastHook {
+  const { setNotice } = useToastContext();
+
+  const showToast = useCallback((props) => {
+    setNotice({
+      ...props,
+    });
```
stephenh, line 14: Suggestion: Really minor but maybe `setNotice(props)`? Not that the copy will be expensive, but just that maintainers might wonder "...why do we need a copy here?", and if there isn't an explicit reason, they might get distracted by it w/o needing to be.

(Alternatively, if the copy is actually required, an in-source comment about why would be great.)

src/components/Toast/useToast.tsx
```diff
@@ -0,0 +1,18 @@
+import { useCallback } from "react";
+import { useToastContext } from "./ToastContext";
+import { ToastNoticeProps } from "./ToastNotice";
+
+export interface UseToastHook {
+  showToast: (props: ToastNoticeProps) => void;
+}
+export function useToast(): UseToastHook {
+  const { setNotice } = useToastContext();
```
stephenh, line 9: Question: Who eventually calls `setNotice(undefined)` or something like that to mean "the toast isn't shown anymore?".

It kinda looks like the caller is expected to pass in an `onClose`, but we don't have a way for them to call `hideToast()`, and it also seems like we should probably just do that for them anyway.

...like I think was expecting the `ToastNotice` `X` `onClick` to call `setNotice(undefined)` to tell Beam itself "don't show this toast anymore" and then call `onClose` as an (optional) notification to the page that we're not showing its toast anymore (which, fwiw I'm wondering if we can just drop, if `Toast` / `ToastNotice` fully owns making the toast go away).

stephenh (commented):
Few questions but the story looks great! Really looking forward to getting this in.

src/components/Toast/ToastNotice.test.tsx
kbesingeryeh: I like that, although these tests cases are weird in that the `render` method was rendering the component twice.. I did some debugging and could not understand why it was rendering twice when I had separate `it` blocks and I am only calling the component once.. I also could not verify it was rendering twice in my stories either.. I will check with Brandon on this too

src/components/Toast/useToast.tsx
kbesingeryeh: yes, that makes sense. I spoke to Brandon about this in office hours last week, so I dropped `ToastNotice` and combined with `Toast` and also dropped the `onClose` so now when you click the X it just calls `setNotice(undefined)`.  makes things clean and simple. thank you!

src/components/index.ts
```diff
@@ -36,4 +35,5 @@ export * from "./Table";
 export { TabContent, Tabs, TabsWithContent } from "./Tabs";
 export type { RouteTab, RouteTabWithContent, Tab, TabWithContent } from "./Tabs";
 export * from "./Tag";
+export * from "./Toast/Toast";
```
stephenh, line 38: Suggestion: Maybe do:

```
export { showToast } from "./Toast/useToast";
```

As I think that's the primary user-facing API for setting the toast.

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { setNotice } = useToastContext();
+  const tid = useTestIds(props, "toast");
+
+  return (
+    <div css={{ ...variantStyles[type], ...Css.df.aic.w100.gap2.px2.py2.$ }} {...tid} role="alert">
+      <span css={Css.fs0.plPx(12).$}>
+        <Icon icon={typeToIcon[type]} {...tid.type} color={Palette.Gray900} />
+      </span>
+      <span {...tid.message}>{message}</span>
+      <span css={Css.pr1.add("lineHeight", 0).add("marginLeft", "auto").$}>
```
stephenh, line 26: Suggestion: I think `Css.pr1.lh(0).ml("auto").$` would work here, instead of needing to use the longer `add` methods.

stephenh (approved):
Cool, few minor suggestions, but :+1: !

src/components/Toast/Toast.stories.mdx
```diff
@@ -0,0 +1,26 @@
...
+};
+
+# Toast
+
+#### Toast component for showing alerts.
+
+# Example
+
+<Canvas>
+  <Story name="Toast">{Template.bind({})}</Story>
```
bdow, line 21: **suggestion/thought:** For the `mdx` stories, we're hoping to use these more as guides on how to use a component. This gets trickier when showing off a component that we'd expect to use via a hook. In that mindset, I'd expect that in the only time we display the "Show code", it'd include something like:
```
const { showToast } = useToast();
return <Button onClick={() => showToast(...args) } label="Trigger Toast" />
```

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <IconButton icon="x" contrast onClick={() => setNotice(undefined)} {...tid.close} color={Palette.Gray900} />
```
bdow, line 27: **nit:** I don't think we want `contrast` here. It makes an odd hover state...
Screen Shot 2022-11-09 at 10 38 51 AM

src/components/Toast/Toast.tsx
bdow, line 26: **question:** Curious, what is the need to set `line-height: 0` here?

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export function Toast(props: ToastProps) {
+  const { message, type } = props;
+  // using toast context to set notice to undefined once the user clicks out of the alert
+  const { setNotice } = useToastContext();
+  const tid = useTestIds(props, "toast");
+
+  return (
+    <div css={{ ...variantStyles[type], ...Css.df.aic.w100.gap2.px2.py2.$ }} {...tid} role="alert">
+      <span css={Css.fs0.plPx(12).$}>
```
bdow, line 22: **thought/suggestion:** Based on the Figma mocks, we do not need the `plPx(12)` here. Appears to me the `px2` on the container would be enough padding.

bdow: Screen Shot 2022-11-09 at 1 50 02 PM

src/components/Toast/Toast.tsx
bdow, line 26: **suggestion:** Based on the Figma mocks, I do not believe we need `pr1` here. The container's padding should be enough...
Screen Shot 2022-11-09 at 1 50 14 PM

bdow, line 26: **suggestion:** I think we can remove the `ml("auto")` if we set `fg1` on the `span` that contains the message.

kbesingeryeh: I was using the same pattern as `Snackbar's` `IconButton`  which also has a `line-height: 0`. By removing the `line-height` it makes the X not as vertically centered aligned
Screen Shot 2022-11-09 at 5 18 10 PM

bdow: Ahh... okay. I remember this issue. Because the icon is an `inline` element, then it tries to inherit some `line-height`, causing the element to be taller than it should. This is a trick to tell it not to do that. I think we could've also made the `span` which contains the `<Icon />` to be `Css.df.aic.$` and that'd achieve the same thing. But either way, was just trying to remember what that was for. Thanks!

src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,27 @@
...
+  },
+});
+
+export function ToastProvider(props: PropsWithChildren<{}>) {
+  const [notice, setNotice] = useState<ToastProps>();
+  const contextValue = useMemo(() => ({ setNotice }), []);
+  return (
+    <ToastContext.Provider value={contextValue}>
+      {props.children}
+      {notice && <Toast message={notice.message} type={notice.type} />}
```
bdow, line 20: **thought:** I don't think this is how we want to deliver this `<Toast />` component. Instead, The `<Toast />` component should be something that the application can place anywhere. Right now we're forcing it to be underneath the rest of the page's content. I think we will probably want the user to be able to just drop in `<Toast />` and then that component is what would conditionally render based on `notice` being set.
I am thinking we might need to have the `ToastContextProps` include both `notice` and `setNotice`. Then we'd set the contextValue as: 
```
const contextValue = useMemo(() => ({ notice, setNotice }), [notice, setNotice]);
```

And in the `<Toast />` component, we'd remove the `ToastProps`, as we'll get everything from the hook now....
```
const { notice, setNotice } = useToast();
return <>{notice && (<div>...</div>)}</>;
```

bdow: This suggestion unfortunately changes up a lot of this PR, and I'm sorry to not have noticed it/caught it until now. But I do think we need to change it so we can place the `Toast` component in the right place in Blueprint.

kbesingeryeh: no worries, I think that makes sense. I'll work on this tomorrow

kbesingeryeh: can we discuss this next week? I started implementing and am a little confused on how `useToast` is supposed to work..

kbesingeryeh: I think I figured it out!! took a lot of rubber ducking hah

src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export type ToastContextProps = {
+  notice: ToastNoticeProps | undefined;
+  setNotice: React.Dispatch<React.SetStateAction<ToastNoticeProps | undefined>>;
+};
+
+export const ToastContext = createContext<ToastContextProps>({
+  setNotice: () => {},
+  notice: { type: "error", message: "" },
```
kbesingeryeh, line 16: I don't think we want to explicitly set `type: error`, but when I tried to add an `undefined/null` type to `ToastTypes`, I then got an error when trying to use the type as an index.

src/components/Toast/ToastContext.tsx
kbesingeryeh: @bdow does the new implementation work?

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,45 @@
+import { Icon, IconKey } from "src/components/Icon";
+import { Css, Palette } from "src/Css";
+import { useTestIds } from "src/utils";
+import { IconButton } from "../IconButton";
+import { useToastContext } from "./ToastContext";
+
+export type ToastTypes = "error" | "warning" | "success" | "info" | "alert";
```
bdow, line 7: **nit (non-blocking):** Would prefer to move `types` to the bottom of the file. This is per our component style guide: https://homebound.atlassian.net/wiki/spaces/ENG/pages/339574863/Component+Style+Guide#Component-Structure

src/components/Toast/useToast.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+    // Then expect it to show
+    expect(r.toast()).toBeInTheDocument();
+  });
+
+  it("can close a toast notice", async () => {
+    // Given an app that can trigger a toast
+    const r = await render(<TestToastComponent />);
+    // When triggering the toast and clicking on the close button
+    click(r.toastButton());
+    await clickAndWait(r.toast_close());
```
bdow, line 22: **question:** Does this need to be a `clickAndWait`, or could it have just been a `click`?

src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,34 @@
...
+});
+
+export function ToastProvider(props: PropsWithChildren<{}>) {
+  const [notice, setNotice] = useState<ToastNoticeProps>();
+  const contextValue = useMemo(() => ({ setNotice, notice }), [setNotice, notice]);
+  return (
+    <ToastContext.Provider value={contextValue}>
+      {props.children}
+      <Toast />
```
bdow, line 27: **request:** Can we remove `<Toast />` here? The application should decided where the `Toast` component will be placed. Beam only needs to provide the ToastContext wrapper around the application.

src/components/Toast/useToast.tsx
```diff
@@ -0,0 +1,17 @@
+import { useCallback } from "react";
+import { ToastNoticeProps, useToastContext } from "./ToastContext";
+
+export interface UseToastHook {
+  showToast: (props: ToastNoticeProps) => void;
+}
+export function useToast(): UseToastHook {
```
bdow, line 7: **nit:** Would prefer the interface to be called `UseToastProps`

src/components/Toast/ToastContext.tsx
bdow: When doing this, you'll need to update the `useToast` hook's `TestToastComponent` to be something like:
```
function TestToastComponent() {
  return (<>
   // Include the Toast component here.
    <Toast />
    <Button onClick={...} />
  </>);
}

File: src/components/Toast/useToast.test.tsx (hunk 16 above)
Comment (kbesingeryeh) [reply src/components/Toast/useToast.test.tsx:22]: no, it doesn't need to be a `click`.  I had this as a `clickAndWait` before when I had `onClose` as a callback. I will remove!

File: src/components/Toast/Toast.tsx [hunk 19]
```diff
@@ -0,0 +1,45 @@
...
+import { IconButton } from "../IconButton";
+import { useToastContext } from "./ToastContext";
+
+export function Toast() {
+  const { setNotice, notice } = useToastContext();
+  const tid = useTestIds({}, "toast");
+  return (
+    <>
+      {notice && (
+        <div css={{ ...variantStyles[notice.type], ...Css.df.aic.w100.gap2.px2.py2.$ }} {...tid} role="alert">
```
bdow, line 13: **request:** Per the figma mocks, it appears they gap should only be `8px`. You could also remove an additional "padding" definition by combining `py2.px2` to `p2`. 
```suggestion
        <div css={{ ...variantStyles[notice.type], ...Css.df.aic.w100.gap1.p2.$ }} {...tid} role="alert">
```

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,45 @@
...
+  alert: "errorCircle",
+  error: "xCircle",
+};
+
+const variantStyles: Record<ToastTypes, {}> = {
+  success: Css.bgGreen100.gray900.$,
+  info: Css.bgLightBlue100.gray900.$,
+  warning: Css.bgYellow200.gray900.$,
+  alert: Css.bgGray200.gray900.$,
+  error: Css.bgRed100.gray900.$,
```
bdow, line 42: **thought:** Maybe remove `gray900` out of here and put it above?

kbesingeryeh: ok got it. will update. ty

src/components/Toast/Toast.tsx
bdow: Looking at this again, Figma says "basic" on it, but it doesn't appear to have the same definition in Figma that we have in Beam.... Probably fine for now. I'd just use `bshBasic` and maybe we update that globally.
Screen Shot 2022-12-01 at 12 08 58 PM

bdow (approved):
Looks great. Just some small styling suggestions, but otherwise good to go!

PR #698, opened 2022-10-24
fix: SC-22792 Add even better menu icon

jgujgu:
Design kept providing me faulty SVGs, but finally I think this one is good.


bdow (approved):
Looks good. I'm curious, though - what has been "faulty" about the previous SVGs?

PR #703, opened 2022-10-26
feat: ButtonMenu to include searching

asanchezHB:


src/components/ButtonMenu.tsx
```diff
@@ -9,31 +10,42 @@ import {
...
+          <TextField placeholder="Search" value={searchTerm} onChange={setSearchTerm} label="" />
+      )}
       <Menu
         ariaMenuProps={menuProps}
         onClose={() => state.close()}
-        items={items}
+        items={filteredItems}
         persistentItems={persistentItems}
         {...tid}
       />
```
bdow, line 51: **thought:** I'm thinking it might be better off having this combination of a Menu + Search be its own component. Fewer things to rerender when the state updates. Right now we re-render the whole ButtonMenu component whenever the `searchTerm` updates, as this components is in charge of that state. I'd be curious to see if it is worth adding `searchable` to the Menu component itself, or if we should make a `SearchableMenu` component that is basically these two things. I'd say take a look and see what makes more sense.

src/components/ButtonMenu.tsx
```diff
@@ -62,3 +74,10 @@ export type ImageMenuItemType = MenuItemBase & {
 export type MenuItem = MenuItemBase | IconMenuItemType | ImageMenuItemType;
 // This is done just to adapt to the React-Aria API for generating Sectioned lists of Menu Items.
 export type MenuSection = MenuItem & { items?: MenuItem[] };
+
+// Searching function the recuisively searches through the items through a built up regex and returns the items that match the search term. Maybe not necessary, as we could use .filter() instead.
+function match(term: string, items: MenuItem[]) {
+  const builtExp = Array.from(term).reduce((prevVal, curVal, i) => `${prevVal}[^${term.substr(i)}]*?${curVal}`, '');
+  const re = RegExp(builtExp);
+  return items.filter(({ label }) => label.toLowerCase().match(re));
+}
```
bdow, line 83: **suggestion:** React-Aria has a `useFilter` hook, which has been helpful on the SelectField for searching through the items. Take a look at how that works here: https://github.com/homebound-team/beam/blob/main/src/inputs/internal/SelectFieldBase.tsx#L97

src/components/ButtonMenu.tsx
bdow: Just took a look at Storybook, might be better doing this directly in the Menu component so we can get the search box to appear per designed within the menu.
Also, I am thinking we might not want to use the `TextField` component for this as the design of the input is very different than what we get from `TextField`. Maybe another text field component that is bespoke for this purpose, "MenuSearchField" perhaps?

src/inputs/internal/MenuSearchField.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { labelProps, inputProps } = useTextField({ ...props }, inputRef);
+
+  return (
+    <div css={Css.df.aic.fg1.bb.bGray200.$}>
+      <span css={Css.ml1.pr1.$}>
+        <Icon icon="search" />
+      </span>
+      <label {...labelProps}>{props.label}</label>
+      <input
+        {...inputProps}
```
bdow, line 30: **question:** Could we have used `<TextField />` here? Then we'd get all the hover/focus/etc styling for free which is missing right now.

asanchezHB: We could, there's a very small tradeoff, unless there's a way around it. If we use `TextFieldBase` component, the border radius on the bottom left and bottom right are set at `4`. In the figma it shown as a straight line, wdyt?

bdow: Yea, we should definitely be using the same `TextFieldBase` here. I think we live with the minor styling issues for now.

bdow: I say that, but I think that is design's decision. I'd go back to Design and see if they are okay with this, or if they want us to rebuild it without the border-radius. We could also include a new prop to `TextFieldBase` that changes the display for this purpose. I'd like to come up with a good name for the prop that is contextual as to why we need it styled slightly differently.

bdow (commented):
To be honest, I'm a bit worried that not having the keyboard navigation for the menu items will be a deal-breaker. If it were up to me and this was not a pressing feature, then I'd ask for keyboard navigation. It is natural for a user to type to filter and use keyboard navigation to select. Asking them to go back between keyboard to filter and mouse to select feels like a miss imo. 
Maybe as Design about this. I'd pose the question as: We can get out of the door with a suboptimal user experience (can't use the keyboard arrows to navigation to menu items while typing to filter), or we take a little more time and do it with full keyboard support.

src/inputs/internal/MenuSearchField.tsx
asanchezHB: sounds good. I think design would be ok with that styling difference but let me see what they say, if it's a dealbreaker we could go down the prop route of and change the styling based off that. But lets see

src/components/ButtonMenu.stories.tsx
```diff
@@ -157,3 +157,21 @@ export function WithTooltip() {
 
   return <ButtonMenu trigger={{ label: "Menu trigger" }} items={menuItems} tooltip="Tool tip text" />;
 }
+
+export function WithSearchListItems() {
+  const menuItems: MenuItem[] = [
+    { label: "Development 1 - Cohort 1", onClick: action("Development 1 - Cohort 1") },
+    { label: "Development 1 - Cohort 2", onClick: action("Development 1 - Cohort 2") },
+    { label: "Development 1 - Cohort 3", onClick: action("Development 1 - Cohort 3") },
+    { label: "Development 2 - Cohort 1", onClick: action("Development 2 - Cohort 1") },
```
chr1sjf0x, line 166: quibble (non-blocking): These examples didn't make it really easy to test the search. The only thing one could use to get options to change was the numbers. 🤷

src/components/ButtonMenu.stories.tsx
```diff
@@ -157,3 +157,21 @@ export function WithTooltip() {
...
+  ];
+
+  return (
+    <ButtonMenu
```
chr1sjf0x, line 170: thought (non-blocking): I wonder if it would be worth having an example of the icon button menu, since that is what we are using.

chr1sjf0x (approved):
Looks good to my no-sense-of-design eye 😆

src/components/ButtonMenu.stories.tsx
asanchezHB: This makes sense, let me update that

src/components/ButtonMenu.stories.tsx
asanchezHB: True, let me make them more distinct

PR #707, opened 2022-10-28
feat: GridTable Expandable Column support

bdow:
Changes include:

Expandable Columns
- New custom 'expandableHeader' row kind that has custom rendering to provide the UX for expanding and collapsing associated columns.
- ExpandableColumHeader component

TableState
- Controls the columns, which are visible, expanded, etc...

EditColumnsButton
- Now uses GridTableApi to show/hide columns instead of changing the GridTable's 'columns' prop.
- Names of columns use a new 'name' prop on GridColumn rather than 'id'. Having this distinction between a unique identifier and a label can be helpful.
- Removes useColumns hook. No longer needed with using the TableState available in GridTableApi. The TableState also takes care of session storage for the selected columns.


src/components/Table/index.ts
```diff
@@ -14,7 +14,6 @@ export * from "src/components/Table/components/SortHeader";
 export { SortHeader } from "src/components/Table/components/SortHeader";
 export { useGridTableApi } from "src/components/Table/GridTableApi";
 export type { GridTableApi } from "src/components/Table/GridTableApi";
-export * from "src/components/Table/hooks/useColumns";
```
0ces, line 17: Is this a breaking change?

src/components/Table/utils/TableState.ts
```diff
@@ -201,6 +211,37 @@ export class TableState {
...
+
+  get visibleColumnIds(): string[] {
+    return [...this.visibleColumns.values()];
+  }
+
+  get expandedColumnIds(): string[] {
+    return [...this.expandedColumns.values()];
+  }
+
+  toggleExpandedColumn(columnId: string) {
```
stephenh, line 237: Question: Does toggling this require `visibleColumns` to be re-calc-d? Is that already happening somehow?

src/components/Table/utils/TableState.ts
```diff
@@ -46,21 +47,30 @@ export class TableState {
   private initialSortState: SortState | undefined;
   private onSort: ((orderBy: any | undefined, direction: Direction | undefined) => void) | undefined;
 
+  // Non-reactive list of our columns
+  public columns: GridColumnWithId<any>[] = [];
+  // An observable set of column ids to keep track of which columns are currently expanded
+  private expandedColumns = new ObservableSet<string>();
+  // An observable set of column ids to keep track of which columns are visible
+  public visibleColumns = new ObservableSet<string>();
+  private visibleColumnStorageKey: string = "";
```
stephenh, line 56: Question: I don't see any `if key === "" don't store`; is the intent to always persist visible columns into this `""` spot?

Fwiw I probably would not have minded skipping the storage key stuff in v1, if anything b/c my suspicion/bias is that we introduced these sort of "remember the page is sorted/filtered/etc by x" for the original S&S view that was very "hop to new page to see details, hit back, hop to a new page, hit back" happy; but given that most of our views have superdrawers to enable working on the page/without leaving the page, I just wonder if the "remember last-used sort/filter/collapse state" is not as important as it used to be.

stephenh (commented):
Still early in the reading, just asking questions as I go. :-)

src/components/Table/utils/TableState.ts
bdow: No, it does not. `visibleColumns` only holds the top level / parent columns. Columns defined within `expandColumns` cannot be made visible/invisible right now. This was done on purpose to limit the scope of the change.

src/components/Table/GridTable.tsx
```diff
@@ -242,16 +268,32 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = {}
     return rows;
   }, [columns, rows, sortOn, sortState, caseSensitive]);
 
-  const hasTotalsRow = rows.some((row) => row.id === "totals");
+  const hasTotalsRow = rows.some((row) => row.id === TOTALS);
+  const hasExpandableHeader = rows.some((row) => row.id === EXPANDABLE_HEADER);
```
stephenh, line 272: Question: These are calc-d outside of the `useMemo` but aren't part of the deps array? Should we just calc them inside the useMemo?

src/components/Table/utils/TableState.ts
bdow: This session storage for visible columns is an existing behavior that the `useColumns` hook was doing for us. But now we don't have the `useColumns` hook and the `EditColumnsButton` utilizes the API.

src/components/Table/GridTable.tsx
bdow: Ahh, yup. Forgot to add `hasExpandableHeader` to that dep array. Will move this as you suggest, no need to have them defined out here.

src/components/Table/utils/TableState.ts
bdow: The storage key also defaults to a value based on the Table's column ids. It is initialized in `setColumns`

src/components/Table/utils/TableState.ts
bdow: I was trying to figure out if it made sense to have `visibleColumns` keep track of both the columns that a user toggles on and off via the "EditColumnsButton" (or similar workflow), and which columns are currently "expanded". It seemed like it could get messy with determining which is expanded + which of the potentially expanded columns are currently hidden. I was treating them as being orthogonal to one another for now. 
Maybe in a follow up we can extend this work so that the "child"/expandable columns can also be hidden via the `EditColumnsButton`. Could be cool to have that button re-render based on which columns are expanded, so you could potentially toggle on/off some of the "child" columns.

src/components/Table/GridTable.stories.tsx
```diff
@@ -1136,3 +1136,99 @@ export function ToggleCustomCollapse() {
...
+    ],
+    [],
+  );
+
+  const columns: GridColumn<ExpandableRow>[] = useMemo(
+    () => [
+      selectColumn<ExpandableRow>({ sticky: "left" }),
+      column<ExpandableRow>({
+        expandableHeader: () => "Address",
+        header: emptyCell,
```
TylerR909, line 1182: comment:

I'm looking at the Story right now. I do find it a bit odd that only 1 of 4 columns is expandable, yet the `expandableHeader: ...` is what's providing the header for 75% of the coincidentally non-expandable table.

But I'm also curious how much of that is just implementation details. Can _I_ make a `{ kind: "header" }` that is also expandable, and a second row that is a "subheader" or something?

src/components/Table/GridTable.stories.tsx
```diff
@@ -1136,3 +1136,99 @@ export function ToggleCustomCollapse() {
     </>
   );
 }
+
+type ExpandHeader = { id: "expandableHeader"; kind: "expandableHeader" };
+type Header = { id: "header"; kind: "header" };
+type ExpandableData = {
```
TylerR909, line 1142: Why doesn't this one need `id`?

Also 98% of places I've seen ID used, with the sole exception of `user:1` below, it's matched -exactly- the `kind`. Why can't we just infer it from `kind` unless they want to specifically override it? `row.id ??= row.kind`.

bdow: The `id` is a required prop for all `GridRowData` types. We currently use a combination of the `kind` + `id` props to create the unique React `key` for each row.
Potentially for these "reserved" sort of rows, we could infer it, but that'd mean making `row.id` optional in some cases, but not all. This could get tricky... 🤔 
I get your point, seems redundant and unnecessary. But it is a required prop for how everything is organized right now. I don't want to get into a big optional row id refactor at the moment, but could be an interesting exploration in the future. Thanks for calling out!

src/components/Table/GridTable.stories.tsx
bdow: That is an interesting take. Hopefully in practice there will be more than just one column that is expandable. Otherwise, seems like an unnecessary UX.
The idea of having a single row define multiple rows is an interesting take too. That might mean reworking some of our types, as we don't want any `GridDataRow` type to include multiple rows. I do think we can improve in this aspect, specifically now that we have 3 rows that should always be fixed in the table header. It'd be cleaner, as you suggest, to have a single:
```
const headerRow = { kind: "tableHead", id: "redundant... : )", data: {...} };

const someColumn = { 
  tableHead: {
    totals: (data) => ({ content: data.totalPrice }),
    expandableHeader: () => ({ content: "Expandable Title" }),
    header: () => ({ content: "Total price" }),
  },
  data: {...}
}
```
I think that is a fairly larger refactor in order to do this, but we could start it without impact the existing tables, which could be a cool way to move forward. Thanks for the suggestion and thoughts.

src/components/Table/GridTable.stories.tsx
```diff
@@ -1136,3 +1136,99 @@ export function ToggleCustomCollapse() {
...
+  data: {
+    firstName: string | undefined;
+    lastName: string | undefined;
+    birthdate: string | undefined;
+    age: number | undefined;
+    occupation: string | undefined;
+    manager: string | undefined;
+  };
+};
+type ExpandableRow = ExpandHeader | Header | ExpandableData;
```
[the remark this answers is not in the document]
bdow: Fair. There is already a `type Row` in this file, which is why I didn't do it. Though, I could move these types into the story function, maybe? Or I could rename this to something else.

src/components/Table/GridTable.stories.tsx
```diff
@@ -1136,3 +1136,99 @@ export function ToggleCustomCollapse() {
     </>
   );
 }
+
+type ExpandHeader = { id: "expandableHeader"; kind: "expandableHeader" };
+type Header = { id: "header"; kind: "header" };
```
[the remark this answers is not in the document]
bdow: I think we do export the `Header` as `SimpleHeader`. We could also do a `SimpleExpandHeader` as well. "Why `Simple...`?", you ask? Because sometimes we may want to include some data in the row types, like:
```
type Header = { id: "header", kind: "header", data: SomeFragment }
```
Then we can use the data in the row to help render the header.

TylerR909 (approved):

src/components/Table/utils/TableState.ts
stephenh: Ah gotcha! I thought `visibleColumns` was going to be something like "if parent column A is expanded, then put its children into visibleColumns". Like more of a calculation of what columns are physically visible right now.

But what you have makes sense.

Agreed to descope on visibleColumns can only be otp-level.

PR #710, opened 2022-10-29
feat: add useBreakpoint

crshannon:
## [SC-24345](https://app.shortcut.com/homebound-team/story/24345/create-usebreakpoint-hook)
- adding `useBreakpoint` which returns  `breakpoints` which is a record of our css breakpoints  from truss to bool's - enables us to no conditionally check for breakpoints in our react components when we cant handle things with the css options.
- updates on browser resize.

example of breakpoints `{print: false, sm: true, md: false, smOrMd: true, ...}`


src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,33 @@
...
+  useEffect(() => {
+    function handleResize() {
+      // Set breakpoints on resize
+      let bps = {} as BreakpointsType;
+      Object.entries(Breakpoints).forEach(([name, bp]) => {
+        bps[name as keyof BreakpointsType] = window.matchMedia(bp.replace("@media screen and ", "")).matches;
+      });
+      setBreakpoints(bps);
+    }
+    window.addEventListener("resize", handleResize);
```
crshannon, line 26: @bdow - checked our the  `useResizeObserver` but it needs a provided target which didnt feel as clean as just watching for a resize window event.

src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @example
+ * const { breakpoints } = useBreakpoint();
+ * if(breakpoints.mdAndDown) {...do something cool}
+ */
+
+type BreakpointsType = Record<BreakpointKey, boolean>;
+
+export function useBreakpoint(): { breakpoints: BreakpointsType } {
+  const [breakpoints, setBreakpoints] = useState({} as BreakpointsType);
```
stephenh, line 15: Thought: It probably doesn't matter, but the `{} as BreakpointsType` means that, until the `useEffect` runs, the very first caller will see `bp.mdAndDown` as `undefined` instead of `false`.

Makes me think to either:

a) Update the type to be `type BreakpointsType = Partial<Record<...>>` so that we're just explicit to the caller that `{}` is a value we'll be using (which seems fine, I don't think it would break the caller's `if (bp.print)` type checks, or

b) Initialize this `useState` with `{ mdAndDown: false, ... }` explicitly.

I'm generally ambivalent between the two, but maybe pick one? And, agreed, it's not a bug per se, but just something I could see be surprising for users if, for whatever reason, they venture outside the stock `if (bp.print)` usage and think like "oh I'm going to Object.values(bp)" or something.

src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  useEffect(() => {
+    function handleResize() {
+      // Set breakpoints on resize
+      let bps = {} as BreakpointsType;
+      Object.entries(Breakpoints).forEach(([name, bp]) => {
+        bps[name as keyof BreakpointsType] = window.matchMedia(bp.replace("@media screen and ", "")).matches;
```
stephenh, line 22: Thought: Kinda looks like leaving in `screen and` will still work:

```
window.matchMedia("screen and (min-width:1px)")
MediaQueryList {media: 'screen and (min-width: 1px)', matches: true, onchange: null}
```

So maybe `@media` should be the only thing we strip?

src/hooks/useBreakpoint.tsx
bdow: I'm fine with that. Curious though, could the "window" or "body" be the target?

src/hooks/useBreakpoint.tsx
```diff
@@ -9,19 +9,23 @@ import { BreakpointKey, Breakpoints } from "../Css";
...
-type BreakpointsType = Record<BreakpointKey, boolean>;
+type BreakpointsType = Partial<Record<BreakpointKey, boolean>>;
+
+function matchMediaBreakpoints(): BreakpointsType {
+  let bps = {} as BreakpointsType;
+  Object.entries(Breakpoints).forEach(([name, bp]) => {
+    bps[name as keyof BreakpointsType] = window.matchMedia(bp.replace("@media ", "")).matches;
+  });
+  return bps;
+}
```
bdow, line 20: **request:** Can we move these types and the `matchMediaBreakpoints` function to the bottom of the file. That way the hook is the first thing users see when they open the hook file. This would also allow the JSDoc to be associated to the hook function

src/hooks/useBreakpoint.tsx
crshannon: good call - updated

src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  useEffect(() => {
+    function handleResize() {
+      // Set breakpoints on resize
+      setBreakpoints(matchMediaBreakpoints());
+    }
+    window.addEventListener("resize", handleResize);
+    handleResize();
+    // Remove event listener on cleanup
+    return () => window.removeEventListener("resize", handleResize);
```
bdow, line 33: **request:** Can we debounce this to reduce the number of times this could run while the user is actively resizing their window? Otherwise we'll trigger _a lot_ of re-renders while resizing.

src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }, []);
+
+  return { breakpoints };
```
bdow, line 36: **thought:** Every time we return breakpoints, then it is a new object, even if the key/value pairs didn't change. Can we ensure that the hook _only_ triggers a re-render of the calling components when there is a change to the matched breakpoints?

crshannon: Done!

stephenh: Afaiu the code as-is wouldn't trigger re-renders, it would just potentially break useMemos, i.e.:

```
const { breakpoints } = useBreakpoints();
const somethingExpensive = useMemo(() => ..., [breakpoints]);
```

...

That said, I wonder if we could just remove the hash and do:

```
const bp = useBreakpoints();
```

Granted, destructuring is super-common for hooks, but seems like we don't really need it here, :shrug: . Defer to @crshannon .

src/hooks/useBreakpoint.tsx
crshannon: added a debounce to reduce the number of re-renders of this

src/hooks/useBreakpoint.tsx
bdow: Why wouldn't it cause the calling component to re-render? When the hook's value changes it is causing the component to re-render as far as I can tell.

crshannon: I am somewhat impartial here but like the de-structuring do we don't need `bp.`  - @bdow do you have any strong opinions on this as the blueprint guru?

src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * if(breakpoints.mdAndDown) {...do something cool}
+ */
+
+type BreakpointsType = Partial<Record<BreakpointKey, boolean>>;
+
+export function useBreakpoint(): { breakpoints: BreakpointsType } {
+  const [breakpoints, setBreakpoints] = useState(matchMediaBreakpoints());
+
+  useEffect(() => {
+    const handleResize = useDebouncedCallback(
```
bdow, line 19: I don't think you can use a hook inside a hook like this. You might need to put this outside of the useEffect

src/hooks/useBreakpoint.tsx
stephenh: @crshannon AFAIU there is not an "extra `bp.`" b/c it just changes from:

```
const { breakpoints } = useBreakpoints();
console.log(breakpoints.smAndMd);
```

To:

```
const breakpoints = useBreakpoints();
console.log(breakpoints.smAndMd);
```

So I happened to use `bp` as a cute/shorter variable name for the non-destructured example, but they really are the same.

If anything the 2nd "not destructured" might actually be "more destructure" friendly b/c then we really could remove the `bp.` or `breakpoints.` by doing:

```
const { mdAndSm } = useBreakpoints();
if (mdAndSm) { ... }
```

bdow: I like the idea of not having to immediately destructure, and doing `const breakpoints = useBreakpoints()` or being able to destructure for the specific breakpoint you care about... `const { mdAndDown } = useBreakpoints();`

stephenh: > Why wouldn't it cause the calling component to re-render? When the hook's value
> changes it is causing the component to re-render as far as I can tell.

Re-renders only happen when `setState`s are called, so merely asking a hook for a value, and getting back a "new" / non-stable return value, does not cause inherently cause _another_ re-render to happen.

If it did, this code as-is would infinite loop/constantly re-render, right?

If I'm following.

stephenh: Ah gotcha, gotcha, I took the "return value is new" too literally :-D, and yep, we're on the same page that deep-equaling old-vs-new `BreakpointsType` before calling `set` is a good idea.

src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,44 @@
...
+      // function
+      () => {
+        setBreakpoints(matchMediaBreakpoints());
```
bdow, line 22: **suggestion:** Per my comment:
> Can we ensure that the hook only triggers a re-render of the calling components when there is a change to the matched breakpoints?
This is where we'd want to ensure something changed before we call `setBreakpoints`. Otherwise we're saying a state changed, even if it didn't really change.

src/hooks/useBreakpoint.tsx
crshannon: note: issue with `const { mdAndDown } = useBreakpoints();` is this now conflicts with the exported css breakpoints if they are being used in the same file.

`import {  mdAndDown } from "@homebound/beam";`

stephenh: @crshannon ah yeah, that's true, destructuring out literally `mdAndDown` may not be a good idea.

crshannon: just updated so we can now use `const breakpoints = useBreakpoints();` (or `const bps = useBreakpoints();` if you fancy) and implemented a `isEqual` check.

src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,41 @@
+import { isEqual } from "lodash";
```
stephenh, line 1: Suggestion: @blimmer has done some performance/bundle optimizations for us, and it turns out these "lodash" imports pull in all of lodash and so end up being something we avoid.

Just doing:

```
import isEqual from "lodash/isEqual";
```

I believe is the kosher/tree-friendly way of importing this. 

(I also wouldn't mind using `import equal from "fast-deep-equal"`, which we use in `useComputed` so is already in Beam/being included in our bundle)

stephenh (approved):
Lgtm!

src/hooks/useBreakpoint.tsx
crshannon: added check before setting so we reduce the unneeded re-renders etc.

src/hooks/useBreakpoint.tsx
crshannon: ah yes, I swear I had (lodash/isEqual) at one point but must have been messing around and lost it but anyway -  switched to using `fast-deep-equal` as to prevent an additional need in the bundle..

stephenh: Great!

bdow (commented):
Hey @crshannon - do you think you could do a quick follow up to write tests for this hook?

src/hooks/useBreakpoint.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const useBreakpoint = (): BreakpointsType => {
+  const [breakpoints, setBreakpoints] = useState(matchMediaBreakpoints());
+
+  const handleResize = useDebouncedCallback(() => {
+    const newBps = matchMediaBreakpoints();
+    if (equal(breakpoints, newBps)) return;
+
+    setBreakpoints(newBps);
+  }, 1000);
```
bdow, line 24: **nit:** 1 second seems like a fairly large delay. Feels like it can be a bit sluggish to respond. How did you land on using 1 second?

PR #712, opened 2022-10-31
feat: (SC-22763) Update button modal

pgamboa07:
Updates needed for this ticket [SC-22763](https://app.shortcut.com/homebound-team/story/22763/inline-editing-predecessors-successors)

Screen Shot 2022-10-31 at 13 39 18

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -2,7 +2,7 @@ import { AriaButtonProps } from "@react-types/button";
 import { ButtonHTMLAttributes, ReactNode, RefObject, useMemo, useRef, useState } from "react";
 import { useButton, useFocusRing, useHover } from "react-aria";
 import { Icon, IconProps, maybeTooltip, navLink, resolveTooltip } from "src/components";
-import { Css, Palette, Xss } from "src/Css";
+import { Css, Palette } from "src/Css";
 import { BeamButtonProps, BeamFocusableProps } from "src/interfaces";
 import { isAbsoluteUrl, isPromise, noop } from "src/utils";
 import { getButtonOrLink } from "src/utils/getInteractiveElement";
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -2,7 +2,7 @@ import { AriaButtonProps } from "@react-types/button";
 import { ButtonHTMLAttributes, ReactNode, RefObject, useMemo, useRef, useState } from "react";
 import { useButton, useFocusRing, useHover } from "react-aria";
 import { Icon, IconProps, maybeTooltip, navLink, resolveTooltip } from "src/components";
-import { Css, Palette } from "src/Css";
+import { Css, Palette, Xss } from "src/Css";
 import { BeamButtonProps, BeamFocusableProps } from "src/interfaces";
 import { isAbsoluteUrl, isPromise, noop } from "src/utils";
 import { getButtonOrLink } from "src/utils/getInteractiveElement";
@@ -23,6 +23,7 @@ export interface ButtonProps extends BeamButtonProps, BeamFocusableProps {
   /** Denotes if this button is used to download a resource. Uses the anchor tag with the `download` attribute */
   download?: boolean;
   contrast?: boolean;
+  xss?: Xss<"borderColor">;
 }
 
 export function Button(props: ButtonProps) {
@@ -35,6 +36,7 @@ export function Button(props: ButtonProps) {
     openInNew,
     download,
     contrast = false,
+    xss,
     ...otherProps
   } = props;
   const asLink = typeof onPress === "string";
@@ -93,6 +95,7 @@ export function Button(props: ButtonProps) {
     css: {
       ...Css.buttonBase.tt("inherit").$,
       ...baseStyles,
+      ...xss,
       ...(isHovered && !isPressed ? hoverStyles : {}),
       ...(isPressed ? pressedStyles : {}),
       ...(isDisabled || asyncInProgress ? { ...disabledStyles, ...Css.cursorNotAllowed.$ } : {}),
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -23,7 +23,7 @@ export interface ButtonProps extends BeamButtonProps, BeamFocusableProps {
   /** Denotes if this button is used to download a resource. Uses the anchor tag with the `download` attribute */
   download?: boolean;
   contrast?: boolean;
-  xss?: Xss<"borderColor">;
+  forceFocusStyles?: boolean;
 }
 
 export function Button(props: ButtonProps) {
@@ -36,7 +36,7 @@ export function Button(props: ButtonProps) {
     openInNew,
     download,
     contrast = false,
-    xss,
+    forceFocusStyles = false,
     ...otherProps
   } = props;
   const asLink = typeof onPress === "string";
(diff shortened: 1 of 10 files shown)


src/components/internal/Popover.tsx
```diff
@@ -25,7 +25,7 @@ export function Popover(props: PopoverProps) {
 
         // If we clicked the trigger element (or within it), then that will call a `state.toggle` for us.
         // Return early if that happens, otherwise we'd call `onClose`, then the trigger would toggle it back open.
-        if (triggerRef.current?.contains(e)) {
+        if (triggerRef.current?.contains(e) || e.closest('[data-overlay-container]')) {
```
bdow, line 28: **question:** Why do we want to return `true` when a user interacts with the `data-overlay-container`?

src/components/internal/OverlayTrigger.tsx
```diff
@@ -76,6 +78,7 @@ export function OverlayTrigger(props: OverlayTriggerProps) {
           disabled={disabled}
           tooltip={tooltip}
           onClick={noop}
+          xss={Css.if(showActiveBorder && state.isOpen).bLightBlue600.$}
```
bdow, line 81: Can we avoid `xss` here and just always supply the `bLightBlue600` if `showActiveBorder` is true? I'm also curious if this border color is the same as the focus ring we'd typically do on Buttons? If it is, maybe we can just use the same focus ring that we already have and force the focus ring to show. Might be worth chatting with design so we don't have multiple ways of displaying a focus sort of style.

src/components/Button.tsx
```diff
@@ -23,6 +23,7 @@ export interface ButtonProps extends BeamButtonProps, BeamFocusableProps {
   /** Denotes if this button is used to download a resource. Uses the anchor tag with the `download` attribute */
   download?: boolean;
   contrast?: boolean;
+  xss?: Xss<"borderColor">;
```
bdow, line 26: **fyi:** I'm hoping we can remove the XSS property here, but just for an fyi... The XSS property is intended to limit the style properties that can be put on an element. As it is right now, we are allowing the user to put any styles on the element via the `xss` property, which we don't really want to do. To make this actually type safe, then we'd need to update the Button function component.
It would look something more like:
```
import { Only, Xss } from "src/Css";

// Create a type to make the same XSS easier to reference
type ButtonXss = Xss<"borderColor">;

// Use a generic in the props for the XSS type.
interface ButtonProps<X> { xss?: X; }

// Have the generic extend our `Only` TS helper which limits the values that can be passed into the XSS based on what we defined.
export function Button<X extends Only<ButtonXss, X>>(props: ButtonProps<X>) {
```

src/components/internal/OverlayTrigger.tsx
bdow: I'm thinking we'd do something more like:
```
<Button showActiveBorder={state.isOpen} />
or
<Button forceFocusStyles={state.isOpen} />
```

src/components/internal/Popover.tsx
bdow: I ask because I think when we return `true` in `shouldCloseOnInteractOutside`, then we are saying "Yes, close this overlay". But in this instance, when we interact with other elements within the Popover, we don't really want it to close. I also think that we should copy the logic used in the Modal component's `shouldCloseOnInteractOutside` method.

src/components/internal/OverlayTrigger.tsx
```diff
@@ -76,6 +78,7 @@ export function OverlayTrigger(props: OverlayTriggerProps) {
           disabled={disabled}
           tooltip={tooltip}
           onClick={noop}
+          forceFocusStyles={showActiveBorder && state.isOpen}
```
pgamboa07, line 81: I kept the `showActiveBorder` flag because we don't want to apply an active border for all cases

src/components/internal/Popover.tsx
```diff
@@ -29,6 +29,10 @@ export function Popover(props: PopoverProps) {
           return true;
         }
 
+        // Do not close the Modal if the user is interacting with the Tribute mentions dropdown (via RichTextField) or with another 3rd party dialog (such as a lightbox) on top of it.
+        return !(e.closest(".tribute-container") || e.closest("[role='dialog']") || e.closest("[role='alert']"));
+
+
         onClose();
         return false;
```
bdow, line 37: The addition here makes it so `onClose` isn't called. Does this still work as expected?

src/components/internal/OverlayTrigger.tsx
bdow: I see you updated to just use the focus styles. Was that something approved by Design?

src/components/internal/Popover.tsx
bdow: (We're now returning before `onClose` can be called or the `return false` can be called.

src/components/internal/Popover.tsx
```diff
@@ -29,6 +29,10 @@ export function Popover(props: PopoverProps) {
           return true;
         }
 
+        // Do not close the Modal if the user is interacting with the Tribute mentions dropdown (via RichTextField) or with another 3rd party dialog (such as a lightbox) on top of it.
```
bdow, line 32: **nit:** This comment is specific for the modal. Can it be updated/removed?

bdow (commented):
The logic inside of the `shouldCloseOnInteractOutside` might be a bit outdated and could be improved. Would be great for us to validate the existing logic still makes sense and work in the updated logic with it.

src/components/internal/OverlayTrigger.tsx
pgamboa07: Yes, it was approved by Lauren

src/components/internal/OverlayTrigger.tsx
bdow: **question** When using the ButtonModal should `showActiveBorder` something we should support for all possible triggers? That'd include IconButton and AvatarButton.

src/components/internal/OverlayTrigger.tsx
```diff
@@ -35,6 +35,7 @@ export interface OverlayTriggerProps {
   /** Prop set the style of the button element */
   variant?: ButtonVariant;
   hideEndAdornment?: boolean;
+  showActiveBorder?: boolean;
```
bdow, line 38: **thought:** Since we define the `showActiveBorder` prop here, I would expect that it impacts any variation of the OverlayTrigger, which means it could be a text button, Icon button, or avatar button.

src/components/internal/OverlayTrigger.tsx
```diff
@@ -58,6 +60,8 @@ export function OverlayTrigger(props: OverlayTriggerProps) {
     isOpen: state.isOpen,
     onClose: state.close,
     placement: (placement ? `bottom ${placement}` : "bottom left") as Placement,
+    shouldUpdatePosition: showActiveBorder || undefined,
```
bdow, line 63: **suggestion:** This shouldn't be necessary. By default `shouldUpdatePosition` is `true`. See https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/overlays/src/useOverlayPosition.ts#L84

src/components/internal/OverlayTrigger.tsx
```diff
@@ -58,6 +60,8 @@ export function OverlayTrigger(props: OverlayTriggerProps) {
...
+    offset: showActiveBorder ? 4 : undefined,
```
bdow, line 64: **praise:** Nice!

bdow (approved):
Looks good! Thanks for working through all the feedback. My only other thought about this is that I'm curious how it'll look using the focus styles/box-shadow for the active stying when in a table. I think it'll be fine, but wouldn't be surprised if the box-shadow accidentally gets cut off... 🤷 If so we'll just need to revisit.

PR #713, opened 2022-10-31
feat: add Icon to Tags (sc-22870)

Andrielson:
image

Changes in Tags include:
- muting the background colors
- adjusting the padding to make them smaller
- supporting the ability to add an icon if desired

### DIFF
--- a/src/components/Tag.test.tsx
+++ b/src/components/Tag.test.tsx
@@ -13,4 +13,10 @@ describe("Tag", () => {
     const r = await render(<Tag text="test" data-testid="testTag" xss={Css.mt1.$} />);
     expect(r.testTag()).toHaveStyleRule("margin-top", "8px");
   });
+
+  it("renders with icon", async () => {
+    const r = await render(<Tag text="text" data-testid="testTag" icon="infoCircle" />);
+    const iconElement = r.container.querySelector(`[data-icon="infoCircle"]`)!;
+    expect(iconElement).toBeInTheDocument();
+  });
 });
--- a/src/components/Tag.test.tsx
+++ b/src/components/Tag.test.tsx
@@ -15,7 +15,7 @@ describe("Tag", () => {
   });
 
   it("renders with icon", async () => {
-    const r = await render(<Tag text="text" data-testid="testTag" icon="infoCircle" />);
+    const r = await render(<Tag text="text" icon="infoCircle" />);
     const iconElement = r.container.querySelector(`[data-icon="infoCircle"]`)!;
     expect(iconElement).toBeInTheDocument();
   });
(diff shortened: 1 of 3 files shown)


src/components/Tag.test.tsx
```diff
@@ -13,4 +13,10 @@ describe("Tag", () => {
     const r = await render(<Tag text="test" data-testid="testTag" xss={Css.mt1.$} />);
     expect(r.testTag()).toHaveStyleRule("margin-top", "8px");
   });
+
+  it("renders with icon", async () => {
+    const r = await render(<Tag text="text" data-testid="testTag" icon="infoCircle" />);
```
bdow, line 18: **nit:** `data-testid="testTag"` is unnecessary as a property as we are not validating against it.

src/components/Tag.tsx
```diff
@@ -16,9 +18,14 @@ export function Tag<X extends Only<Xss<TagXss>, X>>({ text, type, xss, ...otherP
   const tid = useTestIds(otherProps);
 
   return (
-    <span {...tid} css={{ ...Css.dib.tinySb.ttu.px1.pyPx(4).gray900.br4.$, ...typeStyles, ...xss }} title={text}>
+    <span
+      {...tid}
+      css={{ ...Css.dif.tinySb.ttu.aic.gap(0.5).pxPx(6).pyPx(2).gray900.br4.$, ...typeStyles, ...xss }}
```
bdow, line 23: **nit:** Would prefer to use `gapPx(4)` here.

src/components/Tag.tsx
```diff
@@ -16,9 +18,14 @@ export function Tag<X extends Only<Xss<TagXss>, X>>({ text, type, xss, ...otherP
...
+      title={text}
+    >
       {/* Nesting `lineClamp` styles as the padding bottom set would expose the remainder of the text if applied on the same element */}
       {/* Using `lineClamp1` instead of `truncate` as `truncate` requires a width set to properly truncate and `lineClamp` can smartly do it based on the parent's width */}
+      {otherProps.icon && <Icon icon={otherProps.icon} inc={1.5} />}
```
bdow, line 28: **suggestion:** When the text gets really long, then the icon is currently shrinking due to the flex container child having a default `flex-shrink: 1`, which is basically saying "Shrink your size as needed". 
Screen Shot 2022-11-01 at 9 58 54 AM

But we don't really want it to shrink here. What we can do is wrap this in another element and provide a `flex-shrink: 0`, telling it not to shrink. Something like:
```suggestion
      {otherProps.icon && <span css={Css.fs0.$}><Icon icon={otherProps.icon} inc={1.5} /></span>}
```

Andrielson: Thanks for the explanation. It annoyed me, but I wasn't sure how to fix it.

bdow (approved):

PR #714, opened 2022-11-01
feat: SC-8122 new bx columns icon

JonnCharpentier:
new bx columns icon


TylerR909:
for posterity: 
image

JonnCharpentier:
> for posterity: image

What do you mean by that? Is that color change something internal frontend do?

TylerR909:
I just wanted to see what the new icon was without having to run Storybook. Now it's immortalized in this PR. Grabbed the image from your ticket and pasted it in here.

src/components/Icon.stories.tsx
```diff
@@ -48,6 +48,7 @@ export const Icon = (props: IconProps) => {
     "duplicate",
     "filter",
     "columns",
+    "bxcolumns"
```
bdow, line 51: **question:** Do you know if design had a specific reason to use a second "columns" icon? Having two icons to represent the same thing within the same application seems confusing to me. I wouldn't be surprised if the designer was just unaware of the other icon. Or if they are aware, I'd want to double check that we really need a second `columns` icon.

JonnCharpentier: @aeppy

aeppy: @bdow Sure thing, will do!

bdow: I'm also okay with us naming it something else. Looking at the Figma file, it appears we named it `kanban`, which isn't as confusing. One icon for "columns" and on for "kanban" helps me understand their purpose a bit better. This PR had them both named "columns" (granted, one was "bxcolumns"), which is why this was red flag to me. But if they do have different purposes, and are named for those different purposes, then I'm good with that too.

aeppy: Great, good to know. We may not need 2 anyway. I just asked the design team and lauren said she removed the columns icon where she was using it in schedules, and I don't think we use it elsewhere. But I'm waiting to hear back from the rest of the team just to be sure.

bdow: Awesome! I just did a quick check in Blueprint, and it does not appear that we use this icon anywhere at the moment.

JonnCharpentier: @bdow should I replace columns icon with this one?

bdow: That will be up to design. I'd wait for their response.

aeppy: Just waiting to hear back from the rest of the design team if they're cool with replacing it. Only heard from Lauren so far. But yes, that's the plan if everyone agrees.

aeppy: Ok no one else chimed in so I will assume they're ok with it since I believe that's the only place in blueprint it's used. We can go ahead and replace the icon, @JonnCharpentier.

bdow: Awesome! Thanks @aeppy for asking the team about. Can we update the Beam Figma to remove the old Column icon and change the name of the new one from `kanban` to `columns`?

bdow: Or if we want to keep the name as `kanban`, I'm okay with that too. Just want to make sure the naming is consistent between Figma and the implementation.

aeppy: Yep! Will do! I'll keep it as kanban. @JonnCharpentier @bdow

src/components/Icon.tsx
```diff
@@ -49,6 +49,11 @@ export const Icon = React.memo((props: IconProps) => {
  */
 export const Icons = {
   // Actions
+  kanban: (
+    <>
+      <path d="M19.893 3.001H4C2.897 3.001 2 3.898 2 5.001V19.001C2 20.104 2.897 21.001 4 21.001H19.893C20.996 21.001 21.893 20.104 21.893 19.001V8V7V6V5.001V5C21.892 3.897 20.994 3.001 19.893 3.001ZM8 19.001H4V8H8V19.001ZM14 19.001H10V8H14V19.001ZM16 19.001V8H19.893L19.894 19.001H16Z" />
+    </>
```
bdow, line 55: **nit:** I don't think we need the surrounding `<>...</>` element, but not a big deal.

bdow (approved):

PR #717, opened 2022-11-04
fix: SC-21205 Remove autoFocus from contextual modal

jgujgu:


src/components/internal/ContextualModal.tsx
```diff
@@ -12,7 +12,7 @@ export function ContextualModal(props: ContextualModalProps) {
   const { content, title } = props;
   const tid = useTestIds(props, "popup");
   return (
-    <FocusScope restoreFocus autoFocus>
+    <FocusScope restoreFocus>
```
bdow, line 15: **thought:** My concern with removing the auto-focus behavior here is that for keyboard users they'd have to tab through the whole underlying screen before they can tab into this contextual modal now. Wondering if we can do this another way. I think the reason why we didn't run into this auto-focus issue on the Modal is because the first thing that gets focus is the close button.

bdow: I'm going to create a storybook example in here and see how it behaves.

PR #718, opened 2022-11-04
fix: SC-21205 Trigger select popover with click, not focus

jgujgu:
This PR prevents undesirable behavior in the `ContextualModal`. This solution has been discussed with Brandon. We apparently never noticed this before because the normal modal autoFocuses on the close button. However, since a filter is the first focusable element in the PR in question, it revealed this behavior.

https://www.loom.com/share/d7ed9cc253b54ae3a1d277a1a4fedbd8

Related PR: https://github.com/homebound-team/internal-frontend/pull/2970

### DIFF
--- a/src/inputs/internal/SelectFieldInput.tsx
+++ b/src/inputs/internal/SelectFieldInput.tsx
@@ -158,6 +158,7 @@ export function SelectFieldInput<O, V extends Value>(props: SelectFieldInputProp
             maybeCall(onFocus);
           },
           onClick: () => {
+            if (inputProps.readOnly) return;
             buttonRef.current?.click();
           },
           size:


src/inputs/internal/SelectFieldInput.tsx
```diff
@@ -156,6 +156,11 @@ export function SelectFieldInput<O, V extends Value>(props: SelectFieldInputProp
             if (inputProps.readOnly) return;
             setIsFocused(true);
             maybeCall(onFocus);
+          },
+          onClick: () => {
+            if (inputProps.readOnly) return;
+            setIsFocused(true);
```
bdow, line 162: **question** Do we need to `setIsFocused` to true here? I'd assume once a user clicks on the input, then `focus` will also be triggered and that'll call `setIsFocused(true)`.

src/inputs/internal/SelectFieldInput.tsx
```diff
@@ -156,6 +156,11 @@ export function SelectFieldInput<O, V extends Value>(props: SelectFieldInputProp
...
+            maybeCall(onFocus);
```
bdow, line 163: Same question here, is this necessary?

bdow (commented):
Found a blocking issue here: https://www.loom.com/share/42170608fd054dd18b2e75a7cb859917
Thanks for diving into this.

src/inputs/internal/SelectFieldInput.tsx
```diff
@@ -156,7 +156,10 @@ export function SelectFieldInput<O, V extends Value>(props: SelectFieldInputProp
             if (inputProps.readOnly) return;
             setIsFocused(true);
             maybeCall(onFocus);
-            state.open();
+          },
+          onClick: () => {
+            if (inputProps.readOnly) return;
+            buttonRef.current?.click();
```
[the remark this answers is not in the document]
bdow: Hey, I think this is fine. That was an odd behavior happening and I have no clue how to fix it. This works and is very minimal. A little odd to have to do, but that bug was also a little odd... so 🤷 . 
fwiw, I do like the `if (inputProps.readOnly) return;` - no reason to trigger another click if it won't do anything.

bdow (approved):
I'm good with this until we can figure out a more direct way of fixing that bug, but so be it for now. 
FYI, I introduced a breaking change (https://github.com/homebound-team/beam/pull/707) that will block the next Beam bumps in `internal-frontend` until it's been addressed. I have the update locally, but requires on more Beam PR to go out first: https://github.com/homebound-team/beam/pull/720.

PR #719, opened 2022-11-05
fix: Improve Accordion height animation

bdow:
No longer use max-height for animating as it results in inconsistent animation timings depending on size of the content. The proposed way in this PR is to animate the height. We now hide the Accordion's content until it has been expanded. Once expanded and the content is rendered, then a useEffect runs to calculate the contents height. That height is then applied to the wrapping element to animate its height to that of the content. It now requires two renders to complete the animations, but it is much smoother and the second render should be very minimal


stephenh (approved):
:+1:

PR #720, opened 2022-11-08
fix: Support custom storage key for visible columns

bdow:
After attempting upgrade Beam in internal-frontend, I noticed the Project Schedule's EditColumnsButton + useColumns implementation used a custom session storage key to differentiate the visible column states between project stages. This PR adds that ability back and updates tests.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -161,6 +161,11 @@ export interface GridTableProps<R extends Kinded, X> {
    * Expected format is `${row.kind}_${row.id}_${column.id}`.
    */
   activeCellId?: string;
+  /**
+   * Defines the session storage key for which columns are visible. If not provided, a default storage key will be used based on column order and/or `GridColumn.id`
+   * This is beneficial when looking at the same table, but of a different subject (i.e. Project A's PreCon Schedule vs Project A's Construction schedule)
+   */
+  visibleColumnStorageKey?: string;
 }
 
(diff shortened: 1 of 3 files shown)


0ces (approved):

src/components/Table/GridTable.tsx
```diff
@@ -161,6 +161,11 @@ export interface GridTableProps<R extends Kinded, X> {
    * Expected format is `${row.kind}_${row.id}_${column.id}`.
    */
   activeCellId?: string;
+  /**
+   * Defines the session storage key for which columns are visible. If not provided, a default storage key will be used based on column order and/or `GridColumn.id`
+   * This is beneficial when looking at the same table, but of a different subject (i.e. Project A's PreCon Schedule vs Project A's Construction schedule)
+   */
+  visibleColumnsStorageKey?: string;
```
stephenh, line 168: Just random musing, we have enough `fooStorageKey`s floating around that I had just assumed `GridTable` already had one, and so was going to suggestion having a generic `GridTable.storageKey` that would then be used as the basis for all of the internal storage keys.

But turns out `GridTable` only has this one, and I guess the others are just separate components/like filters/etc.

So n/m, but just typing it out anyway, that maybe the next storage key we add to GridTable, instead of `fooStorageKey` / `barStorageKey`, we make a single `storageKey=...` and just assume the user is fine with both storage use cases being driven off the same key.

PR #721, opened 2022-11-09
fix: Expandable Column fixes

bdow:
Two issues addressed:
1. Columns should be expandable to a larger width, they do not necessarily need to show more columns. Updated so 'expandColumns' prop can accept a number value, which is used as the pixels to grow the column to.
2. Virtualized GridTable forgot to account for the expandable header when calculating the total number of rows.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -914,7 +914,7 @@ describe("GridTable", () => {
 
     it("returns expandedColumns value if is a number and column is currently expanded", () => {
       expect(
-        calcColumnSizes([{ id: "c1", w: "10%", expandColumns: "300px" }, { w: 2 }] as any, undefined, undefined, [
+        calcColumnSizes([{ id: "c1", w: "10%", expandedWidth: "300px" }, { w: 2 }] as any, undefined, undefined, [
           "c1",
         ]).join(" "),
       ).toEqual("300px ((100% - 0% - 300px) * (2 / 2))");
@@ -2452,7 +2452,7 @@ describe("GridTable", () => {
               expandableHeader: () => "Client name",
               header: () => "Name",
               data: ({ firstName }) => firstName,
-              expandColumns: "300px",
+              expandedWidth: "300px",
               initExpanded: true,
               w: "200px",
             }),
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -2443,8 +2443,8 @@ describe("GridTable", () => {
       expect(cell(r, 1, 1)).toHaveTextContent("Last name");
     });
 
-    it("can expand columns when `expandColumns` defines a number (pixel width)", async () => {
-      // Given a table with `expandColumns` defined as a number and is initially expanded
+    it("can expand columns when `expandedWidth` defines a value", async () => {
+      // Given a table with `expandedWidth` defined
       const r = await render(
         <GridTable
           columns={[
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -878,25 +878,25 @@ describe("GridTable", () => {
 
   describe("column sizes", () => {
     it("as=virtual defaults to fr widths", () => {
-      expect(calcColumnSizes([{}, {}], undefined, undefined).join(" ")).toEqual(
+      expect(calcColumnSizes([{ id: "c1" }, { id: "c2" }], undefined, undefined, []).join(" ")).toEqual(
         "((100% - 0% - 0px) * (1 / 2)) ((100% - 0% - 0px) * (1 / 2))",
       );
     });
 
     it("as=virtual treats numbers as fr", () => {
-      expect(calcColumnSizes([{ w: 1 }, { w: 2 }] as any, undefined, undefined).join(" ")).toEqual(
+      expect(calcColumnSizes([{ w: 1 }, { w: 2 }] as any, undefined, undefined, []).join(" ")).toEqual(
         "((100% - 0% - 0px) * (1 / 3)) ((100% - 0% - 0px) * (2 / 3))",
       );
     });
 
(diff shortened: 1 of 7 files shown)


src/components/Table/types.ts
```diff
@@ -75,15 +75,20 @@ export type GridColumn<R extends Kinded> = {
...
   /** Flag that will allow to know which hide-able columns are visible on initial load */
   initVisible?: boolean;
-  /** A list of columns that should only be shown when this column is "expanded" */
-  expandColumns?: GridColumn<R>[];
+  /**
+   * Either
+   * 1. A list of columns that should only be shown when this column is "expanded", or...
+   * 2. `number | string` representing width to allow for the column to expand itself larger. This prop is treated the same as the `GridColumn.w` prop.
+   *     Example: Collapsed state shows number of books. Expanded state shows titles of books.
+   */
```
stephenh, line 83: Thought: Huh...seems like maybe this should just be a different key? Like `expandedSize`? 

Even then, the name `expandedSize` would make me think its related to "expanded-as-in-adding-new columns", but AFAICT there are basically totally separate features? They just both happen to use the term "expanded"?

Should we use a totally separate term for this new feature...

bdow: Yea, I was wondering about that too. My hesitation was the simplicity in using the same prop to signal the same/similar behavior. I also thought about renaming `expandColumns` to something else that better noted it could be either more columns of a size value.

stephenh: Is the intent to use this same "expanded-becomes-wider" and "expanded-shows-more-columns" in the same table? Just curious...

Do they want the "expanded-becomes-wider" to be useable in tables that don't even use the "expanded-show-more-columns" feature? Like in a regular table with just a regular `header`, similar to sorting, I can like toggle it expanded-becomes-wider?

src/components/Table/utils/columns.tsx
```diff
@@ -163,15 +168,20 @@ export function assignDefaultColumnIds<T extends Kinded>(columns: GridColumn<T>[
...
+        : Array.isArray(expandColumns)
+        ? expandColumns.map((ec, ecIdx) => ({
+            ...ec,
+            id: ec.id ?? (`${generateColumnId(idx)}_${ecIdx}` as string),
+            // Defining this as undefined to make TS happy for now.
+            // If we do not explicitly set to `undefined`, TS thinks `expandColumns` could still be of type GridColumn<T> (not WithId).
+            // We only support a single level of expanding columns, so this is safe to do.
+            expandColumns: undefined,
+          }))
+        : expandColumns;
```
TylerR909, line 184: May be able to just simplify to `!Array.isArray(expandColumns) ? expandColumns : expandColumns.map(...)`

src/components/Table/types.ts
bdow: Yes, that is the intent. We could have a column that shows a count of "Add-on". When expanded, it'll list out the Add-on names.
The UX is the same between the two, they look no different. The result is whether expanded increases the width of a single column, or adds more columns. Though, if we did use a `expandedSize` prop, then we could do both... allow for "sub-columns" to show when expanded, and define a different width for the column that became expanded, which would be really nice. 
I'll go ahead and update to use a `expandedWidth` for this purpose.

bdow: Updated to use a different property. Thanks for the feedback.

src/components/Table/GridTable.test.tsx
```diff
@@ -2434,6 +2443,32 @@ describe("GridTable", () => {
       expect(cell(r, 1, 1)).toHaveTextContent("Last name");
     });
 
+    it("can expand columns when `expandColumns` defines a number (pixel width)", async () => {
```
stephenh, line 2446: Nit: Update the test name (my fault since I nudged the rename :-) )

src/components/Table/GridTable.tsx
```diff
@@ -232,11 +232,15 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = {}
       tableState.columns
         .filter((c) => tableState.visibleColumnIds.includes(c.id))
         .flatMap((c) =>
-          c.expandColumns && tableState.expandedColumnIds.includes(c.id) ? [c, ...c.expandColumns] : [c],
+          c.expandColumns && Array.isArray(c.expandColumns) && tableState.expandedColumnIds.includes(c.id)
```
stephenh, line 235: Question: Can the `Array.isArray` be removed now that `expandedWidth` is a different property?

src/components/Table/components/Row.tsx
```diff
@@ -118,6 +118,7 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
       {columns.map((column, columnIndex) => {
         // Need to keep track of the expanded columns so we can add borders as expected for the header rows
         const isExpanded = tableState.expandedColumnIds.includes(column.id);
+        const numExpandedColumns = isExpanded && Array.isArray(column.expandColumns) ? column.expandColumns.length : 0;
```
stephenh, line 121: Suggestion: Maybe another that could be undone

src/components/Table/GridTable.tsx
```diff
@@ -232,11 +232,15 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = {}
...
+            ? [c, ...c.expandColumns]
+            : [c],
         ) as GridColumnWithId<R>[],
     [tableState],
   );
 
+  const expandedColumnIds: string[] = useComputed(() => tableState.expandedColumnIds, [tableState]);
```
stephenh, line 242: Suggestion: Super minor maybe put this line right before line 267, since afaict that's the 1st time where this value is needed?

stephenh (approved):
Cool!

src/components/Table/GridTable.tsx
bdow: Yup, definitely. Thanks for catching this one.

PR #727, opened 2022-11-12
feat: [SC-24998]  Create number range filter type

pwillikins:
Created a new number range filter type that allows users to set `min` and/or `max` number range values.  The new filter is set up to render two separate layouts for vertical or horizontal.

- Created the new `NumberRangeFilter` component
- Created unit tests for new component
- Updated the Filter story to include the new number range filter
- Updated `NumberField` to allow new props - `sizeToContent` & `inlineLabel`

Vertical View:
Screen Shot 2022-11-11 at 6 16 30 PM

Horizontal View:
Screen Shot 2022-11-11 at 6 47 16 PM

### DIFF
--- a/src/inputs/NumberField.tsx
+++ b/src/inputs/NumberField.tsx
@@ -105,14 +105,14 @@ export function NumberField(props: NumberFieldProps) {
       type === "percent"
         ? { style: "percent" }
         : type === "basisPoints"
-          ? { style: "percent", minimumFractionDigits: 2 }
-          : type === "cents"
-            ? { style: "currency", currency: "USD", minimumFractionDigits: 2 }
-            : type === "dollars"
-              ? { style: "currency", currency: "USD", minimumFractionDigits: numFractionDigits ?? 2 }
-              : type === "days"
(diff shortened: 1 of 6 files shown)


src/inputs/NumberField.tsx
```diff
@@ -48,6 +48,8 @@ export interface NumberFieldProps {
   hideErrorMessage?: boolean;
   // Typically used for compact fields in a table. Removes border and uses an box-shadow for focus behavior
   borderless?: boolean;
+  inlineLabel?: boolean;
```
stephenh, line 51: Question: As a lazy ask, I assume this `inlineLabel` gets passed through `otherProps` to something else? (Just don't see it used here directly)

stephenh (approved):
:+1:

src/inputs/NumberField.tsx
pwillikins: @stephenh - Yes, the `inlineLabel` would get passed in through `otherProps` here.

bdow (approved):
Nice work! Really great set of tests.

PR #731, opened 2022-11-15
fix: position and z-index of a tooltip on a ButtonModal Component

ipiedrahb:
## FIgma
Screen Shot 2022-11-15 at 10 23 01

## Issue
Screen Shot 2022-11-15 at 10 21 55

## Update
Screen Shot 2022-11-15 at 11 01 23

### DIFF
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -83,7 +83,7 @@ function Popper({ triggerRef, content, placement = "auto" }: PopperProps) {
       ref={popperRef}
       style={styles.popper}
       {...attributes.popper}
-      css={Css.maxw("320px").bgGray900.white.px1.py("4px").br4.xs.z999.$}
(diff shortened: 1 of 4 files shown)


src/components/Tooltip.tsx
```diff
@@ -83,7 +83,7 @@ function Popper({ triggerRef, content, placement = "auto" }: PopperProps) {
       ref={popperRef}
       style={styles.popper}
       {...attributes.popper}
-      css={Css.maxw("320px").bgGray900.white.px1.py("4px").br4.xs.z999.$}
+      css={Css.maxw("320px").bgGray900.white.px1.py("4px").br4.xs.add("zIndex", 999999).$}
```
bdow, line 86: **question:** What is the z-index of the button modal?

bdow: Discovered it is `100000`. This is needed mostly because we use `Popper` right now instead of React-Aria. One day we'll move to not need popper. For now this is fine.

bdow (approved):

PR #733, opened 2022-11-16
chore: Reorganize Table header rows into a single container

bdow:
Whether rendering as virtual, a table, or by default, all header rows (Totals, Expandable Column headers, and regular column header) are contained in a single element. This allows to have more percise styling and eliminates the need to define a 'stickyOffset' per header row when 'stickyHeader' is defined.


stephenh (approved):
Nice! I'm just scanning but seems like a great clean up.

...do we still need `stickyOffset`? I know we've discussed in the past that I'm trigger happy to delete that / thinking that ScrollableParent/Child is a way to avoid that now, but am not 100% sure / close to the details anymore.

bdow:
> stickyOffset

It's not being used, but I'm not as trigger happy as you are about it. Part of me still wants to see our custom scrollable containers go away at some point.

stephenh:
> Part of me still wants to see our custom scrollable containers go away at some point.

Ah yeah, I'd love for PM/UX to stop asking for sticky behavior :-) but seems unlikely; pending that, dunno, imo the scrollable containers are the lesser evil of them or hardcoded offsets, once we hit "sticky the page header ... and the tabs ... and the filters ... and the searchbar" slippery slope anyway.

PR #734, opened 2022-11-17
feat: SuperDrawerHeader

bdow:
Introduces the SuperDrawerHeader component. This component portals itself into the header of the SuperDrawer. Follows the existing pattern of our Modals.

This is a fairly large breaking change as we have completely changed how to define the header of the SuperDrawer. It removes all title/header configuration properties from the openInDrawer method.

### DIFF
--- a/src/components/SuperDrawer/SuperDrawer.stories.tsx
+++ b/src/components/SuperDrawer/SuperDrawer.stories.tsx
@@ -185,8 +185,6 @@ export function OpenWithTitleRightContent() {
           rightContent={<Button label="Manage RFP" onClick={() => {}} />}
         />
       ),
-      // titleLeftContent: <Tag text={"ASSIGNED"} type={"success"} />,
-      // titleRightContent: <Button label="Manage RFP" onClick={() => {}} />,
     });
   }
   useEffect(open, [openInDrawer]);
@@ -210,8 +208,6 @@ export function LongTitle() {
           rightContent={<Button label="Manage RFP" onClick={() => {}} />}
         />
       ),
-      // titleLeftContent: <Tag text={"ASSIGNED"} type={"success"} />,
-      // titleRightContent: <Button label="Manage RFP" onClick={() => {}} />,
     });
   }
   useEffect(open, [openInDrawer]);
@@ -378,15 +374,7 @@ function TestDrawerContent(props: TestDrawerContentProps) {
 
   return (
     <>
-      <SuperDrawerHeader hideControls={hideControls}>
-        <div css={Css.df.jcsb.aic.gap2.$}>
-          <div css={Css.fg1.df.aic.gap2.$}>
-            <h1>{title}</h1>
-            {leftContent}
-          </div>
-          <div css={Css.fs0.$}>{rightContent}</div>
-        </div>
-      </SuperDrawerHeader>
+      <SuperDrawerHeader hideControls={hideControls} title={title} left={leftContent} right={rightContent} />
       <SuperDrawerContent
         actions={
           hasActions
@@ -426,9 +414,7 @@ function TestDetailContent({ book, onPurchase, title }: { book: Book; onPurchase
   const { closeDrawerDetail } = useSuperDrawer();
   return (
     <>
-      <SuperDrawerHeader>
-        <h1>{title}</h1>
-      </SuperDrawerHeader>
+      <SuperDrawerHeader title={title} />
       <SuperDrawerContent
         actions={[
           { label: `Back to "${book.bookTitle}"`, onClick: closeDrawerDetail, variant: "tertiary" },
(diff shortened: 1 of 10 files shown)


src/components/SuperDrawer/SuperDrawer.stories.tsx
```diff
@@ -368,55 +377,70 @@ function TestDrawerContent({ book, hasActions = true }: TestDrawerContentProps)
...
-      <h2 css={Css.xlSb.mb1.$}>{book.bookTitle}</h2>
-      <p css={Css.base.$}>
-        <strong>Description:</strong> {book.bookDescription}
-      </p>
-      <hr css={Css.my1.$} />
-      <div css={Css.df.gap1.aic.$}>
-        <p>
-          <strong>Author:</strong> {book.authorName}
+    <>
+      <SuperDrawerHeader hideControls={hideControls}>
```
stephenh, line 381: Nice! I think this is a much better approach. The previous API was definitely the right one to start with, but now that we're being asked to "do more" / keep it up to date/etc.

src/components/SuperDrawer/SuperDrawer.stories.tsx
```diff
@@ -368,55 +377,70 @@ function TestDrawerContent({ book, hasActions = true }: TestDrawerContentProps)
...
+        <div css={Css.df.jcsb.aic.gap2.$}>
+          <div css={Css.fg1.df.aic.gap2.$}>
```
stephenh, line 383: :thinking: what do you think about some sort of `TableActions` type component that, even within a `SuperDrawerHeader` gives the user some higher-level left/right placeholders to fill in, so that we don't have the h1/gap2/etc. padding copy/pasted across each usage site?

Granted, having the escape hatch for "let me do whatever" is probably necessary too.

...like maybe `SuperDrawerHeader title={...} left={...} right={...}`, which if set use the default/out-of-box padding/spacing, but if title/left/right are not used, then we just do whatever children is?

src/components/SuperDrawer/useSuperDrawer.tsx
```diff
@@ -26,8 +18,6 @@ export interface OpenInDrawerOpts {
 }
 
 export interface OpenDetailOpts {
-  /** Title of the SuperDrawer for this detail page (replaces the original title). */
-  title?: string;
```
stephenh, line 30: Question: So components opened within `openDetail` will be responsible for bringing their own `SuperDrawerHeader` now? I think that makes sense, just checking my understanding.

stephenh (approved):
Nice!

src/components/SuperDrawer/useSuperDrawer.tsx
bdow: Yes, that is correct.

src/components/SuperDrawer/SuperDrawer.stories.tsx
bdow: That is an interesting idea. Would help with some default styling/positioning. Maybe two potential interfaces, one with children, and then one with title/left/right, where title is required... 🤔

bdow: Alrighty - just gave that a shot. Let me know what you think.

src/components/SuperDrawer/components/SuperDrawerHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const { sdHeaderDiv, drawerContentStack: contentStack } = useBeamContext();
+  const firstContent = contentStack.current[0]?.opts as OpenInDrawerOpts;
+  const { onPrevClick, onNextClick } = firstContent ?? {};
+  const currentContent = contentStack.current[contentStack.current.length - 1]?.opts;
+  const isDetail = currentContent !== firstContent;
+  const tid = useTestIds({}, "superDrawerHeader");
+
+  return createPortal(
+    <div css={Css.df.aic.jcsb.gap3.$} {...tid}>
+      {isStructuredProps(props) ? (
```
stephenh, line 32: Praise: Neat! Looks good!

stephenh (approved):

src/components/SuperDrawer/components/SuperDrawerHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+interface SuperDrawerHeaderStructuredProps {
+  title: string;
+  left?: ReactNode;
+  right?: ReactNode;
+  hideControls?: boolean;
+}
+
+export function SuperDrawerHeader(props: SuperDrawerHeaderProps | SuperDrawerHeaderStructuredProps): JSX.Element {
```
stephenh, line 21: Thought: Pretty neat how React props & TS type unions/type-checking play nicely together like this.

PR #736, opened 2022-11-18
feat: HackDay: Add xs loader; Button allow change icons on async

apattersonATX-HB:
[SC-25242](https://app.shortcut.com/homebound-team/story/25242/fe-beam)

Spending Hackday updating some libraries and doing cleanup for Underwriting app. 

The async button changes felt like a natural progression since we're automatically disabling it in flight already.

### DIFF
--- a/src/components/Button.stories.tsx
+++ b/src/components/Button.stories.tsx
@@ -1,7 +1,7 @@
 import { action } from "@storybook/addon-actions";
 import { Meta } from "@storybook/react";
 import { capitalCase } from "change-case";
-import { Button, ButtonSize, ButtonVariant } from "src";
+import { Button, ButtonSize, ButtonVariant, Icon, Loader } from "src";
 import { Css } from "src/Css";
 import { withRouter } from "src/utils/sb";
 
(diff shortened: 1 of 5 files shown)


src/components/Button.stories.tsx
```diff
@@ -151,10 +151,31 @@ export function ButtonWithTooltip() {
...
+      <div css={Css.mt4.$}>
+        <h2>Additional Props may be passed in to further customize in flight look</h2>
+        <Button
+          label={label}
+          endAdornment={ea}
+          icon={icon}
+          inFlightEndAdornment={eaInFlight}
+          inFlightIcon={iconInFlight}
+          inFlightLabel={labelInFlight}
+          onClick={async () => await new Promise((resolve) => setTimeout(resolve, 2000))}
```
bdow, line 176: **thought:** I like this pattern a lot. But it makes me wonder if we really need to have separate props for this... like what if we just expected people to do:
```
const [save, {loading}] = useMutation();
<Button label={loading ? "Uploading..." : "Upload} ... />
```
?
Maybe having these ternary statements here isn't as nice as hiding them behind the curtain within `Button`. But I dunno... Part of me also would rather standardize on some of this... like maybe the label changes, but perhaps we _always_ show that Loader for async buttons? Other than the loader and changing the text, I'm not sure I see the benefit of the `icon` being able to change. 🤔  I think it'd be odd if the loading state for our Button was different depending on who impelments it, so standardizing on always showing that `Loader` would be better imo.

src/components/Button.tsx
```diff
@@ -79,9 +87,11 @@ export function Button(props: ButtonProps) {
...
   const buttonContent = (
     <>
-      {icon && <Icon xss={iconStyles[size]} icon={icon} />}
-      {label}
-      {endAdornment && <span css={Css.ml1.$}>{endAdornment}</span>}
+      {icon && <Icon xss={iconStyles[size]} icon={inFlightIcon && asyncInProgress ? inFlightIcon : icon} />}
+      {inFlightLabel && asyncInProgress ? inFlightLabel : label}
+      {endAdornment && (
+        <span css={Css.ml1.$}>{inFlightEndAdornment && asyncInProgress ? inFlightEndAdornment : endAdornment}</span>
+      )}
```
bdow, line 94: **thought:** This is where I think that maybe we _always_ do the `<Loader />` end adornment if `asyncInProgress`. So it'd be more like:
```suggestion
      {(endAdornment || asyncInProgress) && (
        <span css={Css.ml1.$}>{asyncInProgress ? inFlightEndAdornment : endAdornment}</span>
      )}
```

src/components/Button.stories.tsx
apattersonATX-HB: I'm not picky about leaving `Icon` out and making `Loader` standard for anything async is verrry nice. It works for my use case 😏  

Keeping `inFlightLabel?: string" is great for a HOC. In our use-case we're passing the onClick Fn around so if we can leave out loading that's 1 less prop to define and pass down.

PR #739, opened 2022-11-23
feat: Optional disabled select value tooltips

blambillotte:
Allows for optionally adding a tooltip for the disabled reason on disabled select dropdown values
image


src/inputs/internal/ListBox.tsx
```diff
@@ -29,6 +30,7 @@ export function ListBox<O, V extends Key>(props: ListBoxProps<O, V>) {
     contrast = false,
     positionProps,
     loading,
+    disabledOptionsWithReasons = {},
```
blambillotte, line 33: Fallback here felt cleaner rather than dealing with an actual undefined value downstream for the `ChipSelectField` component that uses the `ListBox` component, but doesn't support disabling options

src/inputs/SelectField.test.tsx
```diff
@@ -148,6 +148,36 @@ describe("SelectFieldTest", () => {
...
+    // Given a Select Field with a disabled option
+    const { age, getByRole } = await render(
+      <SelectField
+        label="Age"
+        value="1"
+        options={options}
+        getOptionLabel={(o) => o.name}
+        getOptionValue={(o) => o.id}
+        data-testid="age"
+        disabledOptions={[["2", "Example Tooltip"]]}
```
stephenh, line 162: Thought: Wdyt about a hash like:

```
disabledOptions={[{ value: "2", reason: "Example Tooltip" }]}
```

?

stephenh (approved):
Cool!

src/inputs/SelectField.test.tsx
blambillotte: yeah that's much cleaner - I'll make that change!

PR #741, opened 2022-11-28
feat: (SC-26002) Add isCollapsedRow to GridTableApi

allangaldinosilva:
This PR update the GridTableApi interface to include a new method to discover when a row is collapsed.

Reference: https://github.com/homebound-team/internal-frontend/pull/3112


stephenh (approved):
Cool, thanks!

PR #745, opened 2022-11-29
fix: Use undefined for simpleHeader.data.

stephenh:
eslint doesn't like the `{}` type.

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -210,7 +210,7 @@ export function NoRowsFallback() {
 }
 
 // Make a `Row` ADT for a table with a header + 3 levels of nesting
-type HeaderRow = { kind: "header"; data: {} };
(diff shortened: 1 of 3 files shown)


bdow (approved):
I'm curious why eslint thinks `{}` is so bad?

stephenh:
Yeah, just copy/pasting their error message:

```
Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
```

I.e. even things like `true` match `{}`:

```
type F = {};
const f: F = true;
```

And the eslint rule is pointing that most programmers don't anticipate that, and write `{}` thinking that "any object literal", but really the `{}` type is more forgiving than that, like it could be `boolean` or `string` or what not.

PR #746, opened 2022-11-30
feat: SC-23012 Expose GridTable sorting in API

jgujgu:


src/components/Table/GridTableApi.ts
```diff
@@ -124,4 +126,8 @@ export class GridTableApiImpl<R extends Kinded> implements GridTableApi<R> {
   public getVisibleColumnIds() {
     return this.tableState.visibleColumnIds;
   }
+
+  public setSortColumn(columnId: string) {
+    return this.tableState.setSortKey(columnId);
+  }
```
bdow, line 132: **question:** I'm curious to see the Figma mocks that resulted in this request. I'm wondering if we should be passing along the ASC/DESC direction as well, or at least make it an optional property.
Specifically I am thinking of an instance where a user sorts the "value" column, which will make it sort ascending. Then they want to sort it descending.
Also, should we be exposing which column is currently sorted and which direction it is sorted? Might be necessary for whatever the other UX is that will be triggering this sort.

kbesingeryeh: @bdow here is the figma design: https://www.figma.com/file/OwnVReuiJqcsn0SxRVTQ3T/Q3-2022-%7C-%F0%9F%9B%A0-Bulk-Schedule-Management?node-id=4939%3A114688&t=Zw3NXBosroxwdvXT-0

PR #747, opened 2022-11-30
feat: SC-22120 Init active row

pgamboa07:
[SC-22120](https://app.shortcut.com/homebound-team/story/22120/maintain-schedule-view-after-refreshing)

**Description**
In order to maintain the scroll position of the table when the page is refreshing, add a initActiveRow prop to GridTable to scroll to index of active row is it is set


blambillotte:
Even though we're not going to use this to maintain scroll position on refresh, I think this would still be a useful addition, right? Being able to deep link someone directly to the position of a record sounds like a nice feature?

PR #752, opened 2022-12-02
feat: Support tooltips on ButtonGroup buttons

bdow:


src/components/ButtonGroup.tsx
```diff
@@ -17,9 +18,11 @@ export type ButtonGroupButton = {
   text?: string;
   onClick?: VoidFunction;
   /** Disables the button. Note we don't support the `disabled: ReactNode`/tooltip for now. */
```
stephenh, line 20: Thought: Wonder if this comment needs updated?

If I'm following, if you set _only_ disabled, it functions as a tooltip, but if you set both disabled and tooltip, then tooltip wins (and the button is still disabled)?

stephenh (approved):

src/components/ButtonGroup.tsx
bdow: I didn't even notice this comment. I will update it.
If you set `disabled={true} tooltip="Tooltip"`, then `tooltip` wins.
If you set `disabled="Disabled reason" tooltip="Tooltip"`, then `disabled` wins.

stephenh: Ah gotcha, sounds good!

PR #753, opened 2022-12-02
fix: [sc-26310] Mark Bound checkbox/switch disabled for readonly; 

apattersonATX-HB:
Also update BoundDateRange to be readonly based on field;


stephenh (approved):
Nice, thanks for the tests!

PR #755, opened 2022-12-02
feat: ButtonGroup 'xs' size

bdow:
In Blueprint, the Schedules page was reaching into the component to make the button group smaller. This caused problems when changing the markup of the ButtonGroup component. Instead of trying to rework those styles reaching into Beam, we are not supporting a smaller size.


stephenh (approved):
Nice! Thanks for cleaning that up!

PR #757, opened 2022-12-05
WIP: Sc 25077 - right pane layout

crshannon:
[sc-25077](https://app.shortcut.com/homebound-team/story/25077/beam-page-layout-with-detail-pane)

- Moving right pane component pattern established within project schedules into beam.

### DIFF
--- a/src/components/Layout/RightPaneLayout/RightPaneContext.tsx
+++ b/src/components/Layout/RightPaneLayout/RightPaneContext.tsx
@@ -1,0 +1,42 @@
+import React, { ReactElement, ReactNode, useContext, useMemo, useState } from "react";
+
+interface OpenRightPaneOpts {
+  content: ReactNode;
+}
+
+export type RightPaneLayoutContextProps = {
+  openInPane: (opts: OpenRightPaneOpts) => void;
+  closePane: () => void;
+  isRightPaneOpen: Boolean;
+  rightPaneContent: ReactNode;
+};
+
+export const RightPaneContext = React.createContext<RightPaneLayoutContextProps>({
+  openInPane: () => {},
+  closePane: () => {},
+  isRightPaneOpen: false,
+  rightPaneContent: null,
+});
+
+export function RightPaneProvider({ children }: { children: ReactElement }) {
+  const [rightPaneContent, setRightPaneContent] = useState<ReactNode>(undefined);
+  const [isRightPaneOpen, setIsRightPaneOpen] = useState<Boolean>(false);
+
+  const context = useMemo(() => {
+    return {
+      openInPane: (opts: OpenRightPaneOpts) => {
+        setRightPaneContent(opts?.content);
+        setIsRightPaneOpen(true);
+      },
+      closePane: () => setIsRightPaneOpen(false),
+      rightPaneContent,
+      isRightPaneOpen,
+    };
+  }, [setRightPaneContent, rightPaneContent, isRightPaneOpen]);
+
+  return <RightPaneContext.Provider value={context}>{children}</RightPaneContext.Provider>;
+}
+
+export function useRightPaneContext() {
+  return useContext(RightPaneContext);
+}
--- a/src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
+++ b/src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
@@ -1,0 +1,38 @@
+import { Meta } from "@storybook/react";
+import { RightPaneLayout } from "src/components/Layout/RightPaneLayout/RightPaneLayout";
+import { Css } from "src/Css";
+import { Button } from "../../Button";
+import { useRightPane } from "./useRightPane";
+
+export default {
+  component: RightPaneLayout,
+  title: "Workspace/Components/Layout/RightPaneLayout",
+} as Meta;
+
+function SampleContent() {
+  const { openRightPane } = useRightPane();
+  return (
+    <div>
+      <Button label={"Open Pane"} onClick={() => openRightPane({ content: <DetailPane /> })} />
+    </div>
+  );
+}
+
+function DetailPane() {
+  const { closeRightPane } = useRightPane();
+  return (
+    <div>
+      <Button label={"Close Pane"} onClick={() => closeRightPane()} />
+    </div>
+  );
+}
+
+export function Examples() {
+  return (
+    <div css={Css.df.fdc.gap2.$}>
+      <RightPaneLayout>
+        <SampleContent />
+      </RightPaneLayout>
+    </div>
+  );
+}
--- a/src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
+++ b/src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
@@ -1,7 +1,16 @@
 import { Meta } from "@storybook/react";
-import { RightPaneLayout } from "src/components/Layout/RightPaneLayout/RightPaneLayout";
-import { Css } from "src/Css";
+import { PropsWithChildren, useMemo } from "react";
+import { zeroTo } from "src/utils/sb";
+import { Css, Palette } from "../../../Css";
 import { Button } from "../../Button";
+import { IconButton } from "../../IconButton";
+import { GridColumn, GridDataRow, GridTable, simpleHeader, SimpleHeaderAndData } from "../../Table";
+import { FullBleed } from "../FullBleed";
+import { PreventBrowserScroll } from "../PreventBrowserScroll";
+import { ScrollableContent } from "../ScrollableContent";
+import { ScrollableParent } from "../ScrollableParent";
+import { RightPaneProvider } from "./RightPaneContext";
+import { RightPaneLayout } from "./RightPaneLayout";
 import { useRightPane } from "./useRightPane";
 
 export default {
@@ -12,7 +21,7 @@ export default {
 function SampleContent() {
   const { openRightPane } = useRightPane();
   return (
-    <div>
+    <div css={Css.bgWhite.h100.$}>
       <Button label={"Open Pane"} onClick={() => openRightPane({ content: <DetailPane /> })} />
     </div>
   );
@@ -21,18 +30,129 @@ function SampleContent() {
 function DetailPane() {
   const { closeRightPane } = useRightPane();
   return (
-    <div>
+    <div css={Css.bgWhite.h100.$}>
       <Button label={"Close Pane"} onClick={() => closeRightPane()} />
     </div>
   );
 }
 
-export function Examples() {
+export function Basic() {
   return (
-    <div css={Css.df.fdc.gap2.$}>
-      <RightPaneLayout>
-        <SampleContent />
-      </RightPaneLayout>
-    </div>
+    <RightPaneProvider>
+      <div css={Css.df.fdc.gap2.h100.$}>
+        <RightPaneLayout paneBgColor={Palette.White}>
+          <SampleContent />
+        </RightPaneLayout>
+      </div>
+    </RightPaneProvider>
+  );
+}
+
+export function GridTableWithRightPane() {
+  return (
+    <TestProjectLayout>
+      <ExamplePageComponent />
+    </TestProjectLayout>
+  );
+}
+
+function ExamplePageComponent() {
+  return (
+    <>
+      <FullBleed>
+        <header css={{ ...Css.py2.bb.bGray200.$ }}>
+          <h1 css={Css.xlSb.$}>Page content fixed to top</h1>
+        </header>
+      </FullBleed>
+      <ScrollableTableExample />
+    </>
+  );
+}
+
+function TestProjectLayout({ children }: PropsWithChildren<{}>) {
+  return (
+    <PreventBrowserScroll>
+      <div css={Css.df.h100.$}>
+        <ScrollableParent xss={Css.px3.$}>{children}</ScrollableParent>
+      </div>
+    </PreventBrowserScroll>
+  );
+}
+
+function ScrollableTableExample({ numCols, numRows }: { numCols?: number; numRows?: number }) {
+  return (
+    <ScrollableContent>
+      <RightPaneProvider>
+        <RightPaneLayout wrapProvider={true} paneBgColor={Palette.White}>
+          <TableExample numCols={numCols} numRows={numRows} />
+        </RightPaneLayout>
+      </RightPaneProvider>
+    </ScrollableContent>
+  );
+}
+
+type Row = SimpleHeaderAndData<{ name: string; value: number }>;
+function TableExample({ numCols = 10, numRows = 100 }: { numCols?: number; numRows?: number }) {
+  const { openRightPane } = useRightPane();
+
+  const rows: GridDataRow<Row>[] = useMemo(
+    () => [
+      simpleHeader,
+      ...zeroTo(numRows).map((i) => ({
+        kind: "data" as const,
+        id: String(i),
+        data: { name: `ccc ${i}`, value: i + 1 },
+      })),
+    ],
+    [numRows],
+  );
+  const columns: GridColumn<Row>[] = useMemo(
+    () =>
+      zeroTo(numCols - 1).map((i) => ({
+        header: `Header ${i + 1}`,
+        data: ({ value }) =>
+          i === 0 ? (
+            <div>
+              <Button
+                label={"Open Pane"}
+                onClick={() => openRightPane({ content: <TestDetailPane value={value} /> })}
+              />
+            </div>
+          ) : (
+            `Cell ${i + 1}x${value}`
+          ),
+        w: "200px",
+        sticky: i === 0 ? "left" : undefined,
+      })),
+    [numCols],
+  );
+
+  return <GridTable as={"virtual"} stickyHeader columns={columns} rows={rows} style={{ rowHeight: "fixed" }} />;
+}
+
+function TestDetailPane({ value }: { value: number }) {
+  const { closeRightPane } = useRightPane();
+
+  return (
+    <ScrollableParent>
+      <div css={Css.relative.p2.bb.$}>
+        <div css={Css.df.jcsb.aic.$}>
+          <h2 css={Css.py2.$}>Detail Pane {value}</h2>
+          <div>
+            <IconButton icon={"x"} onClick={() => closeRightPane()} />
+          </div>
+        </div>
+        <ScrollableContent virtualized={true}>
+          <nav>
+            <ul css={Css.listReset.df.fdc.gap5.mt2.p2.$}>
+              {zeroTo(20).map((i) => (
+                <li key={i}>scroll items</li>
+              ))}
+              <li>Bottom!</li>
+            </ul>
+          </nav>
+        </ScrollableContent>
+      </div>
+    </ScrollableParent>
   );
 }
--- a/src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
+++ b/src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
@@ -83,7 +83,7 @@ function ScrollableTableExample({ numCols, numRows }: { numCols?: number; numRow
   return (
     <ScrollableContent>
       <RightPaneProvider>
-        <RightPaneLayout wrapProvider={true} paneBgColor={Palette.White}>
+        <RightPaneLayout paneBgColor={Palette.White}>
           <TableExample numCols={numCols} numRows={numRows} />
         </RightPaneLayout>
       </RightPaneProvider>
--- a/src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
+++ b/src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
@@ -1,66 +1,59 @@
 import { AnimatePresence, motion } from "framer-motion";
-import React, { ReactElement, ReactNode, useContext, useMemo, useState } from "react";
-import { Css, px } from "../../../Css";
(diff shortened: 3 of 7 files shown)


src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <ScrollableParent xss={Css.px3.$}>{children}</ScrollableParent>
+      </div>
+    </PreventBrowserScroll>
+  );
+}
+
+function ScrollableTableExample({ numCols, numRows }: { numCols?: number; numRows?: number }) {
+  return (
+    <ScrollableContent>
+      <RightPaneProvider>
```
crshannon, line 85: Currently exposing `RightPaneProvider` as in testing on schedules it needs to sit at a higher level than the layout part of the component. Looking for options here as to this being a pattern we are okay with to keep the context provider localized to the pages that need it. Currently the Schedules detail pane is shared between the different views (Table, Gantt, Calendar) and the layout needs to support being passed by the `ScrollingContainer`.

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -0,0 +1,52 @@
...
+export function RightPaneLayout({ children, paneBgColor }: { children: ReactElement; paneBgColor?: Palette }) {
+  const { isRightPaneOpen, rightPaneContent } = useRightPaneContext();
+
+  // Todo pass as prop?
+  const paneWidth = 450;
+
+  return (
+    <div css={Css.h100.df.$}>
+      <>
+        <motion.div
```
crshannon, line 15: @bdow - do we want to wrap the page content in framer here or use css + transitions

bdow: My personal choice would be to use CSS Transitions whenever possible and avoid framer-motion. I think framer motion has some good use cases, like when needing to do multiple animations together, or attempting to animate from `auto` to `100%`. But typically, I don't see the benefit in framer motion over css transitions.

src/components/Layout/RightPaneLayout/RightPaneContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export function RightPaneProvider({ children }: { children: ReactElement }) {
+  const [rightPaneContent, setRightPaneContent] = useState<ReactNode>(undefined);
+  const [isRightPaneOpen, setIsRightPaneOpen] = useState<Boolean>(false);
+
+  const context = useMemo(() => {
+    return {
+      openInPane: (opts: OpenRightPaneOpts) => {
+        setRightPaneContent(opts?.content);
+        setIsRightPaneOpen(true);
```
bdow, line 29: **question:** Can we get away with just one state here? Like if `rightContentPane` isn't null/undefined, then considered the right pane opened?

src/components/Layout/RightPaneLayout/RightPaneContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+      },
+      closePane: () => setIsRightPaneOpen(false),
```
bdow, line 31: **thought:** If we do want to keep both the states (`rightPaneContent` and `isRightPaneOpen`), should we setting `rightPaneContent` back to `undefined` here? 
I could see the argument that you don't want to have the content removed until after the animation to close has finished, so maybe we can figure that out too, but I'd be okay with just having it removed.

src/components/Layout/RightPaneLayout/useRightPane.tsx
```diff
@@ -0,0 +1,27 @@
...
+  /** Closes the right pane */
+  closeRightPane: () => void;
+}
+
+export function useRightPane(): UseRightPaneHook {
+  const { openInPane, closePane } = useRightPaneContext();
+  const openRightPane = useCallback((opts: OpenRightPaneOpts) => {
+    openInPane(opts);
+  }, []);
+  const closeRightPane = useCallback(() => closePane(), []);
```
bdow, line 21: **question:** Do we need to re-create these methods, or can we directly use what we get from the context? It doesn't appear we do anything additional to them, we are just wrapping them in a `useCallback`. Though, they should be stable already as the context memoizes the values. 
...I guess the benefit here is that the we do not define hook dependencies, so even if the context re-creates the `openInPane` or `closePane` methods, then we will not update them here. 🤔  Could that end up causing any issues? 🤔 ... 🚶‍♂️ heading back to the context to comment some more with this thought in mind...

src/components/Layout/RightPaneLayout/RightPaneContext.tsx
bdow, line 31: **thought/suggestion:** What do you think about memoizing the `openInPane` and `closePane` methods separately? Right now they'll be built new every time a dependency changes, which _could_ have other implications further down the tree.
I'd think we could do something like:
```
const openInPane = useCallback(() => ..., [setRightPaneContent]);
const closePane = useCallback(() => ..., [setRightPaneContent]);

const contextValue = useMemo(() => ({ openInPane, closePane, rightPaneContent }), [openInPane, closePane, rightPaneContent]);
```
This way, the `openInPane` and `closePane` methods would only ever be rebuilt when if the setter method changes (which should be never).

src/components/Layout/RightPaneLayout/useRightPane.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  return {
+    openRightPane,
+    closeRightPane,
+  };
```
bdow, line 26: **super nit:** This is a good opportunity to go from 4 lines of code to 1. 
```suggestion
  return { openRightPane, closeRightPane };
```

src/components/Layout/RightPaneLayout/useRightPane.tsx
```diff
@@ -0,0 +1,27 @@
+import { ReactNode, useCallback } from "react";
+import { useRightPaneContext } from "./RightPaneContext";
+
+interface OpenRightPaneOpts {
+  content: ReactNode;
+}
```
bdow, line 6: **suggestion:** This `OpenRightPaneOpts` interface on what properties are accepted by the `openInPane` are duplicated between here and in the Context file. Is there a reason to separate them out, or can we consolidate this?

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -0,0 +1,52 @@
...
+          layout="position"
+          key="rightPaneLayoutPageContent"
+          css={Css.h100.if(!!isRightPaneOpen).overflowX("scroll").mr3.if(!isRightPaneOpen).overflowX("unset").$}
```
bdow, line 18: **question:** Do we need `if(!isRightPaneOpen).overflowX("unset")`?

src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
stephenh: Fwiw I agree ideally a caller could just use `RightPaneLayout` and not have to worry about "I also need to wrap that in `RightPaneProvider`".

I had the same musing, of having `RightPaneLayout` just internally include a new `RightPaneProvider`, but if it needs to be higher up in the tree, maybe we just add `RightPaneProvider` directly to `BeamContext`? 

Granted, we have a ton of `SnackbarProvider` / `ToastProvider` / `AutoSaveStatusProvider` already there :-) but :shrug: so far its been working out well.

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -0,0 +1,52 @@
...
+              data-testid="rightPaneContent"
+              css={Css.bgColor(paneBgColor).h100.maxw(px(paneWidth)).w100.df.fdc.relative.$}
+              // Keeping initial x to offset pane width and space between panel and page content
+              initial={{ x: paneWidth + 24 }}
+              animate={{ x: 0 }}
+              // Custom transitions settings for the translateX animation
+              transition={{ ease: "linear", duration: 0.2 }}
+              exit={{ transition: { ease: "linear", duration: 0.2 }, x: paneWidth }}
+              // Preventing clicks from triggering parent onClick
+              onClick={(e) => e.stopPropagation()}
```
bdow, line 43: **question:** What is the reason for `stopPropagation`?

bdow: Specifically, in what scenario do we want to avoid this the event bubbling?

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
stephenh, line 18: Nit: could use `overflowXScroll`; granted, we don't have `overflowXUnset`.

Although, do you need `overflowX("unset")`? I think if `isRightPaneOption` is false, then `overflowX("scroll")` just won't be output, so you won't need to unset it?

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -0,0 +1,52 @@
...
+              layout="position"
+              key="rightPane"
+              data-testid="rightPaneContent"
+              css={Css.bgColor(paneBgColor).h100.maxw(px(paneWidth)).w100.df.fdc.relative.$}
+              // Keeping initial x to offset pane width and space between panel and page content
+              initial={{ x: paneWidth + 24 }}
+              animate={{ x: 0 }}
+              // Custom transitions settings for the translateX animation
+              transition={{ ease: "linear", duration: 0.2 }}
+              exit={{ transition: { ease: "linear", duration: 0.2 }, x: paneWidth }}
```
bdow, line 41: **thought:** If I'm reading this correctly, it seems like the Right Pane is immediately rendered once `isRightPaneOpen` is true, but it's rendered off screen. Then uses a `translateX` transition to bring it into the viewport, is that right?  If so, I'm thinking that would cause an unnecessary horizontal scrollbar until the transition completes and the pane is completely in the viewport.

bdow: Maybe `AnimatePresence` solves that? I dunno.

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
bdow, line 18: Can we use `overflowXAuto` instead of `scroll`? The benefit is not creating a scrollbar gutter when unnecessary.

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
bdow: Doesn't appear to solve it:
https://www.loom.com/share/6d136b819ea645e7853226ae8cc40ea7

src/components/Layout/RightPaneLayout/RightPaneLayout.stories.tsx
crshannon: I agree, the initial thought was to have them both contained in the singular beam component but the complexity of consuming it in blueprint has proven to be more tricky. I am testing to see how things work with exposing it via the BeamContext which for now seems like the path of least resistance.

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
crshannon: sounds good. I will switch this over - FWIW the `layout="position"` is what I used to resolve the animations being in sync.

src/components/Layout/RightPaneLayout/RightPaneContext.tsx
crshannon: I agree It should be cleared after the close animation has been performed.

src/components/Layout/RightPaneLayout/useRightPane.tsx
crshannon: hm - odd perhaps auto formatting..

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
crshannon: updated to use `overflowXAuto`

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
crshannon: was able to resolve this by using `overflowXHidden`

src/components/Layout/RightPaneLayout/RightPaneContext.tsx
crshannon: Note: framer-motion exposes a `onAnimationComplete` which is nice to check against the pane being open vs closed for clearing content

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
crshannon: Removed framer-motion from the page content wrapper and implemented with css transitions

src/components/BeamContext.tsx
```diff
@@ -87,18 +88,20 @@ export function BeamProvider({ children, ...presentationProps }: BeamProviderPro
...
-            <ToastProvider>
-              <OverlayProvider>
-                {children}
-                {modalRef.current && <Modal {...modalRef.current} />}
-              </OverlayProvider>
-              <SuperDrawer />
-            </ToastProvider>
-          </SnackbarProvider>
-        </AutoSaveStatusProvider>
+        <RightPaneProvider>
```
crshannon, line 91: @stephenh @bdow - relocated `RightPaneProvider` to the beam context, want to make sure you both don't have any concerns with this living at this level. - It feels like the most ideal location for it when testing in blueprint.

stephenh: Yep, I think it's the right place for it.

src/components/Layout/RightPaneLayout/RightPaneContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const RightPaneContext = React.createContext<RightPaneLayoutContextProps>({
+  openInPane: () => {},
+  closePane: () => {},
+  clearPane: () => {},
+  isRightPaneOpen: false,
+  rightPaneContent: null,
+});
+
+export function RightPaneProvider({ children }: { children: ReactNode }) {
+  const [rightPaneContent, setRightPaneContent] = useState<ReactNode>(undefined);
```
stephenh, line 24: Suggestion: Maybe add an in-source comment that we track the pane content separate from is open so that we can animate away the pane content even though the pane is now closed?

Just thinking I'd probably re-wonder this reading/maintaining the code a few months down the road.

stephenh (approved):
Cool! Lgtm!

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Css, Palette, px } from "../../../Css";
+import { useRightPaneContext } from "./RightPaneContext";
+
+export function RightPaneLayout({
+  children,
+  paneBgColor,
+  paneWidth = 450,
+}: {
+  children: ReactElement;
+  paneBgColor?: Palette;
```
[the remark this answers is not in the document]
crshannon: this was just my past design system habits kicking in to make this as dumb as possible. I can remove and set for now and if we need we can add that back in with a default set.

src/components/Layout/RightPaneLayout/RightPaneContext.tsx
crshannon: Added note to this for future reference.

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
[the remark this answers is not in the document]
stephenh: Ah cool; yeah for off-the-shelf/widely-used design systems that customization/flexibility makes a lot of sense.

Admittedly, with Beam I probably have too much fun leaning into the mission of being "opinionated!" to keep the app-side as simple and consistent as possible. :-)

src/components/Layout/RightPaneLayout/useRightPane.tsx
stephenh: Fwiw I'm +1 for keeping the current "this is our public API `useRightPane`" that is separate from the internal `useRightPaneContext`.

crshannon: sounds good to me I agree Id prefer to keep it also

bdow: I'm good with keeping them separate. My question was can this just do:
```
export function useRightPane(): UseRightPaneHook {
  const { openInPane, closePane } = useRightPaneContext();

  return {
    openRightPane: openInPane,
    closeRightPane: closePane,
  };
```
?
Theres not really a reason, at least that I see, to wrap these function calls in another function.

src/components/Layout/RightPaneLayout/useRightPane.tsx
```diff
@@ -0,0 +1,17 @@
...
+  closeRightPane: () => void;
+}
+
+export function useRightPane(): UseRightPaneHook {
+  const { openInPane, closePane } = useRightPaneContext();
+
+  return {
+    openRightPane: (opts: OpenRightPaneOpts) => openInPane(opts),
+    closeRightPane: () => closePane(),
+  };
```
bdow, line 16: **suggestion:** I know pushed you towards not using `onCallback` for `openRightPane` and `closeRightPane` in my previous suggestions, but that was because I was thinking we could directly reference the `openInPane` and `closePane` methods, vs wrapping them in another function to call them. As it is right now the `openRightPane` and `closeRightPane` methods will be considered "new" whenever the calling component re-renders, which can have some side-effects of unnecessary recalculations.
Consider:
```
function MyComponent() {
  const { openRightPane } = useRightPane();
  
  const myCallback = useCallback((id) => {
    openRightPane(id);
  }, [openRightPane]);
  
  const [hovered, setHover] = useHover();

  return (
    <div 
      onClick={myCallback(123)}
      onMouseEnter={() => setHover(true)}
      onMouseLeave={() => setHover(false)}
      css={Css.if(hover).underline.$}
   >
      Click Me!
   </div>
  );
}
```
Whenever we hover over the div, it triggers the component to rerender, reevaluating the `useRightPane` hook, getting a new function/object in memory, which then triggers the `useCallback` to reevaluate, generating a new `myCallback` in memory as well. 

So TLDR: Let's either directly reference the stable `openInPane` and `closePane` methods
```
return { openRightPane: openInPane, closeRightPane: closePane }
```
Or memoize the local methods so they don't trigger unnecessary rerenders downstream.

Sorry if this was over verbose.

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -0,0 +1,51 @@
...
+        >
+          {children}
+        </div>
+        <AnimatePresence>
+          {isRightPaneOpen && (
+            <motion.div
+              layout="position"
+              key="rightPane"
+              data-testid="rightPaneContent"
+              css={Css.bgColor(paneBgColor).h100.maxw(px(paneWidth)).w100.df.fdc.relative.$}
```
bdow, line 35: **question:** I'm wondering if we should dictate that the right pane div is a `flex` parent? Doesn't seem necessary to me to define that here, but curious if there was a reason?

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -0,0 +1,51 @@
...
+  paneWidth?: number;
+}) {
+  const { isRightPaneOpen, rightPaneContent, clearPane } = useRightPaneContext();
+
+  return (
+    <div css={Css.h100.df.overflowXHidden.$}>
+      <>
+        <div
+          css={
+            Css.add("width", `calc(100% - ${paneWidth + 24}px)`)
```
bdow, line 22: **nit (non-blocking):** This could be `Css.w(`calc(100% - ${paneWidth + 24}px)`).$`. (no need of the `Css.add`)

bdow (approved):
The only thing I'd like to get addressed before merging is memoizing the methods returned in the `useRightPane` hook. Otherwise, looks good! Thanks for going through all the feedback.

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
bdow: Also, I am curious about the `maxw(px(paneWidth)).w100`. Could we  just define this to be `wPx(paneWidth)`? Seems like we'd be accomplishing the same thing, right? I feel like I might be missing something 🤔 . 

Ahh... Thinking through when to use the `max-width` + `width: 100%` combination,.. seems like it'd be when the element could be shown in a viewport that is thinner than the defined `paneWidth`. That way we say "take up all the space you can, but stop at 100% of the viewport to not cause a horizontal scrollbar". It's a good little trick for mobile screens, but not sure if it's a concern for this layout. Either way, doesn't need to be changed - the styles just stuck out to me.

src/components/Layout/RightPaneLayout/useRightPane.tsx
crshannon: yeah makes sense to just pass them down

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
crshannon: yeah admittedly this was brought over from the blueprint implementation but thanks for the eyes

PR #759, opened 2022-12-07
fix: Avoid inheriting line-height in ButtonGroup

bdow:
In the below examples, the line-height is being inherited from the ModalHeader component, causing `<span/>` elements that wrap each button to be taller than they should.

#### Before
Screen Shot 2022-12-06 at 8 28 19 PM

#### After
Screen Shot 2022-12-06 at 8 28 43 PM


stephenh (approved):
Huh! Nice fix!

PR #760, opened 2022-12-07
feat: add horizontal support for FormLines

roberthdg:
image

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -525,14 +525,12 @@ export const DataTypeColumns = newStory(
     const dateCol = dateColumn<Row2>({ header: "Date", data: ({ date }) => date });
     const priceCol = numericColumn<Row2>({
       header: "Price",
-      data: ({ priceInCents }) => (
-        <NumberField labelStyle="hidden" label="Price" value={priceInCents} onChange={noop} type="cents" />
-      ),
+      data: ({ priceInCents }) => <NumberField label="Price" value={priceInCents} onChange={noop} type="cents" />,
     });
     const readOnlyPriceCol = numericColumn<Row2>({
       header: "Read only Price",
       data: ({ priceInCents }) => (
-        <NumberField labelStyle="hidden" label="Price" value={priceInCents} onChange={noop} type="cents" readOnly />
+        <NumberField label="Price" value={priceInCents} onChange={noop} type="cents" readOnly />
       ),
     });
     const actionCol = actionColumn<Row2>({ header: "Action", data: () => <IconButton icon="check" onClick={noop} /> });
@@ -565,9 +563,7 @@ export function WrappedHeaders() {
   const rightAlignedColumn = numericColumn<Row2>({
     header: "Right aligned column header",
     data: ({ priceInCents }) => ({
-      content: (
-        <NumberField labelStyle="hidden" label="Price" value={priceInCents} onChange={noop} type="cents" readOnly />
-      ),
+      content: <NumberField label="Price" value={priceInCents} onChange={noop} type="cents" readOnly />,
       sortValue: priceInCents,
     }),
     w: "150px",
@@ -614,11 +610,9 @@ export function ColSpan() {
   const dateCol = dateColumn<ColspanRow>({ header: "Date", data: ({ date }) => date, total: "" });
   const priceCol = numericColumn<ColspanRow>({
     header: "Price",
-    data: ({ priceInCents }) => (
-      <NumberField labelStyle="hidden" label="Price" value={priceInCents} onChange={noop} type="cents" />
-    ),
+    data: ({ priceInCents }) => <NumberField label="Price" value={priceInCents} onChange={noop} type="cents" />,
     total: ({ totalPriceInCents }) => (
-      <NumberField labelStyle="hidden" label="Price" readOnly value={totalPriceInCents} onChange={noop} type="cents" />
+      <NumberField label="Price" readOnly value={totalPriceInCents} onChange={noop} type="cents" />
     ),
   });
   // Use a green background to show the 1st column is flush left
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -526,13 +526,13 @@ export const DataTypeColumns = newStory(
     const priceCol = numericColumn<Row2>({
       header: "Price",
       data: ({ priceInCents }) => (
-        <NumberField hideLabel label="Price" value={priceInCents} onChange={noop} type="cents" />
+        <NumberField labelStyle="hidden" label="Price" value={priceInCents} onChange={noop} type="cents" />
       ),
     });
     const readOnlyPriceCol = numericColumn<Row2>({
       header: "Read only Price",
       data: ({ priceInCents }) => (
-        <NumberField hideLabel label="Price" value={priceInCents} onChange={noop} type="cents" readOnly />
+        <NumberField labelStyle="hidden" label="Price" value={priceInCents} onChange={noop} type="cents" readOnly />
       ),
     });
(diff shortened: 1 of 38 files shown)


src/forms/FormLines.tsx
```diff
@@ -51,6 +53,14 @@ export function FormLines(props: FormLinesProps) {
             const clone = cloneElement(child, { isFirst: firstFormHeading });
             firstFormHeading = false;
             return clone;
+          } else if (horizontalLayout && typeof child === "object" && (child as ReactElement).props?.label) {
```
stephenh, line 56: Thought: Ah! Wow, I'm kinda surprised this works, but yeah I see what you're doing...

I know it's more work, but I was thinking a more robust approach to this change I think would be:

* Rename `PresentationFieldProps.hideLabel` to `labelStyle` with a type of `"inline" | "hidden" | "above" | "left"`
* Remove `TextFieldProps.inlineLabel` (replaced by `labelStyle=inline`)
* Update `TextFieldBase` line ~175 to show the label above-vs-left-vs-hidden based on the `labelStyle`
* Etc. for `Switch`, `SelectField`, etc.

cc @bdow for thoughts/input 

And, again, it will be a larger diff but I think it will let pages that don't necessarily use `FormLines` (which most do not, unfortunately) still use the `labelStyle=top`.

And, being part of `PresentationProps`, callers would ideally be able to set `labelStyle` just once (i.e. via FormLines) and then not have to repeat the `hideLabel` on each of their wrapped form fields

src/forms/FormLines.stories.tsx
```diff
@@ -181,6 +181,25 @@ export function WithFieldProps() {
...
+        label="Unit of Measure"
+        value={1}
+        hideLabel
+        options={[
+          { id: 1, name: "Each" },
+          { id: 2, name: "Square Feet" },
+        ]}
+        onSelect={noop}
+      />
+      <Switch label="Signatory" labelStyle="form" hideLabel selected={true} onChange={noop} />
```
bdow, line 198: Would also be awesome if our `Switch` component and `RadioButtonGroup` and `CheckboxGroup` components could use the same `PresentationFieldProps` to drive the label layout.

src/components/Table/GridTable.stories.tsx
```diff
@@ -526,13 +526,13 @@ export const DataTypeColumns = newStory(
     const priceCol = numericColumn<Row2>({
       header: "Price",
       data: ({ priceInCents }) => (
-        <NumberField hideLabel label="Price" value={priceInCents} onChange={noop} type="cents" />
+        <NumberField labelStyle="hidden" label="Price" value={priceInCents} onChange={noop} type="cents" />
```
bdow, line 529: **FWIW:** We could just remove this property all together. `GridTable` has a `<PresentationProvider />` wrapped around it that should set the `labelStyle: "hidden"` accordingly. That should allow any field which reads this property to inherit it properly from the PresentationContext.

src/inputs/RadioGroupField.tsx
```diff
@@ -67,8 +67,16 @@ export function RadioGroupField<K extends string>(props: RadioGroupFieldProps<K>
...
+    <div
+      css={
+        Css.w("max-content")
+          .maxw(anyDescriptions ? "344px" : "320px")
+          .if(labelStyle === "left").w100.df.fdr.$
+      }
+    >
+      <div css={Css.if(labelStyle === "left").mw50.my("auto").$}>
+        <Label label={label} {...labelProps} {...tid.label} hidden={labelStyle === "hidden"} />
+      </div>
```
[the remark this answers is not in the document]
roberthdg: Without the div, the label wouldn't be vertically centered

image

image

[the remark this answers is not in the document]
bdow: Do we want the label to vertically centered? I kind of assumed it'd look better if it were top-aligned. 
If we do want it to be vertically centered, then we can stil easily apply that by setting `align-items: center` on the parent element. In the example code above I put `aifs` (align-items: flex-start), which could change to `aic` and then you'd have it center aligned.

src/inputs/Switch.tsx
```diff
@@ -16,9 +16,7 @@ export interface SwitchProps {
   /** Input label */
   label: string;
   /** Where to put the label. */
-  labelStyle?: "form" | "inline" | "filter";
-  /** Whether or not to hide the label */
-  hideLabel?: boolean;
+  labelStyle?: "form" | "inline" | "filter" | "hidden" | "left";
```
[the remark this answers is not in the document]
bdow: hmm.. interesting. I'd say don't worry about it for now. But maybe we can add a ```// TODO: Update `labelStyle` to make consistent with other `labelStyle` properties in the library```

src/inputs/RadioGroupField.tsx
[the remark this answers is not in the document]
roberthdg: You're right, for a form with multiple field it looks better top-aligned. I don't think it's been implemented yet in figma, so I'll leave it top-aligned.

roberthdg (commented):
I've made a [pr](https://github.com/homebound-team/internal-frontend/pull/3191) for when this lands, so that we update internal-frontend with the form field changes along with the version update.

src/inputs/CheckboxGroup.tsx
```diff
@@ -33,29 +34,44 @@ export interface CheckboxGroupProps {
...
-          <CheckboxGroupItem
-            key={option.value}
-            {...option}
-            groupState={state}
-            selected={state.value.includes(option.value)}
-          />
-        ))}
+      <div css={Css.if(labelStyle === "left").df.fdc.add("flexFlow", "wrap").$}>
+        {labelStyle !== "hidden" && (
+          <div css={Css.if(labelStyle === "left").mw50.my("auto").$}>
```
roberthdg, line 61: @bdow I think this is causing the dom diff (in this case it does need the vertically centered label to match the figma design)

src/inputs/TextFieldBase.tsx
```diff
@@ -113,13 +111,15 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
...
+    container: Css.df.fdc.w100.maxw(px(550)).relative.if(labelStyle === "left").fdr.gap2.jcsb.aic.$,
     inputWrapper: {
       ...Css[typeScale].df.aic.br4.px1.w100
         .hPx(fieldHeight - maybeSmaller)
         .if(compact)
         .hPx(compactFieldHeight - maybeSmaller).$,
-      ...Css.bgColor(bgColor).gray900.if(contrast).white.$,
+      ...Css.bgColor(bgColor)
+        .gray900.if(contrast)
+        .white.if(labelStyle === "left").maxw50.$,
```
stephenh, line 122: Question: @roberthdg hm, does this `maxw50` handle making all of the labels line up?

Like I'm wondering what happens if we have:

```
| Some really long label   <field> |
| Short label <field>              |
```

Will the "really long label" vs. "really short label" mean that the two `<fields>` are no longer show vertically aligned?

I kind of assume not, which if so is a miss on my part, not thinking of that up-front / before starting this ticket, that to ensure the right-side fields are 100% vertically aligned, we probably need some sort of cross-`TextFieldBase` containers/constraints.

But, I dunno, I'm kind of guessing; @roberthdg  how does vertical alignment work out in the current approach?

stephenh (approved):
Nice, I think this is great. Pretty happy to ship this as-is, even if we have to do follow-up work to handle cross-field vertical alignment.

src/inputs/TextFieldBase.tsx
roberthdg: That `maxw50` would be for the `inputWrapper` (by default it expands to whatever width its available), it could also just be set to w50 to override the default w100.

PR #764, opened 2022-12-08
feat: Remove rowLink from RowStyles.

stephenh:


bdow (approved):
I like it! Gonna be a fairly big breaking change, but I think it makes sense to clean up rowStyles where it's no longer needed.

PR #766, opened 2022-12-09
 fix: make job logs filters responsive

aTrujilloHomeBound:
This is a fix to previously adding a filters prop to fitlers. off by one.


src/components/Filters/Filters.tsx
```diff
@@ -29,15 +29,15 @@ interface FilterProps<F, G extends Value = string> {
...
+  const { filter, onChange, filterDefs, groupBy, vertical = false, numberOfInlineFilters = groupBy ? 2 : 3 } = props;
   const testId = useTestIds(props, filterTestIdPrefix);
 
   const { openModal } = useModal();
   const [pageFilters, modalFilters] = useMemo(() => {
     // Take the FilterDefs that have a `key => ...` factory and eval it
     const impls = safeEntries(filterDefs).map(([key, fn]) => [key, fn(key as string)]);
     // If we have more than numberOfInlineFilters depending on groupby,
-    if (!vertical && impls.length > numberOfInlineFilters) {
+    if (!vertical && impls.length > numberOfInlineFilters + 1) {
```
stephenh, line 40: Thought: both `impls.length` and `numberOfInlineNumbers` are 1-based numbers, so I don't thinkwe need the + 1 to adjust it?

Like if `impls.length` is 3, and `numberOfInlineFilters = 3`, that is fine, we don't need to break out some of the modals.

It's only if `impls.length` is 4 and `numberOfInlineFilters = 3`, that we need to do the break out?

src/components/Filters/Filters.tsx
```diff
@@ -29,15 +29,15 @@ interface FilterProps<F, G extends Value = string> {
 }
 
 function Filters<F, G extends Value = string>(props: FilterProps<F, G>) {
-  const { filter, onChange, filterDefs, groupBy, vertical = false, numberOfInlineFilters = groupBy ? 3 : 4 } = props;
+  const { filter, onChange, filterDefs, groupBy, vertical = false, numberOfInlineFilters = groupBy ? 2 : 3 } = props;
```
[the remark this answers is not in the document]
aTrujilloHomeBound: Yeah, I was under the assumption the default was 4

src/components/Filters/Filters.tsx
aTrujilloHomeBound: With the +1 when there are 4 filters we get the last filter without more filters shown:
Screen Shot 2022-12-09 at 11 13 39 AM

when there is not + 1 and  4 filters we see: 
Screen Shot 2022-12-09 at 11 13 51 AM

I like that it shows the last filter without opening the modal for 1 item

stephenh: Ah gotcha, yeah, I remember that now, don't show "more filters" if there'd be only 1 item to put in it...

I bet that is also why I thought "there are four inline filters by default", e.g. looking at your 1st screenshot and we see four.

So...is the "most intuitive" fix to:

a) keep `numberOfInlineFilters = groupBy ? 3 : 4` (because if you pass 4 filters, with no group by, then yeah you'll see 4 inline filters), and then

b) keep line 40 to `impls.length > numberOfInlineFilters`, 

c) make line 43 be `Object.fromEntries(impls.slice(0, numberOfInlineFilters - 1))`

Because this will handle:

* `numberOfInlineFilters=4`, `impls.length=3`, don't show the more button, all filters on the page
* `numberOfInlineFilters=4`, `impls.length=4`, don't show the more button, all filters on the page
* `numberOfInlineFilters=4`, `impls.length=5`, show the more button and take the 1st three filters (as before)

Like I don't 100% know what bug you're fixing, but is most of the code on `main` right, it's just the `impls.slice` was putting an extra filter on the page? (b/c we weren't `-1`-ing to keep a slot reserved for the more button)

stephenh (approved):

PR #769, opened 2022-12-16
feat: SC-24760 Add checkbox filter

kbesingeryeh:
* Adds a new `checkboxFilter` to Beam's Filter components
* Behaves similar to `toggleFilter`, but using a checkbox instead
Screen Shot 2022-12-16 at 9 02 54 AM

### DIFF
--- a/src/components/Filters/CheckboxFilter.tsx
+++ b/src/components/Filters/CheckboxFilter.tsx
@@ -1,0 +1,58 @@
+import { BaseFilter } from "src/components/Filters/BaseFilter";
+import { Filter } from "src/components/Filters/types";
+import { Checkbox } from "src/inputs";
+import { TestIds } from "src/utils/useTestIds";
+
+export type CheckboxFilterProps<V> = {
+  label?: string;
+  onValue?: V | undefined;
+  offValue?: V | undefined;
+  defaultValue?: V | undefined;
+};
+
+/**
+ * Provides a two-state "on/off" filter.
+ *
+ * By default the on/off values are `on === true` and `off === undefined`.
+ *
+ * You can flip the on/off values by passing `onValue: false`, in which case
+ * `on === false` and off === undefined`.
+ *
+ * Or you can set on/off directly, by passing both `onValue` and `offValue`, even to
+ * non-boolean values, i.e. `onValue: "foo", offValue: "bar"`.
+ */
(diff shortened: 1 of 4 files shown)


src/components/Filters/CheckboxFilter.tsx
```diff
@@ -0,0 +1,58 @@
...
+        label={this.label}
+        onChange={(on) => {
+          setValue(on ? onValue : offValue);
+        }}
+        {...this.testId(tid)}
+      />
+    );
+  }
+
+  get hideLabelInModal() {
```
stephenh, line 55: Ah nice, I forgot we already had a way to hide labels in the modal; that worked out...

stephenh (approved):
Cool, this lgtm!

Fwiw, let's go ahead and ship this, but next time around, I'm curious Kelsie/your-tech lead, if we should nudge UX to ask "...so do you really want _sometimes_ toggles and _sometimes_ checkboxes? Why is that? Can you describe the rules-of-thumb for when to use which one?"

Like I'm wondering if they would have preferred to either a) realize that oh toggles are actually the standard for on/off-sytle filters and we should purposefully stick with that pattern (i.e. the designer unintentionally went off-pattern without realizing it), or b) realize that actually we really don't like the toggle style filter anymore, and we should move everything over to the new checkbox style, so the app is consistent.

src/components/Filters/CheckboxFilter.tsx
kbesingeryeh: Sounds good, thanks for the feedback. I will mention it to the team on Monday. I wanted to take this ticket for Hackday because I noticed design moving to this new pattern of checkbox filters and we've been using the toggle instead because this hadn't been built yet.

stephenh: Cool! Yeah, cleaning up/adding to Beam is great hack day work; sorry, I got on my usual "ux consistency" soapbox, so didn't mean for that to be demotivating.

Since this is something design consistently wants (i.e. hopefully is the new pattern), and sounds like it just never made it "to the top of a squad's feature backlog" (which is common for small-ish Beam updates), then implementing these updates on hack days is a perfect use of the carved-out hack day time imo.

...if you wanted to see if design was cool with just removing the old toggle filter, and just whole-sale cut over to the new checkbox filter, that would also be amazing, but happy to leave that to later/follow up discussions. Really not a big deal to leave the old one as-is.

PR #770, opened 2022-12-16
fix: update styles for required/helper text and selectField dropdown

roberthdg:
image

image

### DIFF
--- a/src/inputs/TextFieldBase.tsx
+++ b/src/inputs/TextFieldBase.tsx
@@ -269,16 +269,30 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
             </div>
           ),
         })}
+        {/* Compound fields will handle their own error and helper text. Do not show error or helper text when 'readOnly' or disabled */}
+        {labelStyle !== "left" && !compound && !inputProps.disabled && !inputProps.readOnly && (
+          <>
+            {errorMsg && !errorInTooltip && (
+              <ErrorMessage id={errorMessageId} errorMsg={errorMsg} hidden={hideErrorMessage} {...tid.errorMsg} />
+            )}
+            {helperText && <HelperText helperText={helperText} {...tid.helperText} />}
+          </>
+        )}
       </div>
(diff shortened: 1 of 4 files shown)


src/inputs/TextFieldBase.tsx
```diff
@@ -170,114 +170,129 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
...
             )}
-            {errorInTooltip && errorMsg && !hideErrorMessage && (
-              <span css={Css.df.aic.pl1.fs0.$}>
-                <Icon icon="error" color={Palette.Red600} tooltip={errorMsg} />
-              </span>
+            {helperText && <HelperText helperText={helperText} {...tid.helperText} />}
+          </>
+        )}
+      </div>
+      {/* alert and helper text for "left" labelStyle */}
```
roberthdg, line 282: There's probably a more elegant way of doing this, but even small changes to the current approach would result in several dom diffs in different places, so this is the _make sure we're don't break anything_ approach

src/inputs/TextFieldBase.tsx
```diff
@@ -170,114 +170,129 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
...
             )}
-            {errorInTooltip && errorMsg && !hideErrorMessage && (
-              <span css={Css.df.aic.pl1.fs0.$}>
-                <Icon icon="error" color={Palette.Red600} tooltip={errorMsg} />
-              </span>
+            {helperText && <HelperText helperText={helperText} {...tid.helperText} />}
+          </>
+        )}
+      </div>
+      {/* Error message and helper text for "left" labelStyle */}
```
roberthdg, line 282: There's probably a more elegant way of doing this, but even small changes to the current approach would result in several dom diffs in different places, so this is the _make sure we don't break anything_ approach

stephenh: Cool, I'm good with a little bit of copy/paste for simplicity.

stephenh (approved):

PR #772, opened 2022-12-17
feat: ReactNode as Button labels

TylerR909:
I started using `<ButtonModal />` to support the Bulk Template Management work and it was working amazingly right up until I had to implement the Cell UX and couldn't pass anything other than a String to the label.

image


stephenh (approved):
Cool! Kinda surprised only test ids need updated, i.e. I expected something in react-aria to want `label` to always be a string for accessibility purposes, but :shrug: .

PR #776, opened 2023-01-03
feat: Add MultiLineSelectField

roberthdg:
https://user-images.githubusercontent.com/24662091/210456406-06fe3f82-4bbb-493f-8d34-7369bedd1fa5.mov

### DIFF
--- a/src/inputs/MultiLineSelectField.tsx
+++ b/src/inputs/MultiLineSelectField.tsx
@@ -79,7 +79,7 @@ export function MultiLineSelectField<O, V extends Value>(
                   onSelect(selectedValues, selectedOptions);
                   setCurrentOptions(options.filter((o) => !selectedOptions.includes(o)));
                   // Display the input field if there are no selected values
-                  if (selectedValues.length === 0) setIsDisplayed(true);
+                  if (selectedOptions.length === 0) setIsDisplayed(true);
                 }}
               />
             </div>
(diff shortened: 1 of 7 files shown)


src/inputs/MultiLineSelectField.tsx
```diff
@@ -0,0 +1,118 @@
...
+              <Button
+                {...tid.deleteSelected}
+                variant="tertiary"
+                label={""}
+                aria-label={`Delete selected ${otherProps.label}`}
+                icon={"x"}
+                onClick={() => {
+                  // Delete the selected value from the array
+                  const [selectedValues, selectedOptions] = options
+                    .filter((o) => values.filter((v) => v !== value).includes(getOptionValue(o)))
```
stephenh, line 70: Suggestion: We're invoking `values.filter(v => v !== value)` inside of a `filter(o => ...)` but not using the `o`, which means we're re-doing the same `values.filter` over and over w/o needing to.

In general it seems like this could be simpler, like something like:

```
onClick={() => {
  const selectedOptions = options.filter(o => getOptionValue(o) !== value);
  const selectedValues = selectedOptions.map(o => getOptionValue(o));
  onSelect(selectedValues, selectedOptions);
```

Granted, we're invoking `getOptionValue` twice for each `o`, but I think the current code is also doing that, and :shrug: `getOptionValue` should be really cheap to invoke.

roberthdg: In this case  `const selectedOptions = options.filter(o => getOptionValue(o) !== value)` would make it select _all_ options instead of the current deleted, I updated it to take into account only selected values
```
const selectedOptions = options.filter( 
   (o) => values.includes(getOptionValue(o)) && getOptionValue(o) !== value,
);
```

stephenh: @roberthdg ah right! Good catch!

stephenh (approved):

PR #778, opened 2023-01-04
chore: lint, vscode fixes

TylerR909:
VSCode users weren't getting any linting or autoformatting help from ESLint because its setup & installation was incomplete. This finishes that installation as well as resolves some blocking lint errors:

1. most notably the invalid `{}` type which exposes a [huge type-safety hole](https://github.com/typescript-eslint/typescript-eslint/issues/2063#issuecomment-675156492) and should not be used. 
1. some rule-of-hooks violations where hooks were being conditionally called 
1. other --fix'able violations

### DIFF
--- a/src/components/IconButton.tsx
+++ b/src/components/IconButton.tsx
@@ -1,8 +1,9 @@
 import { AriaButtonProps } from "@react-types/button";
-import { RefObject, useMemo, useRef } from "react";
+import { RefObject, useMemo } from "react";
 import { useButton, useFocusRing, useHover } from "react-aria";
 import { Icon, IconProps, maybeTooltip, navLink, resolveTooltip } from "src/components";
 import { Css, Palette } from "src/Css";
+import { useGetRef } from "src/hooks/useGetRef";
 import { BeamButtonProps, BeamFocusableProps } from "src/interfaces";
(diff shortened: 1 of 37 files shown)


src/components/IconButton.tsx
```diff
@@ -45,7 +46,7 @@ export function IconButton(props: IconButtonProps) {
   const isDisabled = !!disabled;
   const ariaProps = { onPress, isDisabled, autoFocus, ...menuTriggerProps };
   // eslint-disable-next-line react-hooks/rules-of-hooks
```
stephenh, line 48: Nit could remove now

src/hooks/useGetRef.ts
```diff
@@ -0,0 +1,6 @@
+import { MutableRefObject, RefObject, useRef } from "react";
+
+export const useGetRef = <T extends HTMLElement>(maybeRef: RefObject<T> | undefined): MutableRefObject<T | null> => {
```
stephenh, line 3: Suggestion: Took me awhile to remember what this was for; maybe add an in-source comment like: `/** Used to sometimes accept a caller's ref, or if they did not pass one, use an internal one anyway. */`

stephenh (approved):
Nice!

PR #779, opened 2023-01-04
feat: Allow parent rows to be selected independent of child row selections

bdow:
When adding `GridDataRow.inferSelectedState: false` to a row, then it will no longer derived its selected state based on the state of its children rows. Instead the row must be selected specifically (or via its own parent) in order for it to be considered selected. This also removes the ability for the row to show in the 'indeterminate' state

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -2538,6 +2538,59 @@ describe("GridTable", () => {
       // Then the visible column session storage is defined using the `visibleColumnsStorageKey` prop
       expect(sessionStorage.setItem).toHaveBeenLastCalledWith("testStorageKey", '["name"]');
     });
+
+    it("respects setting inferSelectState to false", async () => {
+      // Given nested rows
+      const rows: GridDataRow<NestedRow>[] = [
+        simpleHeader,
+        {
+          // With one grand parent that sets `inferSelectedState: false`
+          ...{ kind: "parent", id: "p1", inferSelectedState: false, data: { name: "parent 1" } },
+          children: [
+            {
+              ...{ kind: "child", id: "p1c1", data: { name: "child 1" } },
+              children: [
+                { kind: "grandChild", id: "p1c1gc1", data: { name: "grandChild p1c1gc1" } },
+                { kind: "grandChild", id: "p1c1gc2", data: { name: "grandChild p1c1gc2" } },
+              ],
+            },
+          ],
+        },
+      ];
+      const api: MutableRefObject<GridTableApi<NestedRow> | undefined> = { current: undefined };
+      const r = await render(<TestFilterAndSelect api={api} rows={rows} />);
+
+      // When selecting a grand child of the grand parent that sets `inferSelectedState: false`
+      click(cellAnd(r, 3, 1, "select"));
+
+      // Then the header row should be indeterminate
+      expect(cellAnd(r, 0, 1, "select")).toBePartiallyChecked();
+      // And the grand parent row to not be checked.
+      expect(cellAnd(r, 1, 1, "select")).not.toBeChecked();
+      // And the parent row should show indeterminate,
+      expect(cellAnd(r, 2, 1, "select")).toBePartiallyChecked();
+      expect(api.current!.getSelectedRowIds()).toEqual(["p1c1gc1"]);
+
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.test.tsx
```diff
@@ -2538,6 +2538,59 @@ describe("GridTable", () => {
...
       expect(sessionStorage.setItem).toHaveBeenLastCalledWith("testStorageKey", '["name"]');
     });
+
+    it("respects setting inferSelectState to false", async () => {
+      // Given nested rows with two grand parents
+      const rows: GridDataRow<NestedRow>[] = [
+        simpleHeader,
+        {
+          // With one grand parent that sets `inferSelectedState: false`
+          ...{ kind: "parent", id: "p1", inferSelectedState: false, data: { name: "parent 1" } },
```
bdow, line 2548: @ipiedrahb - Once this is available, you'll want to just set `inferSelectedState: false` for the whole house template rows and you should be good to go!

ipiedrahb: Hi @bdow 
Awesome, thank you for the support

src/components/Table/utils/TableState.ts
```diff
@@ -375,15 +376,17 @@ export class TableState {
   }
 
   private getMatchedChildrenStates(children: GridDataRow<any>[], map: Map<string, SelectedState>): SelectedState[] {
-    return children
+    const respectedChildren = children.flatMap(getChildrenForDerivingSelectState);
```
stephenh, line 379: Question: Before this was just doing `children.map` ... but now we're doing `children.flatMap(something that gets children)`, so are we now effectively processing the grandchildren? 

Oh, I guess `getChildrenForDerivingSelectState` returns both the row itself and its children recursively...

Looks like it doesn't matter that we're now processing children + grandchildren vs. before was only doing children?

src/components/Table/components/Row.tsx
```diff
@@ -334,4 +334,6 @@ export type GridDataRow<R extends Kinded> = {
   initSelected?: boolean;
   /** Whether row can be selected */
   selectable?: false;
+  /** Whether this row should infer its selected state based on its children's selected state */
+  inferSelectedState?: false;
```
stephenh, line 338: Thought: I was kind of wondering if this should be a table-level setting instead of setting it onsey/twosy on grid rows, but I suppose it is more flexible, like you could have some parent rows with it false, some with it true... :shrug:

stephenh (approved):

src/components/Table/utils/TableState.ts
bdow: Right, getChildrenForDerivingSelectState returns both the row itself and its children _only_ if the row has `inferSelectedState === false`. Otherwise, we only return the row. So basically peering deeper into the nesting only if necessary.

stephenh: Cool, and just checking, this line 378 was previously _not_ getting deep results, but it is now, and that's fine / no big deal? Probably not if things are passing tests/etc., but just checking.

bdow: Correct. It was _not_ getting deep results previously. But now it does get deep results _only_ if necessary.

src/components/Table/GridTable.test.tsx
```diff
@@ -2538,6 +2538,59 @@ describe("GridTable", () => {
...
+      // Then the header row should be indeterminate
+      expect(cellAnd(r, 0, 1, "select")).toBePartiallyChecked();
+      // And the grand parent row to not be checked.
+      expect(cellAnd(r, 1, 1, "select")).not.toBeChecked();
+      // And the parent row should show indeterminate,
+      expect(cellAnd(r, 2, 1, "select")).toBePartiallyChecked();
+      expect(api.current!.getSelectedRowIds()).toEqual(["p1c1gc1"]);
+
+      // When selecting the grand parent
+      click(cellAnd(r, 1, 1, "select"));
```
stephenh, line 2575: Question: Ah okay, so just curios/clarifying: clicking the GP still "pushes down" the new state to its children/grand children, it's just the "deriving back up" state that we've snipped?

Good tests!

bdow: Yup! That is correct.

PR #780, opened 2023-01-05
fix(grid-table-expanded-cols): persist between rerenders

TylerR909:
For Bulk Template Management, whenever we touch a cell (delete, save) and trigger a new `createColumns()` call, that trickles through GridTable and TableApi and resets/collapses any expanded columns we had.

By the nature of our table, users will be updating items across the row in quick succession, so having to re-expand a column each time they touch it or its children will get tedious.

Commenting out 220/221 appears to fix it but would likely break the initExpanded feature. So I've scoped it to only run when initially setting the columns. This likely means that if a developer is _toggling_ `initExpanded` in _their_ createColumns for some reason, that likely won't have any effect.

### DIFF
--- a/src/components/Table/utils/TableState.ts
+++ b/src/components/Table/utils/TableState.ts
@@ -212,12 +212,13 @@ export class TableState {
   }
 
   setColumns(columns: GridColumnWithId<any>[], visibleColumnsStorageKey: string | undefined): void {
+    const isInitial = !this.columns || this.columns.length === 0;
     if (columns !== this.columns) {
       this.columns = columns;
       this.visibleColumnsStorageKey = visibleColumnsStorageKey ?? camelCase(columns.map((c) => c.id).join());
       this.visibleColumns.replace(readOrSetLocalVisibleColumnState(columns, this.visibleColumnsStorageKey));
       const expandedColumnIds = columns.filter((c) => c.initExpanded).map((c) => c.id);
-      this.expandedColumns.replace(expandedColumnIds);
+      if (isInitial) this.expandedColumns.replace(expandedColumnIds);
     }
   }
 
--- a/src/components/Table/utils/TableState.test.ts
+++ b/src/components/Table/utils/TableState.test.ts
@@ -2,6 +2,22 @@ import { deriveSortState, TableState } from "src/components/Table/utils/TableSta
 import { ASC, DESC } from "src/components/Table/utils/utils";
 
 describe(TableState, () => {
+  it("preserves expanded state", () => {
+    // Given some column definitions and initial state
+    const state = new TableState();
+    const columns = [{ id: "one" }, { id: "two" }];
+    state.setColumns(columns, undefined);
+
+    // When we toggle a column
+    state.toggleExpandedColumn("one");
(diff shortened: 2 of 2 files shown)


src/components/Table/utils/TableState.ts
```diff
@@ -212,12 +212,13 @@ export class TableState {
...
 
   setColumns(columns: GridColumnWithId<any>[], visibleColumnsStorageKey: string | undefined): void {
+    const isInitial = !this.columns || this.columns.length === 0;
     if (columns !== this.columns) {
       this.columns = columns;
       this.visibleColumnsStorageKey = visibleColumnsStorageKey ?? camelCase(columns.map((c) => c.id).join());
       this.visibleColumns.replace(readOrSetLocalVisibleColumnState(columns, this.visibleColumnsStorageKey));
       const expandedColumnIds = columns.filter((c) => c.initExpanded).map((c) => c.id);
-      this.expandedColumns.replace(expandedColumnIds);
+      if (isInitial) this.expandedColumns.replace(expandedColumnIds);
```
bdow, line 221: Makes sense to me... don't blow away the existing `expandedColumns` set when updating, and instead keep the expanded states per what the table already has defined... I can't think of a reason why this would break anything 🤔

bdow (approved):

src/components/Table/utils/TableState.ts
stephenh: I agree, great find @TylerR909 for such a targeted/small slice point to handle this, I would have thought it'd be more work.

Fwiw wouldn't mind a test just b/c I could see this pretty easily regressing (aka me breaking it :-P :-)) if/when we refactor TableState for whatever reasons in the future.

stephenh (approved):
Per re-calling `createColumns` after every cell edit, it makes me curious if we could find a way to do finer-grained re-rendering, i.e. have the columns array stable across any render that doesn't literally change the columns being shown on the screen (and so get the could-be-mutated-after-edits into the cells some other way than the columns array, like a context or store).

But I also wouldn't be surprised if that with virtualization it's actually not a huge deal/noticeable re-render; I think we've been surprised at how much "just virtualize and only render the rows on the screen" has mitigated perf issues that otherwise we had to try really hard to avoid re-renders but really it was b/c we were re-rendering every single row in the table.

bdow:
> Per re-calling `createColumns` after every cell edit...

I was also curious about this. I would think, if anything, our `columns` should be stable, while the rows and data would be what gets updated. Without getting too much into the weeds of it, is there a reason why we need to rebuild the `columns`?

TylerR909:
> I was also curious about this. I would think, if anything, our columns should be stable, while the rows and data would be what gets updated. Without getting too much into the weeds of it, is there a reason why we need to rebuild the columns?

Because the entire table is a giant list of ITIs that we manually have to group up into Parent/Child templates, and the Child/Option templates are what expand/collapse onto the Parent. If they click an "Add Item" Cell, we must regenerate columns. If they click the "+Add Item" button on top then we have to regenerate rows. If they click the "[ ] Spec Level Dependent" box then X items update and X*2 or X*3 items are created/added to the table (Where again we need to recalc columns).

Basically, this entire table is dynamically generated.

PR #782, opened 2023-01-06
fix: [sc-27302] Multiselect filter correctly disables chips in modal;

apattersonATX-HB:
Multiselect filter will now correctly disable options when rendered as a ToggleChipGroup.

### DIFF
--- a/src/inputs/ToggleChipGroup.tsx
+++ b/src/inputs/ToggleChipGroup.tsx
@@ -1,6 +1,7 @@
 import { useRef } from "react";
 import { useCheckboxGroup, useCheckboxGroupItem, useFocusRing, VisuallyHidden } from "react-aria";
 import { CheckboxGroupState, useCheckboxGroupState } from "react-stately";
+import { maybeTooltip } from "src/components";
 import { Label } from "src/components/Label";
 import { Css } from "src/Css";
 import { useTestIds } from "src/utils/useTestIds";
@@ -8,6 +9,8 @@ import { useTestIds } from "src/utils/useTestIds";
 type ToggleChipItemProps = {
   label: string;
   value: string;
+  isDisabled?: boolean;
+  disabledReason?: string;
 };
 
 export interface ToggleChipGroupProps {
@@ -35,6 +38,8 @@ export function ToggleChipGroup(props: ToggleChipGroupProps) {
             groupState={state}
             selected={state.value.includes(o.value)}
             label={o.label}
+            isDisabled={o.isDisabled}
+            disabledReason={o.disabledReason}
             {...tid[o.value]}
           />
         ))}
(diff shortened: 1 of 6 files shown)


src/inputs/ToggleChipGroup.tsx
```diff
@@ -48,33 +53,41 @@ interface ToggleChipProps {
...
+    children: (
+      <label
+        css={{
+          ...Css.relative.dib.br16.sm.px1.cursorPointer.pyPx(4).bgGray200.if(isDisabled).cursorNotAllowed.gray600.pr1.$,
+          ...(isSelected
+            ? {
+                ...Css.white.bgLightBlue700.$,
+                ":hover:not(:isDisabled)": Css.bgLightBlue800.$,
+              }
+            : { ":hover:not(:isDisabled)": Css.bgGray300.$ }),
```
stephenh, line 78: Question: is `:isDisabled` correct? I've seen `:disabled`:

https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled

But not `:isDisabled`. Does this somehow magically work with the `data-disabled` attribute? 

I was expecting per:

https://stackoverflow.com/a/17249001

Something like `:not([data-disabled="true"])`

src/inputs/ToggleChipGroup.tsx
```diff
@@ -1,13 +1,16 @@
...
+import { maybeTooltip } from "src/components";
 import { Label } from "src/components/Label";
 import { Css } from "src/Css";
 import { useTestIds } from "src/utils/useTestIds";
 
 type ToggleChipItemProps = {
   label: string;
   value: string;
+  isDisabled?: boolean;
+  disabledReason?: string;
```
stephenh, line 13: Suggestion: Instead of two separate props, we typically model this as a union, like:

```
export interface BeamButtonProps {
  /**
   * Whether the interactive element is disabled.
   *
   * If a ReactNode, it's treated as a "disabled reason" that's shown in a tooltip.
   */
  disabled?: boolean | ReactNode;
```

That way the user can't do something like `isDisabled=false disabledReason="some error"` and it's ambiguous that maybe we really be disabled, like does a set `disabledReason` override the `isDisabled` is false?

src/inputs/ToggleChipGroup.tsx
apattersonATX-HB: I inserted `:isDisabled` because react-aria's `useCheckboxGroupItem` props extends InputBase  type down the chain which does have the `isDisabled` prop. It'll actually work without using `data-disabled` attribute so maybe it's just a quirk of react-aria?

But yea probably better to do it the traditional CSS way.

src/inputs/ToggleChipGroup.tsx
apattersonATX-HB: Thats true. Had tunnel vision on the filter so hadn't considered `ToggleChipGroup` ever using anything besides a string

stephenh (approved):
Cool, looks good!

PR #786, opened 2023-01-12
feat: Support selected state in ButtonMenu items

bdow:


src/components/internal/Menu.tsx
```diff
@@ -15,10 +15,12 @@ interface MenuProps<T> {
   searchable?: boolean;
   persistentItems?: MenuItem[];
   contrast: boolean;
+  selectedItem: string | undefined;
+  onChange: ((key: string) => void) | undefined;
```
0ces, line 19: **question**: Is there any special reason to write this this way and not something like
```suggestion
  selectedItem?: string;
  onChange?: (key: string) => void;
```

0ces (approved):
Thanks!

src/components/internal/Menu.tsx
bdow: The reason is to force the caller to pass those variables. It is basically saying that this is **not** an optional property, but you can pass undefined.

PR #787, opened 2023-01-12
fix: SelectField to handle boolean values correctly

bdow:


ipiedrahb (approved):
@bdow looks great!

src/inputs/Value.ts
```diff
@@ -26,7 +26,8 @@ export function keyToValue<V extends Value>(key: Key): V {
     } else if (key === "__VALUE:undefined") {
       return undefined as V;
     } else if (key.startsWith("__VALUE:boolean:")) {
-      return (key.split(":")[2] === "true" ? true : false) as V;
+      // boolean values end with `;` (i.e. '__VALUE:boolean:true;').
+      return key.split(":")[2].startsWith("true") as V;
```
stephenh, line 30: Thought: Lol, ah I see! Took me awhile to see in-source the comment "oh damn there is a `;` there"... :-)

Fwiw I wonder if that is the real bug, i.e. none of the the other `__VALUE`s have a trailing `;`:

```
  } else if (typeof value === "number") {
    // Despite using the Key type, react-aria's select manager always returns strings,
    // so tag this value as really being a number.
    return `__VALUE:number:${value}`;
  } else if (typeof value === "boolean") {
    return `__VALUE:boolean:${value};`;
  } else if (value === null) {
    return "__VALUE:null";
  } else if (value === undefined) {
    return "__VALUE:undefined";
```

bdow: ahh shoot. I thought React-Aria was delivering the `__VALUE:boolean:${value};` stuff for us... 🤦

bdow: I might follow up with a `chore` PR to address at the root.

bdow: https://github.com/homebound-team/beam/pull/788

PR #789, opened 2023-01-13
Sc 24855/persist collapse for columns

kbesingeryeh:


src/components/Table/GridTable.test.tsx
```diff
@@ -2604,6 +2604,94 @@ describe("GridTable", () => {
...
+      function Test() {
+        const _api = useGridTableApi<ExpandableRow>();
+        api.current = _api;
+        return <GridTable api={_api} columns={columns} rows={rows} />;
+      }
+
+      // When rendering the table
+      const r = await render(<Test />);
+
+      // Then the column is initially hidden - this is not hiding it
```
kbesingeryeh, line 2686: Not sure if I have my test set up correctly, but I have tried to test that the column is hidden, but its showing in the DOM

src/components/Table/utils/TableState.ts
```diff
@@ -221,8 +221,31 @@ export class TableState {
...
+        // looks at initExpanded and reads localStorage
+        if (isInitial && c.initExpanded) {
+          expandedColumnIds.push(c.id);
+          expandedColumnIds.concat(localStorageColumns!);
+          this.expandedColumns.replace(expandedColumnIds);
+        } else {
+          // subsequent load, ignoring initExpanded
+          // if its a new column or existing column then create new expanded column ids
+          // new expanded column ids merge with local storage
+          const newExpandedColumnIds = this.columns.filter((col) => col.initExpanded).map((c) => c.id);
```
kbesingeryeh, line 238: I don't think I understand this 100%.. I searched through `this.columns` and cannot find anything that would help me determine if the column is new or existing..

bdow: What you'd want to do here is not yet set `this.columns = columns;`. Do that as like a last step. Then you should be able to look at `columns` and see which of the columns do not yet exist in `this.columns` in order to determine which are new.

src/components/Table/GridTable.test.tsx
```diff
@@ -2604,6 +2604,145 @@ describe("GridTable", () => {
...
+      // When setting the column to be visible
+      api.current?.setVisibleColumns(api.current.getVisibleColumnIds().concat("myColumn2"));
+      expect(row(r, 1).childNodes).toHaveLength(2);
+      // wait for promise to resolve
+      await wait();
+
+      // then expect 2 columns + 2 expandable columns to be visible
+      expect(row(r, 1).childNodes).toHaveLength(4);
+    });
   });
```
bdow, line 2746: **praise:** Great tests! Lazy ask, do you know if we already have a test for `expandColumns` that'd be like:

```
it("ignores initExpanded for existing columns on rerender, async() => {
  // Given `columnA` that is initially expanded
  // When collapsing `columnA`
  // And when then triggering new `columnB` to be introduced.
  // Then the `columnA` remains collapsed
});
```

src/components/Table/utils/TableState.ts
```diff
@@ -218,25 +218,61 @@ export class TableState {
...
+      ).then(() => {
+        const newIds = newExpanded.map((c) => c.id);
+        // if there is a difference between list of current expanded columns vs list we just created, then replace
+
+        if (isInitial || isDifferent) {
+          this.expandedColumns.replace(newIds);
+        }
+        //  update our persistCollapse if set
+        if (this.persistCollapse) {
+          sessionStorage.setItem(`expandedColumn_${this.persistCollapse}`, JSON.stringify(newIds));
```
bdow, line 256: **suggestion:** We generate this string four times in this file, so I think it'd be worth making some helpers here. Maybe something along the lines of:
```
sessionStorage.setItem(getColumnStorageKey(this.persistCollapse), JSON.stringify(newIds));

...

function getColumnStorageKey(storageKey: string): string {
  return `expandedColumn_${storageKey}`;
}
```

src/components/Table/utils/TableState.ts
```diff
@@ -218,25 +218,61 @@ export class TableState {
...
-      if (isInitial) this.expandedColumns.replace(expandedColumnIds);
+      // list of local storage columns
+      const localStorageColumns = this.persistCollapse ? readExpandedColumnsStorage(this.persistCollapse) : [];
+      // expandedColumnIds initialized as localStorageColumns
+      const existingExpandedColumnIds = isInitial ? localStorageColumns : this.expandedColumnIds;
+
+      // list of all existing columns
+      const existingColumnIds = this.columns.map((c) => c.id);
+
+      // filter existing columns that are initExpanded to see which are new
```
bdow, line 231: **nit:** The comment here is misleading. We are filtering new columns, not the existing. Maybe something like:
```suggestion
      // filter to find new columns that should be initially expanded
```

src/components/Table/utils/TableState.ts
```diff
@@ -218,25 +218,61 @@ export class TableState {
...
       const result = await column.expandColumns();
       // once we have the loaded columns, add result to local cache
       this.loadedColumns.set(column.id, assignDefaultColumnIds(result));
     }
-    // once column is in local cache, then toggle
-    this.toggleExpandedColumn(column.id);
+    if (column.initExpanded) {
+      // manually toggle expanded column
+      this.toggleExpandedColumn(column.id);
+    }
```
bdow, line 275: **thought/question:** Above there is a code change & comment that says: 
```
await tableState.loadExpandedColumns(column);
// manually calling this as loadExpandedColumns does not toggle
tableState.toggleExpandedColumn(column.id);
```
But it looks like `loadExpandedColumns` does call `toggle` for a column that is `initExpanded`. 
I'm thinking we probably do not want this `toggleExpandedColumn` within `loadExpandedColumns` like your comment suggests. Otherwise for columns that are `initExpanded`, when a user clicks the ExpandableHeader component, then I'm pretty sure we'd call `toggleExpandedColumn` twice. Once in `ExpandableHeader` and another time here. I'd think that'd result not being able to change the collapsed state of the column.

Which code path was this logic created for?

bdow (changes requested):
Looks good. Just a couple questions and potentially a bug with making multiple `toggleExpandedColumn` calls

src/components/Table/utils/TableState.ts
kbesingeryeh: got it

src/components/Table/GridTable.test.tsx
kbesingeryeh: I actually think this should be in a `tableState` test

src/components/Table/utils/TableState.ts
bdow: Gotcha. I feel like there is a way to simplify the logic here. Will think about it/play around with it.

src/components/Table/components/ExpandableHeader.tsx
```diff
@@ -35,6 +35,8 @@ export function ExpandableHeader<R extends Kinded>(props: ExpandableHeaderProps<
         if (isFunction(column.expandColumns)) {
           setIsLoading(true);
           await tableState.loadExpandedColumns(column);
+          // manually calling this as loadExpandedColumns does not toggle
+          tableState.toggleExpandedColumn(column.id);
           setIsLoading(false);
         } else {
           tableState.toggleExpandedColumn(column.id);
```
bdow, line 42: **nit:** Could do something like:
```
        if (isFunction(column.expandColumns)) {
          setIsLoading(true);
          await tableState.loadExpandedColumns(column);          
          setIsLoading(false);
        }
        tableState.toggleExpandedColumn(column.id);
```
De-dupes the two `toggleExpandedColumn` calls. Instead of having one for both conditionals, now we just have one.

src/components/Table/utils/TableState.ts
```diff
@@ -218,25 +218,61 @@ export class TableState {
...
+      }
+
+      const newExpanded = columns.filter((c) => existingExpandedColumnIds.includes(c.id));
+      // if we have existing expanded columns ids, then do a Promise.all() and execute column.expandColumns
+      Promise.all(
+        newExpanded.map(async (c) => {
+          // if the column is initExpanded, manually toggle the column
+          if (c.initExpanded) {
+            this.toggleExpandedColumn(c.id);
+          }
```
bdow, line 248: **thought:** A couple thoughts here:
1. Instead of calling the `toggle`, I think we should be more explicit by doing a `this.expandedColumns.add(c.id)`
2. Though, I also don't think we want to make this call multiple times. Instead we should do a single `replace`... which I think is already happening in the `then` clause... 🤔

src/components/Table/utils/TableState.ts
```diff
@@ -245,6 +281,32 @@ export class TableState {
...
+      const newlyVisibleIds = ids.filter((id) => !this.visibleColumnIds.includes(id));
+      // look through columns where we match ids to the new ids
+      const newExpanded = this.columns.filter((c) => newlyVisibleIds.includes(c.id) && c.initExpanded);
+      // here do a Promise.all() and execute column.expandColumns if necessary.
+      Promise.all(
+        newExpanded.map(async (c) => {
+          // if the column is initExpanded, manually toggle the column
+          if (c.initExpanded) {
+            this.toggleExpandedColumn(c.id);
+          }
```
bdow, line 296: **thought:** Same thoughts as above... I'm kind of surprised we need to do this. Also, I believe every `column` that is in the `newExpanded` array would have `initExpanded: true` due to the filter before. So I think this conditional is unnecessary.

bdow (commented):
Looks good. I'm just surprised by a couple things. I'm going to pull this down locally and see what is going on.

bdow (approved):

PR #790, opened 2023-01-13
fix: Listbox to only take 50% width when labelStyle is left on SelectFields

bdow:


src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -72,6 +73,7 @@ export function SelectFieldBase<O, V extends Value>(props: BeamSelectFieldBasePr
     unsetLabel,
     ...otherProps
   } = props;
+  const labelStyle = fieldProps?.labelStyle ?? "above";
```
stephenh, line 76: Thought: Do we need to check `otherProps.labelStyle`?

stephenh (approved):

src/inputs/internal/SelectFieldBase.tsx
bdow: oh yea,... we probably should... 👍

roberthdg (approved):
cool, thanks!

PR #791, opened 2023-01-19
feat: add disabled flag and tooltip to radio group field options (sc-25841)

Andrielson:
image

Changes in RadioGroupField include:

- add optional `disabled` attribute to `RadioFieldOption`
- add optional `disabledTooltip` attribute to `RadioFieldOption`

### DIFF
--- a/src/inputs/RadioGroupField.test.tsx
+++ b/src/inputs/RadioGroupField.test.tsx
@@ -16,4 +16,37 @@ describe("RadioGroupField", () => {
     );
     click(r.favoriteCheese_a);
   });
+
+  it("should disable only first option", async () => {
+    const r = await render(
+      <RadioGroupField
+        label={"Favorite cheese"}
+        value="a"
+        onChange={() => {}}
+        options={[
+          { value: "a", label: "Asiago", disabled: true },
+          { value: "b", label: "Burratta" },
+        ]}
+      />,
+    );
+    const radioInput = r.container.querySelector(`[data-testid="favoriteCheese_a"]`)!;
+    expect(radioInput).toBeDisabled();
+  });
+
+  it("should disable first option and have a tooltip", async () => {
+    const r = await render(
+      <RadioGroupField
+        label={"Favorite cheese"}
+        value="a"
+        onChange={() => {}}
+        options={[
+          { value: "a", label: "Asiago", disabled: "some reason" },
+          { value: "b", label: "Burratta" },
+        ]}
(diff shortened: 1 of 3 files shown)


src/inputs/RadioGroupField.tsx
```diff
@@ -19,6 +20,10 @@ export interface RadioFieldOption<K extends string> {
   description?: string | (() => ReactNode);
   /** The undisplayed value, i.e. an id of some sort. */
   value: K;
+  /** Disable only specific options */
+  disabled?: boolean;
```
stephenh, line 24: Suggestion: Usually we have a single `disabled` that uses a union type:

```
disabled?: boolean | ReactNode
```

Just avoid things like `disabled=false disabledTooltip=some message` being ambiguous.

src/inputs/RadioGroupField.test.tsx
```diff
@@ -16,4 +16,37 @@ describe("RadioGroupField", () => {
...
+        label={"Favorite cheese"}
+        value="a"
+        onChange={() => {}}
+        options={[
+          { value: "a", label: "Asiago", disabled: true },
+          { value: "b", label: "Burratta" },
+        ]}
+      />,
+    );
+    const radioInput = r.container.querySelector(`[data-testid="favoriteCheese_a"]`)!;
```
stephenh, line 32: Suggestion: Given how our `render` DSL works, this can probably be just:

```
expect(r.favoriteCheese_a).toBeDisabled();
```

I think?

stephenh (approved):
Looks great, thanks!

src/inputs/RadioGroupField.test.tsx
Andrielson: I've tried this approach, but it returns an error:

```
TypeError: Cannot read property 'message' of undefined

    at new JestAssertionError (~/homebound/beam/node_modules/jest-chain/dist/chain.js:10:18)
    at Object.newMatcher [as toBeDisabled] (~/homebound/beam/node_modules/jest-chain/dist/chain.js:31:17)
    at Object.<anonymous> (~/homebound/beam/src/inputs/RadioGroupField.test.tsx:32:31)
```

stephenh: Ah yeah, np that this is merged, but I forgot the `()`, so:

```
expect(r.favoriteCheese_a()).toBeDisabled()
```

PR #792, opened 2023-01-21
fix: dropdown label value defaults

TylerR909:
In BluePrint I find that this is very tedious...

```jsx
<SelectField
  options={options}
  getOptionValue={(o) => o.id}
  getOptionLabel={(o) => o.name}
/>
```

...given that it's `id/value` or `displayName/label/name/value` seemingly 95% of the time. So, let's make getOptionValue/Label optional and try to infer, while still easily letting developers override or provide missing behavior when necessary.

Also, I tried to type-guard this (like if you pass in `{ code: string; id: string }[]` it'd tell you getOptionLabel is required), but our abundant use of Interfaces over Types makes that a significant lift. Interfaces may only extend other Interfaces with statically-known types, so a key that's _conditionally_ required or optional breaks extensions and a few overloads.

### DIFF
--- a/src/inputs/internal/SelectFieldBase.tsx
+++ b/src/inputs/internal/SelectFieldBase.tsx
@@ -487,5 +487,5 @@ export const defaultGetOptionLabel = (o: InferrableOptionLabel): string =>
   o?.name ||
   o?.value ||
   (() => {
-    throw new Error("Unable to determine option label. Please provide a custom getOptionValue function");
+    throw new Error("Unable to determine option label. Please provide a custom getOptionLabel function");
   })();
--- a/src/inputs/internal/SelectFieldBase.tsx
+++ b/src/inputs/internal/SelectFieldBase.tsx
@@ -15,8 +15,8 @@ import { areArraysEqual } from "src/utils";
 export interface BeamSelectFieldBaseProps<O, V extends Value> extends BeamFocusableProps, PresentationFieldProps {
   /** Renders `opt` in the dropdown menu, defaults to the `getOptionLabel` prop. `isUnsetOpt` is only defined for single SelectField */
   getOptionMenuLabel?: (opt: O, isUnsetOpt?: boolean) => string | ReactNode;
-  getOptionValue: (opt: O) => V;
-  getOptionLabel: (opt: O) => string;
+  getOptionValue?: (opt: O) => V;
+  getOptionLabel?: (opt: O) => string;
   /** The current value; it can be `undefined`, even if `V` cannot be. */
   values: V[] | undefined;
   onSelect: (values: V[], opts: O[]) => void;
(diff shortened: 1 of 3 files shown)


src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -15,8 +15,10 @@ import { areArraysEqual } from "src/utils";
 export interface BeamSelectFieldBaseProps<O, V extends Value> extends BeamFocusableProps, PresentationFieldProps {
   /** Renders `opt` in the dropdown menu, defaults to the `getOptionLabel` prop. `isUnsetOpt` is only defined for single SelectField */
   getOptionMenuLabel?: (opt: O, isUnsetOpt?: boolean) => string | ReactNode;
-  getOptionValue: (opt: O) => V;
-  getOptionLabel: (opt: O) => string;
+  /** Defaults to `id ?? value` */
+  getOptionValue?: (opt: O) => V;
+  /** Defaults to `displayName || label || name || value` */
+  getOptionLabel?: (opt: O) => string;
```
stephenh, line 21: Thought: Some of this should already be handled by the `SelectField` overload:

```
export function SelectField<O, V extends Value>(props: SelectFieldProps<O, V>): JSX.Element;
export function SelectField<O extends HasIdAndName<V>, V extends Value>(
  props: Optional<SelectFieldProps<O, V>, "getOptionValue" | "getOptionLabel">,
): JSX.Element;
export function SelectField<O, V extends Value>(
  props: Optional<SelectFieldProps<O, V>, "getOptionLabel" | "getOptionValue">,
```

I.e. it conditionally `Optionalize`s `getOptionValue` and `getOptionLabel` only when `O` has an id and name.

Granted, that was very inflexible, but it is more type-safe; can you try working the `AtLeastOne` trick as a replacement for `O extends HasIdAndName`, so that ideally we get both the increased flexibility of "use any one of these N fields" while not having to fallback on runtime checks for callers not passing the properties when they should?

(I had tried making this more flexible in the past, but didn't have the `AtLeastOne` trick, so it led to ~an egregious amount of overloads, like 1 overload per guessed label, 1 overload per guessed value, and so just gave up.)

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -457,3 +461,31 @@ export function disabledOptionToKeyedTuple(
     return [valueToKey(disabledOption), undefined];
   }
 }
+
+type AtLeastOne<T, U = { [K in keyof T]: Pick<T, K> }> = Partial<T> & U[keyof U];
+
+type InferrableOptionValue = AtLeastOne<{
+  id: any;
+  value: any;
```
stephenh, line 469: Thought: granted, this is not a public API, but would prefer using `Value` instead of `any`.

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -457,3 +461,31 @@ export function disabledOptionToKeyedTuple(
...
+  o?.value ??
+  (() => {
+    throw new Error("Unable to determine option value. Please provide a custom getOptionValue function");
+  })();
+
+type InferrableOptionLabel = AtLeastOne<{
+  displayName: any;
+  name: any;
+  label: any;
+  value: any;
```
stephenh, line 482: Thought: Ideally these would be `string` ... or `string | ReactNode`?

src/inputs/internal/SelectFieldBase.tsx
```diff
@@ -457,3 +461,31 @@ export function disabledOptionToKeyedTuple(
...
+}>;
+export const defaultGetOptionLabel = (o: InferrableOptionLabel): string =>
+  o?.displayName ||
+  o?.label ||
+  o?.name ||
+  o?.value ||
+  (() => {
+    throw new Error("Unable to determine option label. Please provide a custom getOptionLabel function");
+  })();
```
stephenh, line 491: Suggestion: Surprised we don't have `fail` copy/pasted in, but seems worth it?

PR #793, opened 2023-01-24
feat: allow to always show helper text

0ces:
Allow number field to always show the helper text if the prop `alwaysShowHelperText` is passed.

### DIFF
--- a/src/inputs/NumberField.tsx
+++ b/src/inputs/NumberField.tsx
@@ -48,7 +48,7 @@ export interface NumberFieldProps extends Pick<PresentationFieldProps, "labelSty
   // Typically used for compact fields in a table. Removes border and uses an box-shadow for focus behavior
   borderless?: boolean;
   sizeToContent?: boolean;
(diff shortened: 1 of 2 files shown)


src/inputs/NumberField.tsx
```diff
@@ -48,6 +48,8 @@ export interface NumberFieldProps extends Pick<PresentationFieldProps, "labelSty
   // Typically used for compact fields in a table. Removes border and uses an box-shadow for focus behavior
   borderless?: boolean;
   sizeToContent?: boolean;
+  // If set, the helper text will always be shown
+  alwaysShowHelperText?: boolean;
```
stephenh, line 52: Suggestion: Can you add "(usually we hide the helper text if read only)" to the comment, otherwise I'll forget when this is necessary. :-)

stephenh (approved):

src/inputs/NumberField.tsx
0ces: Sure

PR #796, opened 2023-02-01
feat: Add tooltip support to GridTable

bdow:
Updated the Tooltip component's wrapping element to use 'display: contents;'. This allows for styles to flow through the element like it wasn't there, such as flex item positioning.

### DIFF
--- a/src/components/Table/components/cell.tsx
+++ b/src/components/Table/components/cell.tsx
@@ -4,6 +4,7 @@ import { navLink } from "src/components/CssReset";
 import { GridTableApi } from "src/components/Table/GridTableApi";
 import { RowStyle, tableRowStyles } from "src/components/Table/TableStyles";
 import { GridCellAlignment, GridColumnWithId, Kinded, MaybeFn, RenderAs } from "src/components/Table/types";
+import { maybeTooltip } from "src/components/Tooltip";
 import { Css, Properties, Typography } from "src/Css";
 
 /**
@@ -30,6 +31,8 @@ export type GridCellContent = {
   css?: Properties;
   /** Allows cell to reveal content when the user hovers over a row. Content must be wrapped in an element in order to be hidden. IE <div>{value}</div>*/
   revealOnRowHover?: true;
+  /** Tooltip to add to a cell. **Only Works on Headers** */
+  tooltip?: ReactNode;
 };
 
 /** Allows rendering a specific cell. */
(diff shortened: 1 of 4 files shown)


src/components/Tooltip.tsx
```diff
@@ -35,6 +35,8 @@ export function Tooltip(props: TooltipProps) {
         {...triggerProps}
         {...(!state.isOpen && typeof title === "string" ? { title } : {})}
         {...tid}
+        // Add display contents to prevent the tooltip wrapping element from short-circuiting inherited styles (i.e. flex item positioning)
+        css={Css.display("contents").$}
```
blambillotte, line 39: cool! Wasn't aware of that property so checked caniuse on its support: https://caniuse.com/css-display-contents

Maybe worth giving Safari and and Firebox a quick test given the notes?

src/components/Tooltip.tsx
```diff
@@ -61,16 +63,20 @@ export function Tooltip(props: TooltipProps) {
...
   placement?: Placement;
 }
 
 function Popper({ triggerRef, content, placement = "auto" }: PopperProps) {
   const popperRef = useRef(null);
   const [arrowRef, setArrowRef] = useState<HTMLDivElement | null>(null);
+  // Since we use `display: contents;` on the `triggerRef`, then the element.offsetTop/Left/etc all equal `0`. This would make
+  // the tooltip show in the top left of the document. So instead, we target either the first child, if available, or the parent element as the tooltip target.
+  // It is possible there are no children if the element only has text content, which is the reasoning for the parentElement fallback.
+  const targetElement = triggerRef.current ? triggerRef.current.children[0] ?? triggerRef.current.parentElement : null;
```
blambillotte, line 77: nit: maybe a nice use of `firstChild` instead of `children[0]` ?

blambillotte (approved):
Nice! Small question on the CSS property `display: content` browser support, my assumption is Safari is a big deal for iPads 🤔

src/components/Tooltip.tsx
bdow: Great suggestion. Will do!

bdow: I've checked the other browsers, all works fine. For iPads, we need a new pattern for touch devices and tooltips. Right now we show tooltips on hover, but there is no "hover" in touch devices. We'd need an explicit trigger element (like an icon) that when you click it, then it displays the tooltip. Will bring this up with design, but I'm pretty sure they are aware of this issue already - might just not be top of mind.

src/components/Table/components/cell.tsx
```diff
@@ -30,6 +31,8 @@ export type GridCellContent = {
   css?: Properties;
   /** Allows cell to reveal content when the user hovers over a row. Content must be wrapped in an element in order to be hidden. IE <div>{value}</div>*/
   revealOnRowHover?: true;
+  /** Tooltip to add to a cell. **Only Works on Headers** */
```
stephenh, line 34: Question: Is the "only works on headers" still true? I thought scanning the story I saw tooltip on non-header rows.

bdow: You're right. It works everywhere. I was going to just start on the headers, but seemed just as easy to do it everywhere once I thought of using `display: contents` in Tooltip.

stephenh: Sweet!

PR #798, opened 2023-02-02
fix: Update deps array to fix issue with expandedColumns not reflecting correct columns size

bdow:


src/components/Table/hooks/useSetupColumnSizes.ts
```diff
@@ -46,7 +46,7 @@ export function useSetupColumnSizes(
       setTableWidth(width);
       setColumnSizes(calcColumnSizes(columns, width, style.minWidthPx, expandedColumnIds));
     },
-    [setTableWidth, setColumnSizes, columns, style],
+    [setTableWidth, setColumnSizes, columns, style, expandedColumnIds],
```
stephenh, line 49: Fwiw we have a lot of missing deps; webstorm catches them and I fix here and there but we really need that eslint rule.

stephenh (approved):
Cool, makes sense

PR #802, opened 2023-02-03
fix: nothingSelectedText for singleFilter

0ces:
I want this `nothingSelectedText` to be able to work for multiple languages.


bdow (approved):

TylerR909:
Would've been nice to pick up MultiSelectFilter on this too because I'm literally about to open the same exact PR.

PR #803, opened 2023-02-03
fix: use nothingSelectedText on MultiSelectField

TylerR909:
BulkTemplateManagement asks people to select filters first, before loading any data. It's disjoint for the filter to say "All" when Nothing is selected or loaded, and `multiFilter({})` is ignoring nothingSelectedText.

### DIFF
--- a/src/components/Filters/MultiFilter.tsx
+++ b/src/components/Filters/MultiFilter.tsx
@@ -55,7 +55,7 @@ class MultiFilter<O, V extends Value> extends BaseFilter<V[], MultiFilterProps<O
       );
     }
 
-    const { defaultValue, ...props } = this.props;
+    const { defaultValue, nothingSelectedText, ...props } = this.props;
     return (
       <MultiSelectField<O, V>
         {...props}
@@ -67,7 +67,7 @@ class MultiFilter<O, V extends Value> extends BaseFilter<V[], MultiFilterProps<O
         onSelect={(values) => {
           setValue(values.length === 0 ? undefined : values);
         }}
-        nothingSelectedText="All"
+        nothingSelectedText={nothingSelectedText ?? "All"}
         {...this.testId(tid)}
       />
     );


src/components/Filters/MultiFilter.tsx
```diff
@@ -67,7 +67,7 @@ class MultiFilter<O, V extends Value> extends BaseFilter<V[], MultiFilterProps<O
         onSelect={(values) => {
           setValue(values.length === 0 ? undefined : values);
         }}
-        nothingSelectedText="All"
+        nothingSelectedText={nothingSelectedText ?? "All"}
```
stephenh, line 70: Totally get the rationale; I'm assuming the caller is going to do something like "if not loaded, set nothingSelectedText to loading?" 

Makes me wonder if, instead of all of our filters having that conditional logic setting `nothingSelectedText`, should we have an additional `noOptionsAvailableText`? Like something that could be set here in Beam as like "just assume if a filter has no options, surely it is being loaded".

Disclaimer I'm guessing at what you're follow up change on the BP side of things will be.

cc @bdow for thoughts

TylerR909: No, the caller is trying to override "All" with "None" because the user hasn't selected anything.

For BTM the filters don't even render until they've fetched data. Maybe that fetch loaded `[]` but this isn't that case. This is Radhi asking "I haven't selected anything, why does that say `All`"

stephenh: Gotcha

stephenh (approved):

src/components/Filters/MultiFilter.tsx
bdow: I would also contemplate disabling a SelectField that has no options and providing a tooltip explaining why its disabled. I wouldn't really want to assume no options means we're loading them. I'd prefer to move towards teaching the fields how to load their options themselves. Then they'd be aware of whether we have nothing because we are still loading or we really have no options.

TylerR909: Sidebar but Async Loading and Persisted Filters are not playing nicely together. While options are loading (before we have data), persistent filters is fetching from session storage, _seeing_ that there's no data, and `nil`'ing out the options. By the time filters load, selections have already been cleared out.

bdow: Huh, good note. Sounds like async filters are not ready for prime time.

PR #804, opened 2023-02-08
fix: (SC-27545) Close right pane on unmount page

pgamboa07:
[SC-27545](https://app.shortcut.com/homebound-team/story/27545/task-super-drawer-not-closing-when-moving-across-projects-developments)

### DIFF
--- a/src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
+++ b/src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
@@ -14,7 +14,7 @@ export function RightPaneLayout({
 }) {
   const { isRightPaneOpen, rightPaneContent, clearPane, closePane } = useRightPaneContext();
 
-  // Close pane on page unmount because otherwise otherwise the next page that has a right pane will show our stale content
+  // Close pane on page unmount because otherwise the next page that has a right pane will show our stale content
   useEffect(() => {
     return () => {
       closePane();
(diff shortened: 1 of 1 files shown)


src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -12,8 +12,15 @@ export function RightPaneLayout({
   paneBgColor?: Palette;
   paneWidth?: number;
 }) {
-  const { isRightPaneOpen, rightPaneContent, clearPane } = useRightPaneContext();
+  const { isRightPaneOpen, rightPaneContent, clearPane, closePane } = useRightPaneContext();
 
+  // Close pane on page unmount
```
stephenh, line 17: Question: Just curious, but why do we care about closing the pane if we're unmounting anyway? 

Fwiw I think this would be a good rationale to document w/an in-source comment.

Like right the in-source comment is telling me _what_ you're doing, which is fine, but I can read to the code to answer "what" is happening.

What I may not know, as a maintainer, is _why_ we are doing this, so updating the comment to be like:

```
// Close pane on page unmount because otherwise ...something bad...
```

stephenh: Oh, it is probably "otherwise the next page that has a right pane will show our stale content"?

stephenh (approved):

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
pgamboa07: Yes, this is the issue. Comment updated!

stephenh: Great, thanks!

PR #806, opened 2023-02-10
fix: Unsubscribe on unmount.

stephenh:
Not sure why I didn't think of adding this before.

Should fix the "cannot call setState on an unmounted component errors" we see in tests sometimes.

### DIFF
--- a/src/hooks/useComputed.ts
+++ b/src/hooks/useComputed.ts
@@ -52,7 +52,7 @@ export function useComputed<T>(fn: () => T, deps: readonly any[]): T {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, deps);
 
-  // Tell unsubscribe the autorun when we're unmounted
+  // unsubscribe the autorun when we're unmounted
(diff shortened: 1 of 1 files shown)


bdow (approved):
🎉 Hopefully this does remove those warnings/errors.

src/hooks/useComputed.ts
```diff
@@ -49,9 +49,15 @@ export function useComputed<T>(fn: () => T, deps: readonly any[]): T {
         setTick((tick) => tick + 1);
       }
     });
+    return current.runner;
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, deps);
 
+  // Tell unsubscribe the autorun when we're unmounted
+  useEffect(() => {
+    return ref.current.runner;
```
bdow, line 57: question: Shouldn't we be calling the runner to unhook it?

bdow: Oh nevermind, this will be called. I read the syntax incorrectly at first.

PR #809, opened 2023-02-10
feat: Add Pagination component

allangaldinosilva:
# [SC-8167](https://app.shortcut.com/homebound-team/story/8167/beam-new-pagination-component)

- Move pagination component from internal-frontend to Beam

image

### DIFF
--- a/src/components/Pagination.test.tsx
+++ b/src/components/Pagination.test.tsx
@@ -1,17 +1,17 @@
 import { noop } from "src/utils";
 import { click, render } from "src/utils/rtl";
-import { initPageSettings, Pagination, toFirstAndOffset } from "./Pagination";
+import { initPageSettings, pageOptions, Pagination, toFirstAndOffset } from "./Pagination";
 
 describe("Pagination", () => {
   it("can have a data-testid", async () => {
     const r = await render(
-      <Pagination hasNextPage={true} label="" setSettings={noop} settings={initPageSettings} data-testid="pag" />,
+      <Pagination totalCount={10} label="" setSettings={noop} settings={initPageSettings} data-testid="pag" />,
     );
     expect(r.firstElement.firstElementChild!.getAttribute("data-testid")).toEqual("pag");
   });
 
   it("defaults data-testid to pagination", async () => {
-    const r = await render(<Pagination hasNextPage={true} label="" setSettings={noop} settings={initPageSettings} />);
+    const r = await render(<Pagination totalCount={10} label="" setSettings={noop} settings={initPageSettings} />);
     expect(r.firstElement.firstElementChild!.getAttribute("data-testid")).toEqual("pagination");
   });
 
@@ -19,7 +19,7 @@ describe("Pagination", () => {
     // Given a pagination component on page 1 and per page 50
     const setSettings = jest.fn();
     const r = await render(
-      <Pagination hasNextPage={true} label="" setSettings={setSettings} settings={{ page: 1, perPage: 50 }} />,
+      <Pagination totalCount={100} label="" setSettings={setSettings} settings={{ page: 1, perPage: 50 }} />,
     );
     // When click to go to the next page
     click(r.pagination_nextIcon);
@@ -32,7 +32,7 @@ describe("Pagination", () => {
     // Given a pagination component on page 1 and per page 50
     const setSettings = jest.fn();
     const r = await render(
-      <Pagination hasNextPage={true} label="" setSettings={setSettings} settings={{ page: 1, perPage: 50 }} />,
+      <Pagination totalCount={10} label="" setSettings={setSettings} settings={{ page: 1, perPage: 50 }} />,
     );
     // Then perPage field have value of 50
     expect(r.pagination_perPage()).toHaveValue("50");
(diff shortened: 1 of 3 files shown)


src/components/Pagination.tsx
```diff
@@ -0,0 +1,72 @@
...
+  settings: PageSettings;
+  setSettings: Dispatch<SetStateAction<PageSettings>>;
+};
+
+export function Pagination(props: PaginationProps) {
+  const { hasNextPage, label, settings, setSettings } = props;
+  const { perPage } = settings;
+  const pageOptions = Array(5)
+    .fill(0)
+    .map((_, i) => ({ id: (i + 1) * 50, name: ((i + 1) * 50).toString() }));
```
stephenh, line 24: Suggestion: We should stop hard-coding this before we land this in Beam.

I believe we probably need like `PageSettings.totalCount`, and then this page options is basically `totalCount / perPage` --> list of pages they can choose in the drop down.

cc @roberthdg does that sound right?

src/components/Pagination.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { useTestIds } from "src/utils";
+
+export type PageSettings = {
+  page: number;
+  perPage: number;
+};
+
+type PaginationProps = {
+  label: string;
+  hasNextPage: boolean;
```
stephenh, line 14: Thought: We could probably drop `hasNextPage` and infer it from `PageSettings.page` and `totalCount / perPage`, per other comment to add `totalCount` to remove the hard-coded drop down.

src/components/Pagination.test.tsx
```diff
@@ -0,0 +1,77 @@
+import { noop } from "src/utils";
+import { click, render } from "src/utils/rtl";
+import { initPageSettings, Pagination, toFirstAndOffset } from "./Pagination";
+
+describe("Pagination", () => {
```
stephenh, line 5: Praise: Really like the tests! Thanks!

src/components/Pagination.tsx
roberthdg: This `pageOptions` would be the number of items that the user could select to appear per page, not the actual page index (could be renamed to better match the functionality)

image

stephenh: Oh jeez, that makes sense. Thanks @roberthdg !

@allangaldinosilva since this doesn't depend on any props, could you make it a top-level `const` for now and use an in-source comment like `// Make a list of 50/100/150/etc page sizes for the user to chose`.

src/components/Pagination.tsx
stephenh: @allangaldinosilva / @roberthdg I guess in retrospect we just don't need `totalCount`? That's cool, so n/m I guess?

stephenh (approved):

src/components/Pagination.tsx
allangaldinosilva: Done 👍 thanks guys!

src/components/Pagination.tsx
allangaldinosilva: We do, actually we will need it for the new requirement that @roberthdg told, we are gonna show the current page info close to the navigation buttons.

image

I pushed these changes in the last commit, when you get a chance to re-review this I would appreciate.

src/components/Pagination.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const hasNextPage = useMemo(() => page < totalCount / perPage, [page, perPage, totalCount]);
+
+  const tid = useTestIds(props, "pagination");
+  return (
+    <div css={Css.df.bGray300.bt.xs.gray500.px1.ml2.pt2.$} {...tid}>
+      <div css={Css.df.mya.mr2.$} {...tid.perPageLabel}>
+        {label} per page:
+      </div>
+      <div css={Css.wPx(78).$}>
+        <SelectField
```
bdow, line 37: **suggestion:** I think you might want to use the `compact` variant of the `SelectField`. When looking at the mocks, they use a SelectField that has a 32px height, which would signal it is the `compact` style.
https://www.figma.com/file/aWUE4pPeUTgrYZ4vaTYZQU/%E2%9C%A8Beam-Component-Library?node-id=32601%3A99960&t=omqHcE1hZIcy6vQz-0

src/components/Pagination.tsx
```diff
@@ -0,0 +1,79 @@
...
+          disabled={!hasNextPage}
+        />
+      </div>
+    </div>
+  );
+}
+
+export const initPageSettings = { page: 1, perPage: 100 };
+
+export function toFirstAndOffset(page: number, perPage: number) {
```
bdow, line 73: **request:** Is this method being used? I only see it being used in the test.

src/components/Pagination.tsx
```diff
@@ -0,0 +1,79 @@
...
+          color={Palette.LightBlue700}
+          onClick={() => setSettings({ page: settings.page + 1, perPage })}
+          disabled={!hasNextPage}
+        />
+      </div>
+    </div>
+  );
+}
+
+export const initPageSettings = { page: 1, perPage: 100 };
```
bdow, line 71: **request:** Can we remove this from the component? It looks like this is used in Storybook and in tests, but not necessary for the component. Can we just copy it to the test and story file?

src/components/Pagination.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Css, Palette } from "src/Css";
+import { SelectField } from "src/inputs";
+import { useTestIds } from "src/utils";
+
+export type PageSettings = {
+  page: number;
+  perPage: number;
+};
+
+type PaginationProps = {
```
bdow, line 12: **request:** To be consistent with our other components, can we make this an `interface` instead of a type?

src/components/Pagination.tsx
```diff
@@ -0,0 +1,79 @@
...
+  label: string;
+  totalCount: number;
+  settings: PageSettings;
+  setSettings: Dispatch<SetStateAction<PageSettings>>;
+};
+
+// Make a list of 50/100/150/etc page sizes for the user to chose
+export const pageOptions = Array(5)
+  .fill(0)
+  .map((_, i) => ({ id: (i + 1) * 50, name: ((i + 1) * 50).toString() }));
```
bdow, line 22: **request:** Per the [component style guide](https://homebound.atlassian.net/wiki/spaces/ENG/pages/339574863/Component+Style+Guide), can we move this const to the bottom of the component?

bdow (changes requested):
Just a couple requested changes to keep consistency. Otherwise looks good!

src/components/Pagination.tsx
allangaldinosilva: Answering in a Beam context I would say "No", it's a help function used [here](https://github.com/homebound-team/internal-frontend/blob/main/src/routes/libraries/item-catalog/hooks/useItemCatalog.tsx) to set a default `first, offset` values for pagination.

I'm unsure it should live inside Beam... As this task was essentially to "move" the existing Pagination component from `internal-frontend` into `Beam` it seems doesn't be part of Pagination?

src/components/Pagination.tsx
allangaldinosilva: cool, really nice this style guide! Thanks for share 👍

src/components/Pagination.tsx
allangaldinosilva: Done 👍

src/components/Pagination.tsx
allangaldinosilva: Done 👍

src/components/Pagination.tsx
allangaldinosilva: You're right, fixed 👍

src/components/Pagination.tsx
bdow: Interesting... I would almost rather it just live in Blueprint for now. Especially since it's only used once. I don't see it hurting anything right now, so non-blocking. But would like to think about these sort of helper methods.

bdow (approved):

PR #810, opened 2023-02-13
feat: Autocomplete component

bdow:
This is a 100% controlled autocomplete component that can work with third party libraries such as Alogolia (used in Blueprint). Adds storybook and tests


src/inputs/Autocomplete.tsx
```diff
@@ -0,0 +1,136 @@
...
+    Pick<TextFieldBaseProps<any>, "label" | "clearable" | "startAdornment"> {
+  onSelect: (item: T) => void;
+  /** A function that returns how to render the an option in the menu. If not set, `getOptionLabel` will be used */
+  getOptionMenuLabel?: (o: T) => ReactNode;
+  /** A function that returns the string value of the option. Used for accessibility purposes */
+  getOptionLabel: (o: T) => string;
+  /** A function that returns a unique key for an option */
+  getOptionValue: (o: T) => string;
+  /** Called when the input value changes */
+  onChange: (value: string) => void;
```
stephenh: Suggestion: Maybe `onInputChange` or `onTypeAheadChange`? Just thinking to clarify it's the "hey they typed something different, redo your GQL call" callback and not the "they actually picked option #4".

src/inputs/Autocomplete.tsx
```diff
@@ -0,0 +1,136 @@
...
+  /** The current value of the input */
+  value: string | undefined;
```
stephenh, line 24: Thought: Wdyt about keeping this `value` internally controlled, like our own internal `useState`, and then firing the `onChange` but debounced?

That way the caller doesn't have to do their own debounce / keep track of the every-single-key-press value of our text field.

src/inputs/Autocomplete.tsx
bdow: Yup. Adding that specificity into the prop name would be helpful.

src/inputs/Autocomplete.tsx
bdow: I thought about this. I still have another example of this component that uses the `useAsyncList` from React-Aria, which makes this `value` internally controlled. Then it handles its own debouncing, fetching, and list of results. I was thinking of making the component smarter later... like you can either have it completely controlled (as-is right now), or more uncontrolled where you just pass in the fetch function and the rest is handled. 
I can see the benefit of adding the debounce here, but then I wonder if that'll hinder the Algolia search experience. I would assume they must do some sort of debouncing on their end.

bdow: If you feel really strongly about doing the debouncing locally, we can. I just don't see it as a pressing need at the moment. Personally I think having the debouncing happen closer to the fetch call would be ideal. It'd just be more clear when/where the debounce is happening. Otherwise, I think I'd want the prop to maybe be named `onDebouncedInputChange` and allow the user to specify a `props.debounceMs` or something. 
Also makes me wonder if there'd be a time where we wouldn't want to debounce? I guess if the potential options are all local (doesn't require an api fetch), then the debounce wouldn't be necessary. 

Anyways, let me know what you think. I'm leaning towards keeping it as is.

stephenh: Cool, I'm good with this. We can always change it if it becomes obvious we have a ton of callers doing the same debounce.

stephenh (approved):
Pretty awesome how succinct this is/sits well on top of our other infra. Nice job!

src/inputs/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { useFilter } from "react-aria";
+import { Css } from "src/Css";
+import { Autocomplete } from "src/inputs/Autocomplete";
+
+export default {
+  component: Autocomplete,
+} as Meta;
+
+export const Example: StoryFn = Template.bind({});
+Example.parameters = { chromatic: { disableSnapshot: true } };
```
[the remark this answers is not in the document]
bdow: It means we will not take snapshots of this Story ([docs link](https://www.chromatic.com/docs/ignoring-elements#ignore-stories)). The other story includes the input field **and** the menu open, which encompasses everything we'd really want to snapshot. I'm trying to be more cognizant which Stories we _really_ need visually regression tests against. Since everything in this Story is covered by the `MenuOpen` story, then I deemed this one superfluous.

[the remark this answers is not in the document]
stephenh: Ah nice! Yeah, sorry I could intuit "it turns off snapshots", I should have better articulated "why?", but you're reasoning makes sense.

Wdyt about an in-source const like `chromatic: skipSnapshotsForCost` or something with an in-source documentation/rationale for "we're skipping this b/c its covered by other stories" and not something like "we're skipping this b/c the snapshot is flakey"?

PR #815, opened 2023-02-17
feat: Bump react

stephenh:


src/components/Modal/Modal.tsx
```diff
@@ -74,10 +74,14 @@ export function Modal(props: ModalProps) {
 
   const [hasScroll, setHasScroll] = useState(forceScrolling ?? false);
 
-  useResizeObserver(modalBodyRef, ({ target }) => {
-    if (forceScrolling === undefined && !isFixedHeight) {
-      setHasScroll(target.scrollHeight > target.clientHeight);
-    }
+  useResizeObserver({
```
stephenh, line 77: @bdow sanity check on this in particular, I moved over to `react-aria`'s `useResizeObserver`.

bdow: Nice! Interesting that React-Aria provides this util/hook.

src/components/Modal/Modal.tsx
```diff
@@ -74,10 +74,14 @@ export function Modal(props: ModalProps) {
...
+    ref: modalBodyRef,
+    onResize: useCallback(() => {
+      const target = modalBodyRef.current!;
+      if (forceScrolling === undefined && !isFixedHeight) {
+        setHasScroll(target.scrollHeight > target.clientHeight);
+      }
+    }, []),
```
bdow, line 84: **thought/question:** I've never seen an inline `useCallback` like this. Was this something we could do in React 16?

src/components/Snackbar/useSnackbar.tsx
```diff
@@ -39,7 +39,7 @@ export function useSnackbar(): UseSnackbarHook {
   }, []);
 
   const triggerNotice = useCallback(
-    (props) => {
+    (props: any) => {
```
bdow, line 42: **suggestion:** I think we could change this `any` to `TriggerNoticeProps`. 
```suggestion
    (props: TriggerNoticeProps) => {
```

package.json
```diff
@@ -35,19 +35,19 @@
...
+    "framer-motion": "^9.0.4",
     "memoize-one": "^5.2.1",
     "react-aria": "^3.18.0",
     "react-day-picker": "8.0.7",
-    "react-popper": "^2.2.5",
-    "react-router": "^5.2.0",
-    "react-router-dom": "^5.2.0",
+    "react-popper": "^2.3.0",
+    "react-router": "^5.3.4",
+    "react-router-dom": "^5.3.4",
```
bdow, line 50: Would love to get this up to 6 at some point as well. Or maybe even start looking at other router options... TanStack Router?

src/components/Toast/useToast.tsx
```diff
@@ -7,7 +7,7 @@ export interface UseToastProps {
 export function useToast(): UseToastProps {
   const { setNotice } = useToastContext();
   const showToast = useCallback(
-    (props) => {
+    (props: any) => {
```
bdow, line 10: **suggestion:** Could be `ToastNoticeProps`
```suggestion
    (props: ToastNoticeProps) => {
```

src/components/internal/ContextualModal.tsx
```diff
@@ -4,7 +4,7 @@ import { Css } from "src/Css";
 import { useTestIds } from "src/utils";
 
 export interface ContextualModalProps {
-  content: ReactNode;
+  content: ReactNode | ((close: () => void) => ReactNode);
```
bdow, line 7: Huh... surprised this worked to begin with.

src/forms/BoundDateField.test.tsx
```diff
@@ -14,12 +13,14 @@ describe("BoundDateField", () => {
...
-    r.birthday().focus();
+    focus(r.birthday);
     // Then the callback should be triggered
     expect(onFocus).toBeCalledTimes(1);
 
     // When blur is triggered on a checkbox
-    r.birthday().blur();
+    blur(r.birthday());
+    // r.birthday().blur();
+    // });
```
bdow, line 23: **nit:** Can we remove these commented out lines?

src/forms/BoundDateField.test.tsx
```diff
@@ -36,7 +37,7 @@ describe("BoundDateField", () => {
     const r = await render(<BoundDateField field={author.birthday} />);
 
     // When triggering the Date Picker
-    r.birthday().focus();
+    focus(r.birthday);
```
bdow, line 40: **thought:** I am a little concerned about using `focus` here instead of `.focus` as I believe they have different behaviors. Specifically:
`element.focus()`: Actually applies focus to this element in the DOM, which in turn triggers the "focus" event.
`fireEvent.focus(element)`: Does not apply focus to the element, only triggers the focus event as if the element _was_ put in focus.

That nuance can be a bit of a pain to figure out. But, if everything still passes, then great. I am just kind of surprised.

src/inputs/Autocomplete.stories.tsx
```diff
@@ -16,7 +16,7 @@ Example.parameters = { chromatic: { disableSnapshot: true } };
 export const MenuOpen: StoryFn = Template.bind({});
 MenuOpen.play = async ({ canvasElement }: { canvasElement: HTMLElement }) => {
   const canvas = within(canvasElement);
-  await fireEvent.focus(canvas.getByTestId("heroes"));
+  canvas.getByTestId("heroes").focus();
```
bdow, line 19: Ahh! Here's one where we went the other way! Granted, this is in Storybook....

src/inputs/RadioGroupField.tsx
```diff
@@ -155,7 +155,7 @@ function Radio<K extends string>(props: {
         </div>
         {description && (
           <div id={descriptionId} css={Css.sm.gray700.if(disabled).gray400.$}>
-            {description}
+            {typeof description === "function" ? description() : description}
```
bdow, line 158: **thought:** Huh, surprised we allowed the description to be a function that returns a ReactNode... I wonder what this use case was? 🤔

bdow (approved):
Awesome!

src/components/Snackbar/useSnackbar.tsx
stephenh: Sweet, changed!

src/components/Toast/useToast.tsx
stephenh: :+1:

src/forms/BoundDateField.test.tsx
stephenh: Yep, done

src/forms/BoundDateField.test.tsx
stephenh: Yeah, fwiw I'd like for our `focus(...)` to eventually more fully emulate `element.focus()`, but for now `focus(...)` a) works, and b) does the `act` for us, otherwise would need to write:

```
act(() => r.birthday.focus());
```

So, yeah, since it passes the test, seems fine for now.

src/inputs/Autocomplete.stories.tsx
stephenh: Yep, I needed "the real focus" here.

Fwiw not sure what the old code was doing b/c `fireEvent.focus` is not a promise.

stephenh (commented):
Thanks for the thorough review @bdow ! Changes are pushed up.

PR #816, opened 2023-02-21
fix: support precise year selection

0ces:
I think this should already exist in beam, is pretty bad UX having to click a 100 times to get to the desired year.

I added this custom header to the `DatePicker` and `DateRangePicker`, open to design feedback 🙂.

image

### DIFF
--- a/src/components/internal/DatePicker/DatePicker.tsx
+++ b/src/components/internal/DatePicker/DatePicker.tsx
@@ -1,26 +1,34 @@
 import { DayPicker, Matcher } from "react-day-picker";
 import { Day } from "src/components/internal/DatePicker/Day";
-import { Header } from "src/components/internal/DatePicker/Header";
+import { Header, PreciseDateHeader, YearSkipHeader } from "src/components/internal/DatePicker/Header";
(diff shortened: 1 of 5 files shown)


0ces:
Storybook for testing: https://60466f7d43c37c0021788867-tviinpscvu.chromatic.com/?path=/story/workspace-components-datepicker--precise

src/components/internal/DatePicker/DatePicker.tsx
```diff
@@ -1,26 +1,34 @@
 import { DayPicker, Matcher } from "react-day-picker";
 import { Day } from "src/components/internal/DatePicker/Day";
-import { Header } from "src/components/internal/DatePicker/Header";
+import { Header, PreciseDateHeader, YearSkipHeader } from "src/components/internal/DatePicker/Header";
 import { WeekHeader } from "src/components/internal/DatePicker/WeekHeader";
 import { Css } from "src/Css";
 import { useTestIds } from "src/utils";
 import "./DatePicker.css";
 
+export type YearPicker = "default" | "skip" | "precise";
```
bdow, line 9: **suggestion:** I would push for the `precise` version to replace `skip`. We'll need to chat with the designer to see if they are good with that too.

PR #821, opened 2023-03-01
feat: Support vertical alignment on Table component

bdow:


0ces (approved):

src/components/Table/TableStyles.tsx
```diff
@@ -66,6 +66,8 @@ export interface GridStyleDef {
   allWhite?: boolean;
   bordered?: boolean;
   rowHover?: boolean;
+  /** Defines the vertical alignment of the content of the cells for the whole table (not including the 'header' rows). Defaults to `center` */
+  vAlign?: "top" | "center" | "bottom";
```
stephenh, line 70: @bdow I just happened to notice this while reading some random code; I'm honestly kinda surprised that our default is center, maybe a naive question, but should this default to top? For some reason I'm thinking table cell content is almost always vertical aligned top? Seems like I'm missing something...

bdow: Good thought, thanks for revisiting. I think for "fixed" row heights it makes sense to use center alignment as there is no vertical padding in the cells, just a hard coded height. So things could look bad. When first building out the new table styles, the "fixed" row height was the default, which is why we defaulted to center vertically. I do think it could make more sense to default the vertical alignment to "top" when using the "flexible" row height. I'll make a PR to update.

bdow: Yea, we could probably drop it. I see it is only used once in `internal-frontend` to define `vAlign: "top"`.

PR #822, opened 2023-03-01
fix: Ensure setting 'vAlign' triggers new table styles

bdow:


src/components/Table/TableStyles.tsx
```diff
@@ -238,6 +238,7 @@ export function resolveStyles(style: GridStyle | GridStyleDef): GridStyle {
     "allWhite",
     "bordered",
     "rowHover",
+    "vAlign",
```
stephenh, line 241: Suggestion: We could probably do something like:

```
const defKeysRecord: Record<keyof GridStyleDef, boolean> = {
  inlineEditing: true,
  ...
};
const defKeys = Object.keys(defKeysRecord);
```

To get compile errors on new keys being added to `GridStyleDef` but not added here?

stephenh (approved):

PR #824, opened 2023-03-09
feat: support for label style on StaticField

allangaldinosilva:
## SC-30546

- Add support for `labelStyle` prop

### Example
image

Discussion: https://github.com/homebound-team/internal-frontend/pull/3533#discussion_r1131495964

### DIFF
--- a/src/forms/StaticField.tsx
+++ b/src/forms/StaticField.tsx
@@ -1,6 +1,6 @@
 import { useId } from "@react-aria/utils";
 import { ReactNode } from "react";
-import { PresentationFieldProps } from "src/components/PresentationContext";
+import { PresentationFieldProps, usePresentationContext } from "src/components/PresentationContext";
 import { Css, px } from "src/Css";
 import { defaultTestId } from "src/utils/defaultTestId";
 import { useTestIds } from "src/utils/useTestIds";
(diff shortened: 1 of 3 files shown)


src/forms/StaticField.tsx
```diff
@@ -1,25 +1,27 @@
...
 interface StaticFieldProps {
   label: ReactNode;
   value?: string;
   children?: ReactNode;
+  labelStyle?: PresentationFieldProps["labelStyle"];
 }
 
 export function StaticField(props: StaticFieldProps) {
-  const { label, value, children } = props;
+  const { label, labelStyle = "above", value, children } = props;
```
stephenh, line 16: Suggestion: Can you initialize this from the presentation context?

```
  const { fieldProps } = usePresentationContext();
  const { label, 
    labelStyle = fieldProps?.labelStyle ?? "above",
```

Like we do in `TextFieldBase`? That way it will get the right default is someone is using this inside of a FormLines or other PresentationContext

stephenh (approved):
Thanks!

src/forms/StaticField.tsx
allangaldinosilva: Nice, good catch, thank you! 👏

PR #827, opened 2023-03-13
fix: Remove wrong width from Static Field div

allangaldinosilva:
After run Chromic step we could notice a bunch of changes in the PR above, this is because the property `width` set in last commit was wrong.

PR (https://github.com/homebound-team/internal-frontend/pull/3533)

This PR is to remove that and avoid change it in many places.

### DIFF
--- a/src/forms/StaticField.stories.tsx
+++ b/src/forms/StaticField.stories.tsx
@@ -39,3 +39,28 @@ export function StaticField() {
     </FormLines>
   );
 }
+
+export function StaticFieldsGrouped() {
+  return (
+    <div css={Css.w50.$}>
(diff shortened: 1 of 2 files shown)


bdow (approved):
request: Thanks for fixing. Can we add some storybook examples of this component in its various states so we can ensure it looks right before we integrate it into Blueprint, or any other application?

allangaldinosilva:
> request: Thanks for fixing. Can we add some storybook examples of this component in its various states so we can ensure it looks right before we integrate it into Blueprint, or any other application?

Cool @bdow, I've added a new story for grouped fields where without those new changes the `Posted date` field broke into a new line.

PR #828, opened 2023-03-17
feat: support wrap text on table using select field

allangaldinosilva:
# SC-31248

Updated `TextFieldBase` to use wrap PresentationContext property to prevent truncate the field.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -11,7 +11,8 @@ import { TableStateContext } from "src/components/Table/utils/TableState";
 import { emptyCell, matchesFilter } from "src/components/Table/utils/utils";
 import { Css, Palette } from "src/Css";
 import { useComputed } from "src/hooks";
-import { Checkbox, TextField } from "src/inputs";
+import { Checkbox, SelectField, TextField } from "src/inputs";
+import { noop } from "src/utils";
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.test.tsx
```diff
@@ -137,6 +138,36 @@ describe("GridTable", () => {
     expect(cell(r, 2, 0)).toHaveTextContent("2");
   });
 
+  it("wraps readonly select field", async () => {
```
stephenh, line 141: Suggestion: Really minor but maybe call this "does not truncate reaodnly select fields"?

I get it is effectively wrapping, and it's called `wrap` in the code, but I think this name more matches what we're asserting against.

stephenh (approved):
Looks great!

src/components/Table/GridTable.test.tsx
allangaldinosilva: cool, it seems better, thanks 👏

bdow: This also addresses it for any field, not just SelectFields. Thanks to updating it in `TextFieldBase`, we get the fix in DateField, NumberField, TextField, SelectField, MultiSelectField, etc... 🎉

bdow (approved):
Nice work! Thanks for adding the tests and stories!

PR #829, opened 2023-03-17
feat: SC-31238-update-gridtable-totals-row

Levi-Huynh:
[SC-31238](https://app.shortcut.com/homebound-team/story/31238/update-totals-row-in-tables)

Issue: The current totals row in tables is difficult to view in context, and creates a gap between filters and the table itself.

Update the totals row/cell component to:

Change text size to XS/Medium (12px) instead of SM/Medium
Add bg color: Gray100
default to being UNDER the header row instead of ABOVE the header row
40px height instead of 48px

**Previews:**
- ChangeEventsLineItemsTable
- CommitmentEditorLineItems (PO/CCO table)
- Homeowner-contracts LineItemsTable
- Projects Invoices (InvoiceLineItemsTable)
- Invoice steps (SelectedFixItems)
- ItemTemplateItemAllTable
- TradePartnerCommitmentsTab table
- Budget Page Table & Budget Super Drawer

https://user-images.githubusercontent.com/43801393/226452385-4283cf3b-dc78-4f66-9b95-6968ae724e26.mov


src/components/Table/TableStyles.tsx
```diff
@@ -116,7 +116,8 @@ function memoizedTableStyles() {
           ...Css.gray700.xsMd.bgGray200.aic.pxPx(12).whiteSpace("unset").hPx(40).$,
           ...(allWhite && Css.bgWhite.$),
         },
-        totalsCellCss: Css.bgWhite.gray700.smMd.hPx(totalsRowHeight).pPx(12).boxShadow("none").$,
+        totalsCellCss: Css.bgWhite.gray700.bgGray100.xsMd.hPx(totalsRowHeight).pPx(12).boxShadow("none").bb.bl.bGray200
```
bdow, line 119: **question:** Why `bl`? 
Also, the `boxShadow` defined on the `cellCss` key in this object is what creates our borders between the table rows. We ended up setting `boxShadow("none")` here because we didn't want that border between the row for the Totals row. But now we do! I'm guessing we could just remove the `boxShadow("none")` and the other new border styles and get the desired result.
Would you mind trying that out?

bdow (commented):
Really awesome that we can update all of Blueprint's tables with changing only ~3 lines of code. 🎉

src/components/Table/TableStyles.tsx
Levi-Huynh: @bdow 

`boxShadow`: I'll push up the boxShadow("none") and share a preview of that in a bit (but I'm thinking thats prob good to remove as you suggested)

`bl`: I really only added this b/c of how it affects our specs & selections table:
(I dont think this bl really was needed anywhere else).  I'm thinking we can handle that here in beam or have it handled just specifically in the totalCss for this table. Open to your thoughts

https://user-images.githubusercontent.com/43801393/226428278-634831e9-b1f9-4acd-83f5-c0abae8ef9fa.mov

bdow: Thanks for the info. I still do not think the `bl` definition is necessary here. I can see how it makes the table look a bit weird, but that is somewhat intended. It may be unexpected, but those are the intended colors to be using.

bdow (approved):

PR #831, opened 2023-03-20
feat: SC-30972 adding bx link icon

JonnCharpentier:


bdow (approved):

src/components/Icon.stories.tsx
```diff
@@ -22,6 +22,7 @@ export const Icon = (props: IconProps) => {
     "x",
     "loader",
     "linkExternal",
+    "bxLink",
```
bdow, line 25: **nit:** Can we call this something else besides `bxLink`, maybe just `link`? The `bx` prefix comes with the Box icon library set design uses.
Also, can you ask the designer for this to add the icon to Beam's Icon Pages in Figma? https://www.figma.com/file/ePfj5WZKtGfOgABpZcUtRE/%E2%9C%A8Beam-Icons?node-id=5-14&t=RlW8BvpXJ1ucYGU4-0

src/components/Tag.tsx
```diff
@@ -2,7 +2,7 @@ import { Icon, IconKey } from "src/components";
 import { Css, Margin, Only, Xss } from "src/Css";
 import { useTestIds } from "src/utils";
 
-type TagXss = Margin;
```
bdow, line 5: **question:** Where did the need for custom background colors come from regarding our tags? Do we have a ticket or Figma regarding this?
The way Tags were intended to work is that we had a subset of supported background colors that provided a specific meaning. By allowing any background color then we are enabling inconsistency across the application.
If we want to allow a new background color, then I would suggest we add a new `TagType`.

JonnCharpentier: Ididn't need it, I didn't know that the tagType change the color, I need it the info one, I'll remove it on my next PR

bdow: Awesome. Thanks for following up!

PR #833, opened 2023-03-21
fix: Prevent Avatar from squishing by setting min-width

bdow:
Includes storybook example as a regression test.

## Before
Screen Shot 2023-03-21 at 10 16 13 AM

## After
Screen Shot 2023-03-21 at 10 17 59 AM


chr1sjf0x (approved):
TY @bdow

PR #834, opened 2023-03-23
feat: adding support for number fields to be positive only

JonnCharpentier:


src/inputs/NumberField.tsx
```diff
@@ -140,7 +143,7 @@ export function NumberField(props: NumberFieldProps) {
     value: valueRef.current.wip ? valueRef.current.value : value === undefined ? Number.NaN : value / factor,
     // // This is called on blur with the final/committed value.
     onChange: (value) => {
-      onChange(formatValue(value, factor, numFractionDigits, numIntegerDigits));
+      onChange(formatValue(value, factor, numFractionDigits, numIntegerDigits, positiveOnly));
```
bdow, line 146: Thought/suggestion: I think there is another place we call formatValue. Should we do this in both places? I'd be curious on the user experience if we do both vs just what you have.

JonnCharpentier: I dont see any difference on behavior, but still added the change. Thanks

PR #835, opened 2023-03-23
feat: Add right pane column content

crshannon:
[SC-31107](https://app.shortcut.com/homebound-team/story/31107/rightpanelayout-swap-out-right-column-2m)

- add ability to add `rightColumnContent` and have that swap out with the `rightPaneContent`

### DIFF
--- a/src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
+++ b/src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
@@ -36,7 +36,7 @@ export function RightPaneLayout({
           {children}
         </div>
 
-        <div css={Css.if(!!defaultPaneContent).wPx(paneWidth).position("relative").$}>
+        <div css={Css.position("relative").if(!!defaultPaneContent).wPx(paneWidth).$}>
           {defaultPaneContent && (
             <div
               css={


src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
```diff
@@ -7,10 +7,12 @@ export function RightPaneLayout({
   children,
   paneBgColor = Palette.White,
   paneWidth = 450,
+  rightColumnContent,
 }: {
   children: ReactElement;
   paneBgColor?: Palette;
   paneWidth?: number;
+  rightColumnContent?: ReactElement;
```
stephenh, line 15: Suggestion: I'm wondering if we should call this like `defaultContent` or `primaryContent` ... just something to highlight that it's the default/primary content until something else within the pane is opened.

...`defaultPaneContent` ... wdyt?

stephenh (approved):

src/components/Layout/RightPaneLayout/RightPaneLayout.tsx
crshannon: sounds good to me - always open to suggestions, made the change

PR #837, opened 2023-03-30
chore: Rename internal components

bdow:
Rename the internal component SelectFieldBase and SelectFieldInput to ComboBoxBase and ComboBoxInput in preparation for creating a SelectField without the "type to filter". This will make distinguishing between the internal components easier as one will use React-Aria's 'useComboBox' (existing) and 'useSelect'. 

For extra context, the eventual state is that when using the 'SelectField' component that has fewer than X options, then a user will get the SelectField vs the ComboBox, as the type to filter will be unnecessary. We had said always supplying the "type to filter" wouldn't hurt, but that has become untrue with users on iPads. Currently when a user opens up a SelectField, then the input is immediately focused, triggering the iPad's keyboard, and then limiting the amount of space there is to view the ~5-10 options (i.e. Status fields). Supporting a SelectField that doesn't include the type to filter would greatly improve the user experience in these situations

### DIFF
--- a/src/components/Filters/MultiFilter.tsx
+++ b/src/components/Filters/MultiFilter.tsx
@@ -1,7 +1,7 @@
 import { Key } from "react";
 import { BaseFilter } from "src/components/Filters/BaseFilter";
 import { Filter } from "src/components/Filters/types";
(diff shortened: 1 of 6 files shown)


ipiedrahb (approved):
TY @bdow 🍣

PR #838, opened 2023-03-30
feat: Add icon for denoting tooltips in table cells

bdow:
When using the `tooltip` property on `GridCellContent`, the 'infoCircle' icon will be displayed as the tooltip trigger/target. This makes it more discoverable when tooltips are added to our table cells.


stephenh (approved):
Sweet!

PR #839, opened 2023-03-30
feat: Disable type to filter for single SelectFields with 10 or fewer options

bdow:
This provides a better UX for users using iPads where the virtual keyboard gets in the way and is unnecessary for select fields with few options.


chr1sjf0x (approved):
Is this worth having a test case? 🤷

bdow:
> Is this worth having a test case? 🤷

Probably. Will write one up quickly.

stephenh (commented):
Just curious, but what do you think about doing this optimization only for ipads?

Huge disclaimer I've not poked at it with a desktop browser (or ipad), but I'm just wondering if the "sometimes this types ahead" / "sometimes it does not" would throw desktop users off (like tab through, type, tab through, type) that we should leave the behavior as-is for them?

Or do you think this is a UX win for desktop as well?

PR #840, opened 2023-04-06
feat: Update React-Aria and React-Stately versions

bdow:


src/inputs/MultiSelectField.test.tsx
```diff
@@ -48,9 +48,11 @@ describe("MultiSelectFieldTest", () => {
     expect(r.age()).toHaveValue("");
 
     // When blurring the field
-    fireEvent.blur(r.age());
+    // fireEvent.blur(r.age());
+    act(() => r.age().blur());
```
stephenh, line 52: Fwiw if we want to make our rtl-utils `blur` do a `.blur()` instead of `fireEvent.blur` that sgtm.

stephenh (approved):
Nice!

src/components/Table/GridTable.test.tsx
```diff
@@ -1663,8 +1663,8 @@ describe("GridTable", () => {
     click(cell(r, 3, 1).children[0] as any);
 
     // Then the Header and Parent rows should have the `indeterminate` checked value
-    expect(cellAnd(r, 0, 1, "select")).toBePartiallyChecked();
-    expect(cellAnd(r, 1, 1, "select")).toBePartiallyChecked();
+    expect(cellAnd(r, 0, 1, "select")).toHaveAttribute("data-indeterminate", "true");
+    expect(cellAnd(r, 1, 1, "select")).toHaveAttribute("data-indeterminate", "true");
```
bdow, line 1667: React-Aria got rid of setting `aria-checked` which was what `toBePartiallyChecked` looks for. 
See comment: https://github.com/adobe/react-spectrum/issues/4004#issuecomment-1414537911

src/inputs/MultiSelectField.test.tsx
bdow: I forgot about this change here. I revisited this and the failure seemed to be due to call `fireEvent.blur` twice. I took off the second `fireEvent.blur` call and it passes. I do find it odd that `focus` seems to really need to be wrapped in an `act`, while `blur` doesn't seem as strict about it.

PR #841, opened 2023-04-07
feat: support disabledOptions object on mock select field

allangaldinosilva:
Currently our SelectField mock doesn't support disabled options object, this PR adds the ability to disable an option when it is an object.


stephenh (approved):
Thank you!

PR #844, opened 2023-04-11
fix: Prevent React-Aria from resetting combobox input value when menu closes

bdow:


src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -254,6 +254,9 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
   const state = useComboBoxState<any>({
     ...comboBoxProps,
     allowsEmptyCollection: true,
+    // We don't really allow custom values, as we reset the input value once a user `blur`s the input field.
+    // Though, setting `allowsCustomValue: true` prevents React-Aria/Stately from attempting to reset the input field's value as well.
```
stephenh, line 258: Nit: do you mind including when (on close?) they do this problematic resetting?

Also just naively/purely for my own understanding, why is resetting the value on close a bad thing?

stephenh (approved):
I assume this is for the issue Ben pointed out; if so really amazing turn around! Thank you!

src/inputs/internal/ComboBoxBase.tsx
bdow: That is correct, it was when the menu closes.
It is not a bad thing, in fact we do it ourselves. We don't rely on React-Aria/Stately to do it because we've kind of messed up their internal state. Specifically, `useComboBoxState` is only made to support a single selection. But we needed it to support multiple selections. So below this we have `state.selectionManager.state = useMultipleSelectionState`, which is us tricking the `useComboBoxState` to support multiple selections. But internally in `useComboBoxState`, they don't know what we're up to. This creates an improper state within that hook. This currently is only a concern if we expect `useComboBoxState` to control our state, but we completely control it on our side.
All of that said, we might need to move our ComboBox away from React-Aria at some point to avoid these issues popping up as we upgrade. Unless they decide to support a combobox that can handle multiple selections in one of those upgrades!

stephenh: Cool!

PR #848, opened 2023-04-13
feat: Add secondary borderless button variant

kbesingeryeh:
[SC-32614](https://app.shortcut.com/homebound-team/story/32614/add-new-button-variant-to-beam-s-button-modal)

* Add new `secondaryBorderless` variant to Button
Screen Shot 2023-04-13 at 11 09 06 AM

### DIFF
--- a/src/components/Button.stories.mdx
+++ b/src/components/Button.stories.mdx
@@ -45,7 +45,7 @@ export const Template = (args) => {
       <Button variant="tertiaryDanger" label="Delete" />
       <Button variant="danger" label="Delete" />
       <Button variant="text" label="Edit" />
-      <Button variant="secondaryBorderless" label="Text" />
+      <Button variant="secondaryText" label="Text" />
     </div>
   </Story>
 </Canvas>
--- a/src/components/Button.stories.mdx
+++ b/src/components/Button.stories.mdx
@@ -34,7 +34,7 @@ export const Template = (args) => {
 
 # Variant
 
-#### There are 6 different variants for a Button.
+#### There are 7 different variants for a Button.
 
 <Canvas>
   <Story name="Variant">
@@ -45,6 +45,7 @@ export const Template = (args) => {
       <Button variant="tertiaryDanger" label="Delete" />
       <Button variant="danger" label="Delete" />
       <Button variant="text" label="Edit" />
+      <Button variant="secondaryBorderless" label="Text" />
     </div>
   </Story>
 </Canvas>
--- a/src/components/Button.stories.mdx
+++ b/src/components/Button.stories.mdx
@@ -45,7 +45,7 @@ export const Template = (args) => {
       <Button variant="tertiaryDanger" label="Delete" />
       <Button variant="danger" label="Delete" />
       <Button variant="text" label="Edit" />
-      <Button variant="secondaryText" label="Text" />
+      <Button variant="textSecondary" label="Text" />
     </div>
   </Story>
 </Canvas>
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -189,7 +189,7 @@ const variantStyles: (contrast: boolean) => Record<
     disabledStyles: Css.lightBlue300.if(contrast).lightBlue700.$,
     focusStyles: Css.bshFocus.if(contrast).boxShadow(`0 0 0 2px ${Palette.White}`).$,
   },
-  secondaryBorderless: {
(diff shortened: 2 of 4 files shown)


src/components/Button.stories.mdx
```diff
@@ -45,6 +45,7 @@ export const Template = (args) => {
       <Button variant="tertiaryDanger" label="Delete" />
       <Button variant="danger" label="Delete" />
       <Button variant="text" label="Edit" />
+      <Button variant="secondaryBorderless" label="Text" />
```
crshannon, line 48: @kbesingeryeh - just curious on the naming of the variant `secondaryBorderless` - Just did some digging around at other design systems and they call this a `text` variant, but I have also seen "flat" or "Ghost" used..

https://ant.design/components/button
https://m3.material.io/components/all-buttons#9134ac95-678e-49ae-a50a-e71948011b05

kbesingeryeh: yes, we have a `text` variant already but its a blue text, so I was thinking since this pattern follows the `secondary` button variant, to call it `secondaryBorderless` since its the same just without a border.. I could also see `secondaryText` being a good name too. Thoughts?

crshannon: yeah I like `secondaryText` as it falls in line with the `tertiaryDanger` pattern.

src/components/Button.tsx
```diff
@@ -189,6 +189,13 @@ const variantStyles: (contrast: boolean) => Record<
     disabledStyles: Css.lightBlue300.if(contrast).lightBlue700.$,
     focusStyles: Css.bshFocus.if(contrast).boxShadow(`0 0 0 2px ${Palette.White}`).$,
   },
+  secondaryText: {
+    baseStyles: Css.gray900.add("fontSize", "inherit").$,
+    hoverStyles: Css.bgGray100.$,
+    pressedStyles: Css.gray900.$,
+    disabledStyles: Css.bgWhite.gray400.$,
+    focusStyles: Css.gray900.$,
+  },
```
bdow, line 198: **question:** Right now, each of our Buttons has a `contrast` variation. We probably don't need this immediately, but would be good to add a `TODO: Handle contrast variant`.

src/components/ButtonModal.stories.tsx
```diff
@@ -91,3 +91,17 @@ export function ButtonModalWithChipAndTooltip() {
     />
   );
 }
+
+export function ButtonModalWithSecondaryText() {
```
bdow, line 95: **nit:** I don't think this story is necessary. I get that this is _why_ we build out this new Button variant. But I don't think it is worth us doing chromatic diffs against, as the existing ButtonModal using a Button story + the updated Button stories should cover the same case overall. 
We can definitely keep this, but I'd recommend we (disable it from chromatic snapshots)[https://www.chromatic.com/docs/ignoring-elements#ignore-stories], just to be conscience of increasing our Chromatic costs.

src/components/Button.tsx
```diff
@@ -204,4 +211,11 @@ const iconStyles: Record<ButtonSize, IconProps["xss"]> = {
...
 export type ButtonSize = "sm" | "md" | "lg";
-export type ButtonVariant = "primary" | "secondary" | "tertiary" | "tertiaryDanger" | "danger" | "text";
+export type ButtonVariant =
+  | "primary"
+  | "secondary"
+  | "tertiary"
+  | "tertiaryDanger"
+  | "danger"
+  | "text"
+  | "secondaryText";
```
bdow, line 221: **thought/nit:** Maybe `textSecondary`? Just to be inline with how `tertiaryDanger` is the "danger" version of our "tertiary" button. It appears that this one is the "secondary" version of our "text" button vs a "text" version of our "secondary" button.

bdow (approved):
A couple nitpicky things, but otherwise looks good.

src/components/Button.tsx
kbesingeryeh: got it. I did ask Andrea about this, but agreed should probably put a note about coming back to this later

src/components/ButtonModal.stories.tsx
kbesingeryeh: thank you!!

src/components/Button.tsx
bdow: I'm realizing that this whole "it's a x of y vs y of x" thought doesn't really make sense. Good how it is!

PR #850, opened 2023-04-14
feat: Export test helpers for Select, MultiSelect and TreeSelectFields

bdow:
Removes the existing SelectField and MultiSelectField mocks that were extra overhead when adding new features to the actual components.


src/utils/rtl.tsx
```diff
@@ -180,3 +181,111 @@ function getTextFromTableCellNode(node: ChildNode) {
...
+  }
+
+  function maybeOpenAndSelect(optionValue: string) {
+    const listboxId = select.getAttribute("aria-controls");
+    const body = select.closest("body") as HTMLElement;
+    const listbox = body.querySelector(`#${listboxId}`) as HTMLElement;
+    if (!listbox) {
+      _click(select);
+    }
+    const options: NodeListOf<HTMLElement> = body.querySelectorAll("[role=option]");
```
stephenh, line 206: Question: Can we scope this to nodes within the `listbox`?

src/utils/rtl.tsx
```diff
@@ -180,3 +181,111 @@ function getTextFromTableCellNode(node: ChildNode) {
...
+    return select.textContent ?? undefined;
+  }
+
+  const listboxId = select.getAttribute("aria-controls");
+  const body = select.closest("body") as HTMLElement;
+  const listbox = body.querySelector(`#${listboxId}`) as HTMLElement;
+  if (!listbox) {
+    _click(select);
+  }
+  const options: NodeListOf<HTMLElement> = body.querySelectorAll("[role=option]");
```
stephenh, line 244: Question: If we have two `SelectField`s on the page, is this going to bring back selections from both select fields?

src/utils/rtl.tsx
```diff
@@ -180,3 +181,111 @@ function getTextFromTableCellNode(node: ChildNode) {
...
+
+  return Array.from(options)
+    .map((o: HTMLElement) => o.dataset.label ?? o.dataset.key ?? "")
+    .filter((o) => !!o);
+}
+
+function resolveIfNeeded(element: HTMLElement): HTMLElement {
```
stephenh, line 280: Fwiw once the React 18 bump re-lands (I'm waiting on zgavin to land a backend PDF PR) I'm planning on de-lazy-ing all of our `r.foo`s to do immediate lookups, so we won't need the `r.foo()` anymore. Looking forward to it!

src/utils/rtl.test.tsx
```diff
@@ -0,0 +1,282 @@
...
+          { id: "2", name: "Two" },
+          { id: "3", name: "Three" },
+        ]}
+      />,
+    );
+    // Then the getOptions helper returns the correct options
+    expect(getOptions(r.number)).toEqual(["One", "Two", "Three"]);
+
+    // When selecting an option
+    select(r.number, "2");
```
stephenh, line 25: Question: Fwiw what do you think about "Two" instead of "2"? I think that's how our current DSL works, for native selects, so will be less code to change/update to use the id instead? Or am I wrong about that...

src/utils/rtl.test.tsx
```diff
@@ -0,0 +1,282 @@
...
+        }}
+        options={[
+          { id: "1", name: "One" },
+          { id: "2", name: "Two" },
+          { id: "3", name: "Three" },
+        ]}
+      />,
+    );
+    // When selecting an option
+    await selectAndWait(r.number, "Two");
```
stephenh, line 84: Praise: neat test!

src/utils/rtl.test.tsx
bdow: You can do both! If you want to select by the label or value of the element you can. 
But you're right, selecting by the label is how we currently do it. But I kind of liked giving the user the option of choosing how they'd want to do it.

src/utils/rtl.tsx
bdow: It should only return this selectfield's options, as the other select fields shouldn't be opened... but might be safer to query from the listbox rather than the body here. I'll update that.

src/utils/rtl.tsx
bdow: Yes, we can. Makes sense to do. I will update.

src/utils/rtl.test.tsx
stephenh: Ah nice! Agreed supporting both sgtm.

src/utils/rtl.tsx
stephenh: Oh I see, sure only one will be open at once, but yep thanks for updating!

stephenh (approved):

PR #859, opened 2023-05-02
fix: Various minor style fixes

bdow:
Prevent Icon and Avatar buttons from adding more height than necessary. Define color for non-contrast menu items (inherited via MenuSection styles) Add 'align-items: center' for MenuItems wrapped in links Add 'label' prop for IconButton for accessibility purposes

### DIFF
--- a/src/components/internal/OverlayTrigger.tsx
+++ b/src/components/internal/OverlayTrigger.tsx
@@ -80,7 +80,8 @@ export function OverlayTrigger(props: OverlayTriggerProps) {
   );
 
   return (
-    <div css={Css.relative.dib.$}>
+    // Add `line-height: 0` to prevent the Icon button and Avatar buttons from inheriting the line-height, causing them to be taller than they should.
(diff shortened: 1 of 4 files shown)


crshannon (approved):
LGTM

src/components/internal/OverlayTrigger.tsx
```diff
@@ -80,7 +80,8 @@ export function OverlayTrigger(props: OverlayTriggerProps) {
   );
 
   return (
-    <div css={Css.relative.dib.$}>
+    // Add `line-height: 0` to prevent the Icon button and Avatar buttons from inheriting the line-height, causing them to be taller than they should.
+    <div css={Css.dib.add("lineHeight", 0).$}>
```
stephenh, line 84: Nit: We don't have `lh0` but we do have `lh(0)` for next time around (no need to do a follow up). Might be cute to add `lh0` / `lh1` for increments :shrug: .

Also great in-source comment!

PR #861, opened 2023-05-05
feat: re-render expanded columns when the columns change

0ces:
With this we will re-render the expanded columns whenever the columns change

### DIFF
--- a/src/components/Table/utils/TableState.ts
+++ b/src/components/Table/utils/TableState.ts
@@ -228,16 +228,16 @@ export class TableState {
       // (We'll add to this after figuring out which columns are new vs existing.)
       const columnIdsToExpand = isInitial ? localStorageColumns : [];
 
-      // Create a list of all existing column ids. (We ignore the `initExpanded` property for existing columns)
-      const existingColumnIds = this.columns.map((c) => c.id);
-
       // Add any columns to our array that are new columns that should be initially expanded.
-      columnIdsToExpand.push(
-        ...columns.filter((c) => !existingColumnIds.includes(c.id) && c.initExpanded).map((c) => c.id),
-      );
+      columnIdsToExpand.push(...columns.filter((c) => c.initExpanded).map((c) => c.id));
+
+      // Clear the cache to force to re-fetch the expanded columns.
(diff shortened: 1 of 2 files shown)


0ces (commented):
@bdow should we keep the `refetchExpandedColumns`? I'd say no as with this it will be automatically made

src/components/Table/utils/TableState.ts
```diff
@@ -228,16 +228,16 @@ export class TableState {
       // (We'll add to this after figuring out which columns are new vs existing.)
       const columnIdsToExpand = isInitial ? localStorageColumns : [];
```
bdow, line 229: **question:** I was thinking that `columnIdsToExpand` would initialize with `this.expandedColumnIds` if `!isInitial`. Would that work here?

bdow:
> @bdow should we keep the `refetchExpandedColumns`? I'd say no as with this it will be automatically made

Yea, I'm okay with removing this.

src/components/Table/utils/TableState.ts
0ces: Yes, it does work

src/components/Table/utils/TableState.ts
```diff
@@ -226,15 +226,13 @@ export class TableState {
       // On initial render the columns we start with is whatever is in local storage.
       // On any subsequent render, we start with an empty array.
       // (We'll add to this after figuring out which columns are new vs existing.)
-      const columnIdsToExpand = isInitial ? localStorageColumns : [];
-
-      // Create a list of all existing column ids. (We ignore the `initExpanded` property for existing columns)
-      const existingColumnIds = this.columns.map((c) => c.id);
```
bdow, line 232: **question:** Is there a reason why we are no longer ignoring the `initExpanded` for existing columns? The idea behind this was that if a user collapsed a column, then the columns rerender we wouldn't want to expand it s the user explicitly collapsed it.

0ces: No, there is no reason for that change, I'll leave it as it was

bdow (approved):

PR #863, opened 2023-05-08
feat: SC-33938: ScrollShadows component

bdow:
New component that adds shadows to a container to indicate it can be scrolled. It also hides the scrollbar.

Adds stories, but not tests as this is only presentational

### DIFF
--- a/src/components/ScrollShadows.stories.tsx
+++ b/src/components/ScrollShadows.stories.tsx
@@ -6,7 +6,7 @@ export default {
   title: "Components/ScrollShadows",
 };
 
-export function Vertical() {
+export function Examples() {
   return (
     <>
       <h2 css={Css.mb1.lgSb.$}>Vertical scroll</h2>
--- a/src/components/ScrollShadows.stories.tsx
+++ b/src/components/ScrollShadows.stories.tsx
@@ -1,0 +1,41 @@
+import { Css } from "src";
+import { ScrollShadows } from "src/components/ScrollShadows";
+
+export default {
+  component: ScrollShadows,
+  title: "Components/ScrollShadows",
+};
+
+export function Vertical() {
+  return (
+    <>
+      <h2 css={Css.mb1.lgSb.$}>Vertical scroll</h2>
+      <ScrollShadows xss={Css.hPx(200).wPx(400).bgWhite.p1.ba.bGray200.$}>
+        {"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec euismod, nisl eget ultricies aliquam, nunc ".repeat(
+          20,
+        )}
+      </ScrollShadows>
+
+      <div css={Css.mt3.df.fdc.bgWhite.ba.bGray200.bshBasic.br8.overflowHidden.hPx(200).wPx(400).$}>
+        <div css={Css.pPx(20).pb2.bgWhite.fw6.$}>Fixed Header</div>
+        <ScrollShadows xss={Css.pxPx(20).fg1.$}>
+          {"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec euismod, nisl eget ultricies aliquam, nunc ".repeat(
+            20,
+          )}
+        </ScrollShadows>
+      </div>
+
+      <h2 css={Css.mt8.mb1.lgSb.$}>Horizontal scroll</h2>
+      <ScrollShadows xss={Css.wPx(200).bgWhite.p1.ba.bGray200.nowrap.$} horizontal>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec euismod, nisl eget ultricies aliquam, nunc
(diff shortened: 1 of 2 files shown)


src/components/ScrollShadows.tsx
```diff
@@ -0,0 +1,103 @@
...
+  /** Set to true if the container scrolls horizontally */
+  horizontal?: boolean;
+  /** Defines the background color for the shadows */
+  bgColor?: Palette;
+}
+export function ScrollShadows(props: ScrollShadowsProps) {
+  const { children, xss, horizontal = false, bgColor = Palette.White } = props;
+  const { height = "auto", width = "auto" } = xss ?? {};
+  // This is admittedly extremely hacky. It expects the background color to be in the format "rgba(255, 255, 255, 1)".
+  // If we ever change how we define our color palette in Beam, then this will break and will need to be fixed.
```
stephenh, line 18: Suggestion: Makes sense, maybe we can detect a non-rgba value and throw an error? Np if that's not worth it.

src/components/ScrollShadows.tsx
```diff
@@ -0,0 +1,103 @@
...
+              `linear-gradient(${endGradientDeg}deg, ${bgColor} 0%, ${transparentBgColor} 92%)`,
+            ).$,
+          }}
+        />
+      )}
+
+      <div
+        css={{
+          ...xss,
+          ...Css.overflowAuto.fg1.addIn("&::-webkit-scrollbar", { display: "none" }).add("scrollbarWidth", "none").$,
```
stephenh, line 85: Thought: Ah yeah, per our musings about making this a `.scrollbarHidden` truss method, if everyone just uses the `ScrollShadows` component anyway, then seems fine to just use the more verbose `addIn`+`add` for the one line of code. (Unless you're anticipating non-`ScrollShadows` also needing this?)

src/components/ScrollShadows.stories.tsx
```diff
@@ -0,0 +1,41 @@
+import { Css } from "src";
+import { ScrollShadows } from "src/components/ScrollShadows";
+
+export default {
+  component: ScrollShadows,
+  title: "Components/ScrollShadows",
+};
+
+export function Vertical() {
```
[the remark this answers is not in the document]
khalidwilliams: Whoops I see the example down lower. Maybe just changing the name to something more general, like `Basic` or `Vertical and Horizontal`

src/components/ScrollShadows.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }, []);
+
+  // Initially set the state dimensions on render to put the shadows in the correct position
+  useLayoutEffect(() => {
+    scrollRef.current && updateScrollProps(scrollRef.current);
+  }, []);
+
+  const { scrollTop, scrollHeight, clientHeight, scrollWidth, scrollLeft, clientWidth } = scrollProps;
+  const showStartShadow = !horizontal ? scrollTop > 0 : scrollLeft > 0;
+  const showEndShadow = !horizontal ? scrollTop + clientHeight < scrollHeight : scrollLeft + clientWidth < scrollWidth;
```
stephenh, line 48: Praise: Nice showstart/end variables! Makes this pretty readable.

...

Could potentially do:

```
const start = horizontal ? scrollLeft : scrollTop;
const len = horizontal ? clientWidth : clientHeight;
const end = horizontal ? scrollWidth : scrollHeight;
const showStartShadow = start > 0;
const showEndShadow = start + len < end;
```

...could maybe even work ^ into the `useState`, like instead of storing all 6 in the `useState`, just store the three `start`/`len`/`end`, and do the `horizontal`-yes/no check during `updateScrollProps`.

stephenh (approved):
Neat!

khalidwilliams (approved):

src/components/ScrollShadows.tsx
khalidwilliams: What would be the upside of using a `.scrollBarHidden` instead? Or do you mean doing this as written vs adding a new `scrollbarHidden` method to truss?

src/components/ScrollShadows.tsx
bdow: Yea, I like that. I will update.

src/components/ScrollShadows.tsx
bdow: If we added it into `truss`, then it could be used in other places. Though, hopefully whenever someone needs to hide the scrollbars, they'll just use this component.

src/components/ScrollShadows.tsx
bdow: Yea, I like that idea. I will update the state.

bdow: I think I might change this to using two states: `showStartShadow` and `showEndShadow`. That might make this a bit cleaner.

PR #867, opened 2023-05-10
fix: Add zIndex to ScrollShadows

bdow:
Refactors styles in hopes of not regenerating them every time the state changes / component re-renders

### DIFF
--- a/src/components/ScrollShadows.tsx
+++ b/src/components/ScrollShadows.tsx
@@ -1,5 +1,5 @@
 import { useResizeObserver } from "@react-aria/utils";
-import { ReactNode, useCallback, useRef, useState } from "react";
+import { ReactNode, useCallback, useMemo, useRef, useState } from "react";
 import { Css, Palette, Properties, useTestIds } from "src";
 
 interface ScrollShadowsProps {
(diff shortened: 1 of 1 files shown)


src/components/ScrollShadows.tsx
```diff
@@ -22,17 +22,24 @@ export function ScrollShadows(props: ScrollShadowsProps) {
...
-  const endShadowStyles = !horizontal ? Css.bottom0.left0.right0.hPx(40).$ : Css.right0.top0.bottom0.wPx(25).$;
-  const startGradientDeg = !horizontal ? 180 : 90;
-  const endGradientDeg = !horizontal ? 0 : 270;
-
   const [showStartShadow, setShowStartShadow] = useState(false);
   const [showEndShadow, setShowEndShadow] = useState(false);
-
   const scrollRef = useRef<HTMLDivElement>(null);
 
+  const [startShadowStyles, endShadowStyles] = useMemo(() => {
```
stephenh, line 29: Question: Is the idea with the `useMemo` that these were expensive, or unintentionally changing slightly between renders? Might be worth an in-source comment, just b/c if I read this 6 months from now, I think my question would be "this seems cheap, why memo?".

bdow: I've been leaning towards the camp of memoize everything. These are not changing between renders, and didn't stick out as expensive, but they'll most likely never need to change so there is no point in recomputing them. If you have strong opinions, I'm fine with removing the `useMemo`.

src/components/ScrollShadows.tsx
```diff
@@ -22,17 +22,24 @@ export function ScrollShadows(props: ScrollShadowsProps) {
...
+    const transparentBgColor = bgColor.replace(/,1\)$/, ",0)");
+    const commonStyles = Css.absolute.z3.$;
```
bdow, line 31: @stephenh - The zIndex (`z3`) is set here.

stephenh: Ack! I grepped for `zIndex` and forgot our shortcut. :speak_no_evil: :-D Thanks!

kbesingeryeh (approved):

src/components/ScrollShadows.tsx
stephenh: Ah gotcha, cool; just the way the PR description was worded, I misread it as like maybe there was a behavior fix here.

This specific `useMemo` lgtm, but I'm hesitant about "memoize everything", b/c the React docs specifically warn about its over-use, with the hint that often times the overhead of creating/tracking the infra of a `useMemo` hook is higher than the calculation of "just create/combine some object literals" that is being cached.

We're doing enough here that, sure, the `useMemo` is probably cheaper? But who knows.

PR #868, opened 2023-05-11
feat: On MenuItem, react-router NavLink was changed to BEAM NavLink

ipiedrahb:
Context of the change [SC-34164](https://app.shortcut.com/homebound-team/story/34164/cmd-click-on-menu-items-under-libraries-does-not-open-pages-in-a-new-tab), after that, I made some changes
- Replace react-router-dom <NavLink with  react-router-dom <Link `<a role="button"`, that is the reason of the issue (I searched and only this component was using this version)
- Update the storybook to show all the variations on a menuItems

Please see the video [DEMO BUG/FIX](https://www.awesomescreenshot.com/video/17326734?key=5de1eb65374829ac52f8debc163c7dbc)

[Menu Storybook](https://60466f7d43c37c0021788867-rmtwfusvhr.chromatic.com/?path=/story/workspace-components-menu--basic-menu-items)

Screen Shot 2023-05-11 at 16 39 28

### DIFF
--- a/src/components/NavLink.tsx
+++ b/src/components/NavLink.tsx
@@ -25,7 +25,7 @@ export interface NavLinkProps extends BeamFocusableProps {
   buttonRef?: RefObject<HTMLElement>;
 }
 
-type NavLinkVariant = "side" | "global" | "normal";
+type NavLinkVariant = "side" | "global";
 
 export function NavLink(props: NavLinkProps) {
   const {
@@ -84,7 +84,7 @@ export function getNavLinkStyles(variant: NavLinkVariant, contrast: boolean) {
   return navLinkVariantStyles(contrast)[variant];
 }
 
-const baseStyles = Css.df.aic.hPx(32).pyPx(6).br4.smMd.outline0.$;
+const baseStyles = Css.df.aic.hPx(32).pyPx(6).px1.br4.smMd.outline0.$;
 
 const navLinkVariantStyles: (contrast: boolean) => Record<
   NavLinkVariant,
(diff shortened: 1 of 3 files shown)


chr1sjf0x (commented):
TY @ipiedrahb - Looks good to me.

Thoughts @bdow

src/components/NavLink.tsx
```diff
@@ -25,7 +25,7 @@ export interface NavLinkProps extends BeamFocusableProps {
   buttonRef?: RefObject<HTMLElement>;
 }
 
-type NavLinkVariant = "side" | "global";
+type NavLinkVariant = "side" | "global" | "normal";
```
bdow, line 28: **thought:** I'm hesitant to add a new `NavLink` variant. These are suppose to be links for our application navigation, like the Global Nav, or the Project Side Nav. `normal` makes me wonder, why not just use a `<a href="..." />` instead?

src/components/internal/MenuItem.tsx
```diff
@@ -167,9 +167,7 @@ function maybeWrapInLink(
       </span>
     </a>
   ) : (
-    <NavLink to={onClick} className="navLink">
-      {content}
-    </NavLink>
```
bdow, line 172: **suggestion:** I think if we would've just changed this from `NavLink` to `Link` (both imported from `react-route-dom`), then this would work.

bdow: Let me know if that does work. Happy to test it out with you!

bdow: Actually, now I'm seeing that this is probably due to `onAction` happening here... not the link element itself. Going to try testing it as well.

ipiedrahb: Hi @bdow 

I replaced NavLink with Link and works fine!, Please check the [Preview Branch](https://60466f7d43c37c0021788867-tvtxvdqrmf.chromatic.com/?path=/story/workspace-components-menu--basic-menu-items)

chr1sjf0x (approved):
ty @ipiedrahb

PR #869, opened 2023-05-16
fix: resolve issue with ButtonMenu is close then open the menu and use cmd + click to open another tab

ipiedrahb:
[Bug Video](https://www.awesomescreenshot.com/video/17454917?key=adfe53f0d6092a81cd103f2618aabe53)

[Solution Video](https://www.awesomescreenshot.com/video/17454953?key=8c1749092074b807e9ba9867c3373935)

Resource => https://github.com/adobe/react-spectrum/issues/1244#issuecomment-1125813249

Example => https://codesandbox.io/s/react-spectrum-menu-link-fvim4d?file=/src/App.tsx:4803-4862

On internat-frontend => https://www.awesomescreenshot.com/video/17455038

### DIFF
--- a/src/components/internal/Menu.stories.tsx
+++ b/src/components/internal/Menu.stories.tsx
@@ -34,6 +34,21 @@ export function BasicMenuItems() {
   );
 }
 
+export function BasicMenuItemsClose() {
+  return (
+    <ButtonMenu
+      trigger={{ label: "Menu Trigger" }}
+      items={[
+        { label: "Menu item 1", onClick: "?test=true" },
+        { label: "Menu item 2", onClick: "https://google.com" },
+        { label: "Menu item 3", onClick: noop, disabled: true },
+        { label: "Menu item 4", onClick: noop },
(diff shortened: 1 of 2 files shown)


chr1sjf0x:
> The issue can be resolved with the description of this [blog](https://github.com/adobe/react-spectrum/issues/1244#issuecomment-1125813249)

@ipiedrahb What a strange fix. 

Could you please:
1) declare a 'const noopWorkaround = () => {}` with a comment explaining the reason and linking the comment above. 
2) pass that const instead of the anonymous methods you currently have. 

As-is, I'd be concerned someone will just delete those without realizing they are resolving this issue.

ipiedrahb:
Hi @chr1sjf0x 
Yes, it seems strange and many people will implement it like this to solve it.

thank you!

chr1sjf0x (approved):
TY @ipiedrahb

chr1sjf0x (approved):
TY @ipiedrahb

PR #874, opened 2023-05-18
feat(button-group): colors

TylerR909:
Support for

image


src/components/ButtonGroup.tsx
```diff
@@ -73,7 +75,9 @@ function GroupButton(props: GroupButtonProps) {
             }}
             {...tid[defaultTestId(text ?? icon ?? "button")]}
           >
-            {icon && <Icon icon={icon} />}
+            {icon && (
+              <Icon xss={{ marginRight: "4px" }} icon={icon} color={disabled ? undefined : iconColor} inc={iconInc} />
```
TylerR909, line 79: image

stephenh: Looking at the stories, this margin right has made icon-only button groups look off center? Should it be conditional, only if there is text?

stephenh: `Css.if(text).mrPx(4).$` :-)

stephenh (approved):

PR #875, opened 2023-05-18
feat: Expose virtuoso infinite scroll props for GridTable

blambillotte:


src/components/Table/types.ts
```diff
@@ -116,3 +116,10 @@ export type Pin = { at: "first" | "last"; filter?: boolean };
 
 // Use IfAny so that GridDataRow<any> doesn't devolve into any
 export type IfAny<T, Y, N> = 0 extends 1 & T ? Y : N;
+
+export type InfiniteScroll = {
+  /** will be called when the user scrolls to the end of the list with the last item index as an argument.  */
+  onEndReached: (index: number) => void;
+  /** The number of pixels from the bottom of the list to eagerly trigger `onEndReached`. The default is is 500px. */
```
stephenh, line 123: Thought: Wonder if we could auto-detect this? Or does virtuoso already do that? Just thinking 500px seems pretty big, and could be like half the vertical screen for smaller screens? Or is that okay? 

I also wonder if we could provide more opinionated/high-level options, like `infiniteScrollUseCase: main-table | side-nav | something`.

chr1sjf0x (approved):
TY @blambillotte 🍣

stephenh (approved):

src/components/Table/types.ts
blambillotte: Yeah hmm so it's really a bit of a guess at the user's vertical scroll speed and how long we expect a certain response to take, as you could scroll 500px vertically in a fraction of a second. The aim is to have the next page loaded before they reach the end to achieve a smooth experience. Maybe we just default to 0 and let the implementor decide?

src/components/Table/GridTable.tsx
```diff
@@ -157,6 +158,13 @@ export interface GridTableProps<R extends Kinded, X> {
    * This is beneficial when looking at the same table, but of a different subject (i.e. Project A's PreCon Schedule vs Project A's Construction schedule)
    */
   visibleColumnsStorageKey?: string;
+  /**
+   * Infinite scroll is only supported with `as=virtual` mode
+   *
+   ** `onEndReached` will be called when the user scrolls to the end of the list with the last item index as an argument.
+   ** `endOffsetPx` is the number of pixels from the bottom of the list to eagerly trigger `onEndReached`. The default is is 500px.
+   */
+  infiniteScroll?: InfiniteScroll;
```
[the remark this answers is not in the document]
blambillotte: hmm yeah ideally the `onEndReached` func is wrapped in a `useCallback` 🤔 

I originally grouped these props under `infiniteScroll` to try and communicate their expected use, but maybe it just makes more sense to have top level props like `virtualOnEndReached` and `virtualIncreaseBottomViewportBy`? That way the 2nd prop would just be a const and we don't need to worry memoing the value

0ces (approved):

PR #876, opened 2023-05-20
refactor: Refactor Pagination component

stephenh:
Most of my hackday was BE paging, but this changes some of the FE, specifically:

* Combines the previous `settings` and `setSettings` props into a single `page` tuple
* Supports both `pageNumber`+`pageSize` as well as `offset`+`limit`, based on caller pref
* Defaults to `offset`+`limit` since that is what our BE filters want
* Reduces # of page size options, plus makes each increment larger/further away from each other
* Removes the `label` prop as unnecessary
* Fixes some boundary cases in the "x - y of z" on the last page of data / only a single page of data

### DIFF
--- a/src/components/Pagination.stories.tsx
+++ b/src/components/Pagination.stories.tsx
@@ -1,7 +1,6 @@
 import { Meta } from "@storybook/react";
 import { useState } from "react";
-import { noop } from "src/utils";
-import { defaultPage, FormLines, TextField } from "..";
+import { defaultPage, FormLines, StaticField } from "..";
 import { Pagination } from "./Pagination";
 
 export default {
@@ -34,10 +33,10 @@ function RenderPagination({ totalRows }: { totalRows: number }) {
   return (
     <>
       <FormLines labelStyle="left">
-        <TextField label={"Number of items"} readOnly value={String(totalRows)} onChange={noop} />
-        <TextField label={"Current page"} readOnly value={String(settings.pageNumber)} onChange={noop} />
+        <StaticField label="Number of items" value={String(totalRows)} />
+        <StaticField label="Current page" value={String(settings.pageNumber)} />
       </FormLines>
-      <Pagination totalCount={totalRows} setSettings={setSettings} settings={settings} />
+      <Pagination totalCount={totalRows} setPage={setSettings} page={settings} />
     </>
   );
 }
--- a/src/components/Pagination.stories.tsx
+++ b/src/components/Pagination.stories.tsx
@@ -1,11 +1,8 @@
 import { Meta } from "@storybook/react";
 import { useState } from "react";
 import { noop } from "src/utils";
-import { withBeamDecorator } from "src/utils/sb";
-import { FormLines, TextField } from "..";
-import { PageSettings, Pagination } from "./Pagination";
-
-const initPageSettings = { page: 1, perPage: 100 };
+import { defaultPage, FormLines, TextField } from "..";
+import { Pagination } from "./Pagination";
 
 export default {
   title: "Workspace/Components/Pagination",
@@ -18,34 +15,29 @@ export default {
       url: "https://www.figma.com/file/aWUE4pPeUTgrYZ4vaTYZQU/%E2%9C%A8Beam-Design-System?node-id=3214%3A10519",
     },
   },
-  decorators: [withBeamDecorator],
 } as Meta;
 
 export function WithPages() {
-  return <RenderPagination totalRows={999} label="items" />;
+  return <RenderPagination totalRows={999} />;
 }
 
 export function NoPages() {
-  return <RenderPagination totalRows={0} label="items" />;
+  return <RenderPagination totalRows={0} />;
 }
 
 export function OnlyOnePage() {
   return <RenderPagination totalRows={14} />;
 }
 
-function RenderPagination({ totalRows, label }: { totalRows: number; label?: string }) {
-  const [settings, setSettings] = useState<PageSettings>({
-    page: initPageSettings.page,
-    perPage: initPageSettings.perPage,
-  });
-
+function RenderPagination({ totalRows }: { totalRows: number }) {
+  const [settings, setSettings] = useState(defaultPage);
   return (
     <>
       <FormLines labelStyle="left">
         <TextField label={"Number of items"} readOnly value={String(totalRows)} onChange={noop} />
-        <TextField label={"Current page"} readOnly value={String(settings.page)} onChange={noop} />
+        <TextField label={"Current page"} readOnly value={String(settings.pageNumber)} onChange={noop} />
       </FormLines>
-      <Pagination totalCount={totalRows} label={label ?? "items"} setSettings={setSettings} settings={settings} />
+      <Pagination totalCount={totalRows} setSettings={setSettings} settings={settings} />
     </>
   );
 }
--- a/src/components/Pagination.stories.tsx
+++ b/src/components/Pagination.stories.tsx
@@ -1,6 +1,6 @@
 import { Meta } from "@storybook/react";
 import { useState } from "react";
-import { defaultPage, FormLines, StaticField } from "..";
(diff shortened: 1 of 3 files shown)


src/components/Pagination.stories.tsx
```diff
@@ -18,34 +14,29 @@ export default {
       url: "https://www.figma.com/file/aWUE4pPeUTgrYZ4vaTYZQU/%E2%9C%A8Beam-Design-System?node-id=3214%3A10519",
     },
   },
-  decorators: [withBeamDecorator],
 } as Meta;
 
 export function WithPages() {
-  return <RenderPagination totalRows={999} label="items" />;
```
stephenh, line 25: I removed label b/c it seems noisy and needlessly clutters the UI

src/components/Pagination.stories.tsx
```diff
@@ -18,34 +14,29 @@ export default {
...
-  });
-
+function RenderPagination({ totalRows }: { totalRows: number }) {
+  const page = useState(defaultPage);
   return (
     <>
       <FormLines labelStyle="left">
-        <TextField label={"Number of items"} readOnly value={String(totalRows)} onChange={noop} />
-        <TextField label={"Current page"} readOnly value={String(settings.page)} onChange={noop} />
+        <StaticField label="Number of items" value={String(totalRows)} />
```
stephenh, line 36: Used StaticField instead of TextField with noop onChange

src/components/Pagination.test.tsx
```diff
@@ -1,110 +1,105 @@
...
-import { pageOptions, Pagination, toFirstAndOffset } from "./Pagination";
+import { pageOptions, PageSettings, Pagination, toFirstAndOffset } from "./Pagination";
 
-const initPageSettings = { page: 1, perPage: 100 };
+const init: PageSettings = { pageNumber: 1, pageSize: 100 };
 
 describe("Pagination", () => {
   it("can have a data-testid", async () => {
-    const r = await render(
-      <Pagination totalCount={10} label="" setSettings={noop} settings={initPageSettings} data-testid="pag" />,
```
stephenh, line 10: I combined the two `settings= setSettings=` into a single `page=` prop that accepts the tuple of the two. This allows callers to just do:

```
const page = useState(defaultPage);

              <Pagination
                page={page}
                totalCount={totalCount}
              />
```

src/components/Pagination.tsx
```diff
@@ -1,78 +1,85 @@
...
 import { IconButton } from "src/components";
 import { Css, Palette } from "src/Css";
 import { SelectField } from "src/inputs";
 import { useTestIds } from "src/utils";
 
 export type PageSettings = {
-  page: number;
-  perPage: number;
+  pageNumber: number;
+  pageSize: number;
```
stephenh, line 9: Renamed these

src/components/Pagination.tsx
```diff
@@ -1,78 +1,85 @@
...
 export function Pagination(props: PaginationProps) {
-  const { label, settings, totalCount, setSettings } = props;
-  const { perPage, page } = settings;
+  const {
+    page: [page, setPage],
+    totalCount,
+  } = props;
+  const { pageSize, pageNumber } = page;
 
-  const hasNextPage = useMemo(() => page < totalCount / perPage, [page, perPage, totalCount]);
```
stephenh, line 23: This calc is super-cheap so de-useMemo-d

src/components/Pagination.tsx
```diff
@@ -1,78 +1,85 @@
...
-          {perPage * (page - 1) + 1} {hasNextPage ? `- ${perPage * page}` : ""} of {totalCount} {label}
+          {first} {showLast ? `- ${last}` : ""} of {totalCount}
         </div>
         <IconButton
-          {...tid.previousIcon}
           icon="chevronLeft"
-          color={Palette.LightBlue700}
-          onClick={() => setSettings({ page: settings.page - 1, perPage })}
-          disabled={settings.page === 1}
+          color={hasPrevPage ? Palette.LightBlue700 : Palette.Gray200}
```
stephenh, line 59: The color was flipping to gray on disabled

src/components/Pagination.tsx
```diff
@@ -1,78 +1,107 @@
...
+  // Don't both with `0 - 0 of 0`
+  const showLast = totalCount > 0;
 
-  const hasNextPage = useMemo(() => page < totalCount / perPage, [page, perPage, totalCount]);
+  // Convert the new page back to whatever format the caller wants
+  function set(newPage: PageNumberAndSize): void {
+    if ("pageNumber" in props.page[0]) {
+      setPage(newPage as any);
+    } else {
+      setPage(toLimitAndOffset(newPage) as any);
```
bdow, line 43: **question:** Can we do something besides the `as any` here? Wondering if this was something that you added while it was a WIP, or we really just need to do this to make TS happy.

src/components/Pagination.tsx
```diff
@@ -1,78 +1,107 @@
...
-          {perPage * (page - 1) + 1} {hasNextPage ? `- ${perPage * page}` : ""} of {totalCount} {label}
+          {first} {showLast ? `- ${last}` : ""} of {totalCount}
         </div>
         <IconButton
-          {...tid.previousIcon}
           icon="chevronLeft"
-          color={Palette.LightBlue700}
-          onClick={() => setSettings({ page: settings.page - 1, perPage })}
-          disabled={settings.page === 1}
+          color={hasPrevPage ? Palette.LightBlue700 : Palette.Gray200}
```
bdow, line 70: **thought:** Maybe IconButton should auto set the color to `Gray200` when the button is `disabled`.

bdow (approved):

src/components/Pagination.tsx
stephenh: I tried, but nope...not entirely sure why...really seems like it should worked.

src/components/Pagination.tsx
stephenh: Ah yeah, agreed...it looks like it _would_ do that:

```
    <Icon icon={icon} color={color || (isDisabled ? Palette.Gray400 : iconColor)} inc={compact ? 2 : inc} />
```

But we skip that once `color` is set, which we're setting here b/c the default color is gray:

```
  const iconColor = contrast ? contrastIconColor : defaultIconColor;
...
   const defaultIconColor = Palette.Gray900;
```

Not sure why our default icon button color is gray 900? Seems like if its clickable, it should be blue?

src/components/Pagination.tsx
```diff
@@ -1,78 +1,107 @@
...
+  const { totalCount } = props;
+  const [page, setPage] = props.page;
+  const { pageSize, pageNumber } = toPageNumberSize(page);
+
+  const hasPrevPage = pageNumber > 1;
+  const hasNextPage = pageNumber < totalCount / pageSize;
+  // Create the `1 - 100 of 1000` or `0 of 0`
+  const first = pageSize * (pageNumber - 1) + (totalCount ? 1 : 0);
+  const last = Math.min(pageSize * pageNumber, totalCount);
+  // Don't both with `0 - 0 of 0`
```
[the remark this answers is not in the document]
stephenh: Yep, fixed!

src/components/Pagination.stories.tsx
stephenh: Yeah, good idea, I'll put it on the list to go find the canonical pagination figma and make a comment...

bdow: (https://www.figma.com/file/aWUE4pPeUTgrYZ4vaTYZQU/%E2%9C%A8Beam-Component-Library?type=design&node-id=32547-99453&t=pskEtucLTxhgvJE4-0)

stephenh: Ah thanks! Actually looking at those mocks, I don't think they have a per-page `label` anyway, like it just says "items" which imo sounds pretty generic. So I think this actually makes the component more like the figma than less. :tada: :-)

PR #878, opened 2023-05-25
feat: Replace GridTable default styling

bdow:
Default and 'condensed' table styles now use the latest patterns.

Includes some various simplifications

### DIFF
--- a/src/components/Table/TableStyles.tsx
+++ b/src/components/Table/TableStyles.tsx
@@ -36,10 +36,6 @@ export interface GridStyle {
   firstCellCss?: Properties;
   /** Applied to the last cell of all rows, i.e. for table-wide padding or right-side borders. */
   lastCellCss?: Properties;
-  /** Applied to a cell div when `indent: 1` is used. */
-  indentOneCss?: Properties;
-  /** Applied to a cell div when `indent: 2` is used. */
-  indentTwoCss?: Properties;
   /** Applied if there is a fallback/overflow message showing. */
   firstRowMessageCss?: Properties;
   /** Applied on hover if a row has a rowLink/onClick set. */
@@ -179,8 +175,6 @@ export interface RowStyle<R extends Kinded> {
   cellCss?: Properties | ((row: R) => Properties);
   /** Renders the cell element, i.e. a link to get whole-row links. */
   renderCell?: RenderCellFn<R>;
-  /** Whether the row should be indented (via a style applied to the 1st cell). */
-  indent?: 1 | 2;
   /** Whether the row should be a link. */
   rowLink?: (row: R) => string;
   /** Fired when the row is clicked, similar to rowLink but for actions that aren't 'go to this link'. */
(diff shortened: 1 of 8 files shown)


src/components/Table/TableStyles.tsx
```diff
@@ -218,17 +219,6 @@ export const cardStyle: GridStyle = {
   },
 };
 
-/** GridTable as Table utility to apply <tr> element override styles. */
-export function tableRowStyles(as: RenderAs, column?: GridColumn<any>) {
```
bdow, line 222: This moved from being called in _every_ cell to now being applied in the `cellCss` defined in `Row.tsx`.

src/components/Table/components/cell.tsx
```diff
@@ -70,12 +65,7 @@ export const headerRenderFn: (column: GridColumnWithId<any>, as: RenderAs, colSp
   (column, as, colSpan) => (key, css, content, row, rowStyle, classNames: string | undefined, onClick, tooltip) => {
     const Cell = as === "table" ? "th" : "div";
     return (
-      <Cell
-        key={key}
-        css={{ ...css, ...tableRowStyles(as, column) }}
-        className={classNames}
-        {...(as === "table" && { colSpan })}
-      >
+      <Cell key={key} css={{ ...css }} className={classNames} {...(as === "table" && { colSpan })}>
```
bdow, line 68: For whatever reason this could become `css={css}` due to its type being `Properties`. It expected something else, but I didn't feel like digging into it too deep. I'm hoping we'll update these at somepoint to be their own components, so I didn't want to spend too much time on it.

src/components/Table/utils/utils.tsx
```diff
@@ -102,7 +102,7 @@ export function toContent(
         {tooltipEl}
       </>
     );
-  } else if (style.emptyCell && isContentEmpty(content)) {
+  } else if (!isHeader && !isExpandableHeader && style.emptyCell && isContentEmpty(content)) {
```
bdow, line 105: It seems like it'd be odd for an empty header and expandable header to show the `emptyCell` style. It makes more sense in a `totals` row, or in cell within the table's body. Hopefully this eliminates the need to remember to use `header: emptyCell`.

stephenh (approved):
Awesome!

src/components/Table/utils/utils.tsx
bdow: That shouldn't be the case if the value is actually `0`, especially if the value is `$0` (a string). I would guess that is happening locally. GridTable only shows the empty cell dash when it is an empty string, null, or undefined.

This is the code that is used to determine it:
```
const emptyValues = ["", null, undefined] as any[];

function isContentEmpty(content: ReactNode): boolean {
  return emptyValues.includes(content);
}
```

stephenh: Ah! That makes sense, thanks! Yep, I see where we're explicitly passing `undefined` instead of `0`. Will fix it there.

PR #879, opened 2023-05-26
feat: Chips/Chip xss and compact prop

Levi-Huynh:
[SC-35341](https://app.shortcut.com/homebound-team/story/35341/beam-chips-grouped-and-chip-style-improvements)

1. Add "color" | "backgroundColor" to Chip (singular) XSS
2. Add compact prop to change Chips (group) size
3. New "info" type to ChipTypes

Screen Shot 2023-05-26 at 2 55 52 PM

Screen Shot 2023-05-26 at 3 01 29 PM

Screen Shot 2023-05-30 at 11 45 54 AM

### DIFF
--- a/src/components/Chip.test.tsx
+++ b/src/components/Chip.test.tsx
@@ -23,4 +23,9 @@ describe("Chip", () => {
     expect(r.chip().textContent).toBe("Chip text content");
     expect(r.tooltip()).toHaveAttribute("title", "title is different");
   });
+
+  it("can set background color and color with xss prop", async () => {
+    const r = await render(<Chip text="Chip" xss={Css.lightBlue100.bgLightBlue100.$} />);
+    expect(r.chip()).toHaveStyle(`background: ${Css.lightBlue100.bgLightBlue100.$}`);
+  });
 });
--- a/src/components/Chip.test.tsx
+++ b/src/components/Chip.test.tsx
@@ -26,6 +26,6 @@ describe("Chip", () => {
 
   it("can set background color and color with xss prop", async () => {
     const r = await render(<Chip text="Chip" xss={Css.lightBlue100.bgLightBlue100.$} />);
-    expect(r.chip()).toHaveStyle(`background: ${Css.lightBlue100.bgLightBlue100.$}`);
+    expect(r.chip()).toHaveStyle(Css.lightBlue100.bgLightBlue100.$);
   });
 });
--- a/src/components/Chip.tsx
+++ b/src/components/Chip.tsx
@@ -4,7 +4,7 @@ import { maybeTooltip } from "src/components/Tooltip";
 import { Css, Margin, Only, Properties, Xss } from "src/Css";
 import { useTestIds } from "src/utils/useTestIds";
 
-type ChipType = "caution" | "warning" | "success" | "light" | "dark" | "neutral" | "darkMode";
+export type ChipType = "caution" | "warning" | "success" | "light" | "dark" | "neutral" | "darkMode";
 
 // exporting for using in type prop as constant - this could be moved and become a global list for colors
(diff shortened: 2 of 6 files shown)


stephenh (approved):
Lgtm!

src/components/Chip.tsx
```diff
@@ -57,4 +61,5 @@ const typeStyles: Record<ChipType, Properties> = {
   dark: Css.bgGray900.white.$,
   neutral: Css.bgGray200.$,
   darkMode: Css.bgGray700.white.$,
+  info: Css.bgLightBlue100.$,
```
Levi-Huynh, line 64: @bdow i think it would be easier for us to just add the new "info" type directly to the chipType (w/ how we need to use it on the FE as a status tag)

let me know if you want me to remove the "color" | "backgroundColor" xss update or keep it there as an escape hatch for other uses

bdow: Yea, that's fine with me. 
If we add it as a supported type here, then we should also include it in the Beam Figma. Can you work with your designer to get it added if you decide to go this route?

Levi-Huynh: Thanks, @bdow yes will work with Radhi to update it beam figma 👍 

would you like me to remove the "color" | "backgroundColor" xss change or keep that?

bdow: We can keep it. I think it'll be nice to have that escape hatch if we really need a one-off. Thanks!

src/components/Chip.test.tsx
```diff
@@ -23,4 +23,9 @@ describe("Chip", () => {
     expect(r.chip().textContent).toBe("Chip text content");
     expect(r.tooltip()).toHaveAttribute("title", "title is different");
   });
+
+  it("can set background color and color with xss prop", async () => {
+    const r = await render(<Chip text="Chip" xss={Css.lightBlue100.bgLightBlue100.$} />);
+    expect(r.chip()).toHaveStyle(`background: ${Css.lightBlue100.bgLightBlue100.$}`);
```
bdow, line 29: I think this is a false positive. The output of `Css.lightBlue100.bgLightBlue100.$` is:
```
{
  backgroundColor: 'rgba(224,242,254,1)',
  color: 'rgba(224,242,254,1)'
 }
```

I just pulled this down and saw that you can put any value in there, and it'll pass. For example, this passes: `expect(r.chip()).toHaveStyle("background: anything");` I'm not 100% sure why that is the case. Seems like a bug based on the documentation for `toHaveStyle`.

I would suggest updating this test to:
```suggestion
    expect(r.chip()).toHaveStyle(Css.lightBlue100.bgLightBlue100.$);
```
That should work correctly.

src/components/Chip.tsx
bdow: Just realized you probably grabbed "info" from the Tag component with the same background-color, maybe? I'd say feel free to merge this as we already have a precedent for a `type="info"` with this color defined. 👍

src/components/Chips.test.tsx
```diff
@@ -0,0 +1,23 @@
+import { render } from "src/utils/rtl";
+import { Chips } from "./Chips";
+
+describe("Chips", () => {
```
bdow, line 4: **praise:** Oh wow! I am surprised we didn't have a test file for this. I guess we previously thought that the Storybook file was sufficient. Thanks for adding!

bdow (approved):
Looks good! Just the one request about updating the test. Thanks!

src/components/Chip.tsx
Levi-Huynh: Ok sg, yes I did grab "info" from the Tag component as a precedent 👍

PR #881, opened 2023-06-02
feat: DnDGrid component

bdow:
The DnDGrid components supports drag and drop functionality within a CSS Grid element. It works for both touch and non-touch screens. It also includes keyboard interactity, though the attributes to make this fully accessible for screen readers is not yet implemented.

Includes:
- DnDGrid.tsx: Builds the grid element and handles all of the events for moving grid items around.
- useDnDGridItem.tsx: A hook for applying properties on draggable elements for the grid.
- DnDGridItemHandle.tsx: A small component that slightly resembles the IconButton. It handles behaviors for the interactive element. This component does not need to be used - you can build your own drag handle using the dragHandleProps from the useDnDGridItem hook.
- DnDGridContext.tsx: Local context only used between DnDGrid and useDnDGridItem. Is not exported in the build.

Adds new 'move' icon

https://www.loom.com/share/2ba004e4192d474a97b3a1a92a034823

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "format": "prettier --loglevel warn --write \"**/*.{ts,tsx,css,md,mdx}\""
   },
   "dependencies": {
-    "@homebound/activesupport": "^1.1.1",
     "@homebound/form-state": "^2.17.0",
     "@internationalized/number": "^3.0.3",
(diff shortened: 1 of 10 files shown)


JonnCharpentier (approved):
Great work! Thanks

PR #882, opened 2023-06-02
docs: Expand comments for unsetLabel and nothingSelectedText.

stephenh:
Also sneaks in a simplification of MultiSelectField.onSelect.


bdow (approved):
Thanks for adding some info!

PR #883, opened 2023-06-02
feat(button-group): allow JSX

TylerR909:
The text on this already gets rendered as a ReactNode `<button>{text}</button>`, but the Typing is only allowing Strings through. So let _me_ use what _Beam_ is already doing. 

Specifically, I need to be able to make `Reject` Red:

image

Previously I needed to color the icons (#874)

image

Both of these would've been solvable if the Type was being exposed in the same way it's being used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@homebound/beam",
-  "version": "2.282.0",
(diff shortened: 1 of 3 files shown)


stephenh (approved):
Cool, makes sense

PR #890, opened 2023-06-19
feat: add the functionality of the initSelected of children rows on init load

ipiedrahb:
I need to that `initSelected` on rows works on children

https://60466f7d43c37c0021788867-xuabsrkbed.chromatic.com/?path=/story/components-table-gridtable--selectable-children-rows

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -1163,19 +1163,17 @@ export function SelectableRows() {
 }
 
 export function SelectableChildrenRows() {
-  type HeaderRow = { kind: "header"; id: string; data: undefined };
-  type ParentRow = { kind: "parent"; id: string; data: { name: string } };
-  type ChildRow = { kind: "child"; id: string; data: { name: string } };
-  type GrandChildRow = { kind: "grandChild"; id: string; data: { name: string } };
-  type Row = HeaderRow | ParentRow | ChildRow | GrandChildRow;
+  type ParentRow = { kind: "parent"; id: string; data: string };
+  type ChildRow = { kind: "child"; id: string; data: string };
+  type GrandChildRow = { kind: "grandChild"; id: string; data: string };
+  type Row = ParentRow | ChildRow | GrandChildRow;
 
   const selectCol = selectColumn<Row>();
 
   const nameCol: GridColumn<Row> = {
-    header: "Name",
-    parent: ({ name }) => name,
-    child: ({ name }) => name,
-    grandChild: ({ name }) => name,
+    parent: (name) => name,
+    child: (name) => name,
+    grandChild: (name) => name,
     mw: "160px",
   };
 
@@ -1183,31 +1181,26 @@ export function SelectableChildrenRows() {
     <>
       <GridTable
         columns={[collapseColumn<Row>(), selectCol, nameCol]}
-        style={{ rowHeight: "fixed" }}
         rows={
           [
             simpleHeader,
             {
               kind: "parent",
               id: "1",
-              data: { name: "Howard Stark" },
+              data: "Howard Stark",
               initSelected: true,
               inferSelectedState: false,
               children: [
                 {
                   kind: "child" as const,
                   id: "2",
-                  data: {
-                    name: "Tony Stark",
-                  },
+                  data: "Tony Stark",
                   initSelected: false,
                   children: [
                     {
                       kind: "grandChild" as const,
                       id: "5",
-                      data: {
-                        name: "Morgan Stark",
-                      },
+                      data: "Morgan Stark",
                       initSelected: true,
                     },
                   ],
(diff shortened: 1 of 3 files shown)


src/components/Table/utils/TableState.ts
```diff
@@ -105,7 +105,8 @@ export class TableState {
   }
 
   loadSelected(rows: GridDataRow<any>[]): void {
-    const selectedRows = rows.filter((row) => row.initSelected);
+    const childRows = rows.flatMap((r) => (r.children ? flattenRows(r.children) : []));
+    const selectedRows = [...rows, ...childRows].filter((row) => row.initSelected);
```
stephenh, line 109: Can you add a unit test for this behavior as well (or instead)? The story is fine, but this sounds like a behavior change more than a visual change, and so a unit test is better for regression coverage.

src/components/Table/utils/TableState.ts
```diff
@@ -105,7 +105,8 @@ export class TableState {
   }
 
   loadSelected(rows: GridDataRow<any>[]): void {
-    const selectedRows = rows.filter((row) => row.initSelected);
+    const childRows = rows.flatMap((r) => (r.children ? flattenRows(r.children) : []));
```
stephenh, line 108: I think this could be just:

```
const allRows = flattenRows(rows);
```

I.e. you can pass `rows` directly to `flattenRows` instead of re-coding some of `flattenRows` here.

src/components/Table/utils/TableState.ts
ipiedrahb: Hi @stephenh 

I included a unit test with this scenario with the commit

ipiedrahb: Hi @stephenh 

Unit test added :), thank you!

ipiedrahb:
Hi @stephenh 

Feedback is done, thank you!

src/components/Table/GridTable.stories.tsx
```diff
@@ -1162,6 +1162,98 @@ export function SelectableRows() {
...
+                  },
+                  initSelected: true,
+                },
+                {
+                  kind: "child" as const,
+                  id: "7",
+                  data: {
+                    name: "Loki",
+                  },
+                  initSelected: true,
```
stephenh, line 1246: Can you remove the Hela and Loki rows? Afaict they are the same as the Thor row and not adding anything new to the test case?

src/components/Table/GridTable.stories.tsx
```diff
@@ -1162,6 +1162,98 @@ export function SelectableRows() {
   );
 }
 
+export function SelectableChildrenRows() {
+  type HeaderRow = { kind: "header"; id: string; data: undefined };
+  type ParentRow = { kind: "parent"; id: string; data: { name: string } };
+  type ChildRow = { kind: "child"; id: string; data: { name: string } };
+  type GrandChildRow = { kind: "grandChild"; id: string; data: { name: string } };
```
stephenh, line 1169: Can you remove the HeaderRow? Both the test and story are super-long for a 2-line code change, so just looking for ways to make this more succinct.

src/components/Table/GridTable.stories.tsx
```diff
@@ -1162,6 +1162,98 @@ export function SelectableRows() {
...
+              data: { name: "Howard Stark" },
+              initSelected: true,
+              inferSelectedState: false,
+              children: [
+                {
+                  kind: "child" as const,
+                  id: "2",
+                  data: {
+                    name: "Tony Stark",
+                  },
```
stephenh, line 1202: Maybe do `data: { name: "Tony Stark" }` just to save some LOC, same for all of these places where `data` could be one line.

src/components/Table/GridTable.test.tsx
```diff
@@ -2421,6 +2421,138 @@ describe("GridTable", () => {
...
+      "
+      | on | Name         |
+      | -- | ------------ |
+      | on | Howard Stark |
+      | on | Tony Stark   |
+      | on | Morgan Stark |
+      | on | Odin         |
+      | on | Thor         |
+      | on | Hela         |
+      | on | Loki         |
```
stephenh, line 2526: Do you really need 7 rows for this test case? If so, can you put an `// And ..what makes this row different...` before each of the `const rows`. As-is there are ~130 LOC that I can't really tell what is 100% necessary and what is extra rows just creating cruft.

src/components/Table/GridTable.stories.tsx
```diff
@@ -1162,6 +1162,98 @@ export function SelectableRows() {
...
+        style={{ rowHeight: "fixed" }}
+        rows={
+          [
+            simpleHeader,
+            {
+              kind: "parent",
+              id: "1",
+              data: { name: "Howard Stark" },
+              initSelected: true,
+              inferSelectedState: false,
```
stephenh, line 1195: Afaict we never set `inferSelectedState` to true? Is it really necessary for the test? Wondering if we could just delete it.

ipiedrahb: Hi @stephenh 

That is the idea, that  `inferSelectedState` don't affect the selection of the child and grandchild if the parent is set or not

ipiedrahb:
Hi @stephenh 

Feedback is ready! thank you!

stephenh (approved):

PR #895, opened 2023-07-06
feat: Always allow type-ahead filtering for SelectFields

ipiedrahb:
Issue
https://www.loom.com/share/f35fc20e35ef4b7ea487be07112414e8

SelectField Feature ( That is the issue :) )
Screen Shot 2023-07-06 at 10 40 14

Proposal
Screen Shot 2023-07-06 at 10 40 51

Demo
https://www.awesomescreenshot.com/video/18921422?key=af47e6ace29ce29be9feb966f114fbda


src/inputs/SelectField.stories.tsx
```diff
@@ -242,6 +242,7 @@ export function PerfTest() {
       }}
       onBlur={action("onBlur")}
       onFocus={action("onFocus")}
+      allowTypeToFilter
```
stephenh, line 245: Suggestion: @ipiedrahb small change, but can we call this `alwaysTypeToFilter`?

cc @bdow , the ask here (which I think generally makes sense) is to disable Beam's opinionated default of "less than 10 fields don't need a type ahead", with the rationale that @ipiedrahb and @chr1sjf0x are working on a page that has ~4-5 select fields in a row, and it can be jarring to be like "oh 1st select field does type-ahead, tab to the next one it does type-ahead, tab to the 3rd one, gah wait I have to use the mouse now".

Or something like that.

Personally I think this makes sense ... frankly it makes me slightly curious if we should remove the "opinionated less than 10" default anyway, just b/c the same "surprise in behavior change" that they're seeing for these ~3-4 selects in a row could potentially be applied in general to the app itself, i.e. "why does my select field on this page do type-ahead, but on this other page it doesn't?"

And maybe the surprise of the user not knowing which behavior any given field will have outweighs the benefit of the less-than-10-doesn't-filter behavior? Wdyt @bdow ?

bdow: I am good with removing the "less-than-10..." rule. That was initially introduced because of a bug in React-Aria that has since been fixed. FWIW, you _don't_ have to use the mouse. A user can still tab to the field and choose an option via keyboard interaction.
One reason why we have kept it was because the virtual keyboard on touch devices is introduced whenever a user opens the menu. It can be a bit jarring, and unnecessary, to have the keyboard open up when you only have 10 or fewer options to select from.
That said, I'm still good with us removing it. We will be creating a new SelectField UX hopefully soon which could change the behaviors again.

I personally don't think we should _always_ have the type to filter. I like having it be smart enough to decide if filtering is necessary based on the number of options, or allowing the SelectField to opt-in to filtering. Though, I'd question the actual benefit of adding the type to filter for 10 or fewer options, specifically when considering virtual keyboards. It can be kind of annoying for the virtual keyboard to open and potentially shift the page around when doing a simple change like updating the status from "In Progress" to "Completed" for example.

ipiedrahb: Hi @bdow @stephenh 

Thank you for your support here, as indicated, the code was updated to remove this feature

chr1sjf0x (approved):
TY @ipiedrahb

bdow (approved):

src/inputs/SelectField.stories.tsx
stephenh: Hm, the mobile use case is interesting... I didn't think of that, and haven't tried it at all on mobile.

@bdow could mobile just not pop-up the virtual keyboard at all, regardless of the # of options, until the user indicated they wanted to type-to-filter?

bdow: There is a new design and UX for the SelectFields where the filter input is inside of the menu. As we build that out, I will work with design to see the best option to move forward with. I'm thinking we will not focus the filter input when the menu opens, requiring the user to explicitly focus the input and bring up the keyboard. Perhaps we can detect whether it is a touch device and only immediately focus the input for non-touch devices.

bdow:
@ipiedrahb - can you remove the `typeToFilter` usage in `src/inputs/TextFieldBase.tsx`?

ipiedrahb:
> 

Hi @bdow 
I saw it but since it's a different component I checked it and it didn't seem convenient to me, can I update it too?

bdow:
> > 
> 
> Hi @bdow I saw it but since it's a different component I checked it and it didn't seem convenient to me, can I update it too?

Yes, please update that as well.

bdow (approved):

PR #900, opened 2023-07-17
feat: Remove vAlign property from TableStyles

bdow:
Previous vertical alignment for all table styles was 'center' with the option to change it explicitly. Only reason we ever changed it was to set it to "top" when using "rowHeight: flexible". After revisiting the logic, it makes more sense to have the vAlign be set based on the `rowHeight` property, and not necessary to allow for the vAlign property. Now 'rowHeight: fixed' defaults to 'center' vertical alignment, and 'rowHeight: flexible' defaults to 'top' vertical alignment.

### DIFF
--- a/src/components/Table/Table.qa.stories.tsx
+++ b/src/components/Table/Table.qa.stories.tsx
@@ -40,7 +40,6 @@ interface TableStoryProps extends GridTableProps<any, any> {
   displayAs?: "default" | "virtual" | "table";
   totals?: boolean;
   expandable?: boolean;
-  vAlign?: "top" | "center" | "bottom";
 }
 
 export default {
@@ -90,7 +89,6 @@ export default {
     grouped: { name: "Group styles", control: { type: "boolean" }, if: { arg: "nestingDepth", neq: 1 } },
     bordered: { name: "Bordered", control: { type: "boolean" } },
     allWhite: { name: "All White", control: { type: "boolean" } },
-    vAlign: { name: "Vertical Alignment", options: ["center", "top", "bottom"], control: { type: "select" } },
     rowHover: { name: "Row Hover styles", control: { type: "boolean" } },
     totals: { name: "Show Totals row", control: { type: "boolean" } },
     expandable: { name: "With Expandable Columns", control: { type: "boolean" } },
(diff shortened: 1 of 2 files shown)


stephenh (approved):
Awesome, thanks!

bdow:
@stephenh - I'm noticing potential issues thanks to Storybook. Specifically when we have a row that contains both plain text and form fields or other elements the increase the height. Then the elements in the row just off quite a bit... see screenshot that I've marked up with a horizontal guide across the row:

Screenshot 2023-07-17 at 12 52 31 PM


You can notice the `#` column looks quite odd when compared to the rest of the row. Even the Icon has a different height from the form field, so those are not vertically aligned perfectly either.

Here's the storybook diff: https://www.chromatic.com/test?appId=60466f7d43c37c0021788867&id=64b55d592cb5a7a37388b99b

stephenh:
Erg, good find ... although I'm wondering if "align middle" is the best fix for those scenarios, b/c let's say the 4th column was even taller than the 2nd, that should push the `0` and `1` down too far, to the middle...

Afaiu we don't really want them "literally in the middle", we just want the content for this row to be pushed down a little bit...

Is that right?

Granted, that's more annoying b/c now each column content has to know "am I an input field that has its own padding, or am I not an input field, and some other column _is_ and input field, so I need to add a little bit of extra top padding"... :thinking:

bdow:
Exactly. It'd be a bit much to start doing like `textCell()`, and `fieldCell`, `iconCell`, etc in order to get the right spacing for type of content within a cell.
I agree the `align: middle` isn't ideal for all situations, but I think it might be the best for most situations.

stephenh:
I kinda wonder if all cells should have a default top padding, and what's special is that the text field knows to use a negative padding/negative margin, b/c it is the "offender" that is making us reduce the padding that otherwise the other cells would be perfectly happy with?

Granted, not sure how we handle this, b/c we'd only only the _first_ input field of a given cell to get the "nudge up" hack.

stephenh:
I guess I'm pushing b/c I think that one page I found in BP, that is using `vAlign=top`, is probably going to have the same mis-alignment issues that you identified (plain text content being pushed up too high), but it's just volunteering to solve it on its own/one-off...

...wonder if we could also work the "inline editing" knob into things, like only use vAlign=middle if row-height=fixed || inline-rendering=true?

bdow:
I see where you're going. It would be nice to solve this, but what was thought to be a fairly simple change has now ballooned a bit. I don't think I have the time at the moment to test out some approaches and evaluate them. 
I think its worth creating a ticket for this in Beam's backlog and looking at it again when we have more time. I'll reference this PR in the ticket.

bdow:
I've created a Beam ticket for this here: https://app.shortcut.com/homebound-team/story/37541/table-component-default-vertical-alignment
I am closing this PR out for now and will revisit later.

PR #906, opened 2023-07-25
fix: SC-37802 Prevent DatePicker triggering virtual keyboard on touch devices

bdow:
Fixes issue where the DatePicker was often rendering underneath the virtual keyboard, preventing users from using the DatePicker. To fix this, DateFieldBase now identifies if it is being used on a touch device. If it is, then it sets the new `TextFieldBase.preventEdit` property to set the input as "readonly". This forces touch screen users to only use the DatePicker for changing dates.

Additional changes include:
 - Move focus to the DatePicker once it is opened and return focus to input when DatePicker is closed.

Updates tests based on new logic for moving focus around

### DIFF
--- a/src/components/internal/DatePicker/DatePickerOverlay.tsx
+++ b/src/components/internal/DatePicker/DatePickerOverlay.tsx
@@ -8,8 +8,6 @@ interface DatePickerOverlayProps {
 // Small wrapper around DatePicker to provide necessary styling and state handling when displayed as an overlay.
 export function DatePickerOverlay({ overlayProps, children }: PropsWithChildren<DatePickerOverlayProps>) {
   return (
-    // Adds `tabIndex` so clicking within the DatePicker will provide a `e.relatedTarget` on blur and focus events.
-    // This allows for components such as the DateField to conditionally trigger their 'onBlur' prop. E.g. If the user leaves the field to interact with the DatePicker, then don't call onBlur
     <div css={Css.br4.bshModal.$} {...overlayProps}>
       {children}
     </div>
--- a/src/components/internal/DatePicker/DatePickerOverlay.tsx
+++ b/src/components/internal/DatePicker/DatePickerOverlay.tsx
@@ -10,7 +10,7 @@ export function DatePickerOverlay({ overlayProps, children }: PropsWithChildren<
   return (
     // Adds `tabIndex` so clicking within the DatePicker will provide a `e.relatedTarget` on blur and focus events.
     // This allows for components such as the DateField to conditionally trigger their 'onBlur' prop. E.g. If the user leaves the field to interact with the DatePicker, then don't call onBlur
-    <div css={Css.br4.bshModal.$} {...overlayProps} tabIndex={0}>
+    <div css={Css.br4.bshModal.$} {...overlayProps}>
       {children}
     </div>
   );
--- a/src/forms/BoundDateField.test.tsx
+++ b/src/forms/BoundDateField.test.tsx
@@ -12,12 +12,15 @@ describe("BoundDateField", () => {
     const author = createObjectState(formConfig, {});
     const r = await render(<BoundDateField field={author.birthday} onBlur={onBlur} onFocus={onFocus} />);
 
-    // When focus is triggered on a checkbox
+    // When focus is triggered
     focus(r.birthday);
     // Then the callback should be triggered
     expect(onFocus).toBeCalledTimes(1);
 
-    // When blur is triggered on a checkbox
+    // When closing the overlay and putting focus back on the input
+    fireEvent.keyDown(r.birthday_datePicker(), { key: "Escape", code: "Escape" });
+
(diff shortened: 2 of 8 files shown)


src/components/internal/DatePicker/DatePickerOverlay.tsx
```diff
@@ -10,7 +10,7 @@ export function DatePickerOverlay({ overlayProps, children }: PropsWithChildren<
   return (
     // Adds `tabIndex` so clicking within the DatePicker will provide a `e.relatedTarget` on blur and focus events.
     // This allows for components such as the DateField to conditionally trigger their 'onBlur' prop. E.g. If the user leaves the field to interact with the DatePicker, then don't call onBlur
-    <div css={Css.br4.bshModal.$} {...overlayProps} tabIndex={0}>
+    <div css={Css.br4.bshModal.$} {...overlayProps}>
```
stephenh, line 13: Nit: I think the comment above this still mentions tabIndex.

src/forms/BoundDateField.test.tsx
```diff
@@ -12,12 +12,15 @@ describe("BoundDateField", () => {
...
 
-    // When focus is triggered on a checkbox
+    // When focus is triggered
     focus(r.birthday);
     // Then the callback should be triggered
     expect(onFocus).toBeCalledTimes(1);
 
-    // When blur is triggered on a checkbox
+    // When closing the overlay and putting focus back on the input
+    fireEvent.keyDown(r.birthday_datePicker(), { key: "Escape", code: "Escape" });
```
stephenh, line 21: Question: Do we want to explicitly check that the callback has _not_ been triggered yet?

src/inputs/TextFieldBase.tsx
```diff
@@ -56,6 +56,9 @@ export interface TextFieldBaseProps<X>
   hideErrorMessage?: boolean;
   // If set, the helper text will always be shown (usually we hide the helper text if read only)
   alwaysShowHelperText?: boolean;
+  // If set, the field will be treated as "read-only" in order to prevent the virtual keyboard on touch devices.
+  // It cannot be edited directly. It can be changed via other methods, though (i.e. DatePicker).
+  preventEdit?: boolean;
```
stephenh, line 61: Question: So is this different then our existing `readOnly` because the field still _looks_ editable?

src/inputs/TextFieldBase.tsx
```diff
@@ -243,8 +247,7 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
                 {...(errorMsg ? { "aria-errormessage": errorMessageId } : {})}
                 ref={fieldRef as any}
                 rows={multiline ? 1 : undefined}
-                // Make the input field readOnly if the field explicitly sets it to `true`
-                readOnly={inputProps.readOnly}
+                readOnly={preventEdit}
```
stephenh, line 250: Question: Hm, we don't need to check `inputProps.readOnly` at all anymore? I would have expected an `||` or something...

stephenh (approved):

src/inputs/TextFieldBase.tsx
bdow: That's correct.

src/inputs/TextFieldBase.tsx
bdow: We use `inputProps.readOnly` to conditionally show our custom readOnly view. It'll never be true in this code block, so is safe to not consider it here.

src/forms/BoundDateField.test.tsx
bdow: We could. Wouldn't hurt and probably would be a better validate it's working as expected. Will update.

bdow:
@stephenh - I decided to take a look through a few different global attributes of HTML and just stumbled upon [inputmode](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode). This can control which virtual keyboard, if any, is displayed for the user. This is something we can do without needing to target wether its a touch screen or not.
I'm going to update to use this attribute and will request a re-review depending on how much changes.

src/inputs/DateFields/DateFieldBase.tsx
```diff
@@ -325,35 +331,37 @@ export function DateFieldBase(props: DateRangeFieldBaseProps | DateSingleFieldBa
...
-                onSelect={(d) => {
-                  setInputValue(formatDate(d, dateFormats.short) ?? "");
-                  onChange(d);
-                  state.close();
-                }}
-                {...tid.datePicker}
-              />
-            )}
-          </DatePickerOverlay>
+          <FocusScope autoFocus restoreFocus>
```
stephenh, line 337: This `FocusScope` component is neat!

stephenh (approved):
Sweet, great find re `inputmode`!

TIL about global attributes :-)

PR #907, opened 2023-07-25
feat: Move AutoSaveStatus into beam.

stephenh:
It originally existed in form-state b/c we used form-state internals to trigger the auto-saving yes/no behavior.

But since then we've moved the "trigger auto-saving yes/no" to the GraphQL mutation instrumentation in apolloHooks, b/c then all pages got the auto-save behavior and not just those using form state.

So, moving it here so we can remove from form-state.

### DIFF
--- a/src/components/AutoSaveStatus/AutoSaveStatusProvider.tsx
+++ b/src/components/AutoSaveStatus/AutoSaveStatusProvider.tsx
@@ -1,0 +1,85 @@
+import React, { PropsWithChildren, useCallback, useEffect, useMemo, useRef, useState } from "react";
+
+export enum AutoSaveStatus {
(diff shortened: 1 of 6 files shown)


src/components/AutoSaveStatus/AutoSaveStatusProvider.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { PropsWithChildren, useCallback, useEffect, useMemo, useRef, useState } from "react";
+
+export enum AutoSaveStatus {
```
stephenh, line 3: All of this code is just copy/pasted as-is from form-state.

bdow (approved):

PR #910, opened 2023-07-27
feat: Add row state

stephenh:
Replaces the imperatively-managed ~4-5 "maps of ids" with RowStates/RowState observables.

In general, the behavior is now nearly all declarative, written like "I'm selected if x/y/z", and then reactively calculated whenever x/y/z happen, instead of imperatively instrumenting the mutations of x/y/z and figuring out what needs to happen, and effectively re-implementing "I'm selected if ..." across multiple entry/mutation points.


src/components/Table/GridTable.tsx
```diff
@@ -777,36 +758,44 @@ export function filterRows<R extends Kinded>(
   columns: GridColumnWithId<R>[],
   rows: GridDataRow<R>[],
   filter: string | undefined,
-): ParentChildrenTuple<R>[] {
+): [string[], ParentChildrenTuple<R>[]] {
```
stephenh, line 761: I changed this method to also calc/return the flat `filteredRowIds` list while we're at it, otherwise we were re-building `filteredRowIds` while doing the `visibleDataRows` calc.

src/components/Table/components/HeaderState.ts
```diff
@@ -0,0 +1,30 @@
+import { makeAutoObservable, observable } from "mobx";
+import { GridDataRow, SelectedState } from "src";
+import { RowState } from "src/components/Table/components/RowState";
+import { RowStates } from "src/components/Table/components/RowStates";
+
+export class HeaderState implements RowState {
```
stephenh, line 6: This is unused, will remove...

src/components/Table/components/Row.tsx
```diff
@@ -40,7 +40,6 @@ interface RowProps<R extends Kinded> {
   row: GridDataRow<R>;
   style: GridStyle;
   rowStyles: RowStyles<R> | undefined;
-  sortOn: SortOn;
```
stephenh, line 43: Removed this prop b/c we can look it up from the TableState; ideally would like to do more of this, to remove the # of props here, but not for this PR.

src/components/Table/GridTable.test.tsx
```diff
@@ -2469,6 +2471,7 @@ describe("GridTable", () => {
             id: "2",
             data: "Tony Stark",
             initSelected: false,
+            inferSelectedState: false,
```
stephenh, line 2474: I had to update this test b/c w/o `inferSelectedState: false`, the `initSelected: false` recommendation was being overridden.

I believe this test update is fine, unless we want to say something like "setting `initSelected` _also_ means we should consider this row to be `inferSelectedState: false`? 

That's a doable change if we want...wdyt?

src/components/Table/GridTable.test.tsx
```diff
@@ -2693,7 +2696,7 @@ describe("GridTable", () => {
       const rows: GridDataRow<NestedRow>[] = [
         simpleHeader,
         {
-          // With one grandparent that sets `inferSelectedState: false`
+          // With one parent that sets `inferSelectedState: false`
```
stephenh, line 2699: Just updating the G/W/Ts to match the `kind`

src/components/Table/GridTable.tsx
```diff
@@ -300,105 +303,77 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
     const keptSelectedRows: RowTuple<R>[] = [];
-    const filteredRowIds: string[] = [];
-
-    const hasExpandableHeader = rows.some((row) => row.id === EXPANDABLE_HEADER);
 
-    function visit([row, children]: ParentChildrenTuple<R>, level: number, visible: boolean): void {
-      visible && visibleDataRows.push([row, makeRowComponent(row, level)]);
-      // This row may be invisible (because it's parent is collapsed), but we still want
-      // to consider it matched if it or it's parent matched a filter.
-      filteredRowIds.push(row.id);
```
stephenh, line 347: This `visibleDataRows` calc was complicated by having to keep adding rows to `filteredRowIds`, even after we're in a collapsed/invisible section of the tree. So I moved the `filteredRowIds` calc elsewhere, and this got simpler.

src/components/Table/utils/RowState.ts
```diff
@@ -0,0 +1,136 @@
...
+  private get visibleChildren(): RowState[] {
+    // The keptGroup should treat all of its children as visible, as this makes select/unselect all work.
+    if (this.row.kind === KEPT_GROUP) return this.children ?? [];
+    return this.children?.filter((c) => c.isMatched === true) ?? [];
+  }
+
+  /** Pretty toString. */
+  [Symbol.for("nodejs.util.inspect.custom")](): string {
+    return `RowState ${this.row.kind}-${this.row.id}`;
+  }
```
bdow, line 135: **question:** What do we use this for?

src/components/Table/utils/RowStates.ts
```diff
@@ -0,0 +1,110 @@
...
+        state.wasRemoved = false;
+        existing.delete(state);
+      }
+      state.children = row.children?.map((child) => addRowAndChildren(state, child));
+      return state;
+    }
+
+    // Probe for the header row, so we can create it as a root RowState, even
+    // though we don't require the user to model their GridDataRows that way.
+    const header = rows.find((r) => r.kind === HEADER);
```
bdow, line 54: **thought:** We shouldn't have this as an issue right now, but I know it's been asked before... what if we have multiple `kind === "header"`? Clearly right now we only support a single `header` kind. And I think we should continue to do that. But thinking that maybe we should warn the implementer somehow if we find multiple `header`s. Or they'll just see that its not working as expected 🤷

src/components/Table/utils/TableState.ts
```diff
@@ -750,16 +532,8 @@ type ColumnSort = {
 
 export type SortState = {
   current?: ColumnSort;
+  /** The persistent sort is always applied first, i.e. for schedules, probably. */
   persistent?: ColumnSort;
 };
 
 export type SortOn = "client" | "server" | undefined;
-
-function keptSelectionsFilter(row: GridDataRow<any>, matchedRows: ObservableSet<string>) {
```
bdow, line 758: Great to see all the red in here!

bdow (approved):
Awesome. This should start getting easier to maintain after the refactor. Thank you!

bdow (changes requested):
I don't think the `deleteRows` is working properly. Maybe we didn't have tests for that?
I see it is holding on to them when playing in the Storybook file: https://60466f7d43c37c0021788867-vsasfhwdup.chromatic.com/?path=/story/components-table--table

src/components/Table/GridTable.tsx
```diff
@@ -300,105 +303,77 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
-            omitRowHover: "rowHover" in maybeStyle && maybeStyle.rowHover === false,
-            sortOn,
-            hasExpandableHeader,
-            isKeptSelectedRow,
-            isLastKeptSelectionRow,
-          }}
-        />
-      );
-    }
+    ) => (
```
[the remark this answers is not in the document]
stephenh: Lol, yep :see_no_evil: ... in this case, I hated the `const` lambda less than the extra indentation or two; ideally over-time would like to trim as many of these props out as possible, at which point :shrug: maybe `makeRowComponent` goes away :crossed_fingers: :-)

src/components/Table/utils/RowState.ts
stephenh: When the `RowState` ends up in `console.log("something ${rs} something")` then Node will call this ... actually, no, I think it's _not_ string interpolation, like `console.log("this", rs)` goes through this symbol... It was useful for some debugging I was doing, but I should clarify in my head + document exactly where it's used.

src/components/Table/utils/RowStates.ts
stephenh: Yeah; I'm tempted to think that the "first header" will always be important/in control of things like select-all/collapse-all; so if they had two headers, I think that would technically be fine, it's just that the 2nd one wouldn't be able to drive any special select-all behavior.

If they wanted that to happen, it seems like the "2nd header" would be "the real header" and we'd use some other kind, like totals, to be a psuedo-header that comes first...

src/components/Table/Table.qa.stories.tsx
```diff
@@ -173,7 +173,7 @@ export function Table(props: TableStoryProps) {
               label="Delete Rows"
               onClick={() => {
                 setRows((rows) => rows.filter((r) => !api.getSelectedRowIds().includes(r.id)));
-                // api.deleteRows(api.getSelectedRowIds("child"));
+                api.deleteRows(api.getSelectedRowIds("child"));
```
bdow, line 176: Well, how about that. haha! Thats probably my fault from before.

bdow (approved):

PR #911, opened 2023-07-28
feat: Refactor collapse logic into RowState

stephenh:
Also created a `RowStorage` abstraction for reading/writing the collapsed ids to sessionStorage.

### DIFF
--- a/src/components/Table/components/CollapseToggle.test.tsx
+++ b/src/components/Table/components/CollapseToggle.test.tsx
@@ -5,6 +5,7 @@ import { render } from "src/utils/rtl";
 describe("CollapseToggle", () => {
   it("defaults to uncollapsed", async () => {
     const rowState = new TableState();
+    rowState.setRows([{ id: "r:1", kind: "header", data: {} }]);
     const r = await render(
       <TableStateContext.Provider value={{ tableState: rowState }}>
         <CollapseToggle row={{ id: "r:1", kind: "header", data: {} }} />
@@ -20,6 +21,7 @@ describe("CollapseToggle", () => {
     [false, "otherKind", "chevronDown"],
   ])("displays the correct chevron based on context and kind", async (isCollapsed, kind, expectedIcon) => {
     const rowState = new TableState();
+    rowState.setRows([{ id: "r:1", kind: "header", data: {} }]);
     if (isCollapsed) {
       rowState.toggleCollapsed("r:1");
     }
@@ -33,6 +35,7 @@ describe("CollapseToggle", () => {
 
   it("only renders for non-header rows when there are children", async () => {
     const rowState = new TableState();
+    rowState.setRows([{ id: "r:1", kind: "otherKind", data: {} }]);
     const r = await render(
       <TableStateContext.Provider value={{ tableState: rowState }}>
         <CollapseToggle row={{ id: "r:1", kind: "otherKind", data: {}, children: [] }} />
--- a/src/components/Table/utils/RowState.ts
+++ b/src/components/Table/utils/RowState.ts
@@ -122,6 +122,13 @@ export class RowState {
     }
   }
 
+  /** Marks the row as removed from `props.rows`, to potentially become kept. */
+  markRemoved(): void {
+    // The kept group is never in `props.rows`, so ignore asks to delete it
+    if (this.row.kind === KEPT_GROUP) return;
+    this.removed = this.selected && this.removed !== "hard" ? "soft" : "hard";
+  }
+
   toggleCollapsed(): void {
     this.collapsed = !this.collapsed;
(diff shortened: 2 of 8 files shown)


src/components/Table/components/CollapseToggle.test.tsx
```diff
@@ -5,6 +5,7 @@ import { render } from "src/utils/rtl";
 describe("CollapseToggle", () => {
   it("defaults to uncollapsed", async () => {
     const rowState = new TableState();
+    rowState.setRows([{ id: "r:1", kind: "header", data: {} }]);
```
stephenh, line 8: These tests need a row inside the `rowState` now to avoid blowing up

src/components/Table/utils/RowState.ts
```diff
@@ -10,30 +11,30 @@ import { GridDataRow, KEPT_GROUP, reservedRowKinds, SelectedState } from "src";
 export class RowState {
   /** Our row, only ref observed, so we don't crawl into GraphQL fragments. */
   row: GridDataRow<any>;
-  /** Our parent RowState, or the `header` RowState if we're a top-level row. */
-  parent: RowState | undefined;
```
stephenh, line 14: I was able to remove `parent` as unnecessary anymore; just simplifies things to have parents only point to their children, and not try to maintain the two-way connection.

src/components/Table/utils/RowState.ts
```diff
@@ -126,7 +148,11 @@ export class RowState {
   private get visibleChildren(): RowState[] {
     // The keptGroup should treat all of its children as visible, as this makes select/unselect all work.
     if (this.row.kind === KEPT_GROUP) return this.children ?? [];
-    return this.children?.filter((c) => c.isMatched === true) ?? [];
+    // Ignore hard-deleted rows, i.e. from `api.deleteRows`; in theory any hard-deleted
+    // rows should be removed from `this.children` anyway, by a change to `props.rows`,
+    // but just in case the user calls _only_ `api.deleteRows`, and expects the row to
+    // go away, go ahead and filter them out here.
+    return this.children?.filter((c) => c.isMatched === true && c.removed !== "hard") ?? [];
```
stephenh, line 155: This change was necessary since `api.deleteRows` no longer actively remove rows from `parent.children`, b/c that was the one spot that needed `RowState.parent` to go up from the RowState to its parent.

Granted, introducing the notion of soft/hard deleted seems a little complicated, but I think it's worth being able to remove the `RowState.parent` pointer and stop trying to rewrite `children` collections.

src/components/Table/utils/RowStates.ts
```diff
@@ -30,81 +40,124 @@ export class RowStates {
    */
   setRows(rows: GridDataRow<any>[]): void {
     const existing = new Set(this.map.values());
+    const states = this;
     const map = this.map;
 
-    function addRowAndChildren(parent: RowState | undefined, row: GridDataRow<any>): RowState {
```
stephenh, line 35: This `parent` is unnecessary

src/components/Table/utils/RowStates.ts
```diff
@@ -30,81 +40,117 @@ export class RowStates {
...
+      state.children = row.children?.map((child) => addRowAndChildren(child));
       return state;
     }
 
     // Probe for the header row, so we can create it as a root RowState, even
     // though we don't require the user to model their GridDataRows that way.
-    const header = rows.find((r) => r.kind === HEADER);
-    const headerState = header ? addRowAndChildren(undefined, header) : undefined;
+    const headerRow = rows.find((r) => r.kind === HEADER);
+    this.header = headerRow ? addRowAndChildren(headerRow) : undefined;
```
stephenh, line 65: All of this code is basically the same, just it's now assigning the `this.header` and `this.topRows` fields.

src/components/Table/utils/RowState.ts
```diff
@@ -10,30 +11,47 @@ import { GridDataRow, KEPT_GROUP, reservedRowKinds, SelectedState } from "src";
...
+   *
+   * To avoid this, we by default keep selected rows, as "kept rows", to make
+   * extra sure the user wants them to go away.
+   *
+   * Soft-deleted rows are rows that were removed from `props.rows` (i.e. we
+   * suspect are just hidden by a changed server-side-filter), and hard-deleted
+   * rows are rows the page called `api.deleteRow` and confirmed it should be
+   * actively removed.
+   */
+  removed: false | "soft" | "hard" = false;
```
stephenh, line 40: I should have kept this as a separate PR, but I changed `wasRemoved` to a tri-state, so that `api.deleteRows` can mark the hard-deletion + allowed removing the `RowState.parent` field.

PR #913, opened 2023-07-29
fix: Safer selectRow

stephenh:
I think a number of internal-frontend tests had just ~slightly invalid test cases, and were causing `selectRow`s to happen on rows that don't exist.

Going to try this and see how the test count changes.


stephenh:
Nevermind, got things working without this.

PR #914, opened 2023-07-29
fix: Allow selecting parents with no children.

stephenh:
Fixes an assumption some Invoicing v2/expense tables were making, of deriving children like `row.children = someOtherList.map(...)`, and if `someOtherList` was empty, they want the row to be selectable / i.e. not really a parent.


stephenh:
Merging to unblock the beam-bump in internal-frontend.

PR #915, opened 2023-07-29
fix: Call setMatchedRows from useEffect

stephenh:
This doesn't really change any behavior/fix any bugs, it is just more kosher/avoids the React warning.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -401,7 +401,9 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
     visibleDataRows = visibleDataRows.slice(0, filterMaxRows + keptSelectedRows.length);
   }
 
-  tableState.setMatchedRows(filteredRowIds);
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -401,7 +401,9 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
     visibleDataRows = visibleDataRows.slice(0, filterMaxRows + keptSelectedRows.length);
   }
 
-  tableState.setMatchedRows(filteredRowIds);
+  useEffect(() => {
+    tableState.setMatchedRows(filteredRowIds);
+  }, [tableState, filteredRowIds]);
```
stephenh, line 406: Mutating observables directly from render causes those "Cannot update" warnings, so if we do it from a `useEffect`, it'll be fine/albeit might delay an update:

https://github.com/facebook/react/issues/26962#issuecomment-1656742586

bdow (approved):

PR #916, opened 2023-07-30
feat: Refactor column handling into ColumnState

stephenh:
Similar to the same refactoring for RowState/RowStates, does the same for column state.

### DIFF
--- a/src/components/Table/GridTableApi.ts
+++ b/src/components/Table/GridTableApi.ts
@@ -26,36 +26,35 @@ export function useGridTableApi<R extends Kinded>(): GridTableApi<R> {
 /** Provides an imperative API for an application page to interact with the table. */
 export type GridTableApi<R extends Kinded> = {
   /** Scrolls row `index` into view; only supported with `as=virtual` and after a `useEffect`. */
-  scrollToIndex: (index: number) => void;
+  scrollToIndex(index: number): void;
 
   /** Returns the ids of currently-selected rows. */
   getSelectedRowIds(): string[];
   getSelectedRowIds<K extends R["kind"]>(kind: K): string[];
-
   /** Returns the currently-selected rows. */
   getSelectedRows(): GridDataRow<R>[];
+  /** Returns the currently-selected rows of the given `kind`. */
   getSelectedRows<K extends R["kind"]>(kind: K): GridDataRow<DiscriminateUnion<R, "kind", K>>[];
-
-  /** Deselects all rows */
+  /** Set selected state of a row by id. */
+  selectRow(id: string, selected?: boolean): void;
+  /** De-selects all selected rows. */
   clearSelections(): void;
 
+  /** Whether a row is currently collapsed. */
+  isCollapsedRow(id: string): boolean;
+  /** Toggle collapse state of a row by id. */
+  toggleCollapsedRow(id: string): void;
+
   /** Sets the internal state of 'activeRowId' */
-  setActiveRowId: (id: string | undefined) => void;
-
+  setActiveRowId(id: string | undefined): void;
   /** Sets the internal state of 'activeCellId' */
-  setActiveCellId: (id: string | undefined) => void;
+  setActiveCellId(id: string | undefined): void;
 
-  /** Set selected state of a row by id */
-  selectRow: (id: string, selected?: boolean) => void;
+  /** Deletes a row from the table, i.e. so it's not detected as kept. */
+  deleteRows(ids: string[]): void;
 
-  /** Deletes a row from the table */
-  deleteRows: (ids: string[]) => void;
-
-  /** Toggle collapse state of a row by id */
-  toggleCollapsedRow: (id: string) => void;
-  isCollapsedRow: (id: string) => boolean;
-  setVisibleColumns: (ids: string[]) => void;
-  getVisibleColumnIds: () => string[];
+  getVisibleColumnIds(): string[];
+  setVisibleColumns(ids: string[]): void;
 };
 
 // Using `FooImpl`to keep the public GridTableApi definition separate.
@@ -70,6 +69,7 @@ export class GridTableApiImpl<R extends Kinded> implements GridTableApi<R> {
     virtuosoRef: MutableRefObject<VirtuosoHandle | null>,
     rows: GridDataRow<R>[],
   ) {
+    // Technically this drives both row-collapse and column-expanded
     if (persistCollapse) this.tableState.loadCollapse(persistCollapse);
     this.virtuosoRef = virtuosoRef;
   }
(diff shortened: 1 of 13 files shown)


src/components/Table/GridTableApi.ts
```diff
@@ -26,36 +26,35 @@ export function useGridTableApi<R extends Kinded>(): GridTableApi<R> {
 /** Provides an imperative API for an application page to interact with the table. */
 export type GridTableApi<R extends Kinded> = {
   /** Scrolls row `index` into view; only supported with `as=virtual` and after a `useEffect`. */
-  scrollToIndex: (index: number) => void;
```
stephenh, line 29: Made all of these the same syntax, plus a little re-ordering, otherwise no change.

src/components/Table/components/EditColumnsButton.tsx
```diff
@@ -43,28 +43,30 @@ export function EditColumnsButton<R extends Kinded>(props: EditColumnsButtonProp
       : trigger.name,
   );
 
-  const { options } = useMemo(() => {
-    return columns.reduce(
```
stephenh, line 47: This reduce was necessary when it was written, b/c two collections were being created at the same time, but the 2nd collection was removed in an earlier PR, so now it can be just a filter.

src/components/Table/components/EditColumnsButton.tsx
```diff
@@ -43,28 +43,30 @@ export function EditColumnsButton<R extends Kinded>(props: EditColumnsButtonProp
...
+        .map((column) => ({ label: column.name!, value: column.id! })),
+    [columns],
+  );
 
   const selectedValues = useComputed(() => api.getVisibleColumnIds(), [api]);
   const setSelectedValues = useCallback(
-    (values: string[]) => {
+    (ids: string[]) => {
+      // Doesn't `options` already filter us to non-hidden/valid-id columns? I.e. could we just do:
+      // api.setVisibleColumns(ids);
```
bdow, line 67: This sounds right to me. `setSelectedValues` cannot be called with values that do not pass the filter below, due to them having to have passed a similar, if more strict filter above. 
Did you happen to try this and see if the tests still all pass?

src/components/Table/components/Row.tsx
```diff
@@ -133,9 +133,9 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
 
           // Need to keep track of the expanded columns so we can add borders as expected for the header rows
           const numExpandedColumns = isExpanded
-            ? tableState.getExpandedColumns(maybeExpandedColumn)?.length
+            ? tableState.numberOfExpandedChildren(maybeExpandedColumn.id)
               ? // Subtract 1 if the column is hidden on expand, since we're not rendering it.
-                tableState.getExpandedColumns(maybeExpandedColumn).length - (maybeExpandedColumn.hideOnExpand ? 1 : 0)
+                tableState.numberOfExpandedChildren(maybeExpandedColumn.id) - (maybeExpandedColumn.hideOnExpand ? 1 : 0)
```
bdow, line 138: Per the `hideOnExpand` - The amount of customization available in this table is a bit wild.

src/components/Table/utils/ColumnState.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  constructor(private states: ColumnStates, storage: ColumnStorage, column: GridColumnWithId<any>) {
+    this.column = column;
+    // If the user sets `canHide: true`, we default to hidden unless they set `initVisible: true`
+    this.visible = storage.wasVisible(column.id) ?? (column.canHide ? column.initVisible ?? false : true);
+    if (this.visible && (storage.wasExpanded(column.id) ?? column.initExpanded)) {
+      this.expanded = true;
+      this.doExpand();
+    }
+    makeAutoObservable(this, { column: observable.ref });
```
bdow, line 27: Maybe worth an inline comment about what `observable.ref` achieves for us here.

src/components/Table/utils/ColumnState.ts
```diff
@@ -0,0 +1,81 @@
...
+  /** Returns this column, if visible, and its children, if expanded. */
+  get maybeSelfAndChildren(): ColumnState[] {
+    if (!this.visible) {
+      return [];
+    } else if (this.expanded && this.children) {
+      // Maybe do the `hideOnExpand` thing here? Seems cute, but the Row rendering still
+      // needs to do the "look back to the prior column for the expandableHeader cell" logic.
+      // if (this.column.hideOnExpand) {
+      //   return this.children.flatMap((c) => c.selfAndMaybeChildren);
+      // }
```
bdow, line 75: Thanks for adding the comment and contemplating if we could do the `hideOnExpand`. Do you think its worth keeping this comment?

src/components/Table/utils/ColumnStorage.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  loadVisible(storageKey: string): void {
+    this.visibleIds = loadArrayOrUndefined(storageKey);
+    // Unlike the others, where we only store the value on change, we immediately
+    // store this value (but I'm not sure why...), hence using `autorun`.
+    autorun(() => {
+      const columnIds = this.states.allVisibleColumns.map((cs) => cs.column.id);
+      sessionStorage.setItem(storageKey, JSON.stringify(columnIds));
+    });
```
bdow, line 28: **question:** Are you saying we don't update the storage value whenever the user changes its visible columns?

src/components/Table/utils/RowStorage.ts
```diff
@@ -1,4 +1,5 @@
 import { reaction } from "mobx";
+import { loadArrayOrUndefined } from "src";
```
bdow, line 2: **nit:** I have see issues sneak up with importing directly from `src` once we build the package. Seems to cause some circular import issues. Have you tried building and syncing this locally in BP? Maybe worth a try if not.

src/components/Table/utils/TableState.ts
```diff
@@ -147,141 +138,56 @@ export class TableState {
 
   // Updates the list of rows and regenerates the collapsedRows property if needed.
   setRows(rows: GridDataRow<any>[]): void {
+    // Note that because of using `rows: observable.shallow` above, this is always
+    // false, and this logic runs on every render. We can eventually fix this, but it
+    // is convenient b/c it puts no-longer-kept rows back into the right spot in their
+    // parents.
     if (rows !== this.rows) {
```
bdow, line 145: **question:** Do you mean to say "this is always true" (`rows !== this.rows`)? That they'll never be equal due to the `observable.shallow`.

bdow (approved):

bdow:
Looks much cleaner! Thanks for taking the time to comb through all of this

src/components/Table/components/Row.tsx
stephenh: Yeah :-) the "oh wait, the Nth column uses the N-1th column's rendering logic?..." really threw me for a loop. I didn't know about that functionality, but "get it"/makes sense...

src/components/Table/utils/ColumnStorage.ts
stephenh: Ah no, we do update it after every change, it's just that we also update it _immediately_, with whatever the default value is. Our other "persist" stuff didn't bother storing the default state, until the user changed something. Not sure that it matters, it's just different between the two, hence making the in-source comment.

PR #917, opened 2023-08-01
fix: DatePicker to open via click instead of focus

bdow:
The DatePicker now opens when the input is clicked instead of when it is focused. This is to prevent the DatePicker from re-opening when focus is returned to the field after selecting a date.

We were managing a variable to know whether or not the focus was returned to the input due to the DatePicker closing. However, there were some race conditions introduced when using multiple FocusScope components. Specifically, when the DateField’s `onChange` method triggers a modal, then we were attempting to return focus to the input field, and at the same time move focus to modal. React-Aria uses a requestAnimationFrame to restoreFocus, which was causing the race condition between when the focus moves and when it gets restored. By removing the behavior of focus triggering the DatePicker, then we completely avoid the race condition, and we can get rid of our ‘closingDatePicker’ variable for determining when focusing the element should trigger the DatePicker.

Updates many tests due to the new behavior

### DIFF
--- a/src/components/Filters/DateRangeFilter.test.tsx
+++ b/src/components/Filters/DateRangeFilter.test.tsx
@@ -1,7 +1,7 @@
 import { useState } from "react";
 import { FilterDefs, Filters } from "src/components/Filters";
 import { ProjectFilter, taskCompleteFilter } from "src/components/Filters/testDomain";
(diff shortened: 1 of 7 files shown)


stephenh (approved):
Nice! Looks much simpler!

PR #918, opened 2023-08-01
feat: Use observerable.ref for rows as well.

stephenh:
This means the `setRows` really only runs when rows actually change, hence MobX no longer upgrades the rows array to an ObservableArray behind the scenes.

### DIFF
--- a/src/components/Table/utils/RowState.ts
+++ b/src/components/Table/utils/RowState.ts
@@ -153,7 +153,7 @@ export class RowState {
   }
 
   private get visibleChildren(): RowState[] {
-    // The keptGroup is special and it's children are the dynamically kept rows
+    // The keptGroup is special and its children are the dynamically kept rows
     if (this.row.kind === KEPT_GROUP) return this.states.keptRows;
     // Ignore hard-deleted rows, i.e. from `api.deleteRows`; in theory any hard-deleted
     // rows should be removed from `this.children` anyway, by a change to `props.rows`,
--- a/src/components/Table/utils/RowState.ts
+++ b/src/components/Table/utils/RowState.ts
@@ -43,7 +43,7 @@ export class RowState {
   // isDirectlyMatched = accept filters in the constructor and do match here
   // isEffectiveMatched = isDirectlyMatched || hasMatchedChildren
 
-  constructor(states: RowStates, row: GridDataRow<any>) {
+  constructor(private states: RowStates, row: GridDataRow<any>) {
     this.row = row;
     this.selected = !!row.initSelected;
     this.collapsed = states.storage.wasCollapsed(row.id) ?? !!row.initCollapsed;
@@ -115,10 +115,8 @@ export class RowState {
     this.selected = selected;
     // We don't check inferSelectedState here, b/c even if the parent is considered selectable
     // on its own, we still push down selected-ness to our visible children.
-    if (this.children) {
-      for (const child of this.visibleChildren) {
-        child.select(selected);
-      }
+    for (const child of this.visibleChildren) {
+      child.select(selected);
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/components/Table/utils/RowState.ts
```diff
@@ -152,8 +150,8 @@ export class RowState {
   }
 
   private get visibleChildren(): RowState[] {
-    // The keptGroup should treat all of its children as visible, as this makes select/unselect all work.
-    if (this.row.kind === KEPT_GROUP) return this.children ?? [];
+    // The keptGroup is special and it's children are the dynamically kept rows
+    if (this.row.kind === KEPT_GROUP) return this.states.keptRows;
```
stephenh, line 154: Previously I'd been patching up `keptGroupRow.children` inside of every `setRows` call, which is why it was lucky/necessary for `setRows` to keep running all the time.

This makes the keptGroupRow.children reactive/calculated, instead of having to be imperatively updated, which is pretty cute imo. :-)

src/components/Table/utils/RowStates.ts
```diff
@@ -67,24 +67,17 @@ export class RowStates {
...
     }
 
     // Then mark any remaining as removed
     for (const state of existing) state.markRemoved();
 
-    const keptRows = this.keptRows;
-    if (keptRows.length > 0) {
-      // Stitch the current keptRows into the placeholder keptGroupRow
-      this.keptGroupRow.children = keptRows;
-      this.keptGroupRow.row.children = keptRows.map((rs) => rs.row);
```
stephenh, line 80: This is the imperative "keep the kept rows up to date" that was, by happenstance, running all the time, so when I fixed "don't run all the time", we had to get this updated in a different way.

src/components/Table/utils/RowStates.ts
```diff
@@ -148,9 +146,19 @@ export class RowStates {
...
-  private creatKeptGroupRow(): RowState {
+  private createKeptGroupRow(): RowState {
     // The "group row" for selected rows that are hidden by filters and add the children
-    const keptGroupRow: GridDataRow<any> = { id: KEPT_GROUP, kind: KEPT_GROUP, initCollapsed: true, data: undefined };
-    return new RowState(this, keptGroupRow);
+    const keptGroupRow: GridDataRow<any> = {
+      id: KEPT_GROUP,
+      kind: KEPT_GROUP,
+      initCollapsed: true,
+      selectable: false,
```
stephenh, line 155: Added `selectable` false denote the user can't check/uncheck the row.

src/components/Table/utils/RowState.ts
```diff
@@ -111,14 +113,13 @@ export class RowState {
    * child of a selected parent row.
    */
   select(selected: boolean): void {
-    if (this.row.selectable === false) return;
```
stephenh, line 114: Removed this early return so that unselectable rows will still ignore the "set my selected yes/no", but now will also push the new selected state down to their children.

I.e. the kept row is not selectable, but we want the header being selected/unselected to "push through it" to the children kept rows.

src/components/Table/utils/RowState.ts
```diff
@@ -111,14 +113,13 @@ export class RowState {
...
-    this.selected = selected;
+    if (this.row.selectable !== false) {
+      this.selected = selected;
+    }
     // We don't check inferSelectedState here, b/c even if the parent is considered selectable
     // on its own, we still push down selected-ness to our visible children.
-    if (this.children) {
```
stephenh, line 118: Just removed as unnecessary

src/components/Table/utils/RowStates.ts
```diff
@@ -12,7 +12,7 @@ export class RowStates {
   private map = new ObservableMap<string, RowState>();
   storage = new RowStorage(this);
   // Pre-create our keptGroupRow for if/when we need it.
-  private keptGroupRow: RowState = this.creatKeptGroupRow();
+  private keptGroupRow: RowState = this.createKeptGroupRow();
```
stephenh, line 15: Typo fix in the method name

bdow (approved):
awesome! This is looking much cleaner

PR #919, opened 2023-08-03
feat: New color palette

bdow:
Renames "LightBlue" -> "Blue" and "Violet" -> "Purple" 
Updates various component variation colors based on new palette


stephenh (approved):
I don't think I realized how "not blue" our current blue was until seeing it next to the "really blue" new blue. :-) Looks great!

PR #923, opened 2023-08-09
feat: Move sorting into RowState.

stephenh:


src/components/Table/GridTable.tsx
```diff
@@ -270,141 +259,73 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
+        <Row
+          key={`${row.kind}-${row.id}`}
+          {...{
+            as,
+            columns,
+            row,
+            style,
+            rowStyles,
+            columnSizes,
+            level: 1,
```
stephenh, line 295: Er, forgot to fix this temporary hard-coding...

src/components/Table/GridTable.tsx
```diff
@@ -260,6 +245,10 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
     tableState.activeCellId = activeCellId;
   }, [tableState, activeCellId]);
 
+  useEffect(() => {
+    tableState.setSearch(filter);
+  }, [tableState, filter]);
```
stephenh, line 250: Pushing the filter into `TableState` so that `RowState` can reactively read it, i.e. parent rows will re-sort their children each time the search changes.

src/components/Table/GridTable.tsx
```diff
@@ -270,141 +259,73 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
   const expandedColumnIds: string[] = useComputed(() => tableState.expandedColumnIds, [tableState]);
   const columnSizes = useSetupColumnSizes(style, columns, resizeTarget ?? resizeRef, expandedColumnIds);
 
-  // Make a single copy of our current collapsed state, so we'll have a single observer.
-  const collapsedIds = useComputed(() => tableState.collapsedIds, [tableState]);
```
stephenh, line 274: Collapsed-ness is handling within the `RowState.selfAndMaybeChildren` that flattens a row + its children (unless collapased).

The `selfAndMaybeChildren` also does the flattening that we used to do via visits/recursion.

src/components/Table/GridTable.tsx
```diff
@@ -270,141 +259,73 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
-          cellHighlight: "cellHighlight" in maybeStyle && maybeStyle.cellHighlight === true,
-          omitRowHover: "rowHover" in maybeStyle && maybeStyle.rowHover === false,
-          hasExpandableHeader,
-          isKeptSelectedRow,
-          isLastKeptSelectionRow,
-        }}
-      />
-    );
+    boolean,
+  ] = useComputed(() => {
```
stephenh, line 268: Changed this to a `uesComputed` that essentially reacts to `header` `RowState.selfAndMaybeChildren` changing (which itself changes if any top-level rows have their `selfAndMaybeChildren` change, etc)

src/components/Table/GridTable.tsx
```diff
@@ -270,141 +259,73 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
-  // Once our header rows are created we can organize them in expected order.
-  const tableHeadRows = expandableHeaderRows.concat(headerRows).concat(totalsRows);
-
-  const tooManyClientSideRows = filterMaxRows && visibleDataRows.length > filterMaxRows;
-  if (tooManyClientSideRows) {
-    visibleDataRows = visibleDataRows.slice(0, filterMaxRows + keptSelectedRows.length);
-  }
-
-  useEffect(() => {
-    tableState.setMatchedRows(filteredRowIds);
```
stephenh, line 406: We don't push matched-ness into the table state anymore, each `RowState.isMatched` does its own reactive calc, instead of just being a boolean that is set by some outside code.

src/components/Table/GridTable.tsx
```diff
@@ -743,57 +664,3 @@ const VirtualRoot = memoizeOne<(gs: GridStyle, columns: GridColumn<any>[], id: s
...
-    filteredRowIds.push(row.id);
-    return acc.concat([[row, row.children?.reduce(acceptAll, []) ?? []]]);
-  }
-
-  function filterFn(acc: ParentChildrenTuple<R>[], row: GridDataRow<R>): ParentChildrenTuple<R>[] {
-    const matches =
-      reservedRowKinds.includes(row.kind) ||
-      filters.length === 0 ||
-      filters.every((f) =>
-        columns.map((c) => applyRowFn(c, row, api, 0, false)).some((maybeContent) => matchesFilter(maybeContent, f)),
```
stephenh, line 776: This snippet moved into `RowState.isDirectlyMatched`

src/components/Table/GridTable.tsx
```diff
@@ -743,57 +664,3 @@ const VirtualRoot = memoizeOne<(gs: GridStyle, columns: GridColumn<any>[], id: s
...
-      );
-    if (matches) {
-      filteredRowIds.push(row.id);
-      // A matched parent means show all it's children
```
stephenh, line 780: These descriptive comments show up in `RowState.isMatched`

src/components/Table/utils/RowState.ts
```diff
@@ -15,8 +15,6 @@ export class RowState {
   row: GridDataRow<any>;
   /** Our children row states, as of the latest `props.rows`, without any filtering applied. */
   children: RowState[] | undefined = undefined;
-  /** Whether we match a client-side filter; true if no filter is in place. */
-  isMatched = true;
```
stephenh, line 19: This is now a computed getter down below

src/components/Table/GridTable.tsx
```diff
@@ -270,141 +259,73 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
-  const tooManyClientSideRows = filterMaxRows && visibleDataRows.length > filterMaxRows;
-  if (tooManyClientSideRows) {
-    visibleDataRows = visibleDataRows.slice(0, filterMaxRows + keptSelectedRows.length);
-  }
-
-  useEffect(() => {
-    tableState.setMatchedRows(filteredRowIds);
-  }, [tableState, filteredRowIds]);
+    return [tableHeadRows, visibleDataRows, keptSelectedRows, tooManyClientSideRows];
+  }, [as, api, style, rowStyles, maybeStyle, columnSizes, getCount, filterMaxRows]);
```
bdow, line 328: **thought:** I wonder if we can / should do anything more to ensure some of these are stable properties, like `style`, `rowStyles` and `maybeStyle`. I think `style` gets memoized, but `maybeStyle` and `rowStyles` are up to the implementer to do so. 
I think for `maybeStyle` we just need to determine a couple boolean attributes from it. So maybe we should determine those outside of this useComputed and pass those in as primitives.
I think then we only need to worry about `rowStyles`, which I believe I've seen users already are memoizing before passing to GridTable.

src/components/Table/utils/ColumnState.ts
```diff
@@ -1,5 +1,6 @@
 import { makeAutoObservable, observable } from "mobx";
-import { assignDefaultColumnIds, GridColumnWithId } from "src";
+import { GridColumnWithId } from "src/components/Table/types";
+import { assignDefaultColumnIds } from "src/components/Table/utils/columns";
```
bdow, line 3: 👍

bdow (approved):
Amazing! Not looking like a big ol' mess of states any more! Thanks for taking the time to do this

src/components/Table/utils/RowState.ts
```diff
@@ -142,24 +152,53 @@ export class RowState {
...
-      !reservedRowKinds.includes(this.row.kind) &&
+      !this.isReservedKind &&
       !this.isParent &&
       (!this.isMatched || this.removed === "soft")
     );
   }
 
+  get level(): number {
+    // Make the header level -1, so the top-level rows are level 0
+    return !this.parent ? -1 : this.parent.level + 1;
```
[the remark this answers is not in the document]
bdow: Huh, I guess I forgot we're nesting all rows under `header`. I don't think we used to do that, but maybe that's new with this reorganization? I remember we used to have special "if users selects/deselects header -> then just mash all"

src/components/Table/utils/RowStates.ts
```diff
@@ -43,37 +49,32 @@ export class RowStates {
...
       if (!state) {
-        state = new RowState(states, row);
+        state = new RowState(states, parent, row);
         map.set(key, state);
       } else {
+        state.parent = parent;
         state.row = row;
         state.removed = false;
         existing.delete(state);
       }
```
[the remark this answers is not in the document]
stephenh: Good questions! For 1, mobx will basically do that for us, if we do `foo.state = "bar"` and the value was already "bar" (or whatever instance of `parent`), it will not register as a change, and so will not trigger downstream reactions.

For 2, we remove it from existing to basically mark it has claimed, so at the end, anything left in existing, i.e. unclaimed, qualifies as a removed row, i.e. will turn into a kept row.

Maybe the "claimed" terminology would be better?

src/components/Table/utils/RowState.ts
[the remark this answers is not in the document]
stephenh: Ah yeah, we still don't in the applications, i.e. `createRows` all still pass in `props.rows=[header, row1, row2]`, but then when we convert `props.rows` into `RowState`s, we pluck out the `header` and make it the parent of all of the other non-header top-level `RowState`s.

This ends up making a lot of the header's "select/collapse all" logic _mostly_ just work for-free, treating the top-level rows as its children.

But we do have a few special cases where the top-most/header-level behavior (particularly for collapse iirc) works differently than "regular parent/child" behavior, and so we have a few "if header ..." special cases.

But overall it worked out pretty nicely/cutely, and yeah is 100% transparent to the callers.

src/components/Table/GridTable.tsx
stephenh: Agreed! I've been meaning to play with a `Stable<...>` branded type for forever...will nudge it up on the todo list, and see if anything interesting comes of it.

PR #930, opened 2023-08-18
fix: Make the useComputed return value more stable across renders.

stephenh:


src/hooks/useComputed.ts
```diff
@@ -37,21 +37,21 @@ export function useComputed<T>(fn: (prev: T | undefined) => T, deps: readonly an
...
+      const { value: oldValue, hasRan } = current;
       // Always eval fn() (even on 1st render) to register our observable.
       const newValue = fn(oldValue);
+      // If we've already run and the value hasn't changed, don't trigger a re-render
+      //
+      // Also, we avoid a deep equality, b/c if a `useComputed` is returning something complicated/cyclic,
+      // like ReactElement, deep equality will crawl into the guts of React/ReactFiber and cycle/infinite loop.
+      if (hasRan && shallowEqual(newValue, oldValue)) return;
+      // Only change the identity of `current.value` after we've checked that it's not shallow equal
       current.value = newValue;
```
stephenh, line 49: The bug here was that we always assigned `current.value`, before checking if newValue/oldValue were the same effective value (shallow equal).

I.e. if the `fn` created a new array, we'd put `current.value = ...new array...` every time one of our observed values changed, even if later we would not actually call tick.

stephenh: Specifically for the ConfirmLineItems test, this makes the `api.getSelectedRowIds` array stable, which will short-circuit the "selected row ids is a new array!", "createRows useMemos against the selected row ids", "changing the rows props changes the `rowStates.rows` observable", "that recalcs the selected row ids to a new array", loop.

PR #931, opened 2023-08-19
feat: Add GridDataRow.onSelect.

stephenh:
To replace a `useEffect` that was watching `api.getSelectedRows` to drive "when the row is unselected" logic.

### DIFF
--- a/src/components/Table/components/Row.tsx
+++ b/src/components/Table/components/Row.tsx
@@ -376,4 +376,6 @@ export type GridDataRow<R extends Kinded> = {
   selectable?: false;
   /** Whether this row should infer its selected state based on its children's selected state */
   inferSelectedState?: false;
+  /** Fired when this row changes selected state. */
+  onSelect?: (isSelected: boolean) => void;
 } & IfAny<R, AnyObject, DiscriminateUnion<R, "kind", R["kind"]>>;
--- a/src/components/Table/utils/RowState.ts
+++ b/src/components/Table/utils/RowState.ts
@@ -1,4 +1,4 @@
-import { makeAutoObservable, observable } from "mobx";
+import { makeAutoObservable, observable, reaction } from "mobx";
 import { GridDataRow } from "src/components/Table/components/Row";
 import { RowStates } from "src/components/Table/utils/RowStates";
 import { SelectedState } from "src/components/Table/utils/TableState";
@@ -44,6 +44,13 @@ export class RowState {
     this.selected = !!row.initSelected;
     this.collapsed = states.storage.wasCollapsed(row.id) ?? !!row.initCollapsed;
     makeAutoObservable(this, { row: observable.ref }, { name: `RowState@${row.id}` });
+    const onSelect = row.onSelect;
+    if (onSelect) {
+      reaction(
+        () => this.selectedState,
+        (state) => onSelect(state === "checked"),
+      );
+    }
   }
 
   /**
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1547,6 +1547,43 @@ describe("GridTable", () => {
     expect(api.current!.getSelectedRowIds()).toEqual([]);
   });
 
+  it("fires onSelect", async () => {
+    // Given a parent with a child
+    const actions: string[] = [];
+    const rows: GridDataRow<NestedRow>[] = [
+      simpleHeader,
+      {
+        ...{
+          kind: "parent",
+          id: "p1",
+          data: { name: "parent 1" },
+          onSelect: (isSelected) => actions.push(`parent 1 ${isSelected}`),
+        },
+        children: [
+          {
+            kind: "child",
+            id: "p1c1",
+            data: { name: "child p1c1" },
+            onSelect: (isSelected) => actions.push(`child 1 ${isSelected}`),
+          },
+        ],
+      },
+    ];
+    const columns = [selectColumn<NestedRow>({})];
+    function Test() {
+      return <GridTable<NestedRow> columns={nestedColumns} rows={rows} />;
+    }
+    const r = await render(<Test />);
+    // When we select all
+    click(cell(r, 0, 1).children[0] as any);
+    // Then all rows are shown as selected
+    expect(actions).toEqual(["parent 1 true", "child 1 true"]);
+    // And when we unselect all
+    click(cell(r, 0, 1).children[0] as any);
+    // Then they are unselected
+    expect(actions).toEqual(["parent 1 true", "child 1 true", "parent 1 false", "child 1 false"]);
+  });
+
   it("getSelectedRows can see update rows", async () => {
     const api: MutableRefObject<GridTableApi<Row> | undefined> = { current: undefined };
     const _columns = [selectColumn<Row>(), ...columns];


src/components/Table/utils/RowState.ts
```diff
@@ -44,6 +44,13 @@ export class RowState {
     this.selected = !!row.initSelected;
     this.collapsed = states.storage.wasCollapsed(row.id) ?? !!row.initCollapsed;
     makeAutoObservable(this, { row: observable.ref }, { name: `RowState@${row.id}` });
+    const onSelect = row.onSelect;
```
bdow, line 47: What are your thoughts on having a top level GridTableProps.onSelectionChange callback vs a callback per row?

stephenh: Hm, yeah, I think I like that--it keeps "the row is the data" better separated from "the behavior/logic that reacts to the row".

I guess a disclaimer is that we already have a fair amount of "select behavior" on the row, i.e. `initSelected`, `selectable`, and `inferSelectedState`...

Hm. Can't decide... Wdyt?

stephenh: We could probably have both... My thought with the invoicing v2 code was that the "on select" logic would be most clear right now to the code where they're creating the row from the form-state.

Here's the current `useEffect`:

https://github.com/homebound-team/internal-frontend/blob/main/src/routes/projects/invoices/invoice-v2/invoice-steps/SelectFixedItems.tsx#L101

And here's where I thought I'd move the logic, same file, but in createRows:

https://github.com/homebound-team/internal-frontend/blob/main/src/routes/projects/invoices/invoice-v2/invoice-steps/SelectFixedItems.tsx#L262

(I'm already adding an `initSelected` that they had been doing via another `useEffect`.)

bdow: I think having an onSelectionChange as a root prop would be more clear than doing the API + useComputed combination that we have been doing everywhere.

stephenh: Hm, don't disagree re root prop, but not following where the API+useComputed would be needed for the row.onSelect?

The onSelect would be passed the row/data, and so could mutate it.

I guess if, inside the onSelect callback, they wanted to make a GQL call? They'd have to pass the mutation into the createRows fn, but, dunno, would not need the API or useComputed afaiu.

src/components/Table/components/Row.tsx
```diff
@@ -376,4 +376,6 @@ export type GridDataRow<R extends Kinded> = {
   selectable?: false;
   /** Whether this row should infer its selected state based on its children's selected state */
   inferSelectedState?: false;
+  /** Fired when this row changes selected state. */
+  onSelect?: (isSelected: boolean) => void;
```
stephenh, line 380: Ah, fwiw @bdow I meant to include a 2nd opts arg that would pass the row, API, etc. Not sure if that makes a difference or not.

src/components/Table/utils/RowState.ts
stephenh: I went ahead and added a `GridTable.onSelect`; we can tweak/remove one/the other as we/see how they work.

src/components/Table/components/Row.tsx
bdow: Tbh, I don't see the need for a row level `onSelect`. I've always seen our `GridDataRow` objects being generated in a loop, where a row level `onSelect` seems excessive compared to the `onRowSelect` callback.
If we really _do_ need this row level callback, then I think it'd be worth adding the `row` as the 2nd param. Otherwise, it isn't super useful if the GridDataRows are generated within a loop. Though, again, I'd probably flag that as "maybe you should be using onRowSelect".

src/components/Table/GridTable.tsx
```diff
@@ -80,6 +81,13 @@ export type GridSortConfig =
       onSort: (orderBy: string | undefined, direction: Direction | undefined) => void;
     };
 
+/** Allows listening to per-kind row selection changes. */
+export type OnRowSelect<R extends Kinded> = {
+  [K in R["kind"]]?: DiscriminateUnion<R, "kind", K> extends { data: infer D }
+    ? (data: D, isSelected: boolean, opts: { row: GridRowKind<R, K>; api: GridTableApi<R> }) => void
+    : (data: undefined, isSelected: boolean, opts: { row: GridRowKind<R, K>; api: GridTableApi<R> }) => void;
+};
```
bdow, line 89: **question:** Does this force the user to add a callback function per kind? Or can they decide which kinds to add callbacks to? 
I'm wondering what will happen if a user selects a parent row, which in turn selects 10 child rows. Are we going to have 11 callbacks executed - one on the parent and 10 on child rows? It looks that way to me.
Would it be more useful and performant to execute a single callback that maybe includes all selection changes? Maybe an array of tuples `[isSelected, row][]`. Though, I also wonder if it's just worth returning an array of the selected rows vs rows that were deselected.
I'm thinking of use cases where we currently update a form state of selected items (granted this should be a multi-select component, not a table), or when we are enabling/disabling bulk actions based on table selections. Seems like this could be a bit slow to execute a single callback at a time.

src/components/Table/utils/TableState.ts
```diff
@@ -53,6 +53,7 @@ export class TableState {
   private initialSortState: SortState | undefined;
   // The server-side `onSort` callback, if any.
   private onSort: ((orderBy: any | undefined, direction: Direction | undefined) => void) | undefined;
+  onRowSelect: OnRowSelect<any> | undefined = undefined;
```
bdow, line 56: **thought (non-blocking):** Do you think it'd be worth giving `TableState` the actual `Row` kind generic vs always having it be `any`?

stephenh: That ends up happening in the next PR :-)

src/components/Table/GridTable.tsx
stephenh: > Does this force the user to add a callback function per kind? Or can they decide which kinds
> to add callbacks to?

Yep, they can add only what they're interested in; the `?` on the `]]?: ...` line makes each kind key optional.

> Are we going to have 11 callbacks executed - one on the parent and 10 on child rows?

Yes.

> Though, I also wonder if it's just worth returning an array of the selected rows vs
> rows that were deselected.

I think it depends on what the user is doing. If the user just wants "an array of all selected rows", they can use the existing `api.getSelectedRows`.

The code I'm refactoring was doing "when _this_ row changes, clear fields x/y/z on it", so, I dunno, I think the api of "this row changed, do what you want with it" is more intuitive for that use case...

> when we are enabling/disabling bulk actions based on table selections.

Maybe this is why we're talking past each other on "row.onSelect" vs. "props.onSelect" change--I think you're thinking of more bulk/table-level reactions, like "_all_ of my selected rows are now x/y/z, so do this", where-as I'm refactoring this code:

https://github.com/homebound-team/internal-frontend/blob/main/src/routes/projects/invoices/invoice-v2/invoice-steps/SelectFixedItems.tsx#L104

Which is very this-row-focused/scoped in terms of the reaction it wants to drive.

Maybe that is rationale for having "two APIs for the same thing?" We have `row.onSelect` for "I just want to mutate things in this row when it changes state" (without having to deduce that from watching "the full list of rows is now x/y/z"), and we have `props.onSelectionChange` when you want to react to "the entire list of selected rows is now x/y/z".

If so, yeah, we could simplify `onSelectionChange` to just be "here's the full list of rows", which per prior comment is really the same thing as "run a `useComputed`/`useReaction` against the `api.getSelectedRows` observable", but admittedly with a more "react-ish" API, i.e. just give us a callback to call, instead of having to watch the api.

I'm good with that, if you are.

stephenh: If you want to double-down on "we shouldn't have any `GridDataRow` callbacks i.e. no row.onSelect", maybe we're talking about two separate props, i.e.:

* `GridTableProps.onRowSelect` --> fired on individual row is/is not selected (this is the API I want for the refactoring at hand), and
* `GridTableProps.onSelectionChange` --> fired with the complete list of new rows, regardless of kind, which I believe is what you're expecting/asking for

bdow: I see what you're referring to now regarding the `SelectFixedItems` component.
That does seem nice to have the row specific actions on, so I see how we'd benefit there.
To me, there seems to be a bit of a code-smell when having callbacks on a GridDataRow. So I'd be okay with your last suggestion. Then we'd definitely want the `onRowSelect` to return `(isSelected: boolean; row: GridDataRow<R>)` so the user could access the row's data/formState when necessary.
But the code-smell isn't strong enough to absolutely push me that route. I'd be okay with us having it per row. Will be curious to see if that ever becomes a problem or not, though.

stephenh: Cool, pushed up the removal of `GridDataRow.onSelect`, so we have just `GridTableProps.onRowSelect`. We can follow up with a `GridTableProps.onSelectionChange`, which gives the user the whole list, at some point.

bdow (approved):

PR #932, opened 2023-08-19
feat: Add Row Api with visibleChildren, selectedChildren

stephenh:
Add a `GridRowApi` that is the same as `GridTableApi` but adds a per-row `visibleChildren` that cells can use to calc things like sums/totals.

I ended up refactoring a lot of the GridRowProps ... I kinda forget why/if I really needed to, but had been meaning to do it for awhile.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1385,7 +1385,7 @@ describe("GridTable", () => {
   });
 
   it("can access visible child rows", async () => {
-    // Given a parent with a child
+    // Given a parent with two children
     const rows: GridDataRow<NestedRow>[] = [
       {
         ...{ kind: "parent", id: "p1", data: { name: "parent 1" } },
@@ -1395,7 +1395,7 @@ describe("GridTable", () => {
         ],
       },
     ];
-    // And a column where the parent counts the number
+    // And a column where the parent counts the number of visible children
     const columns: GridColumn<NestedRow>[] = [
       ...nestedColumns,
       {
@@ -1423,7 +1423,7 @@ describe("GridTable", () => {
   });
 
   it("can access selected child rows", async () => {
-    // Given a parent with a child
+    // Given a parent with two children
     const rows: GridDataRow<NestedRow>[] = [
       {
         ...{ kind: "parent", id: "p1", data: { name: "parent 1" } },
@@ -1449,7 +1449,7 @@ describe("GridTable", () => {
     expect(cell(r, 0, 2)).toHaveTextContent("parent 1");
     expect(cell(r, 1, 2)).toHaveTextContent("child p1c1");
     expect(cell(r, 2, 2)).toHaveTextContent("child p1c2");
-    // Then we show the number of children
+    // Then we show no children are selected
     expect(cell(r, 0, 3)).toHaveTextContent("0");
     // When we select the 1st child
     click(r.select_1);
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1420,6 +1420,10 @@ describe("GridTable", () => {
     expect(cell(r, 1, 2)).toHaveTextContent("child p1c2");
     // And the parent knows the number of visible children
     expect(cell(r, 0, 3)).toHaveTextContent("1");
+    // And when the parent matches the filter
+    await r.rerender(<GridTable<NestedRow> columns={columns} rows={rows} filter="parent" />);
+    // Then the parent knows both children are shown again
+    expect(cell(r, 0, 3)).toHaveTextContent("2");
   });
 
   it("can access selected child rows", async () => {
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1618,6 +1618,51 @@ describe("GridTable", () => {
     expect(cell(r, 1, 1)).toHaveTextContent("foo2");
   });
 
+  it("getSelectedRows doesn't fire if data has not changed", async () => {
+    const api: MutableRefObject<GridTableApi<Row> | undefined> = { current: undefined };
+    const _columns = [selectColumn<Row>(), ...columns];
+    let computedCount = 0;
+
+    // Given a component using a useComputed against getSelectedRows
+    function Test({ rows }: { rows: GridDataRow<Row>[] }) {
+      const _api = useGridTableApi<Row>();
+      api.current = _api;
+      const selectedNames = useComputed(() => {
+        computedCount++;
+        return _api
+          .getSelectedRows("data")
+          .map((r) => r.data.name)
+          .join(",");
+      }, [_api]);
+      return (
+        <div>
+          <div data-testid="selectedNames">{selectedNames}</div>
+          <GridTable<Row> api={_api} columns={_columns} rows={rows} />
+        </div>
+      );
+    }
+    // And the initial render has computed
+    const r = await render(<Test rows={rows} />);
+    expect(computedCount).toBe(2);
+    // And the user has selected one row
+    click(r.select_1);
+    // And selected rows is initially calc-d
+    expect(r.selectedNames().textContent).toEqual("foo");
+    expect(computedCount).toBe(3);
+    // When we re-render with the different row literals, but stable data
+    await r.rerender(
+      <Test
+        rows={[
+          rows[0],
+          { kind: "data", id: "1", data: rows[1].data as any },
+          { kind: "data", id: "2", data: rows[2].data as any },
+        ]}
+      />,
+    );
+    // Then we didn't recompute
+    expect(computedCount).toBe(3);
+  });
+
   it("can select rows via api", async () => {
     // Given a table with selectable rows
     const rows: GridDataRow<NestedRow>[] = [simpleHeader, { kind: "parent", id: "p1", data: { name: "parent 1" } }];
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -2328,7 +2328,7 @@ describe("GridTable", () => {
     const row1_changed: GridDataRow<Row> = { kind: "data", id: row1.id, data: { name: "one", value: 3 } };
     r.rerender(<GridTable columns={columns} rows={[header, row1_changed, row2]} />);
     // And the original row re-rendered
-    expect(row(r, 1).getAttribute("data-render")).toEqual("3");
+    expect(row(r, 1).getAttribute("data-render")).toEqual("2");
     // But the 2nd added row did not
     expect(row(r, 2).getAttribute("data-render")).toEqual("1");
   });
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1384,6 +1384,44 @@ describe("GridTable", () => {
     expectRenderedRows();
   });
 
+  it("can access visible child rows", async () => {
+    // Given a parent with a child
+    const rows: GridDataRow<NestedRow>[] = [
+      {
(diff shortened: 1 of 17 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -217,8 +216,9 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
   const resizeRef = useRef<HTMLDivElement>(null);
 
   const api = useMemo<GridTableApiImpl<R>>(() => {
+    // Let the user pass in their own api handle, otherwise make our own
     const api = (props.api as GridTableApiImpl<R>) ?? new GridTableApiImpl();
-    api.init(persistCollapse, virtuosoRef, rows);
```
stephenh, line 221: The `rows` param hadn't been used in awhile, so removed it

src/components/Table/GridTable.tsx
```diff
@@ -232,6 +232,8 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
   const { tableState } = api;
 
   tableState.onRowSelect = onRowSelect;
+  // useEffect(() => {
+  // }, [tableState, rows]);
```
stephenh, line 236: Should remove this; I really want to drive this through a `useEffect` to remove the mobx/react warnings, but for some reason it doesn't "just work" like the other "move mutating the observable to useEffects" have.

src/components/Table/GridTable.tsx
```diff
@@ -270,58 +272,49 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
 
   // Flatten, hide-if-filtered, hide-if-collapsed, and component-ize the sorted rows.
   const [tableHeadRows, visibleDataRows, keptSelectedRows, tooManyClientSideRows]: [
-    RowTuple<R>[],
-    RowTuple<R>[],
-    RowTuple<R>[],
```
stephenh, line 275: We didn't need the `[Row, ReactElement]` tuple anymore

src/components/Table/GridTable.tsx
```diff
@@ -270,58 +272,49 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
+          key={rs.key}
           {...{
             as,
-            columns,
-            row,
+            rs,
             style,
             rowStyles,
             columnSizes,
-            level: rs.level,
```
stephenh, line 304: Moved these props to just be TableState/RowState lookups inside of RowImpl.

src/components/Table/GridTable.tsx
```diff
@@ -337,11 +330,8 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
   }, [as, api, style, rowStyles, maybeStyle, columnSizes, getCount, filterMaxRows]);
 
   // Push back to the caller a way to ask us where a row is.
-  const { rowLookup } = props;
-  if (rowLookup) {
-    // Refs are cheap to assign to, so we don't bother doing this in a useEffect
-    rowLookup.current = createRowLookup(columns, visibleDataRows, virtuosoRef);
```
stephenh, line 343: `RowLookup` was the last place using the `[Row, ReactElement]` tuples, so I moved it over to use the `GridTableApi` internally...

Really the `RowLookup` API should be removed, as it's been superceded by `GridTableApi`, but am procrastinating on that one.

src/components/Table/utils/RowState.ts
```diff
@@ -10,11 +12,13 @@ import { applyRowFn, HEADER, KEPT_GROUP, matchesFilter, reservedRowKinds } from
...
  * that uses parent/children easier to write, i.e. selected-ness and collapsed-ness.
  */
-export class RowState {
-  /** Our row, only ref observed, so we don't crawl into GraphQL fragments. */
-  row: GridDataRow<any>;
+export class RowState<R extends Kinded> {
+  /** Our row, not actually observed, b/c each `createRows` calc creates unstable rows. */
+  private _row!: GridDataRow<R>;
+  /** Our data, only ref observed, so we don't crawl into GraphQL fragments. */
+  private _data: unknown;
```
stephenh, line 19: I split these up to better model our desired semantics of `rows` props, which are that:

* If you have a `createRows` that returns 100 new row literals, but
* All of them are just returning the same stable `data` fragment from the GQL cache

Then we won't treat that as a change, and won't re-render.

Prior to this PR, the `RowState.row` would actually change, which would mean things like `api.getSelectedRows` would think "I've not newly selected rows", b/c the row literals had changed, when really the selected _data_s had not changed.

src/components/Table/utils/RowState.ts
```diff
@@ -172,6 +196,24 @@ export class RowState {
     );
   }
 
+  get isLastKeptRow(): boolean {
```
stephenh, line 199: I added a few new methods here, b/c it will help short-circuit the row re-rendering; i.e.

* Without this getter, the RowImpl itself will access keptRows, and anytime kept rows changes, all RowImpls will react, which also means re-render
* With this getter, when keptRows changes, all RowStates will re-calc their isLastKeptRow, but nearly all of them will be the same value, so the RowImpl observer watching this getter won't re-run, hence no re-render

src/components/Table/GridTable.test.tsx
```diff
@@ -1655,6 +1655,51 @@ describe("GridTable", () => {
...
+      return (
+        <div>
+          <div data-testid="selectedNames">{selectedNames}</div>
+          <GridTable<Row> api={_api} columns={_columns} rows={rows} />
+        </div>
+      );
+    }
+    // And the initial render has computed
+    const r = await render(<Test rows={rows} />);
+    expect(computedCount).toBe(2);
```
stephenh, line 1683: Really this should be 1, but I'm not sure why it's 2...that said, the main point of the test is that the `toBe(3)` on line 1700 passes, so giving up on figuring this one out for now

src/components/Table/GridTable.test.tsx
```diff
@@ -2247,7 +2292,7 @@ describe("GridTable", () => {
     const row1_changed: GridDataRow<Row> = { kind: "data", id: row1.id, data: { name: "one", value: 3 } };
     r.rerender(<GridTable columns={columns} rows={[header, row1_changed, row2]} />);
     // And the original row re-rendered
-    expect(row(r, 1).getAttribute("data-render")).toEqual("2");
+    expect(row(r, 1).getAttribute("data-render")).toEqual("3");
```
stephenh, line 2295: This is also odd, I don't know why the `data` identity changing triggered this to re-render twice... that said, the main goal of the test is to make sure the _other_ row, on line 2297, does not render/stays at 1.

...so...treating this as quizzical and something I want to fix, but overall not a blocker to the PR imo.

src/components/Table/GridTable.test.tsx
```diff
@@ -1655,6 +1655,51 @@ describe("GridTable", () => {
     expect(cell(r, 1, 1)).toHaveTextContent("foo2");
   });
 
+  it("getSelectedRows doesn't fire if data has not changed", async () => {
```
stephenh, line 1658: This test is not related to the `rowApi.visibleChildren`, but while working on that I realized `RowState.row` was changing too often (after each `createRows` made new row literals), which in turn was causing `api.getSelectedRows` to be too chatty, so I went ahead and fixed it...

Granted, it kinda blew up the PR...

src/components/Table/GridTable.tsx
```diff
@@ -232,11 +233,21 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
   const { tableState } = api;
 
   tableState.onRowSelect = onRowSelect;
-  tableState.setRows(rows);
```
stephenh, line 235: I didn't really mean to tackle this in this PR, but after some of the other refactorings, on a whim I moved this `setRows` into a `useEffect` and it just worked/all tests passed. Which is great b/c this will avoid the React "render cannot set state in another component" errors.

src/components/Table/components/Row.tsx
```diff
@@ -320,22 +309,17 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
...
+ * on the `data` attribute, b/c our `createRows` methods do not create stable row identities;
+ * i.e. if one row changes, they all change.
  *
- * Note that if you're using virtualization, it can be surprising how unnoticeable broken row
- * memoization is.
+ * However, now RowState.row synthesizes a `row` that only reactively changes when
+ * that row's `data` explicitly changes, so we no longer need a custom `propsAreEqual` here.
  */
 // Declared as a const + `as typeof RowImpl` to work with generics, see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37087#issuecomment-656596623
-export const Row = React.memo(observer(RowImpl), (one, two) => {
```
stephenh, line 334: Also turns out that `observer` internally does a `React.memo`, so we don't need to include one ourselves.

src/components/Table/utils/RowState.ts
```diff
@@ -234,6 +307,28 @@ export class RowState {
     return typeof this.row.pin === "string" || (!!this.row.pin && this.row.pin.filter !== true);
   }
 
+  // mobx will cache this getter for us
+  public get api(): GridRowApi<R> {
+    const rs = this;
+    // Copy the GridTableApi + the getVisibleChildren GridRowApi method
+    return {
+      ...this.states.table.api,
```
bdow, line 315: **thought (non-blocking):** Maybe worth adding a comment here to tie it together that the `this` in the returned table state API methods is going to be the TableState instance due to the binding above, and not the RowState instance.

src/components/Table/utils/RowState.ts
```diff
@@ -234,6 +307,28 @@ export class RowState {
...
+        const children = rs.isCalculatingDirectMatch
+          ? rs.visibleDirectlyMatchedChildren.map((cs) => cs.row)
+          : rs.visibleChildren.map((cs) => cs.row);
+        return !kind ? children : children.filter((r) => r.kind === kind);
+      },
+      getSelectedChildren(kind?: R["kind"]): GridDataRow<R>[] {
+        const children = (rs.children ?? []).filter((cs) => cs.isSelected).map((cs) => cs.row);
+        return !kind ? children : children.filter((r) => r.kind === kind);
+      },
+    } as any;
```
bdow, line 329: **question:** Why do we need this `as any`?

src/components/Table/GridTable.test.tsx
```diff
@@ -1655,6 +1732,51 @@ describe("GridTable", () => {
...
+      <Test
+        rows={[
+          rows[0],
+          { kind: "data", id: "1", data: rows[1].data as any },
+          { kind: "data", id: "2", data: rows[2].data as any },
+        ]}
+      />,
+    );
+    // Then we didn't recompute
+    expect(computedCount).toBe(3);
```
bdow, line 1777: Nice work here! Pretty amazing this table can be smart enough to recognize only the difference in the `data` prop.

bdow (approved):

src/components/Table/GridTableApi.ts
```diff
@@ -123,3 +135,9 @@ export class GridTableApiImpl<R extends Kinded> implements GridTableApi<R> {
     this.tableState.deleteRows(ids);
   }
 }
+
+function bindMethods(instance: any): void {
+  Object.getOwnPropertyNames(Object.getPrototypeOf(instance)).forEach((key) => {
+    if (instance[key] instanceof Function && key !== "constructor") instance[key] = instance[key].bind(instance);
+  });
+}
```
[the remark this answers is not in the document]
stephenh: Yep!

src/components/Table/utils/RowState.ts
bdow: Gotcha. Not a big deal. I think we do that in the TableApi file as well somewhere.

PR #933, opened 2023-08-20
feat: Add Stable/useStable and use for GridTableProps.rows/columns

stephenh:


src/components/Filters/Filters.stories.tsx
```diff
@@ -18,7 +18,7 @@ import {
   Filters,
   GridColumn,
   GridDataRow,
-  GridTable,
+  GridTableUnsafe as GridTable,
```
stephenh, line 21: @bdow I'd added `GridTableUnsafe` as an alias with the old/as-is props, but kinda wondering if maybe we should take this opportunity to introduce `Table` as the new component name, which has the new/safe props, and `GridTable` is the old/deprecated component name.

That would make this roll out a lot easier, b/c all exists clients of `GridTable` wouldn't have to be either a) renamed over to the `GridTableUnsafe` alias, or b) have their non-memoed rows/columns fixed.

...although arguably b) would be a great forcing function, it just means rolling this out will be a lot more work/updates.

Wdyt?

bdow: Yea, I like that idea to rename the "new" to just `Table`. Then we won't block updating Beam on updating existing GridTable implementations.

bdow (approved):
Nice! Would love to do a FE Guild discussion on this.

src/components/Filters/Filters.stories.tsx
stephenh: Cool! Pushed up the "GridTable stays the same, Table is the new safe one".

src/hooks/useStable.ts
```diff
@@ -0,0 +1,74 @@
...
+  [K in keyof Props]: Props[K] extends Stable<infer T> ? T : Props[K];
+};
+
+declare module "react" {
+  // Extend useMemo's return values to be automatically considered stable.
+  //
+  // With this extension, we could almost get away without our own `useStable` hook,
+  // but we can't use the extension mechanism to restrict useMemo's `deps` to be
+  // `ReadonlyArray<StableDep>`.
+  function useMemo<T>(factory: () => T, deps: ReadonlyArray<unknown> | undefined): Stable<T>;
```
stephenh, line 73: @bdow FWIW during demos, answering a question for @khalidwilliams , I'm kinda thinking we should back away from this change, which "upgrades" `useMemo` to produce stable values, and have `useStable` be the only one that can make stable values.

The reason is that, with the TS typing, I'm unable to "upgrade" `useMemo`'s dependency list to be _only_ stable values, and it is precisely this restriction that we'd need to have prevented the bug I fixed.

I.e. the bug I fixed, the `const columns = useMemo(...)` was "already useMemo-d" _but_ the `elevations` dependency was itself an unstable array:

https://github.com/homebound-team/internal-frontend/pull/4159/files?w=1#diff-eb71015b91b3c704dd3488a15c973047da7e637251f2d4a5b2f467c0861c73f9R146

So, really, the fix we need to be 100% bulletproof to performance is not just: a) the user uses `useStable` or `useMemo` for the rows/columns props, but also b) any _dependencies_ the user used for `useStable` or `useMemo` must themselves be stable.

And I cannot "fix" `useMemo` to have this restriction (it would also be a huge breaking change), so I think we'd have to just say "now all `const rows = useMemo(...)`s have to be `const rows = useStable(...)`s" and deal with the fallout.

I'm pretty convinced this is the right direction, even though it's a more disruptive change, b/c all of our `useMemo`s lines need to become `useStable`s--and also make sure their dependencies are stables, but that's the whole point, we need to have the stableness check be transitively applied to dependencies of our dependency.

PR #934, opened 2023-08-21
feat: Add GridTableApi.getVisibleRows.

stephenh:
To move LotSequencePage off of RowLookup, which since it's a ref, is not being set in a test due to the useEffect-ification of GridTable.


stephenh:
Merging to unblock the beam bump in internal-frontend.

PR #935, opened 2023-08-21
fix: Fix missing RowState to GridDataRow map.

stephenh:
Now with a test :-/


stephenh:
...again merging to unblock...

PR #936, opened 2023-08-22
fix: Make getSelectedRowIds stable.

stephenh:
Even if the `data` elements are changing on each render.

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -1745,12 +1745,12 @@ describe("GridTable", () => {
     }
     // And the initial render has computed
     const r = await render(<Test rows={rows} />);
-    expect(computedCount).toBe(2);
+    expect(computedCount).toBe(1);
     // And the user has selected one row
     click(r.select_1);
     // And selected rows is initially calc-d
     expect(r.selectedNames().textContent).toEqual("foo");
-    expect(computedCount).toBe(3);
+    expect(computedCount).toBe(2);
     // When we re-render with the different row literals, but stable data
     await r.rerender(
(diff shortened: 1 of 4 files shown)


src/components/Table/GridTableApi.ts
```diff
@@ -96,19 +104,42 @@ export class GridTableApiImpl<R extends Kinded> implements GridTableApi<R> {
   }
 
   public getSelectedRowIds(kind?: string): string[] {
-    return this.getSelectedRows(kind).map((row: any) => row.id);
+    return this.getSelectedRowIdsImpl(kind ?? undefined);
+  }
+
+  // impl with required param for computedFn
+  private getSelectedRowIdsImpl(kind: string | undefined): string[] {
```
stephenh, line 111: The `computedFn` implementation requires a constant number of parameters on every invocation, which, dunno, not sure why, seems kinda annoying imo, but to keep our current API of "kind is optional", we have to materialized the `kind = undefined` before calling into the computed fn.

src/components/Table/GridTable.test.tsx
```diff
@@ -1745,12 +1745,12 @@ describe("GridTable", () => {
     }
     // And the initial render has computed
     const r = await render(<Test rows={rows} />);
-    expect(computedCount).toBe(2);
+    expect(computedCount).toBe(1);
```
bdow, line 1748: 🎉

bdow (approved):

src/components/Table/GridTable.test.tsx
stephenh: Yeah right!

PR #944, opened 2023-08-30
feat: Add loading spinner option to infinite scroll

blambillotte:
* Adds a new `nextPageLoading` option to `GridTable` `infiniteScroll` prop that will show a loading spinner in the footer when provided the `query.loading` boolean. In an ideal world, the scroll would truly be "infinite" AKA we fetch the next page before a user reaches the end, but a slow connection or very fast scroll could leave an awkward gap where it's unclear if you've reached the true end or are just waiting for the next page. 
* Demo Story: https://60466f7d43c37c0021788867-ivzonzyhpo.chromatic.com/?path=/story/components-table-gridtable--infinite-scroll-with-loader

https://www.loom.com/share/9d6a9dd24ee74b8f9001e1a6dda91a6e?sid=e13b2cc3-ffc0-4294-aed0-2e9bae5c9853

### DIFF
--- a/src/components/Table/types.ts
+++ b/src/components/Table/types.ts
@@ -127,4 +127,6 @@ export type InfiniteScroll = {
   onEndReached: (index: number) => void;
   /** The number of pixels from the bottom of the list to eagerly trigger `onEndReached`. The default is 500px. */
   endOffsetPx?: number;
+  /** When true, will append a loading spinner footer to let the user know the next page is still loading */
+  nextPageLoading?: boolean;
 };
--- a/src/components/Table/types.ts
+++ b/src/components/Table/types.ts
@@ -123,10 +123,8 @@ export type Pin = { at: "first" | "last"; filter?: boolean };
 export type IfAny<T, Y, N> = 0 extends 1 & T ? Y : N;
 
 export type InfiniteScroll = {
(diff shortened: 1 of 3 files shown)


src/components/Table/types.ts
```diff
@@ -127,4 +127,6 @@ export type InfiniteScroll = {
   onEndReached: (index: number) => void;
   /** The number of pixels from the bottom of the list to eagerly trigger `onEndReached`. The default is 500px. */
   endOffsetPx?: number;
+  /** When true, will append a loading spinner footer to let the user know the next page is still loading */
+  nextPageLoading?: boolean;
```
stephenh, line 131: Thought: Wdyt about `onEndReached` being a callback that returns a `Promise<void>`, and we show the spinner until the promise resolves? Then if the callback does an `await fetchMore`, the spinner will auto-show/auto-resolve without needing the `nextPageLoading` prop?

stephenh (commented):
Looks slick!

src/components/Table/types.ts
blambillotte: ah! that's a very good idea - will give it a go

stephenh (approved):
Nice!

PR #946, opened 2023-09-11
feat: [SC-39801] allow independent parent select for GridTable rows

bsholmes:
We have a use case that requires parent rows to be selectable independently of their children, so we should allow that to happen with a simple flag.

This behavior is not handled by the inferSelectedState flag.

[SC-39801](https://app.shortcut.com/homebound-team/story/39801/selecting-add-row-always-selects-and-won-t-let-you-deselect-modify-rows)

### DIFF
--- a/src/components/Table/components/Row.tsx
+++ b/src/components/Table/components/Row.tsx
@@ -360,6 +360,4 @@ export type GridDataRow<R extends Kinded> = {
   selectable?: false;
   /** Whether this row should infer its selected state based on its children's selected state */
   inferSelectedState?: false;
-  /** Whether this row should be selectable as a parent without pushing its selected-ness to its visible children */
-  independentParentSelect?: true;
 } & IfAny<R, AnyObject, DiscriminateUnion<R, "kind", R["kind"]>>;
--- a/src/components/Table/components/Row.tsx
+++ b/src/components/Table/components/Row.tsx
@@ -360,4 +360,6 @@ export type GridDataRow<R extends Kinded> = {
   selectable?: false;
   /** Whether this row should infer its selected state based on its children's selected state */
   inferSelectedState?: false;
+  /** Whether this row should be selectable as a parent without pushing its selected-ness to its visible children */
+  independentParentSelect?: true;
 } & IfAny<R, AnyObject, DiscriminateUnion<R, "kind", R["kind"]>>;
--- a/src/components/Table/utils/RowState.ts
+++ b/src/components/Table/utils/RowState.ts
@@ -168,8 +168,10 @@ export class RowState<R extends Kinded> {
     }
     // We don't check inferSelectedState here, b/c even if the parent is considered selectable
     // on its own, we still push down selected-ness to our visible children.
-    for (const child of this.visibleChildren) {
-      child.select(selected);
+    if(!(this.isParent && this.independentParentSelect)) {
+      for (const child of this.visibleChildren) {
+        child.select(selected);
+      }
     }
   }
 
@@ -223,6 +225,10 @@ export class RowState<R extends Kinded> {
 
   private get inferSelectedState(): boolean {
     return this.row.inferSelectedState !== false;
+  }
+
+  private get independentParentSelect(): boolean {
+    return this.row.independentParentSelect !== false;
   }
 
   /** Returns this row and, if we're not collapsed, our children. */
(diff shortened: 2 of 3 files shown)


src/components/Table/components/Row.tsx
```diff
@@ -360,4 +360,6 @@ export type GridDataRow<R extends Kinded> = {
   selectable?: false;
   /** Whether this row should infer its selected state based on its children's selected state */
   inferSelectedState?: false;
+  /** Whether this row should be selectable as a parent without pushing its selected-ness to its visible children */
+  independentParentSelect?: true;
```
stephenh, line 364: Ah huh! Yeah, I guess `inferSelectedState: false` still pushes down selected-ness to children...

I'm kinda wondering if we could make `inferSelectedState` a tri-state, something like:

* `inferSelectedState: undefined` --> default behavior, a parent will infer checked from children + push down checked into children
* `inferSelectedState: "push-to-children" --> current `false`, a parent will push to children, but children won't infer up to the parent
* `inferSelectedState`: false --> what you're creating, which is that selected-ness doesn't push either way...

I dunno, it looks like we only use `inferSelectedState: false` twice in all of blueprint, one place is this PO page:

https://blueprint.qa-homebound.com/developments/dev:108/purchase-orders?search=

image

And the 2nd is...the page you are updating to use this `independentParentSelect`.

Which means, AFAICT, that the dev POs page would be the only place in the app wanting the current behavior, of "parent does push down to children".

Fwiw about ~2 months ago I refactored the state management here, and being surprised that `inferSelectedState: false` still pushed selected-ness down into the children, hence adding that comment in RowState.ts line 169 that it's "on purpose because the tests say no" (and I didn't want to change semantics while doing that refactoring).

But, I dunno, now I'm thinking maybe we should just keep `inferSelectedState: false` as-is (don't do the fancy tri-state), and just say if you turn on `inferSelectedState: false`, then we won't push down selected-ness into children either.

@bdow do you have any recollection of how much we'd want to keep "push selected-ness into children w/inferSelectedState: false is set"?

bsholmes: Yeah I don't disagree with any of that, but you may be overthinking it. These are two different sets of behavior and having another flag for parent-to-child selection is a simple solution to the issue at hand. if we conflate both behaviors into the existing flag we may have to revisit this if we find a case where we want one behavior and not the other. The tri-state flag would also work but just seems unnecessary.

stephenh: > Yeah I don't disagree with any of that, but you may be overthinking it.

:-) Generally agreed, although I'm worried I should have edited out my tri-state musing (which agreed was over-thinking it), to the even simpler "turns out we don't really want the parent-push-down behavior at all".

Granted, you're right that we might be asked to add this split behavior back, but :shrug: seems like it'd be easy to do later.

I'm also suspecting that most users who set `inferSelectedState: false`, without potentially knowing about `independentParentSelect`, will later have to do follow-up fixes when users/PMs realize they also don't want the same "don't auto-select children" behavior.

> I do think this should be a table wide property, rather than per-row.

Don't disagree with the intuition, but the dev POs page happens to already use this, the top-level rows are "regular inferred states" but the 2nd-level commitments & 3rd-level change orders are "grouped but disconnected / separate-ly selectable":

https://blueprint.qa-homebound.com/developments/dev:108/purchase-orders?search=

So, to your point Brandon, the relationship is more for "logical grouping" of independently-selectable entities and not a strict "the children are authoritatively/fundamentally children of the parent".

Given that articulation, I'm even more ~convinced that `inferSelectedState: false` should turn off behavior "children push up" and "parent push down" behaviors, b/c it turns on the "these are separate things", and it's safer/better to just snip the inference in both directions.

bsholmes: > Given that articulation, I'm even more ~convinced that inferSelectedState: false should turn off both "children push up" and "parent push down" behaviors, b/c it turns on the "these (parent / child) rows are truly separate things", such that it's safer/better to just snip the inference in both directions.

Okay, I'm sold, will make that change.

stephenh (approved):
:+1: thanks!

src/components/Table/utils/RowState.ts
```diff
@@ -166,10 +166,11 @@ export class RowState<R extends Kinded> {
     if (this.row.selectable !== false) {
       this.selected = selected;
     }
-    // We don't check inferSelectedState here, b/c even if the parent is considered selectable
-    // on its own, we still push down selected-ness to our visible children.
-    for (const child of this.visibleChildren) {
-      child.select(selected);
+
+    if(!(this.isParent && this.inferSelectedState)) {
```
stephenh, line 170: I noticed all of the test failures, maybe remove the `!`? :crossed_fingers:

PR #951, opened 2023-09-18
feat: Hackday: make beam modals draggable

bsholmes:
using a custom selector using ids so that only the header is draggable which prevents issues when using input elements in the modal.

Lots of extraneous changes from linting

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "mobx-utils": "^6.0.8",
     "react-aria": "^3.26.0",
     "react-day-picker": "8.0.7",
+    "react-draggable": "^4.4.5",
     "react-popper": "^2.3.0",
     "react-router": "^5.3.4",
     "react-router-dom": "^5.3.4",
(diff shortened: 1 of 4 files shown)


stephenh (approved):
Seems good to me! @bdow any concerns about pulling in `react-draggable`?

(Not saying we should re-implement the dragging natively, for sure we should use a package, but also just balancing "is dragging must have ask from UX atm?" vs. "yet another dependency in package.json".)

bdow:
> Seems good to me! @bdow any concerns about pulling in `react-draggable`?

I don't really have any concerns about `react-draggable`. I do see that `react-aria` has a (useDrag)[https://react-spectrum.adobe.com/react-aria/useDrag.html] hook, though I don't know much about it. Not sure if we could just use that? Maybe that's not enough right out of the box.

bdow (commented):
@bsholmes - did you get sign off from the Design team about this? I would like to get their feedback around the UX of this. Right now, there is no indication that this can be dragged, whether that is by having an explicit "drag handle", or changing the cursor to look like it can grab and is grabbing while dragging. Could be worth providing some user feedback.

bdow (commented):
@bsholmes  - I noticed that there are 3 different elements with the `id="draggable"` on them. It doesn't appear to be causing issues with the react-draggable library, but we should not have duplicate `id` properties within the DOM.

Screenshot 2023-09-19 at 1 49 29 PM

PR #953, opened 2023-09-25
feat: Support multiline inputs for select fields

bdow:
SelectField and MultiSelectField will now support multiline inputs either via `multiline` prop being set to true, or whenever the `PresentationProvider.wrap` prop is set to `true`. E.g. When GridTable is rendered with `rowHeight: "flexible"` (default) then the SelectFields will support wrapping text.

### DIFF
--- a/src/components/Label.tsx
+++ b/src/components/Label.tsx
@@ -27,7 +27,6 @@ export const Label = React.memo((props: LabelProps) => {
 
 /** Used for showing labels within text fields. */
 export function InlineLabel({ labelProps, label, contrast, multiline = false, ...others }: LabelProps) {
-  console.log("multiline", multiline);
   return (
     <label
       {...labelProps}
--- a/src/components/Label.tsx
+++ b/src/components/Label.tsx
@@ -10,6 +10,7 @@ interface LabelProps {
   // If set, it is recommended to wrap in an element with `position: relative;` set, as the label will have an absolute position.
   hidden?: boolean;
   contrast?: boolean;
+  multiline?: boolean;
 }
 
 /** An internal helper component for rendering form labels. */
@@ -25,9 +26,14 @@ export const Label = React.memo((props: LabelProps) => {
 });
 
 /** Used for showing labels within text fields. */
-export function InlineLabel({ labelProps, label, contrast, ...others }: LabelProps) {
+export function InlineLabel({ labelProps, label, contrast, multiline = false, ...others }: LabelProps) {
+  console.log("multiline", multiline);
   return (
-    <label {...labelProps} {...others} css={Css.smMd.nowrap.gray900.prPx(4).add("color", "currentColor").$}>
+    <label
+      {...labelProps}
+      {...others}
+      css={Css.smMd.nowrap.gray900.prPx(4).add("color", "currentColor").asc.if(multiline).asfs.pt1.$}
+    >
       {label}:
     </label>
   );
--- a/src/inputs/internal/ComboBoxInput.tsx
+++ b/src/inputs/internal/ComboBoxInput.tsx
@@ -1,9 +1,18 @@
-import React, { InputHTMLAttributes, Key, LabelHTMLAttributes, MutableRefObject, ReactNode, useState } from "react";
+import React, {
+  ChangeEvent,
+  InputHTMLAttributes,
+  Key,
+  LabelHTMLAttributes,
+  MutableRefObject,
+  ReactNode,
+  useState,
+} from "react";
 import { mergeProps } from "react-aria";
 import { ComboBoxState } from "react-stately";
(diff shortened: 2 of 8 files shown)


src/components/Label.tsx
```diff
@@ -25,9 +26,14 @@ export const Label = React.memo((props: LabelProps) => {
 });
 
 /** Used for showing labels within text fields. */
-export function InlineLabel({ labelProps, label, contrast, ...others }: LabelProps) {
+export function InlineLabel({ labelProps, label, contrast, multiline = false, ...others }: LabelProps) {
+  console.log("multiline", multiline);
```
stephenh, line 30: Nit: snuck in :-)

src/inputs/internal/ComboBoxInput.tsx
```diff
@@ -69,9 +84,19 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
   const showFieldDecoration =
     (!isMultiSelect || (isMultiSelect && !isFocused)) && fieldDecoration && selectedOptions.length === 1;
 
+  // To Discuss: Do we want to make `multiline = true` the default behavior?
+  // If we do, we can infer whether we allow wrapping or not based on the PresentationProp. This will currently only be set for "rowHeight: 'fixed'" GridTable styles.
+  // Or, we can keep our existing default behavior, and allow the user to set `multiline = true` if they want to allow wrapping.
+  // If we change the default then we could have unexpected text wrapping in places like Filters.
+  const multilineProps = wrap ? { textAreaMinHeight: 0, multiline: true } : {};
```
stephenh, line 91: Question: Should this `multiline: true` use the `multiline` local variable? Not sure...

src/inputs/internal/ComboBoxInput.tsx
```diff
@@ -57,9 +68,13 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
     resetField,
     isTree,
     listBoxRef,
+    inputRef,
+    inputWrapRef,
+    multiline = false,
     ...otherProps
   } = props;
 
+  const { wrap = multiline } = usePresentationContext();
```
stephenh, line 77: Kinda curious, seems like we have both `wrap` & `multiline` now...will they always be aligned, i.e. if `wrap=true` then `multiline=true`, and if `wrap=false` then `multiline=false`?

Wondering if like `multiline=false` happens on 73, then they had `wrap=true` set in PC, we'll end up with `wrap=true` but `multiline=false` ... is that okay/a supported configuration?

stephenh (approved):
Thank you! Really appreciate hoping on this/sneaking it in. :pray:

src/inputs/internal/ComboBoxInput.tsx
bdow: Good question. I glazed over this a bit when doing it so will try to make this a bit simpler / easier to read.

src/components/Label.tsx
bdow: Thanks! Removed now.

src/inputs/internal/ComboBoxInput.tsx
bdow: `multiline` should always be `true` if we ever way the text to wrap. There are two conditions in which we want the text to wrap; 1. The caller defined `multiline: true`, or 2. `PresentationContext.wrap` is set to `true`.
In this case it could be that `multiline` is `undefined`, but the PresentationContext property is defined as `true`, so we'd want to still set `multiline: true`.

src/inputs/internal/ComboBoxInput.tsx
bdow: Alrighty, just went back over it and added a comment above a new variable `allowWrap`. This variable is going to be true if either `ComboboxInputProps.multiline` is **true**, or `PresentationContext.wrap` is **true**. i.e.
```
const allowWrap = wrap || multiline;
```
For now we'll always expect that if either is set, then we allow it to wrap. Maybe we'll eventually want the local `multiline` prop to override the presentation provider property if it is explicitly set. But I think for now this is fine.

src/inputs/internal/ComboBoxInput.tsx
```diff
@@ -57,9 +68,16 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
...
     listBoxRef,
+    inputRef,
+    inputWrapRef,
+    multiline = false,
     ...otherProps
   } = props;
 
+  const { wrap = false } = usePresentationContext();
+
+  // Allow the field to wrap whether the caller has explicitly set `multiline=true` or the `PresentationContext.wrap=true`
```
stephenh, line 79: Nice, looks great!

PR #955, opened 2023-09-27
fix: Improve layout for multiline TextFields

bdow:
Adds padding for multiline textfields to match the height of their single line counter parts

### DIFF
--- a/src/inputs/TextFieldBase.tsx
+++ b/src/inputs/TextFieldBase.tsx
@@ -142,9 +142,11 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
           .mhPx(compactFieldHeight - maybeSmaller).$),
     },
     input: {
-      ...Css.w100.mw0.outline0.fg1.bgColor(bgColor).if(multiline).br4.$,
+      ...Css.w100.mw0.outline0.fg1.bgColor(bgColor).$,
       // Not using Truss's inline `if` statement here because `addIn` properties do not respect the if statement.
       ...(contrast && Css.addIn("&::selection", Css.bgGray800.$).$),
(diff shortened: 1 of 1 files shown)


src/inputs/TextFieldBase.tsx
```diff
@@ -142,9 +142,11 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
           .mhPx(compactFieldHeight - maybeSmaller).$),
     },
     input: {
-      ...Css.w100.mw0.outline0.fg1.bgColor(bgColor).if(multiline).br4.$,
+      ...Css.w100.mw0.outline0.fg1.bgColor(bgColor).$,
       // Not using Truss's inline `if` statement here because `addIn` properties do not respect the if statement.
       ...(contrast && Css.addIn("&::selection", Css.bgGray800.$).$),
+      // For "multiline" fields we add top and bottom padding of 7px for compact, or 11px for non-compact, to properly match the height of the single line fields
+      ...(multiline ? Css.br4.h100.pyPx(compact ? 7 : 11).add("resize", "none").$ : Css.truncate.$),
```
stephenh, line 149: Only for my own curiosity, but truncate is new; was it probably just missing before, but is a good idea to have?

stephenh (approved):

src/inputs/TextFieldBase.tsx
bdow: It is actually not new! These styles were being defined in line 250 before, but I decided to move them up here. There was really no reason for them to be separated like that.

stephenh: Oh you're right! Totally missed that. TY!

PR #957, opened 2023-10-03
feat: IconCard & IconCardGroup

0ces:
Individual (states):
image

Group:
image


src/inputs/IconCardGroup.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export function Default() {
+  const [values, setValues] = useState<string[]>(["math"]);
+
+  return (
+    <div>
+      <Chips values={values} />
+      <IconCardGroup
+        label="Icon Card Group"
+        // options={Object.values(IconCardGroupValues).map((value, i) => ({
```
blambillotte, line 27: nit: remove

src/inputs/IconCardGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+export function IconCardGroup(props: IconCardGroupProps) {
+  const { fieldProps } = usePresentationContext();
+  const {
+    options,
+    label,
+    labelStyle = fieldProps?.labelStyle ?? "above",
+    values,
+    errorMsg,
+    helperText,
+    columns = 1,
```
blambillotte, line 44: Question: what do you think about just using a flex container since the box sizes are fixed you'd still get an even grid, but then the columns would be responsive based on the parent container size 🤔

blambillotte (approved):
nice!

src/inputs/IconCardGroup.tsx
0ces: Sure, that was my idea at first and then I just tried to match the `CheckboxGroup` but, yeah I'll leave it as a flex with a flex-wrap

PR #964, opened 2023-10-11
wip

stephenh:


src/inputs/SelectField.test.tsx
```diff
@@ -403,20 +403,28 @@ describe("SelectFieldTest", () => {
   it("supports boolean values properly", async () => {
     // Given a select field with boolean and an undefined values
     const onSelect = jest.fn();
-    const r = await render(
-      <SelectField
-        label="label"
-        value={true}
```
stephenh, line 409: Before this refactoring, this `value` could stay hard-coded and `ComboBoxBase` had its own internal version of values that could drift from the `value` prop.

But now, kinda on accident but I think it makes sense, the internal value is always derived from the `value` prop, so it behaves more like a controlled component.

Which is fine, afaict not a breaking change, but meant these tests that were hard-coding `value=true` needed to change.

src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -117,44 +122,33 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
...
-      selectedKeys: selectedOptions?.map((o) => valueToKey(getOptionValue(o))) ?? [],
-      inputValue: getInputValue(
-        options.filter((o) => values?.includes(getOptionValue(o))),
-        getOptionLabel,
-        multiselect,
-        nothingSelectedText,
-      ),
-      filteredOptions: options,
-      allOptions: options,
-      selectedOptions,
```
stephenh, line 137: `filteredOptiosn`, `allOptions`, and `selectedOptions`, and `selectedKeys` are now all derived from the useMemo-d `options` / `values` or what not.

src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -286,65 +253,34 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
...
-          allOptions: options,
-        };
-      });
-    },
-    // We're primarily only re-setting `allOptions`, and so recalc selected as well, but we don't
-    // want to depend on values/etc., b/c we'll defer to their useEffects to update their state
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [options],
-  );
+  // Reset inputValue when closed or selected changes
```
stephenh, line 268: Kinda surprised, this useEffect ends up being a pretty cute place to handle resetting inputValue as various conditions change.

src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -433,12 +369,11 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
...
-type FieldState<O> = {
-  selectedKeys: Key[];
+type FieldState = {
   inputValue: string;
-  filteredOptions: O[];
-  selectedOptions: O[];
-  allOptions: O[];
+  // We need separate `searchValue` vs. `inputValue` b/c we might be showing the
+  // currently-loaded option in the input, without the user having typed a filter yet.
+  searchValue: string | undefined;
```
bdow, line 376: **praise:** Nice, that does help keep this a bit cleaner

bdow (approved):

PR #965, opened 2023-10-12
feat: Add column names to tableSnapshot.

stephenh:
So we can assert against only a subset of columns, and have less noisy assertions especially for tables like Specs & Selections with ~lots of columns.


blambillotte (approved):
Nice!

PR #974, opened 2023-11-23
feat: add inModal state on useModal

allangaldinosilva:
Add a new state, `isModalOpen`, to the `useModal` hook to track the open/closed state of a modal in a page.

It will initially be used on Search Query improvements addressed on this PR https://github.com/homebound-team/internal-frontend/pull/4385

From my tests, it did not trigger unnecessary re-renders for other components.

cc @bdow / @stephenh


src/components/Modal/useModal.tsx
```diff
@@ -51,6 +52,7 @@ export function useModal(): UseModalHook {
           modalState.current.api.current.setSize(size);
         }
       },
+      inModal: () => isDefined(modalState.current),
```
bdow, line 55: This could be an issue as anything within the BeamContext will think it is "inModal" when the modal is opened, even if it is behind the modal.

src/components/BeamContext.tsx
```diff
@@ -7,9 +7,10 @@ import { SnackbarProvider } from "src/components/Snackbar/SnackbarContext";
 import { SuperDrawer } from "src/components/SuperDrawer/SuperDrawer";
 import { ContentStack } from "src/components/SuperDrawer/useSuperDrawer";
 import { CanCloseCheck, CheckFn } from "src/types";
-import { EmptyRef } from "src/utils/index";
+import { EmptyRef, isDefined } from "src/utils/index";
 import { RightPaneProvider } from "./Layout";
 import { ToastProvider } from "./Toast/ToastContext";
+import { ModalContext, ModalProvider } from "./Modal/ModalContext";
```
bdow, line 13: **request:** Can you remove these unused imports? I may have accidentally left them in when I made changes.

src/components/Modal/Modal.stories.tsx
```diff
@@ -111,6 +112,24 @@ export function ModalForm() {
   );
 }
 
+export function InModalState() {
```
bdow, line 115: **nit:** I don't think this story should be necessary. We should be able to verify it works properly via tests instead. Could we remove this story to save on some snapshots?

src/components/Modal/useModal.test.tsx
```diff
@@ -95,4 +95,27 @@ describe("useModal", () => {
...
+      return <div data-testid="modalContent">Modal Content: InModal? {String(inModal)}</div>;
+    }
+
+    // When rendering the test app
+    const r = await render(<TestApp content={<TestModalContent />} />);
+
+    // Then the test app should not be in a modal
+    expect(r.testApp).toHaveTextContent("Behind Modal: InModal? false");
+    // And the modal content should be in a modal
+    expect(r.modalContent).toHaveTextContent("Modal Content: InModal? true");
```
bdow, line 119: **praise:** Perfect! Nice test

bdow (approved):
Just some small cleanup. Otherwise looks good!

PR #979, opened 2023-12-06
feat: Support 'fullWidth' prop on all text fields

bdow:
Adds support for a `PresentationFieldProps.fullWidth` property. The following components will respect the property:
- FormLines
- TextField
- TextAreaField
- RichTextField
- SelectField
- MultiSelectField
- DateField
- NumberField
- TreeSelectField

Updates the `FormLines` component to default to the `"lg"` size instead of `"full"`. This will hopefully minimize the impact on existing layouts as the previous default width for each field within `FormLines` was 550px. By using the `lg` size, we'll continue to default the fields within `FormLines` to 550px.

### DIFF
--- a/src/inputs/internal/ComboBoxInput.tsx
+++ b/src/inputs/internal/ComboBoxInput.tsx
@@ -90,8 +90,6 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
   const multilineProps = allowWrap ? { textAreaMinHeight: 0, multiline: true } : {};
   useGrowingTextField({ disabled: !allowWrap, inputRef, inputWrapRef, value: inputProps.value });
 
-  console.log("otherProps", otherProps);
-
   return (
     <TextFieldBase
       {...otherProps}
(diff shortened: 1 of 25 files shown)


src/inputs/internal/ComboBoxInput.tsx
```diff
@@ -90,6 +90,8 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
   const multilineProps = allowWrap ? { textAreaMinHeight: 0, multiline: true } : {};
   useGrowingTextField({ disabled: !allowWrap, inputRef, inputWrapRef, value: inputProps.value });
 
+  console.log("otherProps", otherProps);
```
stephenh, line 93: Snuck in

stephenh (approved):
Is the idea that text fields were just getting ridiculous too wide, so 550px is a better default max size? That makes sense...

Per making FormLines be lg, that sgtm but honestly not a ton of places use that yet :shrug: . Should be good though!

bdow:
> Is the idea that text fields were just getting ridiculous too wide, so 550px is a better default max size? That makes sense...

No, `550px` is currently the default max-width. With this PR we are allowing the option to remove the max-width now. It sounds like in some instances we want to allow them to be wider.

stephenh (approved):
> With this PR we are allowing the option to remove the max-width now.

Ah gotcha! Kinda curious but sounds good.

PR #985, opened 2023-12-15
fix: Fix Switch not toggling.

stephenh:


src/inputs/Switch.tsx
```diff
@@ -49,12 +49,13 @@ export function Switch(props: SwitchProps) {
   const { isFocusVisible: isKeyboardFocus, focusProps } = useFocusRing(otherProps);
   const { hoverProps, isHovered } = useHover(ariaProps);
   const tooltip = resolveTooltip(disabled, props.tooltip);
+  const tid = useTestIds(otherProps, label);
 
   return (
-    <div
+    <label
```
stephenh, line 55: A previous styling change changed this top-level element from a `label` to a `div`, but that seems to have broken the `input`-flips-to-true/false because the `input` isn't wrapped in a `label` anymore.

This is a little janky b/c we now have two `label`s in this component, but the tests/stories show that this fixes the regression, so I'm prioritizing getting the fix out vs. a perfect DOM structure that doesn't have duplicate `label`s.

bdow (commented):
Thank you!

PR #987, opened 2023-12-15
feat: [SC-43819] draggable table rows

bsholmes:
Add drag events to GridTable rows in order to enable drag & drop reordering

https://app.shortcut.com/homebound-team/story/43819/implement-drag-drop-reordering-for-the-configure-options-table


src/components/Table/GridTable.stories.tsx
```diff
@@ -1849,3 +1854,146 @@ export function Headers() {
     </div>
   );
 }
+
+/**
+ * Shows how drag & drop reordering can be implemented with GridTable drag events
+ */
+export function DraggableRows() {
```
stephenh, line 1861: Praise: I popped this open in storybook and it looks slick!

src/components/Table/GridTable.stories.tsx
```diff
@@ -1849,3 +1854,146 @@ export function Headers() {
...
+    data: (data, { row }) => row.order,
+    spacer: " ",
+  };
+  const actionColumn: GridColumn<OrderedRow<Data>> = {
+    header: "Action",
+    data: () => <div>Actions</div>,
+    spacer: " ",
+    clientSideSort: false,
+  };
+  const spacerRow: GridDataRow<OrderedRow<Data>> = {
```
stephenh, line 1879: Thought: I think this spacer row is a good idea; afaict it like follows the mouse around to show where the being-dragged row will end up going?

src/components/Table/GridTable.stories.tsx
```diff
@@ -1849,3 +1854,146 @@ export function Headers() {
...
+  };
+
+  return (
+    <GridTable
+      columns={[nameColumn, orderColumn, actionColumn]}
+      onRowDragStart={{ data: onDragStart, spacer: onDragStart }}
+      onRowDragEnd={{ data: onDragEnd, spacer: onDragEnd }}
+      onRowDrop={{ data: onDrop, spacer: onDrop }}
+      onRowDragEnter={{ data: onDragEnter, spacer: onDragEnter }}
+      onRowDragOver={{ data: onDragOver, spacer: onDragOver }}
```
stephenh, line 1995: Thought: I think this makes a ton of sense for prototyping/spiking; wondering if maybe before shipping the final API might be more short & sweet, like something like:

* `onRowMove((row, oldIndex, newIndex) => ...)`

And have all of the low-level drag math, and ideally like management of that spacer row (being injected/moving around within the page's rows), all managed internally for the page... :thinking: 

We could also add a few props to our GridStyle, like `dragSpacerRow: { ...css stuff... }` to hopefully both have our out-of-the-box styles support it already, but also pages could tweak it if/as needed.

src/components/Table/GridTable.stories.tsx
stephenh: Seems like the only thing this doesn't do is scroll the page/table when I get to the top and bottom, but otherwise it seems pretty functionally complete? Very nice! Pretty great that it's just native events too, and not bringing in a new library. Kinda assumed we'd just need that, but this looks good!

src/components/Table/components/Row.tsx
```diff
@@ -372,4 +379,6 @@ export type GridDataRow<R extends Kinded> = {
   selectable?: false;
   /** Whether this row should infer its selected state based on its children's selected state */
   inferSelectedState?: false;
+  /** Whether this row is draggable, usually to allow drag & drop reordering of rows */
+  draggable?: true;
```
bdow, line 383: **nit:** Probably worth making this a boolean otherwise TS will sometimes require us to do `draggable: true as const`. If we allow it to be a boolean value, then we don't need the `as const` suffix.

src/components/Table/components/Row.tsx
```diff
@@ -113,7 +120,7 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
   let expandColumnHidden = false;
 
   return (
-    <RowTag css={rowCss} {...others} data-gridrow {...getCount(row.id)}>
+    <RowTag css={rowCss} {...others} data-gridrow {...getCount(row.id)} draggable={row.draggable}>
```
bdow, line 123: **thought/suggestion:** I think it'd be great to make a specific "drag handle" and test out what that looks like as well. 
If we have a row with many other interactive elements inside of it we could have some potentially conflicting interactions (i dunno if that'd be true, but I'd rather avoid it).

bdow (commented):
Nice spike! I agree with Stephen regarding simplifying the API so not every caller needs to implement all the `onDragFoo` properties.

bdow (commented):
Would also be great to test this out on our `as="table"` and `as="virtual"` variants.
When we use `as="virtual"` we are leveraging the React-Virtuoso library, which has an example of how to integrate with React Beautiful DnD: https://virtuoso.dev/react-beautiful-dnd/

src/components/Table/components/Row.tsx
bsholmes: Yeah I'm planning on adding that as well. Specifically on mobile I think it is necessary since dragging is the default behavior for scrolling through the table.

src/components/Table/GridTable.stories.tsx
chr1sjf0x: It was scrolling for me when dragging to top/bottom, thought there seems to be a small window of where you need to hold the row for it to work 🤔 

praise: Looks great @bsholmes

bsholmes: Yeah the scrolling behavior isn't ideal, and is relying on the ```dataTransfer.effectAllowed = "move"``` stuff so isn't tweakable. We may need to implement our own scrolling behavior, at least on mobile.

src/components/Table/GridTable.stories.tsx
```diff
@@ -1849,3 +1854,146 @@ export function Headers() {
...
+  };
+
+  let rowArray: GridDataRow<OrderedRow<Data>>[] = new Array(26).fill(0);
+  rowArray = rowArray.map((elem, idx) => ({
+    kind: "data",
+    id: "" + (idx + 1),
+    order: idx + 1,
+    data: { name: "" + (idx + 1), value: idx + 1 },
+    draggable: true,
+  }));
```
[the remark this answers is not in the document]
bsholmes: Hmm, that is a good call-out. I'm not sure why that would be happening here but I will look into it before I get a final PR ready.

src/components/Table/GridTable.tsx
```diff
@@ -274,6 +290,111 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
+    });
+  }
+
+  function onDragStart(row: GridDataRow<R>, evt: DragEventType) {
+    if (!row.draggable || !droppedCallback) {
+      return;
+    }
+
+    evt.dataTransfer.effectAllowed = "move";
+    evt.dataTransfer.setData("text/plain", JSON.stringify({ row }));
```
stephenh, line 319: Thought: The row might be a little big, although maybe it doesn't matter? I'd be tempted to just do like `setdata("text/plain", row.id)` because the row ids should be unique.

src/components/Table/components/Row.tsx
```diff
@@ -81,6 +94,8 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
   const levelIndent = style.levels && style.levels[level]?.rowIndent;
 
   const rowCss = {
+    // ...Css.add("transition", "padding-top 0.5s ease-in-out").$,
```
stephenh, line 97: Nit: can remove

src/components/Table/utils/RowState.ts
```diff
@@ -59,6 +67,7 @@ export class RowState<R extends Kinded> {
         _row: false,
         _data: observable.ref,
         isCalculatingDirectMatch: false,
+        isDraggedOver: observable.ref, // allows the table to re-render only this row when the dragged over state changes
```
stephenh, line 70: Question: Given that `isDraggedOver` is a primitive, I _think_ this is unnecessary? Usually the `observable.ref` short-circuits mobx from doing a deep compare, which is how primitives basically work by default.

src/components/Table/GridTable.tsx
```diff
@@ -274,6 +290,111 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
   const expandedColumnIds: string[] = useComputed(() => tableState.expandedColumnIds, [tableState]);
   const columnSizes = useSetupColumnSizes(style, columns, resizeTarget ?? resizeRef, expandedColumnIds);
 
+  function isCursorBelowMidpoint(target: HTMLElement, clientY: number) {
```
stephenh, line 293: Minor: Maybe move to the utils file given afaict the function doesn't need to capture any state.

src/components/Table/GridTable.tsx
```diff
@@ -274,6 +290,111 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
+  function recursiveSetDraggedOver(rows: GridDataRow<R>[], draggedOver: DraggedOver) {
+    rows.forEach((r) => {
+      tableState.maybeSetRowDraggedOver(r.id, draggedOver);
+      if (r.children) {
+        recursiveSetDraggedOver(r.children, draggedOver);
+      }
+    });
+  }
+
+  function onDragStart(row: GridDataRow<R>, evt: DragEventType) {
```
stephenh, line 313: Suggestion: I think `useCallback`-ing all of these functions will be necessary to keep the all rows from re-rendering on each GridTable render. As-is the `Row` props have different values/new functions on each GridTable render.

...it might be "too cute" but fwiw that `RowStates` / `TableState` classes are effectively already cached/1:1 with each GridTable, so if you did something like `TableState.onDragStart` (all of these become methods on `TableState` or `RowStates`), and then inside of of `Row` you could do `onDragStart => tableState.onDragStart(row, evt)`.

I guess you'd need the `setDraggedRow` setter, but you could do something like a `TableState.draggedRow` field, that just gets set... (i.e. skip the `useState` + `useRef` pair, and just use a field on the mobx `TableState` to achieve basically the same thing...

src/components/Table/components/Row.tsx
```diff
@@ -112,8 +129,39 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
...
+      {...getCount(row.id)}
+      // these events are necessary to get the dragged-over row for the drop event
+      // and spacer styling
+      onDrop={(evt) => onDrop?.(row, evt)}
+      onDragEnter={(evt) => onDragEnter?.(row, evt)}
+      onDragOver={(evt) => onDragOver?.(row, evt)}
+      ref={ref}
+    >
+      {row.draggable && (
+        <div
```
stephenh, line 149: Thought: I kinda think we need this div to go inside one of the columns, otherwise we're throwing off the column alignment between the header row (2 divs/columns) & the table rows (3 divs/columns)...

Maybe fine to ship for now/for our usage?

We could potentially give users a `<DragHandle />` component that we just tell them "put this into your 1st column" ... or a `dragColumn` helper like in our `columns.tsx` in beam, there is a `dateColumn`, `actionColumn`, etc... That would be a place to put this code that doesn't throw off the column alignment...

Oh, yeah, the `selectColumn` and `collapseColumn` in particular handle being dropped into any table/without knowing the table's kinds...

stephenh (commented):
The code looks really neat/succinct; I think only actually issue is useCallback-ing the functions to get GridTable perf okay (keep all the rows memoized), by either just-use-useCallback or by moving the function declarations someplace already-stable like RowStates (my preference but likely because I like mobx :-).

src/components/Table/GridTable.stories.tsx
```diff
@@ -1849,3 +1853,144 @@ export function Headers() {
...
+          // insert it at the dropped row index
+          setRows([...insertAtIndex(tempRows, reorderRow, droppedRowIndex + indexOffset)]);
+        }
+      }}
+      rows={[...rows]}
+    />
+  );
+}
+
+export const DraggableWithInputColumns = newStory(
```
chr1sjf0x, line 1924: curious @bsholmes - For these last 2 columns, the dragged item does not remain in the position it is dropped. Is that expected? Am I doing something wrong like dropping things where they are not allowed to be?

chr1sjf0x (commented):
Looks pretty good to me overall. I'll defer to @stephenh on the final 👍 given his suggestions.

src/components/Table/GridTable.stories.tsx
bsholmes: I neglecting to implement the drop event handler in those stories, should be there now

src/components/Table/GridTable.tsx
bsholmes: I added a test to GridTable.test.js to look at row rerenders for draggable rows and I believe they already only rerender on data change, just like the non-draggable rows, with the exception that they need to rerender when I'm setting css for the spacer and when they get reordered.

src/components/Table/components/Row.tsx
bsholmes: Good call, I added a drag handle column as you suggested. The only thing that makes this a little odd is that setting the draggable flag and providing a drop callback won't really be usable unless that column is also provided.

stephenh (approved):

src/components/Table/components/Row.tsx
stephenh: > ...won't really be usable unless that column is also provided.

True...if you wanted, you could do something like "if (any row has draggable || drop callback set) && (!columns has a drag column) throw error("please add a drag column").

That sort of polished UX/DX isn't really common/expected, but pretty neat/appreciated when we do have it. :-)

PR #989, opened 2023-12-20
feat: Add SubmitButton.

stephenh:


src/forms/SubmitButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+};
+
+/** Provides a Button that will auto-disable if `formState` is invalid. */
+export function SubmitButton<T>(props: SubmitButtonProps<T>) {
+  const { form, disabled, onClick, label = "Submit", ...others } = props;
+  const invalidAndTouched = useComputed(() => !form.valid && form.touched, [form]);
+  return (
+    <Button
+      label={label}
+      disabled={disabled || invalidAndTouched}
```
bdow, line 16: **suggestion:** I believe our existing pattern for "when to enable the button" was when the form is `dirty`. Then allowing the user to hit the submit button, which would either make the API call, or highlight invalid fields. I _think_ that still sounds like the right approach, wdyt?

bdow (approved):

src/forms/SubmitButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export type SubmitButtonProps<T> = Omit<ButtonProps, "label"> & {
+  label?: ButtonProps["label"];
+  form: ObjectState<T>;
+};
+
+/** Provides a Button that will auto-disable if `formState` is invalid. */
+export function SubmitButton<T>(props: SubmitButtonProps<T>) {
+  const { form, disabled, onClick, label = "Submit", ...others } = props;
+  const invalidAndTouched = useComputed(() => !form.valid && form.touched, [form]);
```
bdow, line 12: **question:** Any benefit to using the `useComputed` vs wrapping the Button in an `<Observer>`?

stephenh: I think the `useComputed`s just end up being cleaner, as they don't have that ~4-lines of Observer+lambda in the jsx.

src/forms/SubmitButton.tsx
stephenh: Yep, good call; I was staring right at our old `FormActions` doing `!isDirty` and just spaced out what it was doing. :-| :-D Updated!

src/forms/SubmitButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Button
+      label={label}
+      disabled={disabled || invalidAndTouched}
+      onClick={(e) => {
+        // canSave will touch any not-yet-keyed-in fields to show errors
+        if (form.canSave()) {
+          if (typeof onClick === "string") {
+            throw new Error("onClick cannot be a string");
```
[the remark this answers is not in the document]
stephenh: Yep, great catch, fixed

bdow (approved):

PR #991, opened 2024-01-12
feat: Drag & Drop fast follow for card style rows

bsholmes:
Card style rows don't work with padding as spacing, and we can't use margin and still get drop events. So we need a containing element for padding and drag & drop events.

### DIFF
--- a/src/components/Table/GridTable.stories.tsx
+++ b/src/components/Table/GridTable.stories.tsx
@@ -2020,3 +2020,48 @@ export function DraggableNestedRows() {
     />
   );
 }
+
+export function DraggableCardRows() {
+  const dragColumn = dragHandleColumn<Row>({});
+  const nameColumn: GridColumn<Row> = {
+    header: "Name",
+    data: ({ name }) => ({ content: <div>{name}</div>, sortValue: name }),
+  };
+
+  const actionColumn: GridColumn<Row> = {
+    header: "Action",
+    data: () => <div>Actions</div>,
+    clientSideSort: false,
+  };
+
+  let rowArray: GridDataRow<Row>[] = new Array(26).fill(0);
+  rowArray = rowArray.map((elem, idx) => ({
+    kind: "data",
+    id: "" + (idx + 1),
+    order: idx + 1,
+    data: { name: "" + (idx + 1), value: idx + 1 },
+    draggable: true,
+  }));
+
+  const [rows, setRows] = useState<GridDataRow<Row>[]>([simpleHeader, ...rowArray]);
+
+  // also works with as="table" and as="virtual"
+  return (
+    <GridTable
+      columns={[dragColumn, nameColumn, actionColumn]}
+      onRowDrop={(draggedRow, droppedRow, indexOffset) => {
+        const tempRows = [...rows];
+        // remove dragged row
+        const draggedRowIndex = tempRows.findIndex((r) => r.id === draggedRow.id);
+        const reorderRow = tempRows.splice(draggedRowIndex, 1)[0];
+
+        const droppedRowIndex = tempRows.findIndex((r) => r.id === droppedRow.id);
+
+        // insert it at the dropped row index
+        setRows([...insertAtIndex(tempRows, reorderRow, droppedRowIndex + indexOffset)]);
+      }}
+      rows={[...rows]}
+      style={cardStyle}
+    />
+  );
+}
--- a/src/components/Table/components/Row.tsx
+++ b/src/components/Table/components/Row.tsx
@@ -146,12 +146,7 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
       onDragOver={(evt) => onDragOver?.(row, evt)}
       ref={ref}
     >
-      <RowTag
-        css={rowCss}
-        {...others}
-        data-gridrow
-        {...getCount(row.id)}
-      >
+      <RowTag css={rowCss} {...others} data-gridrow {...getCount(row.id)}>
         {isKeptGroupRow ? (
           <KeptGroupRow as={as} style={style} columnSizes={columnSizes} row={row} colSpan={columns.length} />
         ) : (
@@ -171,7 +166,8 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
             const numExpandedColumns = isExpanded
               ? tableState.numberOfExpandedChildren(maybeExpandedColumn.id)
                 ? // Subtract 1 if the column is hidden on expand, since we're not rendering it.
-                  tableState.numberOfExpandedChildren(maybeExpandedColumn.id) - (maybeExpandedColumn.hideOnExpand ? 1 : 0)
+                  tableState.numberOfExpandedChildren(maybeExpandedColumn.id) -
+                  (maybeExpandedColumn.hideOnExpand ? 1 : 0)
                 : 0
               : 0;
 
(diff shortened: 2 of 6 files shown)


src/components/Table/components/Row.tsx
```diff
@@ -132,245 +135,234 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
...
 
   return (
-    <RowTag
-      css={rowCss}
-      {...others}
-      data-gridrow
-      {...getCount(row.id)}
+    // we need to use padding for the drag & drop spacing or else drop events don't fire
+    // and we need a containing element for padding to work correctly with the card style row
+    <div
```
bsholmes, line 140: Wrapping rows in this container seems to be the only sane way to make card style rows work with drag & drop. It doesn't seem to have broken anything wrt rows/tables, at least as far as Beam is concerned. But this is somewhat risky in terms of breaking implementations that rely on Beam. Let me know if this is likely to mess something up.

stephenh: Huh, isn't `RowTag` basically already providing a wrapping div around the cells? It seems odd that we'd need two wrapping divs...

Ah, I get it, both dnd & cards were using padding, but when dragging we want the dnd padding to be "on top" / separate from the card padding?

I'm okay-ish with this as a short-term fix, but yeah let's try and only ad the extra div is dnd is actively being used, just to be extra sure we don't mess existing callers.

Afaiu does dnd use padding to provide the "spacer" effect? Kinda wondering if we could use a psuedo element for that (like the `::before` thing), which I personally haven't done a ton of. Not sure how we'd listen to drop events on that, but maybe we could hook up the drop handlers to the root table div, and that would pick up the `::before` element, maybe encode what got "dropped" using a `data-something` attribute on the psuedo element.

cc @bdow for thoughts.

bsholmes: The padding for the spacer goes inside the border of the card, so instead of looking like it's going above or below the row, it looks like it's going inside the card. I will look into the before/after pseudo-elements but I'm pretty sure they won't handle events properly. I could make the container conditional to draggable rows, but theoretically it should be able to work either way.

stephenh: > The padding for the spacer goes inside the border of the card, so instead of
> looking like it's going above or below the row, it looks like it's going inside
> the card.

Yeah, makes sense...

> theoretically it should be able to work either way.

Yeah. Ngl I'm surprised it still works, but I get it, because we're like hand-calcing so much of the column widths/etc; even if it just comes down to DOM aesthetics of avoiding the extra div :shrug: dunno, seems simple enough to avoid/make conditional.

bsholmes: conditional it is

src/components/Table/components/Row.tsx
```diff
@@ -93,8 +92,13 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
   const rowStyleCellCss = maybeApplyFunction(row as any, rowStyle?.cellCss);
   const levelIndent = style.levels && style.levels[level]?.rowIndent;
 
-  const rowCss = {
+  const containerCss = {
     ...Css.add("transition", "padding 0.5s ease-in-out").$,
+    ...(rs.isDraggedOver === DraggedOver.Above && Css.add("paddingTop", "35px").$),
+    ...(rs.isDraggedOver === DraggedOver.Below && Css.add("paddingBottom", "35px").$),
```
bdow, line 98: **nit:** Truss has a lot of great shorthand ways or writing styles that hopefully make this sort of thing less verbose and easier. You could easily do these padding styles via: `Css.ptPx(35).$`.

src/components/Table/components/Row.tsx
```diff
@@ -93,8 +92,13 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
   const rowStyleCellCss = maybeApplyFunction(row as any, rowStyle?.cellCss);
   const levelIndent = style.levels && style.levels[level]?.rowIndent;
 
-  const rowCss = {
+  const containerCss = {
     ...Css.add("transition", "padding 0.5s ease-in-out").$,
```
bdow, line 96: **nit/question:** This timing feels a bit slow. Have you been able to work with Design to define the proper behaviors/timings for the drag and drop?

src/components/Table/GridTable.stories.tsx
```diff
@@ -2020,3 +2020,48 @@ export function DraggableNestedRows() {
...
+    draggable: true,
+  }));
+
+  const [rows, setRows] = useState<GridDataRow<Row>[]>([simpleHeader, ...rowArray]);
+
+  // also works with as="table" and as="virtual"
+  return (
+    <GridTable
+      columns={[dragColumn, nameColumn, actionColumn]}
+      onRowDrop={(draggedRow, droppedRow, indexOffset) => {
```
bdow, line 2052: **question:** I'm a little late to the game here, but wouldn't `draggedRow` and `droppedRow` be the same row?
It is probably too late to go back and update the implementation at this point, but just spit-balling here; Could we have gotten away with a callback like:
```
onRowDrop: (row: GridDataRow, newIndex: number) => void
```
This would give the user the information that a certain row was moved to `newIndex`.

Or many even:
```
onRowDrop: (newRows: GridDataRow[]) => void
```
If the developer only ever needs to just update their `row` property, this might make their implementations simpler.

src/components/Table/components/Row.tsx
bsholmes: @stephenh should only affect draggable rows now

bdow (approved):
I noticed some potential re-rendering perf issues that might come up. I've noted it in the loom here:
https://www.loom.com/share/4834d010e0c34bbcbb7f011deb27c3e6

Another question I had was if we ever looked at inserting a simple dom element into the position that is being dragged over vs adjusting padding on an existing element? I can see how the virtualized tables may not like us adding/removing dom elements on the fly, but would have been cool to check it out. We did something similar to that in the DnDGrid. All the dragging and dropping logic happens outside of React state there, which allows for zero rerenders during the drag and drop process.
React-Virtuoso (which we use in GridTable) provides a really good example of how they worked with React Beautiful DnD: https://virtuoso.dev/react-beautiful-dnd/. It looks like a super simple implementation.

src/components/Table/components/Row.tsx
bdow: I was also thinking the pseudo elements would be the solution here instead of a wrapping element. I'm surprised they don't.

src/components/Table/GridTable.stories.tsx
bsholmes: draggedRow would be the row that was dragged initially, droppedRow would be the row we just dropped the draggedRow onto. We could potentially simplify this callback if we only want it to support reordering, but potentially it could support other kinds of drag & drop behavior and in those cases we want the handler to have info about both rows.

bdow: Oh, yes. That is a bit confusing of a name then. I would've assume the row being "dropped" is the row I was dragging. Maybe `targetRow` is a better name?
Interesting point about other drag and drop behaviors. Are we being asked to support dropping into another row in order to make the `draggedRow` a child of the `target/droppedRow`?
I'd be tempted to support multiple callbacks for that... like `onReorder` for the current use case, and .... I dunno what else? Would we want to allow rows to drop into each other to nest them or something? I'd be tempted to solution only the problem in front of us instead of envision future scenarios. 
I'm fine with leaving it as is for now.

src/components/Table/components/Row.tsx
bsholmes: Nope, sure haven't, but I made it a bit snappier

src/components/Table/components/Row.tsx
stephenh: @bsholmes nice, thank you!

stephenh (approved):

PR #992, opened 2024-01-12
fix: multiselect read-only

0ces:
This PR fixes `TreeSelectField` and `MultiSelectField` read-only renders.

### DIFF
--- a/src/inputs/internal/ComboBoxBase.tsx
+++ b/src/inputs/internal/ComboBoxBase.tsx
@@ -404,8 +404,8 @@ function getInputValue<O>(
   multiselect: boolean,
   nothingSelectedText: string,
 ) {
-  return selectedOptions.length === 1
-    ? getOptionLabel(selectedOptions[0])
+  return selectedOptions.length > 0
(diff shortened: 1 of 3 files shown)


src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -404,8 +404,8 @@ function getInputValue<O>(
   multiselect: boolean,
   nothingSelectedText: string,
 ) {
-  return selectedOptions.length === 1
-    ? getOptionLabel(selectedOptions[0])
+  return selectedOptions.length > 0
+    ? selectedOptions.map(getOptionLabel).join(", ")
```
blambillotte, line 408: Question: I think this affects more than just the read-only field view? I know only showing the count is a pain point we have, but I'd be hesitant to push a change like this through since the read-only view is the specific problem we're having 🤔

0ces: Yeah, this actually affects more than just the read-only mode, I'll change it so it only changes on read-only

blambillotte (approved):

PR #996, opened 2024-01-24
fix: TreeSelectField correct number of options selected

0ces:
imagen

### DIFF
--- a/src/inputs/TreeSelectField/utils.ts
+++ b/src/inputs/TreeSelectField/utils.ts
@@ -14,6 +14,7 @@ export type TreeFieldState<O> = {
   filteredOptions: LeveledOption<O>[];
   selectedKeys: Key[];
   selectedOptions: NestedOption<O>[];
+  selectedOptionsLabels: string[];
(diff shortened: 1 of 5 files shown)


src/inputs/TreeSelectField/utils.ts
```diff
@@ -14,6 +14,7 @@ export type TreeFieldState<O> = {
   filteredOptions: LeveledOption<O>[];
   selectedKeys: Key[];
   selectedOptions: NestedOption<O>[];
+  selectedOptionsLabels: string[];
```
stephenh, line 17: Suggestion: Maybe an add in-source comment that with trees the selected labels could be "p1, p2c3" even though the selected options are "p1, p1c1, p1c2, p2c3", something like that? If I'm following why this is necessary, kinda guessing...

stephenh (approved):
No idea what our test coverage of `TreeSelectField` looks like, but if we have a test to copy/paste, a test would be nice. If no existing tests, that's fine to skip for now.

PR #999, opened 2024-02-04
feat: Add more overloads for SelectField options.

stephenh:
`getOptionValue` is now optional for `options` with ids or codes.

`getOptionLabel` is now optional for `options` with name, displayName, or labels.

### DIFF
--- a/src/inputs/SelectField.test.tsx
+++ b/src/inputs/SelectField.test.tsx
@@ -4,6 +4,7 @@ import { useState } from "react";
 import { SelectField, SelectFieldProps, Value } from "src/inputs";
 import { HasIdAndName, Optional } from "src/types";
 import { blur, click, focus, getOptions, render, select, wait } from "src/utils/rtl";
+import { AuthorHeight } from "src/forms/formStateDomain";
 
 describe("SelectFieldTest", () => {
   it("can set a value", async () => {
@@ -17,7 +18,6 @@ describe("SelectFieldTest", () => {
         options={options}
         getOptionLabel={(o) => o.name}
         getOptionValue={(o) => o.id}
-        data-testid="age"
         onBlur={onBlur}
         onSelect={onSelect}
       />,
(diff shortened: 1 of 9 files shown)


src/inputs/SelectField.test.tsx
```diff
@@ -123,6 +119,38 @@ describe("SelectFieldTest", () => {
     expect(r.age).toHaveValue("Unassigned");
   });
 
+  it("can use option codes", async () => {
+    const options = [
+      { code: AuthorHeight.SHORT, name: "Shortish" },
+      { code: AuthorHeight.TALL, name: "Tallish" },
```
TylerR909, line 125: lol

src/utils/options.ts
```diff
@@ -1,2 +1,20 @@
...
+
+// This `any` is currently on purpose to ignore type errors in ChipSelectField
+export function defaultOptionValue<O>(opt: O): any {
+  if (typeof opt !== "object" || !opt) fail(`Option ${opt} has no id or code`);
+  // Use `in` because returning undefined is fine
+  return "id" in opt ? opt.id : "code" in opt ? opt.code : fail(`Option ${JSON.stringify(opt)} has no id or code`);
+}
+
+export function defaultOptionLabel<O>(opt: O): any {
+  if (typeof opt !== "object" || !opt) fail(`Option ${opt} has no id or code`);
```
TylerR909, line 11: hmm... how destructive would it be to allow strings. Like, I'd _like_ to be able to do

```
options={["One", "Two", "Three"]}
// or
options={Object.values(SomeEnum)}
```

where ID/Val would be the same. Seems unnecessary to map an Enum over to an Object just for the sake of this. But, I'm sure a lot of infra expects an Object so makes sense it would be rough to support.

TylerR909 (approved):

src/utils/options.ts
stephenh: Yeah, that's a fair ask. Will think about it...

PR #1009, opened 2024-03-05
fix: Clear search terms and reset dropdown options when dropdowns re-open

khalidwilliams:
# SC-47696

https://www.loom.com/share/27ec8874b4f549b2813dd8563ea7b68f?sid=d55983ea-d3f2-4679-a658-03a3fc9cbffd


blambillotte (approved):
Change makes sense to me, but CC @bdow for your thoughts

bdow (approved):

PR #1010, opened 2024-03-18
feat: add onSearch to ComboBox

0ces:
Added a `onSeach` callback to the `ComboBoxBase` component to be able to use that search to query options dynamically

### DIFF
--- a/src/inputs/internal/ComboBoxBase.tsx
+++ b/src/inputs/internal/ComboBoxBase.tsx
@@ -61,6 +61,8 @@ export interface ComboBoxBaseProps<O, V extends Value> extends BeamFocusableProp
   hideErrorMessage?: boolean;
   /* Allows input to wrap to multiple lines */
   multiline?: boolean;
+  /* Callback for user searches */
+  onSearch?: (search: string) => void;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -285,7 +288,12 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
         inputValue: getInputValue(selectedOptions, getOptionLabel, multiselect, nothingSelectedText, isReadOnly),
       }));
     }
-  }, [state.isOpen, selectedOptions, getOptionLabel, multiselect, nothingSelectedText, isReadOnly]);
+  }, [state.isOpen, selectedOptions, getOptionLabel, multiselect, nothingSelectedText, isReadOnly, onSearch]);
+
+  // Call on search callback when the user types in the input field
+  useEffect(() => {
+    onSearch?.(fieldState.searchValue ?? "");
+  }, [onSearch, fieldState.searchValue]);
```
stephenh, line 296: Question: We should debounce this on the beam side of the API? cc @bdow wdyt?

@0ces just curious, what use case is this for?

0ces: I did debounce it but on the BP side, I'll change that

PR #1019, opened 2024-05-10
fix: Better support for min-width on Table columns

bdow:


src/components/Table/utils/columns.tsx
```diff
@@ -135,32 +135,71 @@ export function calcColumnSizes<R extends Kinded>(
...
+  let adjustedTotalFr = totalFr;
+  if (tableWidth) {
+    columns.forEach(({ w, mw }) => {
+      const frUnit = getFrUnit(w);
+      if (mw === undefined || frUnit === undefined) return;
+
+      const mwPx = Number(mw.replace("px", ""));
+      const calcedWidth =
+        (tableWidth - (claimedPercentages / 100) * tableWidth - adjustedClaimedPixels) * (frUnit / adjustedTotalFr);
+      // If the calculated width is less than the minWidth, then adjust the claimedPixels and totalFr accordingly
```
stephenh, line 161: It sounds like we're basically saying "you're not FR based anymore, you're actually pixel based"? If so, maybe work that into the in-source comment?

src/components/Table/components/Row.tsx
```diff
@@ -176,13 +176,6 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
           const applyFirstContentColumnStyles = !isHeader && isFirstContentColumn;
           foundFirstContentColumn ||= applyFirstContentColumnStyles;
 
-          if (column.mw) {
```
stephenh, line 179: Makes a ton of sense to have this handled in the same place as the fr / size handling; great change!

stephenh (approved):

PR #1020, opened 2024-05-10
fix: Fallback message for virtual table was exceeding table header width

bdow:
For virtualized tables we subtract 20px from the total width to account for a vertical scrollbar, which helps prevent horizontal scrollbars. The fallback message was not respecting that adjusted width.

| Before | After |
| --- | --- | 
| Screenshot 2024-05-10 at 4 16 32 PM | Screenshot 2024-05-10 at 4 16 47 PM |

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -2,7 +2,7 @@ import memoizeOne from "memoize-one";
 import { runInAction } from "mobx";
 import React, { MutableRefObject, ReactElement, useEffect, useMemo, useRef, useState } from "react";
 import { Components, Virtuoso, VirtuosoHandle } from "react-virtuoso";
-import { Loader } from "src/components";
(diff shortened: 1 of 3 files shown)


src/components/Table/GridTable.tsx
```diff
@@ -486,7 +486,7 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
       <PresentationProvider fieldProps={fieldProps} wrap={style?.presentationSettings?.wrap}>
         {/* If virtualized take some pixels off the width to accommodate when virtuoso's scrollbar is introduced. */}
         {/* Otherwise a horizontal scrollbar will _always_ appear once the vertical scrollbar is needed */}
```
stephenh, line 488: Hm, maybe move this comment into the helper method, to explain the `- 20px`? Could go either way...

stephenh (approved):

PR #1021, opened 2024-05-10
feat: Add new icons

khalidwilliams:


src/forms/BoundCheckboxField.test.tsx
```diff
@@ -63,6 +63,33 @@ describe("BoundCheckboxField", () => {
     click(r.isAvailable);
     // Then the callback should be triggered with the current value
     expect(autoSave).toBeCalledWith(false);
+    expect(autoSave).toBeCalledTimes(1);
```
blambillotte, line 66: Was this a merge conflict or something? Changes seem unrelated?

src/components/Icon.tsx
```diff
@@ -813,6 +813,21 @@ export const Icons = {
       fillRule="evenodd"
     />
   ),
+  bxTime: (
```
blambillotte, line 816: question: where does this `bx` prefix come from? Don't see that anywhere else 🤔

src/components/Icon.tsx
```diff
@@ -813,6 +813,21 @@ export const Icons = {
...
+    <>
+      <path
+        d="M12 2C6.486 2 2 6.486 2 12C2 17.514 6.486 22 12 22C17.514 22 22 17.514 22 12C22 6.486 17.514 2 12 2ZM12 20C7.589 20 4 16.411 4 12C4 7.589 7.589 4 12 4C16.411 4 20 7.589 20 12C20 16.411 16.411 20 12 20Z"
+        fill="#484848"
```
blambillotte, line 820: I believe we usually remove any color attributes and let the icons get colored by the parent CSS
```suggestion
        
```

src/components/Icon.tsx
```diff
@@ -813,6 +813,21 @@ export const Icons = {
...
+      />
+      <path d="M13 7H11V13H17V11H13V7Z" fill="#484848" />
+    </>
+  ),
+  bxLockOpenAlt: (
```
blambillotte, line 825: same naming nit and `fill` nit here - I do see we have an `outline` convention which is maybe what the `alt` is pointing to from another icon lib?

src/components/Icon.tsx
khalidwilliams: I pulled it from the name in the figma file, wasn't sure if it was a new convention there. I can remove it, it's been a while since I added Beam icons

src/forms/BoundCheckboxField.test.tsx
khalidwilliams: Ah yep sorry, this slipped in from some tests I was doing for a different ticket

blambillotte (approved):
lgtm - one suggestion would be to add them to the storybook (if it doesn't happen automatically?)

blambillotte:
> lgtm - one suggestion would be to add them to the storybook (if it doesn't happen automatically?)

nvm I see the updates

PR #1022, opened 2024-05-13
fix: Form Labels to not extend full width of field

bdow:
Kind of an annoying issue that I've see is you try to click away from a SelectField to close the menu, but instead the menu flickers and opens back up. This was due to clicking above the field, which happened to be still clicking on the label element which was taking up the full width of its container. When clicking a label element, then the focus is set to the associated field. By adding 'align-self: flex-start', then the label element no longer extends the full width of the container.

See Loom for example:
https://www.loom.com/share/7b2dd9d22c1a457f991aca884a45731b

### DIFF
--- a/src/components/Label.tsx
+++ b/src/components/Label.tsx
@@ -27,7 +27,13 @@ function LabelComponent<X extends Only<Xss<LabelXss>, X>>(props: LabelProps<X>) 
     <label
       {...labelProps}
       {...others}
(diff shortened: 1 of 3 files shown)


chr1sjf0x (approved):

stephenh (approved):
Wow, great attention to detail!

I'm surprised setting the align-self makes it not stretch :shrug:

PR #1023, opened 2024-05-15
fix: Allow click events to pass through tooltip wrapped elements

bdow:
Thanks @bsholmes for the solution! 🎉

### DIFF
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -36,7 +36,8 @@ export function Tooltip(props: TooltipProps) {
         {...(!state.isOpen && typeof title === "string" ? { title } : {})}
         {...tid}
         // Add display contents to prevent the tooltip wrapping element from short-circuiting inherited styles (i.e. flex item positioning)
-        css={Css.display("contents").$}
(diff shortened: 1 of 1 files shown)


src/components/Tooltip.tsx
```diff
@@ -36,7 +36,8 @@ export function Tooltip(props: TooltipProps) {
         {...(!state.isOpen && typeof title === "string" ? { title } : {})}
         {...tid}
         // Add display contents to prevent the tooltip wrapping element from short-circuiting inherited styles (i.e. flex item positioning)
-        css={Css.display("contents").$}
+        // Once the element is `:active`, allow pointer events (i.e. click events) to pass through to the children.
+        css={Css.display("contents").addIn(":active", Css.add("pointerEvents", "none").$).$}
```
bdow, line 40: I tried to write a test for this, but couldn't figure out how to get a click event to bubble up in tests. After 5-10minutes of searching and tryings out, I figured a manual test was good enough for now given the state of Beam soon being sunsetted.

chr1sjf0x (approved):
TY @bdow

PR #1025, opened 2024-05-17
feat: hackday multiselect shows selected options in field

apattersonATX-HB:
All multiselect fields will show selected options within the field. Tab functionality is preserved. It's not a full rework like what's laid out in [this old Figma](https://www.figma.com/design/GEXYjIEDKFRjoFbnrt515r/New-Multiselect-Component-%E2%9C%A8?node-id=5-14&t=QHpFO2gb9HhNE2yB-0) but should make some people happy.

https://github.com/homebound-team/beam/assets/108819408/a4ac3709-630c-4834-863d-e6b3e13c9b1f

- This also adds a list tooltip to the selected count.
  - Screenshot 2024-05-17 at 3 01 33 PM

Couple spots I could use some design help
  - 
- I set the placeholder container width to `100%` to fill the space. It does introduce some weirdness when fields extend far. How should we handle this?
  - Screenshot 2024-05-17 at 2 49 02 PM
- When packed into a tight table the scrollbars obstruct the view and aren't very viually appealing. Granted with the tooltip on selected count now you can get more info at a glance than we had before but it still bugs me.
  - Screenshot 2024-05-17 at 2 57 39 PM

### DIFF
--- a/src/inputs/TextFieldBase.test.tsx
+++ b/src/inputs/TextFieldBase.test.tsx
@@ -1,5 +1,6 @@
 import { TextFieldBase } from "src/inputs/TextFieldBase";
 import { render } from "src/utils/rtl";
+import { act } from "@testing-library/react";
 
 describe(TextFieldBase, () => {
   it("shows error and helper text", async () => {
@@ -24,5 +25,31 @@ describe(TextFieldBase, () => {
     expect(r.test).toBeDisabled();
     expect(r.query.test_errorMsg).not.toBeInTheDocument();
     expect(r.query.test_helperText).not.toBeInTheDocument();
+  });
+
+  it("handles unfocusedPlaceholder correctly", async () => {
+    // When TextFieldBase is first rendered
+    const r = await render(
+      <TextFieldBase
+        inputProps={{}}
+        unfocusedPlaceholder={"Unfocused placeholder text"}
+        label="Test"
+        errorMsg="Error"
+        helperText="Helper"
+      />,
+    );
+
+    // The unfocused placeholder container is rendered
+    expect(r.test_unfocusedPlaceholderContainer).toBeInTheDocument();
+    // And is visible
+    expect(r.test_unfocusedPlaceholderContainer).not.toHaveStyleRule("position", "absolute");
+
+    // And when we focus the field
+    act(() => {
+      r.test.focus();
+    });
+
+    // Then the unfocused placeholder container is visually hidden
+    expect(r.test_unfocusedPlaceholderContainer).toHaveStyleRule("position", "absolute");
   });
 });
--- a/src/inputs/TextFieldBase.test.tsx
+++ b/src/inputs/TextFieldBase.test.tsx
@@ -1,6 +1,5 @@
 import { TextFieldBase } from "src/inputs/TextFieldBase";
-import { render } from "src/utils/rtl";
-import { act } from "@testing-library/react";
+import { render, focus } from "src/utils/rtl";
 
 describe(TextFieldBase, () => {
   it("shows error and helper text", async () => {
@@ -45,9 +44,7 @@ describe(TextFieldBase, () => {
     expect(r.test_unfocusedPlaceholderContainer).not.toHaveStyleRule("position", "absolute");
 
     // And when we focus the field
-    act(() => {
-      r.test.focus();
-    });
+    focus(r.test);
 
     // Then the unfocused placeholder container is visually hidden
     expect(r.test_unfocusedPlaceholderContainer).toHaveStyleRule("position", "absolute");
--- a/src/inputs/TextFieldBase.tsx
+++ b/src/inputs/TextFieldBase.tsx
@@ -58,6 +58,8 @@ export interface TextFieldBaseProps<X>
   hideErrorMessage?: boolean;
   // If set, the helper text will always be shown (usually we hide the helper text if read only)
   alwaysShowHelperText?: boolean;
+  // Replaces empty input field and placeholder with node
+  // IE: Multiselect renders list of selected items in the input field
   unfocusedPlaceholder?: ReactNode;
 }
 
@@ -256,7 +258,7 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
                   {...tid.unfocusedPlaceholderContainer}
                   css={{
                     ...Css.df.fdc.w100.pyPx(2).maxh100.overflowAuto.$,
-                    ...(isFocused && visuallyHidden),
+                    ...(isFocused && Css.visuallyHidden.$),
                   }}
                 >
                   {unfocusedPlaceholder}
@@ -275,7 +277,7 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
                   ...fieldStyles.input,
                   ...(inputProps.disabled ? fieldStyles.disabled : {}),
                   ...(showHover ? fieldStyles.hover : {}),
-                  ...(unfocusedPlaceholder && !isFocused && visuallyHidden),
+                  ...(unfocusedPlaceholder && !isFocused && Css.visuallyHidden.$),
                   ...xss,
                 }}
                 {...tid}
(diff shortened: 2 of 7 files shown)


aTrujilloHomeBound (approved):

aTrujilloHomeBound:
Great Job!

src/inputs/TextFieldBase.test.tsx
```diff
@@ -25,4 +26,30 @@ describe(TextFieldBase, () => {
...
+
+    // The unfocused placeholder container is rendered
+    expect(r.test_unfocusedPlaceholderContainer).toBeInTheDocument();
+    // And is visible
+    expect(r.test_unfocusedPlaceholderContainer).not.toHaveStyleRule("position", "absolute");
+
+    // And when we focus the field
+    act(() => {
+      r.test.focus();
+    });
```
bdow, line 50: **nit/question:** Would the `focus()` helper from our `rtl-utils` work for this test?

src/inputs/TextFieldBase.tsx
```diff
@@ -176,6 +178,12 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
     e.target.select();
   }, onFocus);
 
+  // Simulate clicking `ElementType` when using an unfocused placeholder
+  function handleUnfocusedPlaceholderClick(e: React.MouseEvent<HTMLDivElement>) {
+    e.stopPropagation();
+    fieldRef.current?.click();
+  }
```
bdow, line 185: **question:** Did you try simulating the `focus` event instead of `click`? I don't know if there's a big difference in the approach, but I kind of felt like simulating focus would've followed at least my mental model a little more. I.e. "User clicks X and we want to move focus to Y". Rather than "User clicks X, but let's pretend they clicked Y". Something about simulating a click feels odd to me, but probably fine. In my mind, the "just move focus" approach aligns better with how a label would interact with a form field, - when a user clicks a label, the field is focused.
Curious if there was a reason to go for the `click` over `focus`.

src/inputs/TextFieldBase.tsx
```diff
@@ -237,11 +245,23 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
               }}
               {...hoverProps}
               ref={inputWrapRef as any}
+              onClick={unfocusedPlaceholder ? handleUnfocusedPlaceholderClick : undefined}
```
bdow, line 248: **thought:** Ahh, this is interesting. I think we actually have this "problem" currently. If a user clicks within the `inputWrap`, but not on the `input` then the field doesn't get focused. 🤔 I don't think its worth trying to address that here for all cases, but I see what you're doing.

src/inputs/TextFieldBase.tsx
```diff
@@ -237,11 +245,23 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
...
             >
               {labelStyle === "inline" && label && (
                 <InlineLabel multiline={multiline} labelProps={labelProps} label={label} {...tid.label} />
               )}
               {startAdornment && <span css={Css.df.aic.asc.fs0.br4.pr1.$}>{startAdornment}</span>}
+              {unfocusedPlaceholder && (
+                <div
+                  {...tid.unfocusedPlaceholderContainer}
+                  css={{
+                    ...Css.df.fdc.w100.pyPx(2).maxh100.overflowAuto.$,
```
bdow, line 258: **suggestion:** There is a variable in this component called `multiline`. It gets set to true when we want to allow the multiselect field's value to wrap to multiple lines. We should take that into account here.

src/inputs/TextFieldBase.tsx
```diff
@@ -255,6 +275,7 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
                   ...fieldStyles.input,
                   ...(inputProps.disabled ? fieldStyles.disabled : {}),
                   ...(showHover ? fieldStyles.hover : {}),
+                  ...(unfocusedPlaceholder && !isFocused && visuallyHidden),
```
bdow, line 278: **question:** Will these return `false` ever? I'm kind of surprised we can do `...(false)` 🤔 I'm probably missing something. If it works, it works! 🎉

src/inputs/TextFieldBase.tsx
```diff
@@ -310,3 +331,7 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
     </>
   );
 }
+
+// Css that would be applied if using react-aria <VisuallyHidden />
+const visuallyHidden = Css.add("clip", "inset(50%)").add("clipPath", "").add("border", 0).hPx(1).mPx(-1).wPx(1).nowrap
+  .p0.overflowHidden.absolute.$;
```
bdow, line 337: **suggestion:** This could be a good style to add to our `trussConfig` at the root of this repo. Then we could do something like `Css.visuallyHidden.$`

src/inputs/internal/ComboBoxInput.tsx
```diff
@@ -92,23 +92,28 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
   const multilineProps = allowWrap ? { textAreaMinHeight: 0, multiline: true } : {};
   useGrowingTextField({ disabled: !allowWrap, inputRef, inputWrapRef, value: inputProps.value });
 
+  const chipLabels = isTree ? selectedOptionsLabels || [] : selectedOptions.map((o) => getOptionLabel(o));
+
   return (
     <TextFieldBase
       {...otherProps}
       {...multilineProps}
+      unfocusedPlaceholder={showNumSelection && <Chips compact={otherProps.compact} values={chipLabels} />}
```
bdow, line 101: **question** Is this going to be displaying Chips for single select fields?

bdow (changes requested):
Fantastic!! 🎉 
A couple small suggestions/requests. Otherwise really great.

src/inputs/TextFieldBase.tsx
```diff
@@ -58,6 +58,7 @@ export interface TextFieldBaseProps<X>
   hideErrorMessage?: boolean;
   // If set, the helper text will always be shown (usually we hide the helper text if read only)
   alwaysShowHelperText?: boolean;
+  unfocusedPlaceholder?: ReactNode;
```
stephenh, line 61: Suggestion: I can't immediately guess what this prop does; it does seem like a good name, but maybe add a comment describe what it does with an "i.e. for multi-select this (...the chips...?)"

stephenh (commented):
Just the one question, otherwise I'll defer to @bdow , but this is great!

src/inputs/TextFieldBase.tsx
apattersonATX-HB: yes it does, updated

src/inputs/internal/ComboBoxInput.tsx
apattersonATX-HB: Nope. `showNumSelection` specifes `> 1`

src/inputs/TextFieldBase.tsx
apattersonATX-HB: Good call, updated

src/inputs/TextFieldBase.tsx
apattersonATX-HB: added!

src/inputs/TextFieldBase.tsx
apattersonATX-HB: This also works the same on the `unfocusedPLaceholder` container. We can move it there if that makes more sense.

src/inputs/TextFieldBase.tsx
apattersonATX-HB: Yes, when we used `focus()` it doesn't open the popup on first click. I didnt dive too deep into the component tree to find what makes `buttonProps` special regarding the onClick though 🤔

src/inputs/TextFieldBase.test.tsx
apattersonATX-HB: sho does!

src/inputs/TextFieldBase.tsx
bdow: Let's leave it as is. It hasn't been called out as an issue and I rather not introduce too many additional interaction handlers if we don't really need them.

bdow (approved):

src/inputs/internal/ComboBoxInput.tsx
apattersonATX-HB: circling back to this all select fields will show a chip now

PR #1028, opened 2024-05-21
feat: MaxLines component

bdow:
MaxLines helps truncate long text vertically based on the number of lines the user wants to initially render. It provides a toggle to "Show More" / "Show Less".
The component listens for changes to its size & children to ensure it updates its need for truncation accordingly.


allangaldinosilva (approved):
Looks awesome! Thanks Brandon 👏

PR #1029, opened 2024-05-21
fix: ComboBoxInput.tsx uses chips only if multiselect

apattersonATX-HB:
- Fixes single selects using chips. I missed a check for `isMultiSelect`
- Reverts pagination to original size

### DIFF
--- a/src/components/Pagination.tsx
+++ b/src/components/Pagination.tsx
@@ -50,7 +50,7 @@ export function Pagination(props: PaginationProps) {
       <div css={Css.df.mya.mr2.$} {...tid.pageSizeLabel}>
         Page size:
       </div>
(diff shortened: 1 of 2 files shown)


src/components/Pagination.tsx
```diff
@@ -50,7 +50,7 @@ export function Pagination(props: PaginationProps) {
       <div css={Css.df.mya.mr2.$} {...tid.pageSizeLabel}>
         Page size:
       </div>
-      <div css={Css.wPx(95).$}>
+      <div css={Css.wPx(78).$}>
```
bdow, line 53: **question:** Why did this need to change?

bdow (approved):

PR #1033, opened 2024-05-25
feat(accordion): clickable title

TylerR909:
In DesignPackage, we want to be able to select "All these slots" by clicking on a header, like so:

image

This PR adds an optional `titleOnClick` function that, when provided, slightly alter the behavior of the Accordion:

1. Clicking the title no longer performs expand/collapse
    - We still want whitespace between the icon and title to be clickable, so it has to `fg0` instead of the span's `fg1`
    - Stop propagation so w don't trigger expand's onClick

### DIFF
--- a/src/components/Accordion.tsx
+++ b/src/components/Accordion.tsx
@@ -105,7 +105,7 @@ export function Accordion<X extends Only<AccordionXss, X>>(props: AccordionProps
       }}
     >
       {titleOnClick ? (
-        <div {...focusProps} aria-controls={id} aria-expanded={expanded} css={Css.df.jcsb.$}>
+        <div {...focusProps} aria-controls={id} aria-expanded={expanded} css={Css.df.$}>
           <button {...tid.title} disabled={disabled} css={{ ...touchableStyle, ...Css.fg0.$ }} onClick={titleOnClick}>
             {title}
           </button>
--- a/src/components/Accordion.tsx
+++ b/src/components/Accordion.tsx
@@ -79,11 +79,6 @@ export function Accordion<X extends Only<AccordionXss, X>>(props: AccordionProps
   }, [expanded, setContentHeight]);
   useResizeObserver({ ref: contentRef, onResize });
 
-  const expandOrCollapse = useCallback(() => {
-    setExpanded((prior) => !prior);
-    if (setExpandedIndex) setExpandedIndex(index);
-  }, [index, setExpandedIndex]);
-
   return (
     <div
       {...testIds.container}
@@ -105,7 +100,10 @@ export function Accordion<X extends Only<AccordionXss, X>>(props: AccordionProps
           ...(isFocusVisible && Css.boxShadow(`inset 0 0 0 2px ${Palette.Blue700}`).$),
           ...xss,
         }}
-        onClick={expandOrCollapse}
+        onClick={() => {
+          setExpanded((prior) => !prior);
+          if (setExpandedIndex) setExpandedIndex(index);
+        }}
       >
         {titleOnClick ? (
           <button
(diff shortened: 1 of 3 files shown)


src/components/Accordion.tsx
```diff
@@ -98,11 +101,24 @@ export function Accordion<X extends Only<AccordionXss, X>>(props: AccordionProps
...
         }}
         onClick={() => {
-          setExpanded(!expanded);
+          setExpanded((prior) => !prior);
           if (setExpandedIndex) setExpandedIndex(index);
         }}
       >
-        <span css={Css.fg1.tl.$}>{title}</span>
+        {titleOnClick ? (
+          <button
```
bdow, line 109: Suggestion: We should try to avoid nesting interactive elements. Per HTML standards, nesting of interactive elements is not permitted- it's possible, just not advisable. 
> Content model:
> Phrasing content, but there must be no interactive content descendant and no descendant with the tabindex attribute specified.

https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element

Please work with design to determine a way forward where we do not nest interactive elements. We could have buttons side by side, or just make the chevron the click target for expanding/collapsing. 

I used to be pretty adamant with the old design crew about avoiding these patterns, and we've been successful so far. I'd like us to keep up with enforcing HTML standards back onto design and help educate them. 

Fwiw, here's what GPT had to say about it:

Nesting a `<button>` element inside another `<button>` element is invalid HTML according to the HTML5 specification. This is because:

1. **Semantic Incorrectness**: Buttons are meant to be interactive elements, and nesting one interactive element inside another can create confusion about which element should be interactive.

2. **Accessibility Issues**: Screen readers and other assistive technologies may have difficulty interpreting nested buttons, leading to poor user experiences for individuals relying on these tools.

3. **Browser Compatibility**: Different browsers may handle nested buttons inconsistently. Some browsers may ignore the nested button, while others may render it in unexpected ways, leading to unpredictable behavior.

4. **Event Handling Confusion**: Nesting buttons can create complications in event handling. For instance, click events might not propagate as expected, causing issues in the functionality of the nested buttons.

To maintain valid, accessible, and predictable HTML, it's best to avoid nesting `<button>` elements inside each other.

bdow (dismissed):

src/components/Accordion.tsx
```diff
@@ -76,51 +79,62 @@ export function Accordion<X extends Only<AccordionXss, X>>(props: AccordionProps
...
-      >
-        <span css={Css.fg1.tl.$}>{title}</span>
-        <span
-          css={{
-            ...Css.fs0.$,
-            transition: "transform 250ms linear",
-            transform: expanded ? "rotate(180deg)" : "rotate(0deg)",
-          }}
+      {titleOnClick ? (
+        <div {...focusProps} aria-controls={id} aria-expanded={expanded} css={Css.df.jcsb.$}>
```
stephenh, line 108: I think because of the `fg0/fg1`, this `jcsb` is unnecessary.

src/components/Accordion.tsx
```diff
@@ -76,51 +79,62 @@ export function Accordion<X extends Only<AccordionXss, X>>(props: AccordionProps
...
+          <button {...tid.title} disabled={disabled} css={{ ...touchableStyle, ...Css.fg0.$ }} onClick={titleOnClick}>
+            {title}
+          </button>
+          <button {...tid.toggle} disabled={disabled} css={{ ...touchableStyle, ...Css.fg1.jcfe.$ }} onClick={toggle}>
```
stephenh, line 112: I think personally I would switch the fg0/1 and have the chevon CTA be smaller, but :shrug: could really see that going either way/defer to you/design.

stephenh (approved):

PR #1034, opened 2024-05-31
feat: Pagination to support custom page size options

bdow:


src/components/Pagination.tsx
```diff
@@ -20,12 +20,14 @@ export const defaultPage: OffsetAndLimit = { offset: 0, limit: 100 };
 interface PaginationProps {
   page: readonly [PageNumberAndSize, Dispatch<PageNumberAndSize>] | readonly [OffsetAndLimit, Dispatch<OffsetAndLimit>];
   totalCount: number;
+  pageSizes?: number[];
```
stephenh, line 23: Thought: This is perfectly fine ofc, albeit my usual soapbox that:

a) the UX should just choose the largest page size reasonable and not bother letting the user swap to a smaller page size, and

b) ideally we'd choose a "most reasonable largest page size" and just use it across the app, albeit maybe with very rare overrides for lists of particularly expensive rows (seems more "Beam ethos" of reasonable defaults / as few per-page decisions as possible).

:-)

stephenh (approved):

src/components/Pagination.tsx
bdow: a) - I think that'd be great. This sort of functionality is definitely a nice-to-have for the user and probably rarely used. If it is used, then I'd guess its just to increase the number rather than decrease.

b) - My take here is that it probably depends on what we're paginating. If we have have a flat list of table rows, then I could see 100 being a reasonable first page size. However, if we have a 4 column grid of fairly tall cards, I could understand why we'd want to limit this. 

I think the actual use-case is for a sort of inline-table. https://www.figma.com/design/D7rrEQt2SS6EPfZs7qQjuv/Q1-2024-%7C-Product-Offering?node-id=7055-55508&t=LbEzEtXumwJd1hdE-0
Granted, I would probably push back on UX and just have a "See all" CTA which takes the user to a new page or opens a drawer, or something to display all rows for this table. 🤷

stephenh: Ah true, inline tables having smaller page sizes makes sense.

PR #1036, opened 2024-06-03
feat: Bump truss

stephenh:


update-truss-diff.txt
bdow, line 1: **question:** Was this a scratch file you were using while updating? Or is this now a codegen'd file. Not sure if its really worth having.

update-truss.sh
bdow, line 1: Oh, I see. Fancy!

bdow (approved):
Should we keep this `update-truss.sh` file? I'm hoping we don't need to run it often, or ever again.

update-truss.sh
stephenh: Yep! Kudos ChatGTP helped a lot with the bash scripting. :-)

Fwiw after running this mapping, I assumed I'd gotten everything, but the build is failing with ~20-30 things I missed (I think they were some of the removals), so will need to make another pass for those...

update-truss-diff.txt
stephenh: > Should we keep this update-truss.sh file? I'm hoping we don't need to run it often,
> or ever again.

Agreed wrt the mapping file, but I was thinking the infra might be nice, i.e. to apply to `internal-frontend` and any other apps when bumping.

I guess the `update-truss.sh` itself could just live in the truss repo, and also probably the mapping itself? Since it'd apply to multiple apps. I'll work on moving it over there.

PR #1041, opened 2024-06-05
fix: Nudge the modal close icon to the middle.

stephenh:
The `x` used to be top aligned:

image


chr1sjf0x (approved):

stephenh:
Turns out this looks weird in some of the other modal stories, so just going to close for now

PR #1045, opened 2024-06-11
fix: TreeSelectField improvements

bdow:
Improvements includes:
- Configurable 'chipDisplay' to only show the leaf, root, or all options in the field and tooltip.
- Supports disabled leaf nodes.
- Dedupes selections if there are multiple of the same option within different trees
- Respects changes to 'values' prop

### DIFF
--- a/src/components/Filters/TreeFilter.test.tsx
+++ b/src/components/Filters/TreeFilter.test.tsx
@@ -19,8 +19,8 @@ describe("TreeFilter", () => {
     click(r.getByRole("option", { name: "Grandparent 0" }));
     click(r.getByRole("option", { name: "Child 1-0-0" }));
     click(r.getByRole("option", { name: "Parent 1-1" }));
-    // Then the filter's value is not empty because filter has persisted
-    expect(r.filter_tree).toHaveValue("All");
+    // Then the filter's value is empty now that we're making selections
+    expect(r.filter_tree).toHaveValue("");
     // Then the filter is set to only return the "root" values that are selected.
     expect(r.value).toHaveTextContent('{"tree":["gp:0","child:1-0-0","parent:1-1"]}');
   });
--- a/src/inputs/TreeSelectField/TreeSelectField.tsx
+++ b/src/inputs/TreeSelectField/TreeSelectField.tsx
@@ -544,7 +544,8 @@ function TreeSelectFieldBase<O, V extends Value>(props: TreeSelectFieldProps<O, 
 
         setFieldState((prevState) => ({
           ...prevState,
-          inputValue: nothingSelectedText,
+          // Since we reset the list of options upon selection changes, then set the `inputValue` to empty string to reflect that.
+          inputValue: "",
           filteredOptions: initialOptions.flatMap((o) => levelOptions(o, 0)),
           selectedKeys: [...selectedKeys],
           selectedOptions,
--- a/src/inputs/TreeSelectField/TreeSelectField.tsx
+++ b/src/inputs/TreeSelectField/TreeSelectField.tsx
@@ -464,7 +464,7 @@ function TreeSelectFieldBase<O, V extends Value>(props: TreeSelectFieldProps<O, 
           for (const { option, parents } of maybeOptions) {
             // If the option has children, then we need to unselect those children as well
             if (option.children && option.children.length > 0) {
-              // Ensure we do not impact children that are disabled. They should be able to toggled on/off
+              // Ensure we do not impact children that are disabled. They shouldn't be able to toggled on/off
               const childrenKeys = option.children
                 .flatMap(flattenOptions)
                 .map((o) => valueToKey(getOptionValue(o)))
--- a/src/inputs/TreeSelectField/TreeSelectField.tsx
+++ b/src/inputs/TreeSelectField/TreeSelectField.tsx
@@ -21,6 +21,7 @@ import { ComboBoxInput } from "src/inputs/internal/ComboBoxInput";
 import { ListBox } from "src/inputs/internal/ListBox";
 import {
   findOption,
+  findOptions,
   flattenOptions,
   LeveledOption,
   NestedOption,
@@ -72,7 +73,7 @@ export interface TreeSelectFieldProps<O, V extends Value> extends BeamFocusableP
   /** Which selected options to display as chips in the input field when not focused.
    * @default root */
   chipDisplay?: "all" | "leaf" | "root";
-  disabledOptions?: (V | { value: V; reason: string })[];
+  disabledOptions?: V[];
 }
 
 export function TreeSelectField<O, V extends Value>(props: TreeSelectFieldProps<O, V>): JSX.Element;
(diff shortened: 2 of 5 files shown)


src/inputs/TreeSelectField/TreeSelectField.tsx
```diff
@@ -68,6 +69,10 @@ export interface TreeSelectFieldProps<O, V extends Value> extends BeamFocusableP
   defaultCollapsed?: boolean;
   /** Allow the field's height to grow up to a predefined height (currently 72px), then scroll. */
   multiline?: boolean;
+  /** Which selected options to display as chips in the input field when not focused.
+   * @default root */
+  chipDisplay?: "all" | "leaf" | "root";
```
stephenh, line 74: Neat!

src/inputs/TreeSelectField/TreeSelectField.tsx
```diff
@@ -157,29 +164,53 @@ function TreeSelectFieldBase<O, V extends Value>(props: TreeSelectFieldProps<O,
...
-    // If a user is filtering, then do not provide level to the options as the various paddings may look quite odd.
-    const actualLevel = filtering ? 0 : level;
-    return [
-      [o, actualLevel],
-      ...(o.children?.length && !collapsedKeys.includes(valueToKey(getOptionValue(o)))
-        ? o.children.flatMap((oc: NestedOption<O>) => levelOptions(oc, actualLevel + 1, filtering))
-        : []),
-    ];
-  }
+  const levelOptions = useCallback(
```
stephenh, line 167: Question: I know this name was already here, but is this basically `flattenOptions`? Or `deepFlattenOptions`? As a potentially better name...

src/inputs/TreeSelectField/TreeSelectField.tsx
```diff
@@ -230,7 +267,33 @@ function TreeSelectFieldBase<O, V extends Value>(props: TreeSelectFieldProps<O,
...
+  useEffect(() => {
+    // if the values does not match the values in the fieldState, then update the fieldState
+    const selectedKeys = fieldState.selectedOptions.map((o) => valueToKey(getOptionValue(o)));
+    if (
+      values &&
+      (values.length !== selectedKeys.length || !values.every((v) => selectedKeys.includes(valueToKey(v))))
+    ) {
+      setFieldState(initTreeFieldState());
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
stephenh, line 295: Why is this disabled? Wouldn't we want to call the most-recent/most-correct `initTreeFieldState`, instead of the one with all of the captured-at-the-last-`values` render?

Kinda related, I generally think of `useCallback` being used to memoized functions that will respond to user input, i.e. "on click do something", and that "do something" needs to "capture over" some of our props.

It seems like most of the `useCallback`s here are not "react to user events (callbacks)" but just "regular helper methods", such that I wonder if they should just be helper methods at the bottom of the file, like instead of:

```
const fnThatUsesFoo = useCallback((a, b) => ..., [foo]);
```

We make a function at the bottom of the file:

```
function fnThatUsesFoo((foo, a, b) => ... {})
```

My suspicion is that would be cleaner, but I'm still kinda glancing/reading around.

src/inputs/TreeSelectField/TreeSelectField.tsx
```diff
@@ -157,29 +164,53 @@ function TreeSelectFieldBase<O, V extends Value>(props: TreeSelectFieldProps<O,
...
+  // This could be a parent of the option, or the option itself.
+  const getTopLevelSelections = useCallback(
+    (o: NestedOption<O>, selectedKeys: Set<Key>): NestedOption<O>[] => {
+      // If this element is already selected, return early. Do not bother looking through children.
+      if (selectedKeys.has(valueToKey(getOptionValue(o)))) return [o];
+      // If this element has no children, then look through the children for top level selected options.
+      if (o.children) return [...o.children.flatMap((c) => getTopLevelSelections(c, selectedKeys))];
+      return [];
+    },
+    [getOptionValue],
```
stephenh, line 191: I.e. for this function, if we added `getOptionValue` as a param, it could be a top-level function at the bottom of the file. Granted, callers would have to a) pass `getOptionValue` and b) add `getOptionValue` to their own deps array (if applicable) but that doesn't seem too bad.

Like if there were ~4-5+ deps here, then maybe it's more worth it, to let the caller just depend on the `useCallback`-d function and not know/care what it's non-parameter deps / deps-that-come-from-props are.

src/inputs/TreeSelectField/TreeSelectField.tsx
bdow: Yea, I think that's probably right. I can look to clean up.

src/inputs/TreeSelectField/TreeSelectField.tsx
```diff
@@ -157,29 +164,53 @@ function TreeSelectFieldBase<O, V extends Value>(props: TreeSelectFieldProps<O,
...
+        if (maybeOption.length === 0) return [];
+        const options: NestedOption<O>[] = [];
+        for (const { option } of maybeOption) {
+          options.push(option, ...(option.children?.flatMap(flattenOptions) ?? []));
+        }
+        return options;
       }) ?? [];
+    // Dedupe selectedOptions
+    const selectedOptions: NestedOption<O>[] = maybeDuplicatedSelectedOptions.filter(
+      (o, idx, self) => idx === self.findIndex((t) => getOptionValue(o) === getOptionValue(t)),
```
stephenh, line 212: Thought: this is technically an O(n^2) dedup, which if we get to ~1000 elements in the list will be slow.

Just curious, but what's the use case for letting clients send in dup values, and wanting TreeSelectField to be responsible for deduping them?

bdow: Yea, I need to look at this closer. Admittedly I was throwing this together as quick as possible to help unblock the digital orchestration team.

src/inputs/TreeSelectField/TreeSelectField.tsx
bdow: I disabled this initially because our only current use case of needing to reinitialize the state was when values change (that was the direct ask). We could do better. I'll see what else should trigger a re-initialization of the state.

src/inputs/TreeSelectField/TreeSelectField.tsx
bdow: The user case for letting clients send in duplicated values... tbh I'm not sure. Good push to revisit the need for this here.

src/inputs/TreeSelectField/TreeSelectField.tsx
stephenh: :+1: fwiw I generally believe/follow the logic of "only re-init when `values` changes" and having that be good enough, my concern was just that the lambda called a potentially-stale `initTreeFieldState`, such that the re-init wouldn't actually use the new `values` / other new props that might be relevant in its calculation/state creation.

So if this deps array is just `[values, initTreeFieldState]`, that's probably fine?...

bdow: Yea, I get what you mean. I could potentially extract the logic for determining which values should be selected from the `initTreeFieldState` method and do that separately.
There's a decent amount of logic for determining the values to be selected for the use case where the caller only passes the leaf node ids. We need to check to see if the parent nodes should also be considered selected. Or if they only pass in the parent node ids, then we need to mark all children as selected.
This probably could use some more thought and refactoring to how/where we determine parent states. In this field we the parent selection state is only ever a reflection of its children's state (i.e. you can't have a parent selected but not its children).
I was hoping not to go too far into refactoring this component, but now you got me itching. 😄

src/components/Filters/TreeFilter.test.tsx
```diff
@@ -19,8 +19,8 @@ describe("TreeFilter", () => {
     click(r.getByRole("option", { name: "Grandparent 0" }));
     click(r.getByRole("option", { name: "Child 1-0-0" }));
     click(r.getByRole("option", { name: "Parent 1-1" }));
-    // Then the filter's value is not empty because filter has persisted
-    expect(r.filter_tree).toHaveValue("All");
+    // Then the filter's value is empty now that we're making selections
+    expect(r.filter_tree).toHaveValue("");
```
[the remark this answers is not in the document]
bdow: I've added some tests here now.

src/inputs/TreeSelectField/TreeSelectField.tsx
bdow: I ended up refactoring this. I believe the set up should be much better now.

src/inputs/TreeSelectField/TreeSelectField.tsx
```diff
@@ -392,71 +425,81 @@ function TreeSelectFieldBase<O, V extends Value>(props: TreeSelectFieldProps<O,
...
-            const childrenKeys = option.children.flatMap(flattenOptions).map((o) => valueToKey(getOptionValue(o)));
-            [key, ...childrenKeys].forEach(removedKeys.add, removedKeys);
-          }
+          const maybeOptions = findOptions(fieldState.allOptions, key, getOptionValue);
+          if (maybeOptions.length === 0) return;
+
+          for (const { option, parents } of maybeOptions) {
+            // If the option has children, then we need to unselect those children as well
+            if (option.children && option.children.length > 0) {
+              // Ensure we do not impact children that are disabled. They should be able to toggled on/off
```
blambillotte, line 467: question: "should *not* be able to be toggled"?

bdow: hah yes! The comments are wrong but, I think the logic is right. I'm pretty sure I wrote a test for this. Will update the comment. Thanks!

bdow: I don't have a test for this. I'll write one really quick.

PR #1048, opened 2024-06-19
feat: Adding support for multiple levels of nesting on grid tables, to keep adding marging based on the level of nesting index

Grandlouis:
example: 
imagen

### DIFF
--- a/src/components/Table/TableStyles.tsx
+++ b/src/components/Table/TableStyles.tsx
@@ -235,7 +235,8 @@ export const cardStyle: GridStyle = {
   levels: new Proxy({} as Record<string, { rowIndent: number }>, {
     get(_target, level: string) {
       const parsedLevel = parseInt(level);
-      if (Number.isNaN(parsedLevel)) return undefined;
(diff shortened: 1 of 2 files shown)


bdow (approved):

chr1sjf0x (approved):
🍣

src/components/Table/TableStyles.tsx
```diff
@@ -231,10 +231,17 @@ export const cardStyle: GridStyle = {
   },
   rowHoverColor: "none",
   nonHeaderRowHoverCss: Css.bshHover.bcGray700.$,
-  levels: {
-    1: { rowIndent: 24 },
-    2: { rowIndent: 48 },
-  },
+  // this will allow having N amount of nested childs without having to define each level margin
+  levels: new Proxy({} as Record<string, { rowIndent: number }>, {
```
stephenh, line 235: Thought: The proxy is cute/creative, but if we end up blessing this, we should probably just update the `GridStyle.levels` type to be a union of `Record<string, { rowIndent }> | `(level: number) => { rowIndent }`.

We control the types, so no need to torture ourselves with fulfilling them in esoteric ways, vs. more directly declaring the behavior we want.

PR #1050, opened 2024-06-26
fix: Add page break styles for table rows

bdow:


JonnCharpentier (approved):

stephenh (approved):
Thanks for helping out, as always!

PR #1057, opened 2024-08-05
feat: allow to change selected color on toggle chip group

JonnCharpentier:


src/inputs/ToggleChipGroup.tsx
```diff
@@ -90,8 +95,10 @@ function ToggleChip(props: ToggleChipProps) {
           ...Css.relative.dib.br16.sm.px1.cursorPointer.pyPx(4).bgGray200.if(isDisabled).cursorNotAllowed.gray600.pr1.$,
           ...(isSelected
             ? {
-                ...Css.white.bgBlue700.$,
-                ":hover:not([data-disabled='true'])": Css.bgBlue800.$,
+                ...Css.white.if(xss?.backgroundColor !== undefined).bgColor(xss?.backgroundColor).else.bgBlue700.$,
+                ":hover:not([data-disabled='true'])": Css.if(xss?.backgroundColor !== undefined).bgColor(
+                  xss?.backgroundColor,
```
stephenh, line 100: Suggestion: I think these would probably be clearer to write as:

```
Css.bgColor(xss?.backgroundColor ?? Palette.white)
```

Instead of the `if(...)` syntax

stephenh (approved):

PR #1060, opened 2024-08-14
feat: Add truck and union icon

kbesingeryeh:
Add Truck and Union Icon


src/components/Icon.tsx
```diff
@@ -600,6 +600,16 @@ export const Icons = {
   cube: (
     <path d="M19.406 4.08601L10.406 0.0860059C10.146 -0.0289941 9.853 -0.0289941 9.593 0.0860059L0.593 4.08601C0.573 4.09501 0.559 4.11001 0.539 4.12101C0.511 4.13501 0.481 4.14401 0.455 4.16101C0.433 4.17601 0.416 4.19501 0.395 4.21101C0.367 4.23201 0.34 4.25501 0.315 4.27901C0.274 4.31801 0.238 4.36001 0.205 4.40501C0.185 4.43301 0.164 4.45801 0.146 4.48601C0.115 4.53901 0.0919999 4.59401 0.0699999 4.65101C0.0609999 4.67801 0.047 4.70301 0.039 4.73001C0.015 4.81701 0 4.90701 0 5.00001V15C0 15.396 0.232 15.753 0.594 15.914L9.594 19.914C9.724 19.972 9.862 20 10 20C10.139 20 10.273 19.961 10.402 19.904L10.406 19.914L19.406 15.914C19.768 15.753 20 15.396 20 15V5.00001C20 4.60401 19.768 4.24701 19.406 4.08601ZM10 2.09501L16.538 5.00001L10 7.90501L8.692 7.32401L3.463 5.00001L10 2.09501ZM2 14.351V6.53901L9 9.65001V17.461L2 14.351ZM11 17.461V9.65001L18 6.53901V14.351L11 17.461Z" />
   ),
+  union: (
```
blambillotte, line 603: thought: do we have a more descriptive name for this icon? Seems like it's a `hardhat` ?

blambillotte: ah do we have this already?
image

kbesingeryeh: sure, i can add that. I just took the name from Figma. I think hardhad is the inverse so I actually need to add that one 
Screenshot 2024-08-14 at 3 58 56 PM

kbesingeryeh: yeah, i'll just change the name to `outlineHardHat` and `hardHat`

blambillotte: Ah, maybe double check with Andrea that we need 2 flavors of this icon? 
I see that Edwar added the inverse version to local blueprint: https://github.com/homebound-team/internal-frontend/blob/main/src/components/Icon.tsx#L83
Would be nice if we can just use the outline version if nobody has strong opinions

kbesingeryeh: The truck icon is also on the Front-end, but not on Beam. I will just go ahead and close this out. Not sure if its worth adding to Beam

PR #1061, opened 2024-08-21
fix: TreeSelectField and treeFilter fixes

bdow:
Updates/fixes include:
1. TreeSelectField can reset values when outside component sets values to undefined
2. Can set individual groups to be default collapsed
3. Increase min-width of ListBox from 200 to 320px wide per Design feedback

### DIFF
--- a/src/inputs/TreeSelectField/TreeSelectField.test.tsx
+++ b/src/inputs/TreeSelectField/TreeSelectField.test.tsx
@@ -3,7 +3,7 @@ import { TreeSelectField } from "src/inputs";
 import { NestedOption } from "src/inputs/TreeSelectField/utils";
 import { HasIdAndName } from "src/types";
 import { noop } from "src/utils";
-import { blur, click, focus, render, wait } from "src/utils/rtl";
+import { blur, click, focus, getSelected, render, wait } from "src/utils/rtl";
 import { useState } from "react";
 
 describe(TreeSelectField, () => {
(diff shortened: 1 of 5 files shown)


src/inputs/TreeSelectField/TreeSelectField.test.tsx
```diff
@@ -682,6 +682,38 @@ describe(TreeSelectField, () => {
...
+        options={getNestedOptions()}
+        label="Favorite League"
+        getOptionValue={(o) => o.id}
+        getOptionLabel={(o) => o.name}
+      />,
+    );
+    // Then the options are initially selected
+    expect(getSelected(r.favoriteLeague)).toEqual(["MLB", "NBA"]);
+
+    // When we re-render with values set to undefined (simulating an outside component's "clear" action, i.e. Filters)
```
stephenh, line 701: Love the comment!

stephenh (approved):

src/inputs/TreeSelectField/TreeSelectField.tsx
```diff
@@ -94,8 +94,16 @@ export function TreeSelectField<O, V extends Value>(
...
   const [collapsedKeys, setCollapsedKeys] = useState<Key[]>(
-    Array.isArray(options) && defaultCollapsed ? options.map((o) => getOptionValue(o)) : [],
+    !Array.isArray(options)
+      ? []
+      : defaultCollapsed
+      ? options.map((o) => getOptionValue(o))
+      : options
+          .flatMap(flattenOptions)
+          .filter((o) => o.defaultCollapsed)
+          .map((o) => getOptionValue(o)),
```
JonnCharpentier, line 104: issue: usually if we haven't load the options, default to an empty array, this will provoke the `option.defaultCollapsed` not be applied
```
treeFilter({
        label: "Locations",
        filterBy: "all",
        getOptionValue: (o) => o.id,
        options: locationOptions ?? [],
      }),
```

JonnCharpentier (changes requested):
I pull this branch locally and ran `yarn watch-beam`, and the mindWidth is not getting applied: 
image

bdow:
> I pull this branch locally and ran `yarn watch-beam`, and the mindWidth is not getting applied: 

@JonnCharpentier - Did you run with a `--force` on `internal-frontend`? When I test the width in Storybook here, it works fine. Maybe we can VC today and verify together. Let's catch up on Slack.

JonnCharpentier (approved):

PR #1063, opened 2024-08-31
feat: Add GridTable csv export.

stephenh:
As easy as:

```
const api = useGridTableApi();

<Button onClick={() => api.downloadToCsv("report.csv")} label="CSV" />
```

Currently assumes all of the data is already available client-side -- not really sure how this would work with server-side driven pagination.

Also assumes we want all of the columns/rows included, and that they're directly mapped from the `content` / `value` keys. We could also add a `csvValue` to let columns provide CSV-specific values.

image

### DIFF
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -12,7 +12,7 @@ import { TableStateContext } from "src/components/Table/utils/TableState";
 import { emptyCell, matchesFilter } from "src/components/Table/utils/utils";
 import { Css, Palette } from "src/Css";
 import { useComputed } from "src/hooks";
-import { Checkbox, SelectField, TextField } from "src/inputs";
+import { SelectField, TextField } from "src/inputs";
 import { noop } from "src/utils";
 import {
   cell,
@@ -27,7 +27,6 @@ import {
   wait,
   withRouter,
 } from "src/utils/rtl";
-import { Button } from "src";
 
 // Most of our tests use this simple Row and 2 columns
 type Data = { name: string; value: number | undefined | null };
--- a/src/components/Table/GridTable.test.tsx
+++ b/src/components/Table/GridTable.test.tsx
@@ -3793,13 +3793,19 @@ describe("GridTable", () => {
     let api: GridTableApi<Row> | undefined;
 
     const columns: GridColumn<Row>[] = [
-      // Given column one returns JSX, but defines a `sortValue`
+      // Given a column returns JSX, but defines a `sortValue`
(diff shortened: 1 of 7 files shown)


src/components/Table/GridTable.test.tsx
```diff
@@ -3799,22 +3833,6 @@ function Collapse({ id }: { id: string }) {
   );
 }
 
-function Select({ id }: { id: string }) {
```
stephenh, line 3802: This was no longer used

src/components/Table/GridTableApi.ts
```diff
@@ -175,10 +199,78 @@ export class GridTableApiImpl<R extends Kinded> implements GridTableApi<R> {
...
+    link.setAttribute("download", fileName);
+    link.style.visibility = "hidden";
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
+  }
+
+  public copyToClipboard(): Promise<void> {
+    // Copy the CSV content to the clipboard
+    return navigator.clipboard.writeText(this.generateCsvContent().join("\n"));
```
bdow, line 221: **nit:** This can throw an error if copying to the clipboard is not allowed. I'm not sure how this would impact the rest of the application if the error is thrown (hopefully not take it down completely), but makes me wonder if we should wrap in a try/catch and report to DataDog and maybe a `window.alert(e.message)`.

src/components/Table/components/Row.tsx
```diff
@@ -198,14 +199,18 @@ function RowImpl<R extends Kinded, S>(props: RowProps<R>): ReactElement {
             currentColspan -= 1;
             return null;
           }
-          const maybeContent = applyRowFn(column as GridColumnWithId<R>, row, rowApi, level, isExpanded, {
+
+          // Combine all our drag stuff into a mini-context/parameter object...
+          const dragData: DragData<R> = {
```
bdow, line 204: **praise:** Thanks for cleaning up

bdow (approved):
Nice! This is a fairly small PR for something that seems like a bigger lift from the outside.

src/components/Table/GridTableApi.ts
stephenh: Yeah, afaiu it would return a rejected Promise, instead of immediately throwing, so I was thinking the caller could just deal with it -- but I suppose centralized error handling would be nice, so I added a q&d window.alert for now (but still returning a rejected promise, with the idea that the caller UI could show like a red "x" instead of a green "check").

In terms of logging, I think dd should be logging all of our rejected promises, which is what this would be atm... :thinking:

PR #1070, opened 2024-09-18
feat: Auto call onAddNew when the user searchs and there is no match

JonnCharpentier:
This change allow us to automatically switch to `Add mode` when the user search and there are not matches in the provided options

### DIFF
--- a/src/inputs/internal/ComboBoxBase.tsx
+++ b/src/inputs/internal/ComboBoxBase.tsx
@@ -312,6 +312,11 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
 
   const [debouncedSearch] = useDebounce(searchValue, 300);
 
+  useEffect(() => {
+    if (!debouncedSearch || !onAddNew) return;
+    if (filteredOptions.length === 1 && filteredOptions[0] === addNewOption) onAddNew(debouncedSearch);
+  }, [debouncedSearch, filteredOptions, onAddNew]);
+
   // Reset inputValue when closed or selected changes
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -312,6 +312,11 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
 
   const [debouncedSearch] = useDebounce(searchValue, 300);
 
+  useEffect(() => {
+    if (!debouncedSearch || !onAddNew) return;
+    if (filteredOptions.length === 1 && filteredOptions[0] === addNewOption) onAddNew(debouncedSearch);
+  }, [debouncedSearch, filteredOptions, onAddNew]);
```
bdow, line 318: **thought:** This feels like we're assuming too much. It's possible the user has a typo, or just wants to see what options are available. They may not want to go straight to the Add New callback. 
**question:** Is this coming from product feedback?

bdow: So if I have a typo, then I'll automatically move to the "Add new" text field. I'll try to hit backspace to fix my input, but then I'm no longer in the select field. As a user I'll have to understand that not matching anything in the list will automatically change my field. And then know that I can return to the list by the two methods you described. 
I don't think that is going to be clear to our users. 
I disagree with the purposed UX and I think it's going to make for a poor user experience. If this is really want Dave wants and thinks that it is the correct thing, then let's push it through.

JonnCharpentier: I'll talk to Dave about it

bdow (approved):
🙈

PR #1072, opened 2024-09-23
feat: Add select on focus  to text field 

JonnCharpentier:


src/inputs/TextFieldBase.tsx
```diff
@@ -61,6 +61,7 @@ export interface TextFieldBaseProps<X>
   // Replaces empty input field and placeholder with node
   // IE: Multiselect renders list of selected items in the input field
   unfocusedPlaceholder?: ReactNode;
+  selectOnFocus?: boolean;
```
stephenh, line 64: Hey @JonnCharpentier , just curious, but I'm getting back into things--what ticket/request was this change for? I'm wondering if we could add an in-source comment to this, like:

```
/** Disable select on focus, i.e. ...description of the use case you're supporting... */
```

JonnCharpentier: This change is for this ticket: https://app.shortcut.com/homebound-team/story/56770/improve-add-new-attribute-ux

Basically Dave want us to auto swap into the text field when the user search don't have matches, but also auto-focus the field for the user to continue typing into the text field, this change optionally removes the Select all on focus, so that when we swap into the text field, the field is focus but not selected.

stephenh: Okay, that makes sense -- the comment as-is does not really explain "why", imo it is just restating "what" happens; can you reword it to be:

```
/** Allow focusing without selecting, i.e. to let the user keep typing after we've pre-filled text + called focus, like the Add New component. */
```

PR #1077, opened 2024-10-14
fix: email icon to use `clipRule` instead of `clip-rule`;

apattersonATX-HB:
This was getting really annoying and I needed to release a new beam version anyway to use a `calendar-x` icon that was added as a `chore` 😃 

Screenshot 2024-10-14 at 11 53 25 AM


stephenh (approved):
Nice, thanks!

PR #1079, opened 2024-10-16
docs: Expand useComputed docs.

stephenh:
Specifically on what fn vs. deps are.

### DIFF
--- a/src/hooks/useComputed.ts
+++ b/src/hooks/useComputed.ts
@@ -13,7 +13,43 @@ interface Current<T> {
   hasRan: boolean;
 }
 
-/** Evaluates a computed function `fn` to a regular value and triggers a re-render whenever it changes. */
+/**
+ * Evaluates a computed function `fn` to a regular value and triggers a re-render whenever it changes.
+ *
+ * Some examples:
+ *
+ * ```ts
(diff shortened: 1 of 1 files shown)


JonnCharpentier (approved):

src/hooks/useComputed.ts
```diff
@@ -13,7 +13,43 @@ interface Current<T> {
...
+ * Note that the difference between the `fn` and the `deps` is:
+ *
+ * - `fn` is "which values we are watching in the observable" (i.e. store or `formState`), and
+ * - `deps` is "which observable we're watching" (i.e. store or `formState`)
+ *
+ * So the `deps` array shouldn't overlap with any of the "watched values" of the `fn` lambda,
+ * other than the root observer itself (which admittedly should rarely change, i.e. our stores
+ * are generally global-ish/very stable, but can change if the user switches pages i.e. "from
+ * editing author:1 t editing author:2").
+ */
```
JonnCharpentier, line 52: @stephenh  is something like this a good approach? 
`const item = useComputed(() => items.find((i) => i.id === formState.itemId.value), [formState, items]);`
Where Items is not an observable?

PR #1085, opened 2024-11-06
feat: add modal prop to force modal interaction;

apattersonATX-HB:
Orchestration has been asked to combine the dynamic schedule draft mode with the standard view. The draft pages were built as individual routes using a shared store to pass task changes between them. In combining these pages, we need to warn users their changes will be lost before allowing them to navigate away. Since we're using a shared store and not just a shared form I want to update Blueprints `<NavigationContext />` `navigationCheckCallbacks` to support more robust nav checks. I've gotten that working, but since the inner `<NavigationPrompt />` from `react-router-navigation-prompt` passes its own callback when a cancel action happens, routing breaks if a user clicks outside of our modal or closes with the `x` icon since the NavigationPrompt inner onCancel callback is never called. 

Context (branch is WIP):
- [NavigationContext.tsx changes](https://github.com/homebound-team/internal-frontend/compare/main...sc-60329/schedules-merge-draft-view#diff-3e5dd94f92d2a5b4d5e1e0b362d7d8e2d9ec52af58eb9cfd833a0e195d56ed3d)
- [Using new callback type](https://github.com/homebound-team/internal-frontend/compare/main...sc-60329/schedules-merge-draft-view#diff-832d5b6b2b80a4d541fbe0eb1a55eaf856777b136a6f2e4540880035f544be87R82)

TLDR: Needed a way to force users to interact with the modal.

https://github.com/user-attachments/assets/6e5c631c-e7e9-4ffc-8c1f-4299afb15c29

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -31,6 +31,13 @@ export interface ModalProps {
   api?: MutableRefObject<ModalApi | undefined>;
   /** Adds a border for the header. */
   drawHeaderBorder?: boolean;
+  /**
+   * Hides `x` icon and removes ability to close modal by clicking outside of it.
+   *
+   * Relies on you to provide a way to close the modal, i.e. a cancel or confirm button.
+   * Useful if you definitely need to force the user to make a choice.
+   * */
+  forceModalInteraction?: boolean;
 }
 
 export type ModalApi = {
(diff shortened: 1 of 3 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -31,6 +31,13 @@ export interface ModalProps {
   api?: MutableRefObject<ModalApi | undefined>;
   /** Adds a border for the header. */
   drawHeaderBorder?: boolean;
+  /**
+   * Hides `x` icon and removes ability to close modal by clicking outside of it.
+   *
+   * Relies on you to provide a way to close the modal, i.e. a cancel or confirm button.
+   * Useful if you definitely need to force the user to make a choice.
+   * */
+  forceModalInteraction?: boolean;
```
stephenh, line 40: Thought: I think my first guess at naming this would have been `allowClosing` (defaults true, users can set it to false)... Maybe still a slight preference for that, but not totally against `forceModalInteraction`; wdyt?

stephenh (approved):

src/components/Modal/Modal.tsx
apattersonATX-HB: Names are hard, and i cant come up with anything better 🙃 Will update to `allowClosing`

PR #1086, opened 2024-11-07
fix: expose `inc` prop on OverlayTriggerProps

apattersonATX-HB:
I've had to write this

```
      <ButtonMenu
        // @ts-ignore: it works but prop isn't exposed
        trigger={{ icon: "verticalDots", inc: 2.3 }}
```

more than a few times lately 😄


stephenh (approved):
Makes sense! The inc: 2.3...makes me think it'd be great to have more logical/aliases sizes like small/big/but shrug.

PR #1088, opened 2024-11-21
Add truncate to multiline

kbesingeryeh:


src/inputs/TextFieldBase.tsx
```diff
@@ -169,7 +169,7 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
       // Not using Truss's inline `if` statement here because `addIn` properties do not respect the if statement.
       ...(contrast && Css.addIn("&::selection", Css.bgGray800.$).$),
       // For "multiline" fields we add top and bottom padding of 7px for compact, or 11px for non-compact, to properly match the height of the single line fields
-      ...(multiline ? Css.br4.pyPx(compact ? 7 : 11).add("resize", "none").$ : Css.truncate.$),
+      ...(multiline ? Css.br4.pyPx(compact ? 7 : 11).add("resize", "none").truncate.$ : Css.truncate.$),
```
kbesingeryeh, line 172: I had to add `truncate` to this `css block`, but I'm wondering why if its not `multiline` we would truncate? `multiline` wasn't really working before when testing on the front-end using `Beam` locally

Screenshot 2024-11-21 at 1 39 52 PM

PR #1090, opened 2024-12-05
fix: allow searching in collapsed TreeSelectField

0ces:
Allows to filter options on a tree select field even if the parent option is collapsed.

### DIFF
--- a/src/inputs/TreeSelectField/TreeSelectField.test.tsx
+++ b/src/inputs/TreeSelectField/TreeSelectField.test.tsx
@@ -394,6 +394,32 @@ describe(TreeSelectField, () => {
     expect(r.getByRole("option", { name: "WNBA" })).toBeVisible();
   });
 
+  it("can filter options even if parent if collapsed", async () => {
+    // Given a TreeSelectField with nested options
+    const r = await render(
+      <TreeSelectField
+        onSelect={noop}
+        options={getNestedOptions()}
+        label="Favorite League"
+        values={[]}
+        getOptionValue={(o) => o.id}
+        getOptionLabel={(o) => o.name}
+      />,
+    );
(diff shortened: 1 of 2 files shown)


src/inputs/TreeSelectField/TreeSelectField.tsx
```diff
@@ -325,6 +325,7 @@ function TreeSelectFieldBase<O, V extends Value>(props: TreeSelectFieldProps<O,
   // Update the filtered options when the input value changes
   const onInputChange = useCallback(
     (inputValue: string) => {
+      console.log({ inputValue });
```
stephenh, line 328: Nit: snuck in

src/inputs/TreeSelectField/TreeSelectField.test.tsx
```diff
@@ -394,6 +394,32 @@ describe(TreeSelectField, () => {
...
+      />,
+    );
+    // When opening the options
+    click(r.favoriteLeague);
+    // Then the child options are visible
+    expect(r.getByRole("option", { name: "MLB" })).toBeVisible();
+    // When we collapse the parent option
+    click(r.treeOption_collapseToggle_basketball);
+    // And typing in the filter input
+    fireEvent.input(r.favoriteLeague, { target: { value: "nba" } });
```
stephenh, line 416: Kinda wonder why this isn't using `type`... :shrug:

stephenh (approved):
Makes sense!

src/inputs/TreeSelectField/TreeSelectField.test.tsx
0ces: I just copied from one of the tests above, using `type` doesn't work :/

PR #1091, opened 2024-12-05
fix: gridtable smooth scroll default; expose options;

apattersonATX-HB:
Title

Found we needed an update in https://github.com/homebound-team/internal-frontend/pull/6048

### DIFF
--- a/src/components/Table/GridTableApi.ts
+++ b/src/components/Table/GridTableApi.ts
@@ -1,8 +1,16 @@
 import { comparer } from "mobx";
 import { computedFn } from "mobx-utils";
 import { MutableRefObject, useMemo } from "react";
-import { FlatIndexLocationWithAlign, VirtuosoHandle } from "react-virtuoso";
-import { applyRowFn, createRowLookup, GridRowLookup, isGridCellContent, isJSX, MaybeFn } from "src/components/index";
+import { VirtuosoHandle } from "react-virtuoso";
+import {
+  applyRowFn,
+  createRowLookup,
+  GridRowLookup,
+  GridTableScrollOptions,
+  isGridCellContent,
+  isJSX,
+  MaybeFn,
+} from "src/components/index";
 import { GridDataRow } from "src/components/Table/components/Row";
 import { DiscriminateUnion, Kinded } from "src/components/Table/types";
 import { TableState } from "src/components/Table/utils/TableState";
@@ -29,7 +37,7 @@ export function useGridTableApi<R extends Kinded>(): GridTableApi<R> {
 /** Provides an imperative API for an application page to interact with the table. */
 export type GridTableApi<R extends Kinded> = {
   /** Scrolls row `index` into view; Defaults "smooth" behavior; only supported with `as=virtual` and after a `useEffect`. */
-  scrollToIndex(index: number | FlatIndexLocationWithAlign): void;
+  scrollToIndex(index: GridTableScrollOptions): void;
 
   /** Returns the currently-visible rows. */
   getVisibleRows(): GridDataRow<R>[];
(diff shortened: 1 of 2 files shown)


src/components/Table/GridTableApi.ts
```diff
@@ -28,8 +28,8 @@ export function useGridTableApi<R extends Kinded>(): GridTableApi<R> {
 
 /** Provides an imperative API for an application page to interact with the table. */
 export type GridTableApi<R extends Kinded> = {
-  /** Scrolls row `index` into view; only supported with `as=virtual` and after a `useEffect`. */
-  scrollToIndex(index: number): void;
+  /** Scrolls row `index` into view; Defaults "smooth" behavior; only supported with `as=virtual` and after a `useEffect`. */
+  scrollToIndex(index: number | FlatIndexLocationWithAlign): void;
```
apattersonATX-HB, line 32: Exposed all the options virtuoso offers since they seemed pretty standard.

stephenh: Historically we've avoided exposing 3rd party types to callers, b/c it leads to them getting coupled to ideally-internal dependencies that make it harder to change/swap out over the long-term.

Also, ideally Beam is making opinionated / "just works" defaults that pages that "want to scroll" get w/o having to tweak/ask for different behavior.

Is there something specific in `FlatIndex...` that pages need / "should be allowed" to override/pass? Fwiw kinda hoping not yet, but maybe?

If so, I'd like to pull those out into a Beam-specific type.

apattersonATX-HB: Ah thats what looked off about it. Updated to use a custom type.

Specifically the only two that I think make sense are `behavior` and `position`.

src/components/Table/types.ts
```diff
@@ -12,6 +12,18 @@ export type Direction = "ASC" | "DESC";
 export type MaybeFn<T> = T | (() => T);
 export type GridCellAlignment = "left" | "right" | "center";
 
+export type GridTableScrollOptions =
+  | number
+  | {
+      /** The index of the row to scroll to */
+      index: number;
+      behavior?: "auto" | "smooth";
```
stephenh, line 20: Just as a lazy/ocd ask, when would a page not want smooth? Kinda wonder if we should document that as a recommendation in the source, i.e.:

```
/** Defaults to `"smooth'`, but can be overridden to `"auto"` for ...tldr of the use case... */
```

...this is admittedly too obsessive, but sometimes if there are very obvious use-cases for each, instead of `behavior: auto | smooth`, an API like `behavior: use-case-1 | use-case-2` can be interesting, to make the API "more declarative" and hide even more impl details from pages.

stephenh (approved):

src/components/Table/types.ts
apattersonATX-HB: Not toooooo obsessive 😆 

We'd want to consider using `"auto"` when we have a very large table and the scrolling animation becomes excessive. To some users it can be disorienting. I'll admit i've only seen that use case specifically in the mobile world but if we're running a gridtable as cards on a tablet it may come up.

For placement `center | end` I think of grid table as cards again. I can see product requesting it whatever we're scrolling to be centered 🤔 

Good callout to add that use case into the docs for sure 👍🏽

stephenh: > "very large tables should be auto"

That makes sense.

Fwiw that could potentially be a Beam-driven decision as well, at some point, i.e. "if you have more than 1k rows => scrolling is auto, otherwise scrolling is smooth", and the GridTable could "just do that", and remove this opt, instead of asking each page to decide "when should I use smooth vs. auto?".

Little bit of a soapbox, but I think "what is Beam" can get pulled into/distracted by "are the atomic UI components off-the-shelf MUI or react-aria or fully custom" or what not, but imo the core mission/value-add of Beam isn't "how the low-level components look / are they custom / are they off-the-shelf", but it's these "opinionated default", "as minimal as possible" APIs that simplify & codify behavior for application pages.

(I.e. like this "big tables get auto scrolling, small tables get smooth scrolling" straw man--to make our application pages as "dumb" / "declarative" / "low code" as possible.

Granted, this approach assumes that UX is also bought into the "each application page makes as few layout/UX decisions as possible" approach, to which the answer has historically been "kinda". :-)

PR #1092, opened 2024-12-06
feat: palette fill as a presentation prop; getInputStylePalette fn prop;

apattersonATX-HB:
[SC-61998](https://app.shortcut.com/homebound-team/story/61998/houston-pilot-feedback-update-beam-to-support-coloring-dropdowns) | [Figma](https://www.figma.com/design/xQ3iNLHJCFEJeOi7boONY4/Q1-Q4-2024-%7C-Dynamic-Schedules?node-id=11535-84519&node-type=frame&t=wrjRlakQknbMmW1j-0)

https://github.com/user-attachments/assets/61fdf544-1949-475c-b01d-f805504a7f61

Initially the goal was to only have select fields show color but since the underlying `<TextFieldBase` that actually controls colors takes props from presentationFieldProps it was pretty easy to hook up the components starting from there. So technically our `<MultiselectField also supports `getInputStylePalette` & `inputStylePalette` and our `<TextFields` support `inputStylePalette` as well practically for free.

### DIFF
--- a/src/components/PresentationContext.tsx
+++ b/src/components/PresentationContext.tsx
@@ -2,6 +2,8 @@ import { createContext, PropsWithChildren, useContext, useMemo } from "react";
 import { GridStyle } from "src/components/Table";
 import { Typography } from "src/Css";
 
+export type InputStylePalette = "success" | "warning" | "caution" | "info";
+
 export interface PresentationFieldProps {
   numberAlignment?: "left" | "right";
   /** Sets the label position or visibility. Defaults to "above" */
@@ -23,10 +25,12 @@ export interface PresentationFieldProps {
   errorInTooltip?: true;
   /** Allow the fields to grow to the width of its container. By default, fields will extend up to 550px */
   fullWidth?: boolean;
+  /** Changes bg and hoverBg; Takes priority over `contrast`; Useful when showing many fields w/in a table that require user attention; In no way should be used as a replacement for error/focus state */
+  inputStylePalette?: InputStylePalette;
 }
 
 export type PresentationContextProps = {
-  fieldProps?: PresentationFieldProps;
+  fieldProps?: Omit<PresentationFieldProps, "inputStylePalette">;
   gridTableStyle?: GridStyle;
   // Defines whether content should be allowed to wrap or not. `undefined` is treated as true.
   wrap?: boolean;
--- a/src/components/PresentationContext.tsx
+++ b/src/components/PresentationContext.tsx
@@ -30,6 +30,7 @@ export interface PresentationFieldProps {
 }
 
 export type PresentationContextProps = {
+  /** `inputStylePalette` omitted because it is too dependent on the individual field use case to be controlled at this level  */
   fieldProps?: Omit<PresentationFieldProps, "inputStylePalette">;
   gridTableStyle?: GridStyle;
   // Defines whether content should be allowed to wrap or not. `undefined` is treated as true.
--- a/src/inputs/SelectField.stories.tsx
+++ b/src/inputs/SelectField.stories.tsx
@@ -411,7 +411,7 @@ function TestSelectField<T extends object, V extends Value>(
   const [inputStylePalette, setInputStylePalette] = useState<InputStylePalette | undefined>();
 
   // @ts-ignore: Hacking around type props within the testSelectField instead of the SB Template
-  const shouldUseStylePalette: boolean = !!props.options?.some((o) => o.name.includes("style palette when selected"));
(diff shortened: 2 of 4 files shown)


src/components/PresentationContext.tsx
```diff
@@ -2,6 +2,8 @@ import { createContext, PropsWithChildren, useContext, useMemo } from "react";
 import { GridStyle } from "src/components/Table";
 import { Typography } from "src/Css";
 
+export type InputStylePalette = "success" | "warning" | "caution" | "info";
```
apattersonATX-HB, line 5: Mirrors our `coloredChips` and does leave room for more bespoke customization in the future if the need arises

src/components/PresentationContext.tsx
```diff
@@ -23,10 +25,12 @@ export interface PresentationFieldProps {
   errorInTooltip?: true;
   /** Allow the fields to grow to the width of its container. By default, fields will extend up to 550px */
   fullWidth?: boolean;
+  /** Changes bg and hoverBg; Takes priority over `contrast`; Useful when showing many fields w/in a table that require user attention; In no way should be used as a replacement for error/focus state */
+  inputStylePalette?: InputStylePalette;
 }
 
 export type PresentationContextProps = {
-  fieldProps?: PresentationFieldProps;
+  fieldProps?: Omit<PresentationFieldProps, "inputStylePalette">;
```
apattersonATX-HB, line 33: Not something we want set at a context level

src/inputs/SelectField.stories.tsx
```diff
@@ -223,6 +232,39 @@ export const Contrast = Template.bind({});
...
+  return undefined;
+}
+
+const standardColoredSelectArgs = {
+  options: coloredOptions,
+  // @ts-ignore
+  getInputStylePalette,
+};
+
+export const Colored = Template.bind({});
```
apattersonATX-HB, line 250: Copy pasta

src/components/PresentationContext.tsx
```diff
@@ -23,10 +25,12 @@ export interface PresentationFieldProps {
   errorInTooltip?: true;
   /** Allow the fields to grow to the width of its container. By default, fields will extend up to 550px */
   fullWidth?: boolean;
+  /** Changes bg and hoverBg; Takes priority over `contrast`; Useful when showing many fields w/in a table that require user attention; In no way should be used as a replacement for error/focus state */
+  inputStylePalette?: InputStylePalette;
```
apattersonATX-HB, line 29: I like this since there's been a lot of discussion around how we can show to users what changes they've made at a glance, mostly in dynamic schedules. With this we can pass something in based on a form state for example if asked

apattersonATX-HB:
🕯️🕯️🕯️ Summoning @bdow to poke holes 🕯️🕯️🕯️

src/components/PresentationContext.tsx
stephenh: Makes sense, that it's not a context-level thing, but then wonder if we can find a better spot for declaring it, then `PresentationFieldProps`?

Hm, I guess `PresentationFieldProps` does seem to be "common across text fields and select fields", and it isn't used by other things that aren't getting supported in this PR... :thinking: so sgtm...

Could make your PR comment an in-source comment :shrug:

apattersonATX-HB: > "common across text fields and select fields",
Yup! Thats why its here. Comment updated

src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -20,6 +20,8 @@ export interface ComboBoxBaseProps<O, V extends Value> extends BeamFocusableProp
   getOptionMenuLabel?: (opt: O, isUnsetOpt?: boolean, isAddNewOption?: boolean) => string | ReactNode;
   getOptionValue: (opt: O) => V;
   getOptionLabel: (opt: O) => string;
+  /** Sets an input style based on the option(s) selected if `inputStylePalette` is not set */
+  getInputStylePalette?: (values: V[] | undefined) => InputStylePalette | undefined;
```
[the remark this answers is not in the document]
apattersonATX-HB: Exactly. I thought `getInputStylePalette` lined up with our other label getters and would cut down on extra code the caller needs to write to support something like `[Good, Better]` or `value passes some regExp combination` but its _basically_ the same in practice so i'll remove it in favor of just relying on updating `inputStylePalette` values

stephenh (approved):
Cool! Lgtm with the `getInputStylePalette` => dumber `inputStylePalette` change.

PR #1095, opened 2024-12-11
feat: adding icon bg color param

JonnCharpentier:


src/components/Icon.tsx
```diff
@@ -11,7 +11,7 @@ export interface IconProps extends AriaAttributes, DOMProps {
   /** The size of the icon in increments, i.e. 1 == 8px, default is 3 == 24px. */
   inc?: number;
   /** Styles overrides */
-  xss?: Xss<Margin | "visibility" | "flexShrink">;
+  xss?: Xss<Margin | "visibility" | "flexShrink" | "backgroundColor">;
```
stephenh, line 14: I think a bgColor: Palette prop would be better/more of a guardrail until we really have a strong need for "any random color".

stephenh (approved):

PR #1096, opened 2024-12-13
feat: New Caution button variant

JonnCharpentier:


stephenh (approved):
Nice! Ngl I'm surprised at how small the diff is; kudos @bdow !

PR #1099, opened 2024-12-20
feat: [Sc-63907] Skip scrollToIndex if row is already in view

blambillotte:
* Example use case on the schedule where the `scrollIntoView` is helpful when moving a task to a date that is outside of the current viewport, but we don't want it to move if it's still visible after a given change. https://www.loom.com/share/a58929dc5354403f855628e60212c1e5
* This somewhat hand-rolled implementation is due to virutoso not exposing this option, some context on this thread where the preferred option is to use an intersection observer, but that seems like overkill given there may be 50+ rows rendered on a large screen https://github.com/petyosi/react-virtuoso/issues/118

### DIFF
--- a/src/components/Table/types.ts
+++ b/src/components/Table/types.ts
@@ -22,8 +22,6 @@ export type GridTableScrollOptions =
        * How to position the row in the viewport
        */
       align?: "start" | "center" | "end";
-      /** Determines if the row should be re-scrolled to even if it's already in view (scrolled to top) */
-      forceRescroll?: boolean;
     };
 
 /**
--- a/src/components/Table/types.ts
+++ b/src/components/Table/types.ts
@@ -22,6 +22,8 @@ export type GridTableScrollOptions =
        * How to position the row in the viewport
        */
       align?: "start" | "center" | "end";
+      /** Determines if the row should be re-scrolled to even if it's already in view (scrolled to top) */
+      forceRescroll?: boolean;
     };
 
 /**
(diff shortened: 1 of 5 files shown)


src/components/Table/types.ts
```diff
@@ -22,6 +22,8 @@ export type GridTableScrollOptions =
        * How to position the row in the viewport
        */
       align?: "start" | "center" | "end";
+      /** Determines if the row should be re-scrolled to even if it's already in view (scrolled to top) */
+      forceRescroll?: boolean;
```
stephenh, line 26: @blambillotte wdyt about just having this always be true, and not exposed as an option?

I.e. cue my usual soapbox about beam having as-few-props-as-possible and ideally we just provide the "best" / same behavior all for all pages.

blambillotte: yeah my initial thought was to just swap the default, but didn't get any engagement on that question in [this thread](https://homeboundinc.slack.com/archives/C01JAFC0RAN/p1734732420998599) ha
Happy to make the "don't jump"/new behavior the default and just not have a prop for it if everyone is good with that

stephenh: Yep, I think between "same behavior everywhere" bias + "no one voted / seemed to care in that thread", let's move ahead w/no customization.

blambillotte: Cool okay - updated!

stephenh (approved):

PR #1101, opened 2025-01-09
fix: SC-64282: Add back in early return for onSearch/debounce

kbesingeryeh:
A bug was reported on the `ChangeLog` page, where the project filter search was not working
[SC-64282](https://app.shortcut.com/homebound-team/story/64282/projects-filter-not-working-properly-on-change-log)

When I added back in the `debouncedSearch` early return, the project filter was fixed, but it caused a bug with the regular select field per the slack threads below and PRs
* Slack Thread: https://homeboundinc.slack.com/archives/G01QEA5CXL1/p1734450733938479 
* PRs:
   * https://github.com/homebound-team/beam/pull/1017/files#diff-4cfc44d3175cb7b1fc2f316109423cede94d022b834134087a6a15bdbe4f296eL280
   * https://github.com/homebound-team/beam/pull/1098/files

I still think this needs an early return, so I changed it to look at the `onSearch` prop and do an early return from that check.

I tested on front-end and everything looks ok on the Change Log Page

LOOM: https://www.loom.com/share/eefefa2904fb4d70b5515168cbad294a

### DIFF
--- a/src/inputs/internal/ComboBoxBase.tsx
+++ b/src/inputs/internal/ComboBoxBase.tsx
@@ -327,7 +327,6 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
   // Call on search callback when the user types in the input field
   useEffect(() => {
     onSearch?.(debouncedSearch ?? "");
-    console.log("debouncedSearch", debouncedSearch);
   }, [onSearch, debouncedSearch]);
 
   // For the most part, the returned props contain `aria-*` and `id` attributes for accessibility purposes.
--- a/src/inputs/internal/ComboBoxBase.tsx
+++ b/src/inputs/internal/ComboBoxBase.tsx
@@ -306,6 +306,7 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
 
   // Reset inputValue when closed or selected changes
   useEffect(() => {
+    if (debouncedSearch) return;
     if (state.isOpen && multiselect) {
       // While the multiselect is open, let the user keep typing
       setFieldState((prevState) => ({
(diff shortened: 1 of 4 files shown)


src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -306,6 +306,8 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
 
   // Reset inputValue when closed or selected changes
   useEffect(() => {
+    // if we are using the onSearch functionality with debounceSearch, we don't want to reset the input value
+    if (onSearch && debouncedSearch) return;
```
apattersonATX-HB, line 310: question: Do we need to check against `debouncedSearch` too? `onSearch` should be enough?

apattersonATX-HB (approved):
Makes sense to me

src/inputs/internal/ComboBoxBase.tsx
kbesingeryeh: yes, good question! I originally had it without `debounced`, but when I was testing, I noticed after I made a selection and hit the "clear button" to clear the filter, it wouldn't reset so I added the extra check

stephenh, line 310: Fwiw @apattersonATX-HB / @kbesingeryeh we've had enough back/forth/churn on this issue, that it would be amazing to have some unit test coverage in this PR.

Looking back at Brandon's last PR that was futzing with this, it was in ~April 2024 when "beam was dead" meant we skimped on test coverage :grimacing: 

That said, I'm sure this edge case is trickier than usual to get covered/exposed in a unit test -- but maybe you could give it a shot? Disclaimer I haven't looked at the existing combo box/select field tests, and whether we already have some "intricate" tests (i.e. key here, focus there, see the list do x/y/z) to copy/paste as a starting point--hopefully! :crossed_fingers:

kbesingeryeh: yes, I can work on adding a test!

apattersonATX-HB: Thanks for the extra context @stephenh and thanks for the tests @kbesingeryeh !

Beam was dead, but it has risen!!!!

kbesingeryeh: @stephenh @apattersonATX-HB I spent some time trying to add a test and it gets tricky as the issue lies in how the internal state of `ComboBoxBase` was resetting the input when we didn't have an early return from the `debounce`. We don't currently have any tests for `ComboBoxBase` - only the components that use it as a base.

I added a story for the search functionality in a `multi-filter` and a test for the search functionality, but I don't think those would have caught this one-off issue with the user typing and internal state resetting. Let me know if this is enough otherwise I may need some additional help

apattersonATX-HB: Ah I think the test you need to setup involves mocking a gql call that'll update the filter options? Or do a search more? If we don't have that test setup in place now (@stephenh do we?) just go ahead and merge this in and add a follow up PR since the setup might be a pain 🙏🏽

PR #1103, opened 2025-01-14
fix: Add test and update early return

kbesingeryeh:
[SC-64494](https://app.shortcut.com/homebound-team/story/64282/projects-filter-not-working-properly-on-change-log#activity-64494)

* Projects filter on `Change Log` page was not working properly and kept resetting itself


src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -307,7 +307,10 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
   // Reset inputValue when closed or selected changes
   useEffect(() => {
     if (state.isOpen && multiselect) {
-      // While the multiselect is open, let the user keep typing
+      // While the multiselect is open, let the user keep typing unless they are using a debounce, then we need to
+      // early return to avoid resetting the input value
+      if (debouncedSearch) return;
```
kbesingeryeh, line 312: @stephenh @apattersonATX-HB 

let me know what ya'll think. I added back in the `early return` but inside of this `if` block since we need to be able to clear it out later

apattersonATX-HB: It's weird we're only resetting the input in multiselect cases so i'd like to also update our [SelectFieldProps](https://github.com/homebound-team/beam/blob/50d95108c3fa83acb9092468b364ce74ab220ac6/src/inputs/SelectField.tsx#L7-L7) type def to also omit `onSearch` and see what it breaks, if anything, in BP so we're covering our bases there.

That aside, this is fine imo. `onSearch` callback only gets called while searching which only happens when the field is open anyway 🤔

apattersonATX-HB: nvm on removing onSearch, i thought part of the issue was devs using onSearch to load up options instead of lazy loading them via `options` params but looks like that isnt setup to include `searchValue` in the cb

src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -320,11 +323,13 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
         inputValue: getInputValue(selectedOptions, getOptionLabel, multiselect, nothingSelectedText, isReadOnly),
       }));
     }
-  }, [state.isOpen, selectedOptions, getOptionLabel, multiselect, nothingSelectedText, isReadOnly]);
+  }, [state.isOpen, selectedOptions, getOptionLabel, multiselect, nothingSelectedText, isReadOnly, debouncedSearch]);
 
   // Call on search callback when the user types in the input field
   useEffect(() => {
-    onSearch?.(debouncedSearch ?? "");
+    if (onSearch && debouncedSearch) {
```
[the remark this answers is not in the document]
kbesingeryeh: I was playing around with different edge cases, but it doesn't make change the behavior if I define it before I call `onSearch` or not - I can remove it though !

src/inputs/internal/ComboBoxBase.tsx
stephenh: Okay, I had to open up the code locally and stare at this for a bit :sweat_smile: but I think your change makes sense @kbesingeryeh ; my only suggestion might be to put `&& !debouncedSearch` into the `if state.isOpen && multiselect && !debouncedSearch` and update the comment to a very pedantic very like:

```
// When a single-select is open, with "inputValue=Blue", when "Blue" is actually
// selected (our value changes), the box will close immediately, so the inputValue
// will get reset in the `!isOpen` case.
//
// When a multi-select is open, with "inputValue=Blue", when "Blue" is actually
// selected (our value changes), we leave the box open, and we want to reset
// the "inputValue=(empty) so the user can now type "inputType=Green".
// 
// ...unless the multi-select is using onSearch, in which case we want to let them
// only explicitly trigger a new `onSearch`, i.e. they might want to keep typing
// "Blueish".
```

@kbesingeryeh do I have the scenario right with the ^ comment?

Fwiw, now that I've written it out, if this description is correct, do we know why we want "if you've typed Blue, and pick something ...sometimes we clear (no onSearch)/sometimes we don't (yes onSearch)"?

It doesn't seem like "this control is/is not using a backend-driven search" should necessarily change it's UX behavior to the user, about when we do/do not reset the input value/search value? :thinking:

kbesingeryeh: yes, the above scenario sounds right... I was not close with the original ask of adding the `onSearch`, so I'm not 100% sure, I think it should only clear when the user either clears all the filters or unselects the project they have searched and selected for

src/inputs/internal/ComboBoxBase.tsx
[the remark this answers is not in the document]
kbesingeryeh: @apattersonATX-HB @stephenh I added another test in `SelectField.test.tsx`. Is this good for approval?

stephenh (approved):

src/inputs/internal/ComboBoxBase.tsx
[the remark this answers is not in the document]
stephenh: Yep, lgtm!

PR #1107, opened 2025-01-30
fix: Handle "Select All" when only unselectable child rows are present

pwillikins:
We've encountered an issue where users cannot select or deselect all rows when a table contains only un-selectable rows, whether grouped or not. For example, if a table includes a header row, a group row, and a single un-selectable child row, selecting all leaves the group unselected and the header in an indeterminate state. This fix ensures that the header and/or groups are marked as `checked` even when only un-selectable rows are present.


pwillikins:
Closing to re open with semantic commit/title.

PR #1115, opened 2025-02-13
feat: add undo icon

zovington:


src/components/Icon.tsx
```diff
@@ -849,6 +849,9 @@ export const Icons = {
   pin: (
     <path d="M12 22L13 20V17H18C18.553 17 19 16.553 19 16V14.414C19 13.888 18.786 13.372 18.414 13L17 11.586V8C17.553 8 18 7.553 18 7V4C18 2.897 17.103 2 16 2H8C6.897 2 6 2.897 6 4V7C6 7.553 6.448 8 7 8V11.586L5.586 13C5.213 13.372 5 13.888 5 14.414V16C5 16.553 5.448 17 6 17H11V20L12 22ZM8 4H16V6H8V4ZM7 14.414L8.707 12.707C8.895 12.52 9 12.266 9 12V8H15V12C15 12.266 15.105 12.52 15.293 12.707L17 14.414V15H7V14.414Z" />
   ),
+  undo: (
```
blambillotte, line 852: nit: is it possible to have design resize this icon to be better sized within the 24x24 standard?
image
Looks like this is just taken from the `undoCircle` interior, but then isn't really sized or positioned quite right on its own?

zovington: Great call out, after chatting with the designer I'm going to close this to just use undoCircle

PR #1118, opened 2025-03-06
feat: Move to esm.

stephenh:
This updates Beam to be a dual CJS/ESM package.

Ideally we'd just go to ESM-only, since vite is perfectly capable (and prefers) pulling in ESM packages, even for downstream projects that are technically still commonjs (i.e. internal-frontend), *but* until downstream test suites also execute in ESM (*cough* Jest *cough*), we still need a CommonJS set of artifacts as well.

In theory there are not any "*hard* breaking changes", but:

* Any `@homebound/beam/dist/...something...` import will now hard break (and was not really supported before anyway) because we use `tsup` to create bundled files
* An exception is `@homebound/beam/rtl` is now a blessed alternate entry point/bundle that includes our testing utilities
* The `@homebound/beam/dist/utils/sb.tsx` import didn't really have anything in it, so is no longer public
* Because `tsup` is doing bundling, it also does path rewriting, so we can remove the `ts-patch` / `transform-paths` tsc plugin


src/utils/rtl.tsx
```diff
@@ -1,35 +1,10 @@
...
-  RenderResult,
-  render as rtlRender,
-  Wrapper,
-} from "@homebound/rtl-utils";
-import { prettyDOM } from "@testing-library/react";
+import { RenderResult, render as rtlRender, Wrapper } from "@homebound/rtl-utils";
 import { ReactElement } from "react";
-import { BeamProvider } from "src/components";
-import { fail } from "src/utils/index";
-export {
```
stephenh, line 21: I split up "the rtl.tsx that Beam's own tests use" (this file) and "rtlUtils.tsx the utilities that we bundle for downstream projects to use" (the next file).

Previously they were combined and should not have been.

src/utils/rtlUtils.tsx
```diff
@@ -0,0 +1,262 @@
...
+  type as _type,
+  typeAndWait as _typeAndWait,
+  wait as _wait,
+  allowAndWaitForAsyncBehavior,
+  RenderResult,
+} from "@homebound/rtl-utils";
+import { prettyDOM } from "@testing-library/react";
+import { fail } from "src/utils/index";
+
+export {
```
stephenh, line 17: This is the code from `rtl.tsx` that is explicitly packaged/exposed as `@homebound/beam/rtl` for usage in downstream tests

src/utils/withTestMock.tsx
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * This defaults to true in test suites, but can be disabled if you
+ * _really_ want to test against the real component.
+ */
+export function setUseMockComponents(value: boolean) {
+  useMock = value;
+}
+
+// HOC that conditionally swaps components
+export function withTestMock<P extends object>(Component: ComponentType<P>, MockComponent: ComponentType<P>) {
```
stephenh, line 16: The bundling means our "trix doesn't work in jsdom" mocking in `internal-frontend` doesn't work anymore:

```
-jest.mock("@homebound/beam/dist/inputs/RichTextField", () => {
-  return require("@homebound/beam/dist/inputs/RichTextField.mock");
-});
```

So had to move the "swap test component" mocking into beam proper.

package.json
```diff
@@ -14,23 +13,34 @@
     "!dist/**/*.{stories,test}.*",
     "!dist/setupTests.*"
   ],
-  "engines": {
-    "node": "~20.9.0"
```
blambillotte, line 18: Question: is it not still helpful to specify the node version here? Thought this was needed to get compatibility warnings when adding new packages?

blambillotte (approved):
🎉

package.json
stephenh: Ah yeah, I had removed it while just futzing around getting things to work; will add back with the bumped version

blimmer: Do we require all end-users to be on Node 22 (exactly)? It feels like the `engines` property we're specifying is too restrictive for a public package. Typically, you'll see something like this:

```json
"engines": {
  "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
}
```

PR #1122, opened 2025-03-10
feat: [SC-67743] BoundForm standardized component

blambillotte:
Spike to explore the concept of a `BoundForm` which codifies how a standard `form-state` "bound" form should be presented as expressed via "config" rather than relying on the correct combination of `BoundFoo` input and layout components. 

Standard form layout designs: https://www.figma.com/design/OfJJm1CXHRGkrHNcPu2grY/2025-UX-Standardization?node-id=513-19012&t=UZAEe6GPqDn3sXsT-0

## Simple Example
With the exception of more complex inputs like select fields that require passing in loaded select options/etc, simple inputs don't require any props to be passed to their input functions as the `FormState` `field` will be automatically passed to the `BoundFoo` input component based on the field name key. 

```tsx
<BoundFormComponent
   inputRows={[
      { firstName: textField(), middleInitial: textField(), lastName: textField() },
      { bio: textAreaField() },
   ]}
   formState={formState}
/>

```

image

Additionally, the form input functions are simply wrapping the `Bound_` component equivalents and accept all of the same props. However we can selectively `Omit<...>` unsupported props to drive further consistency. 

```tsx
const inputConfig: BoundFormInputConfig<AuthorInput> = [
      { firstName: textField(), middleInitial: textField(), lastName: textField() },
      { bio: textAreaField() },
      // We can support any custom JSX node, TODO to come up with a better example
      // { height: <div>Example custom component</div> },
      {
        favoriteSport: selectField({
          options: sportsOptions,
          // Opportunity here to not require label & value getters if using conventional `HasNameAndId`
          getOptionValue: (o) => o.id,
        }),
        favoriteColors: multiSelectField({
          options: colorOptions,
          getOptionValue: (o) => o.id,
        }),
        heightInInches: numberField({ label: "Height (in inches)" }),
        birthday: dateField(),
      },
      { isAvailable: checkboxField({ label: "Is Retired" }) },
];
```

```tsx
<BoundFormComponent inputConfig={inputConfig} formState={formState} />

```

### Additional Advantages
There's also an opportunity to bake in UX nice-to-haves that generally aren't worth the time on each feature like improved loading states. Because we express the shape of the form via the config, we can match the loading state skeleton loaders to the form layout. [Demo](https://60466f7d43c37c0021788867-foluwyvnfz.chromatic.com/?path=/story/forms-boundform--loading-bound-form)


src/forms/BoundForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+            // Otherwise, render the custom node directly
+            return <Fragment key={key.toString()}>{fieldFnOrCustomNode}</Fragment>;
+          })}
+        </FieldGroup>
+      ))}
+    </>
+  );
+}
+
+export function selectField<O, V extends Value>(props: Omit<BoundSelectFieldProps<O, V>, "field">) {
```
blambillotte, line 74: Would need to make wrappers for the existing `BoundFoo` components out there.

src/forms/BoundForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+// Potential TODO: add type overloads for the different HasIdIsh/HasNameIsh combinations, maybe there's a generic way to introspect those types?
+export function boundSelectField<O, V extends Value>(props: Omit<BoundSelectFieldProps<O, V>, KeysToOmit>) {
+  return (field: FieldState<any>): BoundFieldInputFnReturn => ({
+    component: <BoundSelectField field={field} {...props} />,
+    minWith: "200px",
+  });
+}
+
+export function boundMultiSelectField<O, V extends Value>(props: Omit<BoundMultiSelectFieldProps<O, V>, KeysToOmit>) {
```
blambillotte, line 123: One option for these field wrappers is to export them in a config object that would give us namspacing something like `boundFormField.select` ? Or maybe just call these `boundFormFoo` -> `boundFormMultiSelectField` ?

src/forms/BoundForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+ *    <BoundFormComponent
+        inputRows={[
+          { firstName: boundTextField(), middleInitial: boundTextField(), lastName: boundTextField() },
+          { bio: boundTextAreaField() },
+        ]}
+        formState={formState}
+      />
+ * ```
+ */
+export function BoundForm<F>(props: BoundFormProps<F>) {
```
blambillotte, line 56: open to other naming suggestions here too since the expectation is a "page" level component may render multiple of these to split a form into sections. Could call it `BoundFormFields` or something like that to make it clear it doesn't have to be the whole form?

src/forms/BoundForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+  return (field: FieldState<any>): BoundFieldInputFnReturn => ({
+    component: <BoundSelectField field={field} {...props} />,
+    minWith: "200px",
+  });
+}
+
+export function boundMultiSelectField<O, V extends Value>(props: Omit<BoundMultiSelectFieldProps<O, V>, KeysToOmit>) {
+  return (field: FieldState<any>): BoundFieldInputFnReturn => ({
+    component: <BoundMultiSelectField field={field} {...props} />,
+    minWith: "200px",
```
blambillotte, line 126: Probably will need to add a `maxWidth` as well for the naturally shorter inputs like checkboxes and switches? Though they could have long names so we could optionally just remove the `flex-grow` from them so they just size themselves vs. evenly dividing their width
image

src/forms/BoundForm.tsx
stephenh: I was kinda thinking of `...Layout` as the name for some of these new layout components, i.e. `DetailLayout`, `FormLayout`, etc...

Historically, the "bound" term has meant like "data binding", and I almost wouldn't mind `BoundForm` / `FormLayout` being as decoupled as possible from form-state, i.e. something like:

```tsx
      <FormLayout
        rows={[
          <BoundTextField field=fs.firstName>,
          <BoundCheckboxField field=fs.isAvailable>,
          <div>other random jsx that invariably designers will ask for<div>,
        ]}
      />,
```

That said, I could see passing in `FormLayout formState={fs}` being useful for things like "is the form valid" / "driving the submit button" / that sort of thing...

Disclaimer I've not actually tried the `rows={[<BoundTextField...>]}` syntax, but I think it would work...

Dunno, wdty about the ^ strawman, that ironically-but-purposefully is more decoupled from "everything I'm laying out shalt be a form-state-driven field"?

src/forms/BoundForm.tsx
```diff
@@ -28,8 +28,13 @@ import { FormLines } from "./FormLines";
 type BoundFieldInputFnReturn = { component: ReactNode; minWith: string };
 type BoundFieldInputFn<F> = (field: ObjectState<F>[keyof F]) => BoundFieldInputFnReturn;
 
-type BoundFormRowInputs<F> = {
-  [K in keyof F]: BoundFieldInputFn<F> | ReactNode;
+const reactNodePrefix = "reactNode" as const;
+type CustomReactNodeKey = `${typeof reactNodePrefix}${string}`;
```
TylerR909, line 32: I suspect you'd be able to create something like

```ts
type CustomReactNodeKey<T extends string> = `${typeof reactNodePrefix}${T}`;
```

then maybe that opens the door for BoundFormRowInputs to generate the type

```ts
// given some config `{ firstName: string }`
// inputs become: `{ firstName: BoundFieldInput<F>, firstNameReactNode?: ReactNode }`
```

blambillotte: Sounds good, ended up giving that a try and having the enumerated list does help with discoverability of the feature, given we still allow any `reactNode*` prefixed key as valid for when we need custom JSX that doesn't live in the form at all. 

image


image

src/forms/BoundForm.tsx
blambillotte: Addressed this by changing the default switch style to `inline` per design feedback as well as adding a `min-content` rule for allowing variable sized inputs to size themselves vs. going with a hardcoded min. 
https://github.com/homebound-team/beam/pull/1122/commits/12ce84de87541deaecafb7f0165a2b0cb798fdd6

src/forms/BoundForm.tsx
```diff
@@ -0,0 +1,260 @@
...
+  return (field: FieldState<any>): BoundFieldInputFnReturn => ({
+    component: <BoundSelectField field={field} {...props} />,
+    minWith: "200px",
+  });
+}
+
+export function boundMultiSelectField<O, V extends Value>(props: Omit<BoundMultiSelectFieldProps<O, V>, KeysToOmit>) {
+  return (field: FieldState<any>): BoundFieldInputFnReturn => ({
+    component: <BoundMultiSelectField field={field} {...props} />,
+    minWith: "200px",
```
stephenh, line 149: Is the `minWith` vs. `minWidth` a typo?

src/forms/BoundForm.tsx
```diff
@@ -0,0 +1,260 @@
...
+type TReactNodePrefix<S extends string> = `${typeof reactNodePrefix}${CapitalizeFirstLetter<S>}`;
+
+type CustomReactNodeKey = `${typeof reactNodePrefix}${string}`;
+
+type BoundFormRowInputs<F> = Partial<{
+  [K in keyof F]: BoundFieldInputFn<F>;
+}> & {
+  [K in CustomReactNodeKey]: ReactNode;
+} & {
+  [K in keyof F as TReactNodePrefix<K & string>]: ReactNode;
```
stephenh, line 46: If both `K in CustomReactNodeKey` and `K in keyof F as ...` both map to `ReactNode`s, isn't the first `K in CustomReactNodeKey` a superset of the the 2nd, i.e. we probably don't need this line 46, because anything the user would do like `reactNodeFirstName` (that "matches Author.firstName") already matched on `reactNode...whatever you want..." anyway?

stephenh: Oh, I thin you're saying in the other comment that it helps discoverability...makes sense.

stephenh (approved):

src/forms/BoundForm.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+describe("BoundForm", () => {
+  it("renders rows of fields", async () => {
+    // Given a formState with configured fields
+    const formState = createObjectState(formConfig, { isAvailable: true, firstName: "John", lastName: "Doe" });
+
+    // When rendered with two rows of inputs
+    const r = await render(
+      <BoundForm
+        formState={formState}
+        inputRows={[{ firstName: boundTextField(), lastName: boundTextField() }, { isAvailable: boundCheckboxField() }]}
```
stephenh, line 21: Maybe call `inputRows` just `rows`?

src/forms/BoundForm.tsx
blambillotte: yeah exactly, got some good input from Tyler on that above - wish there was a way to tell Intellisense that `reactNode*` is a valid option from the list vs it being a missing option that happens to not type error.

src/forms/BoundForm.tsx
blambillotte: 🤦

src/forms/BoundForm.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+  radioGroupExample?: string | null;
+  richTextExample?: string | null;
+  switchFieldExample1?: boolean | null;
+  switchFieldExample2?: boolean | null;
+  toggleChipGroupField?: string[] | null;
+  treeSelectExample?: string[] | null;
+};
+
+const inputConfig: BoundFormInputConfig<AuthorInput> = [
+  { firstName: boundTextField(), middleInitial: boundTextField(), lastName: boundTextField() },
```
stephenh, line 177: @blambillotte I was trying to articulate what I like/don't like about using "config is a map of input key to UI element"...

I was trying to think, well, "why does Filters use this approach today?" and I'm thinking it's mostly because `Filters` manages both the "filter state / filter input" (crafting the `AuthorFilter` to put on the wire) as well as the "filter layout" (where the UI widgets go).

For our forms, we have a slightly different context because form-state is already handling the "form state / form input" crafting, i.e. creating the `AuthorInput` to put on the wire.

So that convention of "ofc our form(state) config is keyed by / matches the shape of `AuthorInput`" works well for `useFormState` / `FormConfig` (as with `Filters`).

But for `BoundForm` / `FormLayout`, imo that convention of "the layout config is based around ...keys of the `AuthorInput`..." is probably less necessary, and it can be a little bit more "you just want (blobs of whatever JSX) laid out as a form".

So, I dunno, I don't feel strongly enough about this to push :-1: on the `firstName: ...` key, but still wanted to type out, for my own thinking, why I probably wouldn't have gone that way as the initial API/DSL.

Happy to keep this approach and see how it goes!

blambillotte: Ah okay fair enough - I think I'm good to give this API a try and see what we learn with the first implementation

stephenh: Cool, sgtm!

PR #1123, opened 2025-03-12
feat: [SC-67744] FormPageLayout Component

blambillotte:
Scaffolds the `FormPageLayout` component including:
* Dynamic `BoundForm` sections based on config
* Standardized header with "action" buttons, breadcrumbs and title
* Grid layout placeholders for side `Nav` and `SideBar` components to be worked on in another ticket

image

## Example form from `internal-frontend` conversion

image

image


src/components/Layout/PageHeaderBreadcrumbs.tsx
```diff
@@ -0,0 +1,25 @@
+import { Fragment } from "react";
```
blambillotte, line 1: This component was mostly copied from blueprint, but removes the `xss` option and swaps the `chevronRight` icon for `/` : https://github.com/homebound-team/internal-frontend/blob/main/src/routes/layout/PageHeader.tsx#L105-L134

src/forms/BoundForm.tsx
```diff
@@ -76,7 +76,7 @@ export function BoundForm<F>(props: BoundFormProps<F>) {
 
   return (
     <div {...tid}>
-      <FormLines labelSuffix={{ required: "*" }} width="full" gap={4}>
+      <FormLines labelSuffix={{ required: "*" }} width="full" gap={3.5}>
```
blambillotte, line 79: Given this `labelSuffix` is now the default, does it make sense to remove here?

chr1sjf0x: I'd think so 🤔

src/components/Layout/Layout.stories.tsx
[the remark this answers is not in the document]
blambillotte: Ah so I just lumped my new component into this general `Layout` story that was existing - any issues with that?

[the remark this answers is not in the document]
chr1sjf0x: Sure, that' fine 👍

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,175 @@
...
+      ))}
+    </article>
+  );
+}
+
+function LeftNav<F>(props: Pick<FormPageLayoutProps<F>, "formSections">) {
+  const tids = useTestIds(props);
+
+  return (
+    <aside css={Css.gr(2).gc("2 / 3").sticky.topPx(headerHeightPx).px3.df.fdc.gap1.$} {...tids}>
```
TylerR909, line 142: til `<aside />` exists

src/components/Layout/Layout.stories.tsx
```diff
@@ -150,6 +164,83 @@ function ExamplePageComponent() {
...
+  lastName: { type: "value", rules: [required] },
+  birthday: { type: "value", rules: [required] },
+  heightInInches: { type: "value" },
+  favoriteSport: { type: "value" },
+  favoriteColors: { type: "value" },
+  bio: { type: "value" },
+  isAvailable: { type: "value" },
+  favoriteShapes: { type: "value" },
+};
+
```
TylerR909, line 243: I wish we could start assuming `type: "value"`. With the _rare_ exception of some `list` these appear to be 99% of all of our form inputs. Lots of boilerplate.

TylerR909 (approved):
Cool!

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,175 @@
...
+ */
+const headerHeightPx = 120;
+
+export function FormPageLayout<F>(props: FormPageLayoutProps<F>) {
+  const { formSections, formState } = props;
+
+  const tids = useTestIds(props, "formPageLayout");
+
+  const gridColumns =
+    "minMax(0, auto) minMax(100px, 250px) minMax(350px, 1000px) minMax(min-content, 300px) minMax(0, auto)";
```
stephenh, line 41: Lazy ask, but could you in-source comment names for each of these columns?

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  return (
+    // This page is `fixed` to the full screen to allow it to act as a full screen modal while content is mounted below
```
stephenh, line 44: "while content is mounted below"? can you clarify what other content this would be? is this like a table with 100 rows, and so opening this "page" to edit row 1, doesn't throw away the table itself/loosing your scroll position/etc?

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,175 @@
...
+    // Adding "align-items: start" allows "position: sticky" to work within a grid for the sidebars
+    <div
+      css={Css.fixed.top0.bottom0.left0.right0.z(1000).oya.bgWhite.dg.gtc(gridColumns).gtr("auto 1fr").cg3.ais.$}
+      {...tids}
+    >
+      <PageHeader {...props} {...tids.pageHeader} />
```
stephenh, line 50: I wouldn't mind pulling out the PageHeader-only props and only passing those...maybe we should have a `pageHeader` prop in `FormPageLayoutProps` instead of inlining/repeating its props?

stephenh: Oh, I didn't realize PageHeader was an internal helper function, and not a reused component. This makes more sense then.

src/components/Layout/Layout.stories.tsx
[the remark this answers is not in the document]
stephenh: Yeah, I think long-term `FooLayout` -> `FooLayout.stories.tsx` is probably better, matches the convention of `(component).stories.tsx`

src/components/Layout/Layout.stories.tsx
stephenh: Yep; that's what this was trying to fix:

https://github.com/homebound-team/form-state/blob/main/src/configBuilders.test.ts#L6

In general, the libraries I build often like to have "ASTs of their data", and I start out with "just let the user fill out that AST by hand" (they are "just POJOs" ==> easy in JS right?), but ASTs by their nature are not usually succinct, so adding a "builder DSL" at some point is a reasonable next-step/makes sense.

src/components/Layout/FormPageLayout.tsx
blambillotte: yeah exactly, will add more context to the comment

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,181 @@
...
+                />
+              )}
+              {submitAction && (
+                <Button
+                  label={submitAction.label}
+                  onClick={submitAction.onClick}
+                  variant="primary"
+                  disabled={!formState.valid || submitAction.disabled}
+                  tooltip={submitAction.tooltip}
+                  {...tids.submitAction}
```
bdow, line 108: **nit:** 👋 Super small thing, but I'm pretty sure the button will default the test-id to the value of the label. Right now it'll be `data-testid="submitAction"`. Otherwise, if the label was "Save", then it'd be `data-testid="save"` (if I recall correctly). Really minor, more of an "in case you didn't know".
Awesome to see forms to start having defined layouts. Looks great!
Hope you guys are doing well!!

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  // The grid columns are defined as: "left-gutter, left-nav, form-content, right-sidebar, right-gutter"
+  const gridColumns =
+    "minMax(0, auto) minMax(100px, 250px) minMax(350px, 1000px) minMax(min-content, 300px) minMax(0, auto)";
+
+  return (
+    // This page is `fixed` to the full screen to allow it to act as a full screen modal while content is mounted below
+    // since this layout will be replacing most superdrawers/sidebars, we keep the listing mounted below to preserve the users's
+    // scroll position & filters
+    // Adding "align-items: start" allows "position: sticky" to work within a grid for the sidebars
```
bdow, line 49: This one used to get me all the time. "WTF is this not sticking!??!" 😄

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const tids = useTestIds(props);
+
+  return (
+    <header css={Css.gr(1).gc("2 / 5").sticky.top0.hPx(headerHeightPx).bgWhite.z5.$} {...tids}>
+      <div css={Css.py2.px3.df.jcsb.aic.$}>
+        <div>
+          {breadCrumb && <PageHeaderBreadcrumbs breadcrumb={breadCrumb} />}
+          <h1 css={Css.xl3Sb.$} {...tids.pageTitle}>
+            {pageTitle}
+          </h1>
```
bdow, line 76: **question/thought:** Looks like we're moving away from auto-saving forms, is that a fair assumption? Though, does appear the submitAction is optional, so maybe not? Makes me wonder if it'd be worth slapping the `AutoSaveIndicator` component near the page title somewhere? 
I'm sure this pageTitle section will be getting expanded in the future, assuming implementers will need to add things like Tags or the like to suffix the title.

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,181 @@
...
+    // scroll position & filters
+    // Adding "align-items: start" allows "position: sticky" to work within a grid for the sidebars
+    <div
+      css={Css.fixed.top0.bottom0.left0.right0.z(1000).oya.bgWhite.dg.gtc(gridColumns).gtr("auto 1fr").cg3.ais.$}
+      {...tids}
+    >
+      <PageHeader {...props} {...tids.pageHeader} />
+      <LeftNav formSections={formSections} {...tids.nav} />
+      <FormSections formSections={formSections} formState={formState} {...tids} />
+      <SidebarContent />
```
bdow, line 57: **thought:** A bit of a UX observation; I don't know yet what the `SidebarContent` is going to include, but Blueprint had often been criticized for too much going on and users not understanding what they're expected to do on this page. The concern with the Sidebar baked into this component where we clearly want them to focus on filling out a form, is that we're adding to the user's cognitive load in a way that might distract.

One other note is that this sidebar may make it more difficult to be responsive, unless it stacks underneath the form. Not sure if that is worth caring about, but does feel like if we're building new layouts, it'd be good to have a responsive mindset going forward.

src/components/Layout/FormPageLayout.tsx
blambillotte: Good to hear from you Brandon! 
And yeah that's a good callout on the testid, probably better to let the label drive it 🤔

src/components/Layout/FormPageLayout.tsx
```diff
@@ -0,0 +1,181 @@
...
+const headerHeightPx = 120;
+
+function FormPageLayoutComponent<F>(props: FormPageLayoutProps<F>) {
+  const { formSections, formState } = props;
+
+  const tids = useTestIds(props, "formPageLayout");
+
+  // The grid columns are defined as: "left-gutter, left-nav, form-content, right-sidebar, right-gutter"
+  const gridColumns =
+    "minMax(0, auto) minMax(100px, 250px) minMax(350px, 1000px) minMax(min-content, 300px) minMax(0, auto)";
```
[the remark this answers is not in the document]
blambillotte: yeah good question - so I landed on this when we were struggling with the scroll bar positioning back when I was using the 1st grid row to fix the header in place, you'd then get a weird floating scroll bar for the scrollable content that was attached to the inner container vs acting like a full page scroll all the way to the right. 
The header is now `sticky` so the scrollable content uses the full page scroll so we might be able to do the `margin: auto` container setup now 🤔

src/components/Layout/FormPageLayout.tsx
blambillotte: Yeah all good points - the new sidebar inspiration is coming from Dribbble: https://dribbble.com/shots/25855464-Financial-Dashboard-Design

image

We have been thinking about responsive design with "down to ipad size" being the real goal - for example the underlying `BoundForm` component will gracefully wrap in the inputs as the form column gets squeezed: https://main--60466f7d43c37c0021788867.chromatic.com/?path=/story/forms-boundform--bound-form

I know the right panel has been a pain in terms of optimizing for wider screens, but users just like it too much to get rid of. The nice thing about these page templates is we can spend more time working out some of the responsive solutions like making the sidebar more of a position absolute flyover (like the superdrawer) when we don't have enough space, etc.

src/components/Layout/FormPageLayout.tsx
stephenh: Hm, yeah, I mean if "you really can't/shouldn't use this FormPageLayout w/o the onSave prop set", we should probably make it required.

Otherwise, the naive future me will see the "oh, onSave is optional? that's valid? I wonder what use cases this component serves that involves onSave not being set ... oh, it can't". :-) 

Granted, a few "..."-to-think-about are fine, but they have a tendency to stack up as codebases hang around, so any chance we can get to remove them is :+1:

PR #1124, opened 2025-03-17
feat: Add asterisk to required fields by default.

### LINKED ISSUE #1100: Add asterisk to the label of required fields
Beam has had infra to add "required field suffix" or "optional field suffix" for awhile now, but we've been too gun shy to enable it.

Seems like the majority is that "required fields get a star", so we should just roll that out as the default.

stephenh:
Fixes #1100.


stephenh:
@blambillotte yep! Definitely will scan through the chromatic diffs on the bump.

PR #1126, opened 2025-03-21
feat: [SC-67747] Sidebar nav for `FormPageLayout`

blambillotte:
Adds the sidebar nav links as specified in the figma with `scrollIntoView` and `IntersectionObserver` handling for highlighting the active section: https://www.figma.com/design/aWUE4pPeUTgrYZ4vaTYZQU/%E2%9C%A8Beam-Component-Library?node-id=48123-60336&t=Gqb23rq7wqesxWZ8-0

**Demo**
https://60466f7d43c37c0021788867-udguwfvmfs.chromatic.com/?path=/story/components-layout-formpagelayout--single-column

---

image


src/components/Layout/FormPageLayout.tsx
```diff
@@ -141,16 +169,69 @@ function FormSections<F>(props: Pick<FormPageLayoutProps<F>, "formSections" | "f
...
     </aside>
   );
 }
 
+// Use inset box shadow rather than thick border to avoid the button text reflowing when the border is applied
+const activeStyles = Css.baseBd.boxShadow(`inset 3px 0px 0 0px ${Palette.Blue600}`).$;
+const hoverStyles = Css.bgBlue50.baseBd.blue900.boxShadow(`inset 3px 0px 0 0px ${Palette.Blue900}`).$;
+const defaultFocusRingStyles = Css.relative.z2.bshFocus.$;
+
+function SectionNavLink<F>(props: { sectionWithRef: SectionWithRefs<F>; activeSection: string | null }) {
```
blambillotte, line 202: I originally started implementing this as a more generic `LeftNavLink` component, but it didn't really make sense on it's own since we don't have any examples of this type of link style outside this "scroll to section" concept 🤷 
Maybe we pull the internals out of this at a later point if we do?

src/components/Layout/FormPageLayout.tsx
```diff
@@ -38,6 +42,19 @@ function FormPageLayoutComponent<F>(props: FormPageLayoutProps<F>) {
...
+  // and the `IntersectionObserver` behaviors between the `LeftNav` and `FormSections` children
+  const sectionsWithRefs = useMemo(
+    () =>
+      formSections.map((section, id) => ({
+        section,
+        ref: createRef<HTMLElement>(),
+        // Unique key for each section to use in the observer
+        sectionKey: `section-${section.title ?? id}`,
+      })),
+    [formSections],
```
stephenh, line 55: Per prior conversations, it seems unlikely this `formSections` is ever going to be stable, but :shrug: also unclear how many re-renders the top-level FormPageLayout is really going to have anyway :thinking:

stephenh (approved):
Looks great! Love all the in-source comments

src/components/Layout/FormPageLayout.tsx
blambillotte: yeah I guess we'll see, don't really see why it would need to change 🤔 even some of the "poll interval" type components that are in the app header won't be parent components to this so I'm guessing not an issue

PR #1127, opened 2025-04-01
feat: [SC-67748] RightSidebar component

ksk5280:
Adds RightSidebar component that can be opened in the FormPageLayout.

### Design
[Figma](https://www.figma.com/design/OfJJm1CXHRGkrHNcPu2grY/2025-UX-Standardization?node-id=1029-14532&t=j6WSP5ZoPcsE59K1-0)

### Demo
[Storybook](https://60466f7d43c37c0021788867-grzawopecl.chromatic.com/?path=/story/components-layout-formpagelayout--form-page-layout)


src/components/RightSidebar.tsx
```diff
@@ -0,0 +1,58 @@
...
+              inc={3.5}
+            />
+          ))}
+        </div>
+      </div>
+      <div
+        css={{
+          ...Css.wPx(selectedIcon ? 380 : 0).relative.topPx(-48).z0.$,
+          transition: selectedIcon ? "transform 300ms ease-in" : "transform 300ms ease-out",
+          transform: selectedIcon ? "translateX(0)" : "translateX(100%)",
```
ksk5280, line 40: I attempted adding a sidebar transition. It only works when the sidebar is displayed but not when it's hidden because the sidebar content is hidden immediately when the x icon is clicked. I didn't quite figure this out within the timebox I set for myself, but please give input if you have a better way to set up this transition.

ksk5280: update this to use framer motion and it's working better now

src/components/IconButton.stories.tsx
```diff
@@ -121,8 +130,16 @@ export function IconButtonLink() {
...
-    <div css={{ "& button": args.contrast ? iconButtonContrastStylesHover : iconButtonStylesHover }}>
-      <IconButton {...args} />
+    <div
+      css={{
+        "& button": args.contrast
+          ? iconButtonContrastStylesHover
+          : args.circle
+            ? iconButtonCircleStylesHover
+            : iconButtonStylesHover,
+      }}
```
blambillotte, line 140: question: I'd expect the component to just own these styles, was it obvious when working on this why this story is overriding in this way?

src/components/IconButton.tsx
```diff
@@ -107,10 +121,15 @@ export function IconButton(props: IconButtonProps) {
 
 const defaultIconColor = Palette.Gray900;
 const contrastIconColor = Palette.White;
+const circleIconColor = Palette.Gray700;
 const iconButtonStylesReset = Css.bcTransparent.bss.bgTransparent.cursorPointer.outline0.dif.aic.jcc.transition.$;
 const iconButtonNormal = Css.hPx(28).wPx(28).br8.bw2.$;
 const iconButtonCompact = Css.hPx(18).wPx(18).br4.bw1.$;
+const iconButtonCircle = Css.br100.wPx(48).hPx(48).bcGray300.ba.bw1.df.jcc.aic.onHover.bgBlue100.bcBlue200.$;
```
blambillotte, line 128: nit: Can we move the `onHover` styles down to `iconButtonCircleStylesHover`?

src/components/IconButton.tsx
```diff
@@ -107,10 +121,15 @@ export function IconButton(props: IconButtonProps) {
...
 export const iconButtonStylesHover = Css.bgGray200.$;
 export const iconButtonContrastStylesHover = Css.bgGray700.$;
+export const iconButtonCircleStylesHover = Css.bgBlue100.bcBlue200.$;
```
blambillotte, line 131: Question: any reason these are not set via the `isHovered` above?

src/components/Layout/FormPageLayout.tsx
```diff
@@ -28,17 +29,18 @@ type FormPageLayoutProps<F> = {
...
+  rightSideBarContent?: SidebarContentProps[];
 };
 
 /** In order to make the multiple stacked sticky elements work (Header, then sidebar below) we need to set the header height.
  * The alternate solution was to have multiple defined grid areas for the fixed header vs body below, but this becomes more complex
  * to manage when adding in a max content width container for the page while keeping the page scrollbar to the far right of the page.
  * Rather than wrapping the page in a max-width div, we use "gutter" columns `minMax(0, auto)` that kick in when all other columns have met their max widths.
  */
-const headerHeightPx = 120;
+export const headerHeightPx = 120;
```
blambillotte, line 40: nit: Is the export stale now? Hopefully this can stay as a local implementation detail

src/components/RightSidebar.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const tid = useTestIds({}, "rightSidebar");
+  const width = 380;
+
+  return (
+    <>
+      <div css={Css.df.jcfe.relative.pr3.$}>
+        <div css={Css.df.gap2.z1.$}>
+          {content.map(({ icon }) => (
+            <IconButton
+              key={`${icon}-${selectedIcon}`}
```
blambillotte, line 28: thought: since `selectedIcon` isn't a stable value (expected to change on selection), I would think you wouldn't want to include that as the `key` here - was this an intentional addition vs just `icon`?

src/components/RightSidebar.tsx
```diff
@@ -0,0 +1,70 @@
...
+            transition={{ ease: "linear", duration: 0.2 }}
+            exit={{ transition: { ease: "linear", duration: 0.2 }, x: width }}
+          >
+            <div css={Css.wPx(width).relative.topPx(-48).z0.$}>
+              {selectedIcon && (
+                <>
+                  <div css={Css.absolute.leftPx(-24).top0.$}>
+                    <IconButton circle onClick={() => setSelectedIcon(undefined)} icon="x" inc={3.5} />
+                    {/* vertical line */}
+                    <div css={Css.absolute.topPx(48).leftPx(23).h("calc(100vh - 168px)").wPx(1).bgGray300.$} />
```
blambillotte, line 57: Question: where does this `168px` number come from? I'm wondering if we can think of a way to size this based on the content container provided by `FormPageLayout`. It has access to its local `headerHeightPx` so might be less brittle defining via the page?

blambillotte: Or this could just be a `border-left` on the container of the content and it will be sized based on the content?

src/components/IconButton.stories.tsx
stephenh: I believe this is a storybook hack to force showing hover styles w/o actually triggering a hover with the `play` / etc. 

I.e. I saw quite a bit of these when futzing when emotion-vs-fela selectors, and the button was always "wrap the real component in a div" + "setup div > (whatever element tag): ...consts of hover styles i want to snapshot...".

I don't love it, but it is current practice...

src/components/Layout/FormPageLayout.tsx
```diff
@@ -28,17 +29,18 @@ type FormPageLayoutProps<F> = {
   submitAction?: ActionButtonProps;
   cancelAction?: ActionButtonProps;
   tertiaryAction?: ActionButtonProps;
+  rightSideBarContent?: SidebarContentProps[];
```
stephenh, line 32: Thought: maybe just call this `rightSideBar`, i.e. the `...Content` is kind of implied by being within the props API.

src/components/IconButton.stories.tsx
blambillotte: Oh right 🤔 - seems like maybe there are more options now that you can do the `play` interactions with elements: https://www.chromatic.com/docs/hoverfocus/

src/components/RightSidebar.tsx
ksk5280: It unsets the `active` state of the button. Without it, the button stays gray after you select a different icon.

src/components/Layout/FormPageLayout.tsx
ksk5280: good catch, I can remove the `export`

src/components/Layout/FormPageLayout.tsx
ksk5280: sounds good, I will update this

src/components/RightSidebar.tsx
blambillotte: Ah okay makes sense, can you add an in-source comment for future us about that?

src/components/IconButton.tsx
ksk5280: Initially, I thought that all the ternaries were confusing and it seemed more straightforward to add the `onHover` on the `iconButtonCircle`, but then I added this `iconButtonCircleStylesHover` to use in the story. I agree that it doesn't make sense to have this in 2 places, so I can add `iconButtonCircleStylesHover` to the `isHovered` instead.

src/components/RightSidebar.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const width = 380;
+
+  return (
+    <>
+      <AnimatePresence>
+        {selectedIcon === undefined && (
+          // Display the icons vertically
+          <motion.div
+            key="iconsWithoutSidebar"
+            animate={{ transition: { delay: 0.2, opacity: 1 } }}
```
ksk5280, line 29: this delay doesn't seem to do anything. was trying to add a delay so the icons don't show up right away when you close the sidebar

PR #1128, opened 2025-04-02
fix: Enable SubmitButton on new entities.

stephenh:
When duplicating an entity, we might have a form that is valid but also not dirty, b/c technically the user hasn't touched anything yet, but we've prefilled existing values, and still want want to allowing hitting Save.

### DIFF
--- a/src/forms/BoundToggleChipGroupField.tsx
+++ b/src/forms/BoundToggleChipGroupField.tsx
@@ -17,7 +17,16 @@ export function BoundToggleChipGroupField(props: BoundToggleChipGroupFieldProps)
   const testId = useTestIds(props, field.key);
   return (
     <Observer>
-      {() => <ToggleChipGroup label={label} values={field.value || []} onChange={onChange} {...testId} {...others} />}
+      {() => (
+        <ToggleChipGroup
+          label={label}
+          required={field.required}
+          values={field.value || []}
+          onChange={onChange}
+          {...testId}
+          {...others}
+        />
+      )}
     </Observer>
   );
 }
--- a/src/forms/FormStateApp.tsx
+++ b/src/forms/FormStateApp.tsx
@@ -22,29 +22,41 @@ import {
   FieldGroup,
   FormDivider,
   StaticField,
+  SubmitButton,
 } from "src/forms";
 import { BoundCheckboxGroupField } from "src/forms/BoundCheckboxGroupField";
 import { BoundTreeSelectField } from "src/forms/BoundTreeSelectField";
 import { FormLines } from "src/forms/FormLines";
-import { AuthorInput } from "src/forms/formStateDomain";
+import { AuthorInput, jan1 } from "src/forms/formStateDomain";
 import { useComputed } from "src/hooks";
 import { CheckboxGroupItemOption, NestedOption } from "src/inputs";
 import { HasIdAndName } from "src/types";
 
 export function FormStateApp() {
-  // Simulate getting the initial form state back from a server call
-  const queryResponse = useMemo(
-    () => ({
-      firstName: "a1",
-      books: [...Array(2)].map((_, i) => ({
-        id: String(i),
-        title: `b${i}`,
-        classification: { number: `10${i + 1}`, category: `Test Category ${i}` },
-      })),
-    }),
-    [],
-  );
   const [readOnly, setReadOnly] = useState(false);
+  const [isNew, setIsNew] = useState(false);
+  const [enableOn, setEnableOn] = useState<"valid" | "dirty">("dirty");
+  // Simulate getting the initial form state back from a server call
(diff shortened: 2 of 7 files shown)


src/forms/BoundToggleChipGroupField.tsx
```diff
@@ -17,7 +17,16 @@ export function BoundToggleChipGroupField(props: BoundToggleChipGroupFieldProps)
   const testId = useTestIds(props, field.key);
   return (
     <Observer>
-      {() => <ToggleChipGroup label={label} values={field.value || []} onChange={onChange} {...testId} {...others} />}
+      {() => (
+        <ToggleChipGroup
+          label={label}
+          required={field.required}
```
stephenh, line 23: Passing along `required` now, to get the `*` asterisk in the label

src/forms/FormStateApp.tsx
```diff
@@ -22,29 +22,40 @@ import {
...
 import { FormLines } from "src/forms/FormLines";
-import { AuthorInput } from "src/forms/formStateDomain";
+import { AuthorInput, jan1 } from "src/forms/formStateDomain";
 import { useComputed } from "src/hooks";
 import { CheckboxGroupItemOption, NestedOption } from "src/inputs";
 import { HasIdAndName } from "src/types";
 
 export function FormStateApp() {
-  // Simulate getting the initial form state back from a server call
-  const queryResponse = useMemo(
```
stephenh, line 36: This moved down a few lines, and has a `isNew` conditional, so reformatted, but it's largely the same

src/forms/FormStateApp.tsx
```diff
@@ -63,113 +74,75 @@ export function FormStateApp() {
...
-        },
-        { id: "g:4", name: "Action Comedy" },
-      ],
-    },
-    { id: "g:5", name: "Comedy", children: [{ id: "g:6", name: "Comedy Drama" }] },
-  ];
-
   return (
-    <Observer>
-      {() => (
```
stephenh, line 110: Removed the `Observer` wrapper since it was only used for the form buttons, which now use SubmitButton

src/forms/SubmitButton.test.tsx
```diff
@@ -4,35 +4,50 @@ import { jest } from "@jest/globals";
...
 
-    // But once we make it valid
+    // ANd once we make it valid
     act(() => author.firstName.set("f2"));
     // We can click and submit
     click(r.submit);
     expect(onClick).toHaveBeenCalled();
   });
 
-  it("disables if the form is dirty but overridden via prop", async () => {
```
stephenh, line 35: Moved this "disables when prop set" test down, so that the "When editing" and "when creating" tests are by each other

src/inputs/CheckboxGroup.tsx
```diff
@@ -20,6 +21,7 @@ export interface CheckboxGroupItemOption {
 
 export interface CheckboxGroupProps extends Pick<PresentationFieldProps, "labelStyle"> {
   label: string;
+  required?: boolean;
```
stephenh, line 24: Hooking up required to drive the * label

src/forms/SubmitButton.test.tsx
```diff
@@ -4,35 +4,50 @@ import { jest } from "@jest/globals";
...
     act(() => author.firstName.set(""));
-    expect(r.submit).toBeEnabled();
+    // Then we stay disabled
+    expect(r.submit).not.toBeEnabled();
     // Even though it's invalid, so clicking submit still won't do anything
     click(r.submit);
     expect(onClick).not.toHaveBeenCalled();
 
-    // But once we make it valid
+    // ANd once we make it valid
```
blambillotte, line 31:
```suggestion
    // And once we make it valid
```

src/forms/SubmitButton.tsx
```diff
@@ -12,16 +13,22 @@ export function SubmitButton<T>(props: SubmitButtonProps<T>) {
...
   return (
     <Button
       label={label}
-      disabled={disabled || !dirty}
+      disabled={disabled || !canSubmit}
       onClick={(e) => {
         // canSave will touch any not-yet-keyed-in fields to show errors
+        state.clicked = true;
         if (form.canSave()) {
           void onClick(e);
```
blambillotte, line 33: question: will we still get the loading behavior on the button if the potentially async `onClick` isn't returned?

blambillotte (approved):
Looks good! 
Is it worth dropping this into the `FormPageLayout` in this PR or should we make a follow up ticket?

stephenh:
@blambillotte ah yeah, good point about using the FormPageLayout for this test harness; I didn't initially think of that, b/c I typically use this for more "state-driven" behavior poking that layout things (as you can tell by the currently terrible layout :-D), but no reason it can't do both.

But yeah, will leave that for a separate PR

PR #1132, opened 2025-04-08
feat: Updates to `FormPageLayout` and `BoundForm`

blambillotte:
A couple of small changes I noticed while implementing a `FormPageLayout`
* Addresses feedback on making the `submitAction` prop required from here: https://github.com/homebound-team/beam/pull/1123#discussion_r2024891966
* Fixes a type issue where the mapped type for `reactNode*` keys were not made optional if an input type for the form was required
* Also allows the checkbox to show a disabled tooltip
* Swaps out the normal submit action button with `SubmitButton` for the new standard disabled state behavior.

### DIFF
--- a/src/forms/BoundForm.stories.tsx
+++ b/src/forms/BoundForm.stories.tsx
@@ -168,6 +168,7 @@ const genres: NestedOption<HasIdAndName>[] = [
 ];
 
 type AuthorInput = BaseAuthorInput & {
+  // iconCardSelection is required to test against a mapped type regression on `BoundFormRowInputs`
   iconCardSelection: boolean | null;
   iconCardGroupExample?: string[] | null;
   multiLineSelectExample?: string[] | null;
--- a/src/forms/BoundForm.stories.tsx
+++ b/src/forms/BoundForm.stories.tsx
@@ -98,7 +98,13 @@ export function LoadingBoundForm() {
     if (loadedData) return;
 
     setTimeout(() => {
-      setLoadedData({ firstName: "John", middleInitial: "C", lastName: "Doe", bio: "Some bio" });
+      setLoadedData({
+        firstName: "John",
+        middleInitial: "C",
+        lastName: "Doe",
+        bio: "Some bio",
+        iconCardSelection: true,
+      });
     }, 1000);
   }, [loadedData]);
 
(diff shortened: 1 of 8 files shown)


src/inputs/CheckboxBase.tsx
```diff
@@ -45,43 +47,48 @@ export function CheckboxBase(props: CheckboxBaseProps) {
...
-        <div css={Css.ml1.mtPx(-2).$}>
-          {label && <div css={{ ...labelStyles, ...(isDisabled && disabledColor) }}>{label}</div>}
-          {description && <div css={{ ...descStyles, ...(isDisabled && disabledColor) }}>{description}</div>}
-          {errorMsg && <ErrorMessage errorMsg={errorMsg} {...tid.errorMsg} />}
-          {helperText && <HelperText helperText={helperText} {...tid.helperText} />}
-        </div>
-      )}
-    </Tag>
-  );
+  return maybeTooltip({
```
blambillotte, line 58: use hide whitespace to remove most of this diff

src/forms/SubmitButton.tsx
```diff
@@ -1,6 +1,7 @@
 import { ObjectState } from "@homebound/form-state";
 import { useLocalObservable } from "mobx-react";
-import { Button, ButtonProps, useComputed } from "src";
```
blambillotte, line 3: `FormPageLayout.test` did not like `Button` imported in this way

src/forms/BoundForm.stories.tsx
```diff
@@ -162,7 +168,7 @@ const genres: NestedOption<HasIdAndName>[] = [
 ];
 
 type AuthorInput = BaseAuthorInput & {
-  iconCardSelection?: boolean | null;
```
blambillotte, line 165: While the codegened input types generally have all keys as optional, it's common for implementors to extend those types with known required keys in the form component. This was causing a type error since the `reactNode*` mapped type was not also wrapped in `Partial<>`

stephenh: Ah, so the lack of `?` is specifically to cover the mapped type regressing? That makes sense/is a great idea; maybe leave an in-source comment so a future maintainer doesn't wonder why its missing, and puts it back.

stephenh (approved):
:+1:

PR #1134, opened 2025-04-11
feat: Card component

ksk5280:
* Add Card component per [Figma Design](https://www.figma.com/design/1F6M3mCD9FAiXmpiJIrbx8/Cards?node-id=5012-476&p=f&m=dev)

* [Storybook Demo](https://60466f7d43c37c0021788867-yihzrybmyc.chromatic.com/?path=/story/components-card--plan-card)

image

image

image


src/components/TableCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+    imageFit = "contain",
+    type = "card",
+    bordered = false,
+    disabled: isDisabled = false,
+    buttonMenuItems,
+    tag,
+  } = props;
+  const { hoverProps, isHovered } = useHover({ isDisabled });
+  const isList = type === "list";
+  const imageSize = isList ? 96 : bordered ? 224 : 256;
```
blambillotte, line 43: Question: I'm wondering if we can avoid hardcoding these width values given the parent card already has a width set. Ideally the image container can just size itself accordingly to fill the remaining width up to the parent's padding so the browser can ensure we have the correct alignment for various screen resolutions, etc (though if we need to set a max-height for oddly shaped images not growing the height of the card that would make sense). 

For example, the hardcoded with of 224 is slightly too large and overflows into the parent container padding making it no longer aligned with the text below.
image

Removing the hardcoded with then fixes that
image

src/components/TableCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+type CardTag = {
+  text: string;
+  type?: TagType;
+};
+
+export interface TableCardProps {
+  title: string;
+  subtitle: string;
+  metadata: ReactNode;
+  image: string;
```
blambillotte, line 18: nit: poking around beam it seems like we tend to use `src` most often for this type of thing, and `imgSrc` when it could be ambiguous (like in this case)

src/components/TableCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+    <div css={styles} {...hoverProps}>
+      {/* Image */}
+      <div css={Css.wPx(imageSize).hPx(imageSize).ba.br8.bcGray300.overflow("hidden").df.asc.jsc.relative.$}>
+        {/* overlay on hover */}
+        <div
+          css={{
+            ...Css.absolute.top0.left0.w100.h100.add("opacity", 0).$,
+            ...(isHovered && !isList && imageHoverStyles),
+          }}
+        />
```
blambillotte, line 66: thought to take or leave: what you have is 👍 but it might be cool to explore if a CSS filter would do the trick here to give us fine grain control over the behavior and skip dealing with an absolute positioned div, etc. https://css-tricks.com/almanac/properties/f/filter/
Been around for a while now so have very good browser support finally

src/components/TableCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+type CardTag = {
+  text: string;
+  type?: TagType;
+};
+
+export interface TableCardProps {
+  title: string;
+  subtitle: string;
+  metadata: ReactNode;
```
blambillotte, line 17: nit: Maybe just call this `content` since it's just a block to put whatever details you want?

khalidwilliams, line 17: Thought: I wonder if there will be any use cases where we'll want to be able to just have a card with a title/subtitle/image. I haven't seen it in any designs but we could make the prop optional. That would also be an easy change down the line so fine to leave it as is now

src/components/TableCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+  tag?: CardTag;
+}
+
+export function TableCard(props: TableCardProps) {
+  const {
+    title,
+    subtitle,
+    metadata,
+    image,
+    imageFit = "contain",
```
blambillotte, line 34: Cool idea thinking about the image fit behavior - I'm wondering though how practical exposing this to the caller is though? Like the expectation is these cards are going to be used (probably as a list) to present user uploaded images of various sizes where the implementor won't be able to pre-optimize for what that image is going to be. Unless you have a specific use case in mind where a page will surely need `cover` image fit type?

src/components/TableCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+  // contain displays entire image, cover fills the space
+  imageFit?: "contain" | "cover";
+  type?: TableCardType;
+  bordered?: boolean;
+  disabled?: boolean;
+  buttonMenuItems?: MenuItem[];
+  tag?: CardTag;
+}
+
+export function TableCard(props: TableCardProps) {
```
blambillotte, line 28: I was wondering about the "Table" naming for this component in the Figma so added a comment for Oz - seems like these cards aren't actually related to tables?

src/components/TableCard.tsx
ksk5280: It seems like I can make this optional, but I'll double check.

`content` makes me think it's a more encompassing part of the component instead of just a small extra details section. What do you think about calling it something that has more of an indication that it just adds some extra information, like `details`, `detailContent`, `additionalContent`? idk...

src/components/TableCard.tsx
ksk5280: good callout on the name. I'll wait to see if Stephen has input on what to call this

src/components/TableCard.tsx
ksk5280: Yeah, good point. Obviously for items we wouldn't want part of the image cut off, but it seemed like maybe for the plan images they might want them to cover the whole area (from the design example), but perhaps it does make more sense to just leave it as contain to keep it simple. I can ask Oz about his expectation.

This is the plan image with contain vs cover. I was just trying to get something working to match the design.

image
image

src/components/TableCard.tsx
ksk5280: yes good catch! I can remove the hardcoded width

src/components/TableCard.tsx
ksk5280: thank you for suggestion! I was able to apply a brightness filter instead

src/components/TableCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+type CardTag = {
+  text: string;
+  type?: TagType;
+};
+
+export interface TableCardProps {
+  title: string;
+  subtitle: string;
+  detailContent?: ReactNode;
+  imgSrc: string;
```
ksk5280, line 18: @blambillotte @khalidwilliams 

I just noticed that I didn't handle the empty state. I just wanted to double check, but I suppose that I should update this card to handle that state as well?

image

bdow:
Hey! Just passing through. Hope you and family are doing well, @ksk5280!
Question: What makes this a **Table**Card? I don't really see anything table-esque about it. I'd push to just call it a "Card".

src/components/TableCard.tsx
ksk5280: I renamed the component to `Card` instead of `TableCard`

bdow:
> Hey! Just passing through. Hope you and family are doing well, @ksk5280! Question: What makes this a **Table**Card? I don't really see anything table-esque about it. I'd push to just call it a "Card".

Whoops! I just saw a minute ago you changed exactly this, haha!

ksk5280:
> Hey! Just passing through. Hope you and family are doing well, @ksk5280! Question: What makes this a **Table**Card? I don't really see anything table-esque about it. I'd push to just call it a "Card".

hi @bdow, thanks for dropping in! I hope you and yours are doing well also!

Haha yes, I think you posted your comment at the same time that I pushed up the name change. :)

src/components/TableCard.tsx
khalidwilliams: Hmm I missed that too. If it's not too much time / work to add in it's probably worth it, but I don't think any of our team's work requires the empty state, so it would be fine to add in a follow up later IMO.

blambillotte: Ehh maybe leave for whoever needs it next? IMO the null state is more of a separate component than actually being a null state of the card itself since it doesn't really share any properties other than being sized similar?

src/components/Card.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Card, CardProps } from "./Card";
+
+describe("Card Component", () => {
+  const defaultProps = {
+    title: "Test Card",
+    subtitle: "Test Subtitle",
+    imgSrc: "plan-exterior.png",
+    detailContent: <div>Detail Content</div>,
+    tag: { text: "Active", type: "success" },
+  } as CardProps;
```
blambillotte, line 13: nit/suggestion: this may be a better place for `satisfies` vs. casting with `as` since I believe `as` won't error if there are invalid keys the way these options would:

```ts
  const defaultProps = {
    title: "Test Card",
    subtitle: "Test Subtitle",
    imgSrc: "plan-exterior.png",
    detailContent: <div>Detail Content</div>,
    tag: { text: "Active", type: "success" },
  } satisfies CardProps;

// alternatively
  const defaultProps: CardProps = {
    title: "Test Card",
    subtitle: "Test Subtitle",
    imgSrc: "plan-exterior.png",
    detailContent: <div>Detail Content</div>,
    tag: { text: "Active", type: "success" },
  };
```

src/components/Card.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+  { ...Css.fdc.$ },
+);
+
+export const BorderedListCard = createCardStory(
+  { ...baseArgs, type: "list", bordered: true },
+  hoverPlayFn({ click: true }),
+  { ...Css.fdc.$ },
+);
+
+function hoverPlayFn({ click: shouldClick }: { click: boolean }): PlayFunction {
```
blambillotte, line 77: nice!

blambillotte (approved):

src/components/TableCard.tsx
ksk5280: Yeah I agree, I wasn't sure if it made sense to include or not. Thanks for the input!

PR #1136, opened 2025-04-14
fix: Attempting to fix ScrollShadow chromatic snapshots

bdow:
The idea behind this fix is to use data-chromatic="ignore", as defined in https://www.chromatic.com/docs/ignoring-elements/. I'm thinking it'd be better to always have these elements present to be ignored. Not sure if even with the ignore attribute, having the shadow elements sometimes render and sometimes not would cause the same inconsistencies (which is the current issue).

### DIFF
--- a/src/components/ScrollShadows.tsx
+++ b/src/components/ScrollShadows.tsx
@@ -51,16 +51,12 @@ export function ScrollShadows(props: ScrollShadowsProps) {
       setShowStartShadow(start > 0);
       setShowEndShadow(start + boxSize < end);
     },
-    // TODO: validate this eslint-disable. It was automatically ignored as part of https://app.shortcut.com/homebound-team/story/40033/enable-react-hooks-exhaustive-deps-for-react-projects
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [],
+    [horizontal],
   );
 
   // Use a ResizeObserver to update the scroll props to determine if the shadows should be shown.
   // This executes on render and subsequent resizes which could be due to content/`children` changes (such as responses from APIs).
-  // TODO: validate this eslint-disable. It was automatically ignored as part of https://app.shortcut.com/homebound-team/story/40033/enable-react-hooks-exhaustive-deps-for-react-projects
-  // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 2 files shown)


chr1sjf0x (approved):
TY @bdow 🍣 

Does this mean you are coming back?? 😆

src/components/ScrollShadows.tsx
```diff
@@ -73,8 +69,8 @@ export function ScrollShadows(props: ScrollShadowsProps) {
       }
       {...tid}
     >
-      {showStartShadow && <div css={startShadowStyles} />}
-      {showEndShadow && <div css={endShadowStyles} />}
+      <div css={{ ...startShadowStyles, opacity: showStartShadow ? 1 : 0 }} data-chromatic="ignore" />
+      <div css={{ ...endShadowStyles, opacity: showEndShadow ? 1 : 0 }} data-chromatic="ignore" />
```
stephenh, line 73: Is this going to ignore everything _within_ the children / wrapped by the scroll shadow as well

If so, that seems undesirable?

stephenh: Ah right, I missed that these just sit above/below the children, and don't fully wrap them.

I guess there is, right, some overlap:

image

I'm wondering how the chromatic pixel diffing thing will pick that up, b/c naively I would have thought they look at "only pixels on the screen" for their diffs, but this insinuates that they can somehow map these divs back to the pixel they display?

...or maybe they just replace their content with "a big black screen" before taking the screenshot? And that's how the pixels get ignored?

:shrug: makes sense to try! Ty!

stephenh (approved):

src/components/ScrollShadows.tsx
bdow: Here's the page that explains it a bit more: https://www.chromatic.com/docs/ignoring-elements/#dimension-changes-in-ignored-elements-still-trigger-diffs

> Adding the data-chromatic="ignore" attribute or specifying a selector with the ignoreSelectors option instructs the diffing algorithm to disregard pixels within the bounding box of the ignored element. However, dimension changes of this element still trigger a change.

bdow:
I'm going to go ahead and merge this in. Hoping it'll work, but I won't really be able to tell. Let me know if these finally stop going away in the internal-frontend chromatic tests 😄

PR #1140, opened 2025-05-01
feat: [SC-68262] Grid table layout

blambillotte:
A layout component that combines a table with a header, actions buttons and filters. Story for this branch: https://sc-68262-table-page-layout--60466f7d43c37c0021788867.chromatic.com/?path=/story/components-layout-gridtablelayout--query-table-layout

This component can render either a `GridTable` or wrapped `QueryTable` based on the provided props:
 - For static data or custom handled loading states, use `rows` to pass in the data directly:
```tsx
<GridTableLayout
  tableProps={{
    rows: [...],
    columns: [...],
  }}
/>
```
 - To take advantage of data/loading/error states directly from an Apollo query, use `query` and `createRows`:
```tsx
<GridTableLayout
  tableProps={{
    query,
    createRows: (data) => [...],
    columns: [...],
  }}
/>
```

- For filtering, search and group by, this component is also intended to be used with the `useGridTableLayoutState` which  combines the various hooks into a single prop while also providing access to the filter/search states to the parent component:
```tsx
  const layoutState = useGridTableLayoutState({
    persistedFilter: {
      filterDefs,
      storageKey: "grid-table-layout",
    },
    search: "client",
  });

  const query = useExampleQuery({ filter: { ...layoutState.filter, search: layoutState.searchString } });


return <GridTableLayout
  tableProps={{...}}
  layoutState={layoutState}
/>
```

### DIFF
--- a/src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
+++ b/src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
@@ -30,7 +30,7 @@ export function GridTableLayout() {
       filterDefs,
       storageKey: "grid-table-layout",
     },
-    useSearch: "client",
+    search: "client",
   });
 
   return (
@@ -63,7 +63,7 @@ export function QueryTableLayout() {
       filterDefs,
       storageKey: "grid-table-layout",
     },
-    useSearch: "server",
+    search: "server",
   });
 
   // In this example, we set up a server-side search that uses the `searchString` from the layout state.
--- a/src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
+++ b/src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
@@ -1,0 +1,82 @@
+import { Meta } from "@storybook/react";
+import { useMemo } from "react";
+import { checkboxFilter, multiFilter } from "src/components/Filters";
+import { GridColumn } from "src/components/Table/types";
+import { simpleHeader, SimpleHeaderAndData } from "src/components/Table/utils/simpleHelpers";
+import { withBeamDecorator, withRouter } from "src/utils/sb";
+import { GridTableLayout as GridTableLayoutComponent, useGridTableLayoutState } from "./GridTableLayout";
+
+export default {
+  component: GridTableLayout,
+  decorators: [withBeamDecorator, withRouter()],
+} satisfies Meta;
+
+type Data = { name: string | undefined; value: number | undefined };
+type Row = SimpleHeaderAndData<Data>;
+
+export function GridTableLayout() {
+  const nameColumn: GridColumn<Row> = { header: "Name", data: ({ name }) => name };
+  const valueColumn: GridColumn<Row> = { header: "Value", data: ({ value }) => value };
+  const actionColumn: GridColumn<Row> = { header: "Action", data: () => <div>Actions</div> };
+
+  const filterDefs = useMemo(() => {
+    return {
+      primary: multiFilter({
+        options: [
+          { value: "primary", label: "Primary" },
+          { value: "secondary", label: "Secondary" },
+        ],
+        getOptionLabel: (tp) => tp.label,
+        getOptionValue: (tp) => tp.value,
+        label: "Preference",
+      }),
+      status: multiFilter({
+        options: [
+          { label: "Active", value: "active" },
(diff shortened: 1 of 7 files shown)


src/components/Layout/GridTableLayout/QueryTable.tsx
```diff
@@ -0,0 +1,77 @@
+import { useMemo } from "react";
```
blambillotte, line 1: Copied from https://github.com/homebound-team/internal-frontend/blob/main/src/components/QueryTable.tsx
with only a few type changes to avoid actual apollo client dependencies or importing `PageInfo` GQL type

zgavin (approved):

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+  );
+}
+
+export function QueryTableLayout() {
+  const filterDefs = useMemo(() => getFilterDefs(), []);
+
+  const layoutState = useGridTableLayoutState({
+    persistedFilter: {
+      filterDefs,
+      storageKey: "grid-table-layout",
```
stephenh, line 64: Tangent thought: I think we've eventually like these persistedFilter/etc things to "just work" and infer their storageKey/etc based on URL path.

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+    },
+    useSearch: "server",
```
stephenh, line 66: Maybe just `search`? `use` as a prefix is a little react-hook-y

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+  });
+
+  // In this example, we set up a server-side search that uses the `searchString` from the layout state.
+  // in combination with the "QueryTable" behavior for loading/error states.
+  const query = useExampleQuery({ filter: { ...layoutState.filter, search: layoutState.searchString } });
```
stephenh, line 71: Does seem like the ordering of hooks is pretty nuanced here, i.e.:

* The query is going to need the filters as input ... so the `useLayoutState` must come first
* But any "row building" needs to use the backend query data ... so we'll never be able to put "rows" into `useLayoutState`

Which is fine for our current API that is passing the rows into `tableProps`, but just in general it seems like "someday TableLayoutState would like to know _something_ about the rows for some interesting behavior" would not be unheard of, and seems like that would be difficult to pull off.

Not because of your approach, which is working really well, but just because of the strict evaluation order of react hooks.

src/components/Layout/GridTableLayout/QueryTable.tsx
stephenh: Oh nice! I forgot `QueryTable` hadn't made it into Beam yet :+1:

stephenh (approved):

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
blambillotte: yeah that is an interesting point to consider, would be cool if we can spend more focused time on our standard filter behaviors. We know we have to come back and update the UX as well since I totally punted on the new filter design to get this PR out quickly

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
blambillotte: Yeah for sure, this is one of those times where I miss the days of having a redux store that you can just share a central state that any level of the hierarchy can access/set and not deal with extra provider wrappers. The rules of hooks was also a real pain in this implementation as well since we can't conditionally call a hook, but we need the groupBy, filter & search behaviors to be opt-in.

stephenh: Yeah. Fwiw I like "non-global mobx stores that are just passed as props" for that sort of thing; it's what the GridTable headless state is managed with and has ended up being pretty clean.

Granted, still has to pass it as props, but ngl I kinda like knowing/seeing that dependency, vs even contexts which are "okay not truly global variables" but still "dependencies that you can't see".

PR #1141, opened 2025-05-08
feat: Add scroll and fixed icons to RightSideBar component

kbesingeryeh:
[SC-70171](https://app.shortcut.com/homebound-team/story/70171/fe-cost-summary-sidebar-investigate-independent-scrolling)

LOOM VIDEO: https://www.loom.com/share/6004d1f2ba2a47fe9a8d988befaf84ac

### DIFF
--- a/src/components/RightSidebar.test.tsx
+++ b/src/components/RightSidebar.test.tsx
@@ -1,3 +1,4 @@
+import { waitFor } from "@homebound/rtl-utils";
 import { click, render } from "src/utils/rtl";
 import { RightSidebar, SidebarContentProps } from "./RightSidebar";
 
@@ -9,7 +10,7 @@ describe("RightSidebar", () => {
     ] as SidebarContentProps[];
 
     // Given a right sidebar component with content
-    const r = await render(<RightSidebar content={content} />);
+    const r = await render(<RightSidebar content={content} headerHeightPx={120} />);
 
     // Expect the sidebar to render the icons and not render content initially
     expect(r.comment).toBeInTheDocument();
@@ -20,6 +21,11 @@ describe("RightSidebar", () => {
     click(r.comment);
     expect(r.rightSidebar_content).toHaveTextContent("Comments");
     expect(r.rightSidebar_content).not.toHaveTextContent("History");
+
+    // wait for transition to finish
+    await waitFor(() => {
+      expect(r.queryAllByTestId("time")).toHaveLength(1);
+    });
 
     click(r.time);
     expect(r.rightSidebar_content).toHaveTextContent("History");
--- a/src/components/RightSidebar.tsx
+++ b/src/components/RightSidebar.tsx
@@ -12,13 +12,14 @@ export type SidebarContentProps = {
 
 export type RightSidebarProps = {
   content: SidebarContentProps[];
+  headerHeightPx: number;
 };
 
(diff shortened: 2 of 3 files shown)


kbesingeryeh:
LOOM VIDEO: https://www.loom.com/share/6004d1f2ba2a47fe9a8d988befaf84ac

src/components/RightSidebar.tsx
```diff
@@ -48,32 +48,36 @@ export function RightSidebar({ content }: RightSidebarProps) {
...
-                {/* vertical line */}
-                <div css={Css.absolute.topPx(48).leftPx(23).h("calc(100vh - 168px)").wPx(1).bgGray300.$} />
-              </div>
-
-              {/* Horizontal icons when opened */}
-              <div css={Css.df.aic.jcfe.gap2.mb3.$}>
-                <IconButtonList content={content} selectedIcon={selectedIcon} onIconClick={setSelectedIcon} />
+            <>
+              {/* Sticky header section */}
+              <div css={Css.sticky.top0.bgWhite.z999.$}>
```
kbesingeryeh, line 55: I had some difficulty getting all the `icons` to align horizontally and make only the left one sticky, so I ended up just putting them in the same `div` and making them both sticky. I think it should be ok, plus it looks nicer visually to see both stick

src/components/RightSidebar.test.tsx
```diff
@@ -21,6 +22,11 @@ describe("RightSidebar", () => {
     expect(r.rightSidebar_content).toHaveTextContent("Comments");
     expect(r.rightSidebar_content).not.toHaveTextContent("History");
 
+    // wait for transition to finish
+    await waitFor(() => {
```
kbesingeryeh, line 26: there was a scenario in which the animation was rendering two `time` buttons in the DOM when switching icons, so I added this `waitFor` to ensure we actually only have one

src/components/RightSidebar.tsx
```diff
@@ -48,32 +49,36 @@ export function RightSidebar({ content }: RightSidebarProps) {
...
-                {/* vertical line */}
-                <div css={Css.absolute.topPx(48).leftPx(23).h("calc(100vh - 168px)").wPx(1).bgGray300.$} />
-              </div>
-
-              {/* Horizontal icons when opened */}
-              <div css={Css.df.aic.jcfe.gap2.mb3.$}>
-                <IconButtonList content={content} selectedIcon={selectedIcon} onIconClick={setSelectedIcon} />
+            <>
+              {/* Sticky header section */}
+              <div css={Css.sticky.top0.bgWhite.$}>
```
blambillotte, line 56: thought: Rather than having to do another level of sticky headers within a sticky container, what do you think about just setting the max height on the content container and letting the top icon container just sit on top like a normal div 🤔 

image

I think there's also an added benefit to scrollbar more correctly aligning to the content vs. extending beyond the top of the content as it does in your loom video

blambillotte: Nevermind the `scrollbar-gutter` property, I was just playing with something (as obviously the padding needs to be adjusted in my example)

blambillotte: Though I guess each is a compromise, downside of my approach is the icons aren't aligned on the right with the content 🤔 
I guess maybe not worth it, something we can't spend too much time on right now ha

blambillotte (approved):

PR #1143, opened 2025-06-27
feat: [SC-70480] Enable Banner/Snackbar in FormPageLayout

roberthdg:
### Description 

`FormPageLayout` is displayed as a full-screen modal, which prevents notifications from being visible during user actions (via `Snackbar`) or query errors (via `BannerNotice`). We need to add support for both types of notifications.

image

### DIFF
--- a/src/components/Layout/FormPageLayout.tsx
+++ b/src/components/Layout/FormPageLayout.tsx
@@ -9,6 +9,7 @@ import { useDebouncedCallback } from "use-debounce";
 import { Button, ButtonProps } from "../Button";
 import { Icon, IconKey } from "../Icon";
 import { RIGHT_SIDEBAR_MIN_WIDTH, RightSidebar, SidebarContentProps } from "../RightSidebar";
+import { Toast } from "../Toast/Toast";
 import { HeaderBreadcrumb, PageHeaderBreadcrumbs } from "./PageHeaderBreadcrumbs";
 
 type FormSection<F> = {
@@ -42,7 +43,6 @@ const maxContentWidthPx = 1600;
 
 function FormPageLayoutComponent<F>(props: FormPageLayoutProps<F>) {
   const { formSections, formState, rightSideBar } = props;
-
   const tids = useTestIds(props, "formPageLayout");
 
   // Create a ref for each section here so we can coordinate both the `scrollIntoView`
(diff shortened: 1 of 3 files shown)


src/components/Layout/FormPageLayout.tsx
```diff
@@ -85,11 +86,12 @@ export const FormPageLayout = React.memo(FormPageLayoutComponent) as typeof Form
 
 function PageHeader<F>(props: FormPageLayoutProps<F>) {
   const { pageTitle, breadCrumb, submitAction, cancelAction, tertiaryAction, formState } = props;
-
+  const { notice } = useToastContext();
   const tids = useTestIds(props);
 
   return (
-    <header css={Css.gr(1).gc("1 / 4").sticky.top0.hPx(headerHeightPx).bgWhite.z5.$} {...tids}>
```
roberthdg, line 92: @blambillotte removing`hPx` allows to effectively "push down" components below the header when the `BannerNotice` is rendered, but it also disrupts the scroll behavior of the left nav section:

https://github.com/user-attachments/assets/1dfaad8a-d656-453b-84c7-998148a00537

We could probably make it work even with the banner rendered, but that would require making `headerHeightPx` dynamic and maybe pass down a ref to `Banner` to recalc the current height of that component. But I think would overcomplicate things, since it's only a temporary state (and IMO the banner is more disruptive to the UI than the scroll being partially hidden).

I'd like to know your thoughts though.

blambillotte: hmm yeah tricky! I tend to agree that needing to use javascript to do DOM sizing, etc is something to avoid if we can. This seems like a fine solution though happy to look at it with you to consider options - I did notice the right panel isn't growing to fill the full viewport height like it was before so there was some other style regression at some point on this layout

blambillotte (approved):

src/components/Snackbar/Snackbar.tsx
```diff
@@ -12,7 +12,7 @@ export function Snackbar({ notices, offset }: SnackbarProps) {
   return (
     <div
       {...tid.snackbarWrapper}
-      css={Css.fixed.z999.bottomPx(offset.bottom ?? defaultOffset.bottom).left3.df.fdc.aifs.gapPx(12).$}
+      css={Css.fixed.z9999.bottomPx(offset.bottom ?? defaultOffset.bottom).left3.df.fdc.aifs.gapPx(12).$}
```
stephenh, line 15: Can you in-source comment why we're using `z9999` i.e. what we need to be "on top of"?

PR #1154, opened 2025-09-15
feat: Update button styles

kbesingeryeh:
[SC-69976](https://app.shortcut.com/homebound-team/story/69976/beam-update-buttons)

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -164,20 +164,14 @@ const variantStyles: (contrast: boolean) => Record<
     focusStyles: Css.bshFocus.if(contrast).boxShadow(`0 0 0 2px ${Palette.Blue400}`).bgGray700.white.$,
   },
 
-  danger: {
-    baseStyles: Css.bgRed600.white.$,
-    hoverStyles: Css.bgRed700.$,
-    pressedStyles: Css.bgRed800.$,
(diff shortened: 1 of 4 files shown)


src/components/Button.tsx
```diff
@@ -164,21 +164,21 @@ const variantStyles: (contrast: boolean) => Record<
...
+    pressedStyles: Css.gray900.if(contrast).bgWhite.gray900.$,
     disabledStyles: Css.gray400.if(contrast).gray700.$,
-    focusStyles: Css.boxShadow(`0px 0px 0px 2px ${Palette.White}, 0px 0px 0px 4px ${Palette.Red500}`)
+    focusStyles: Css.boxShadow(`0px 0px 0px 2px ${Palette.White}, 0px 0px 0px 4px ${Palette.Gray900}`)
       .if(contrast)
-      .boxShadow(`0px 0px 0px 2px ${Palette.Red500}`).$,
+      .boxShadow(`0px 0px 0px 2px ${Palette.Gray500}`).$,
   },
 
-  danger: {
```
blambillotte, line 177: Didn't realize the ticket was calling for the removal of the red `Danger` buttons - that seems like a risky change in terms of design language steering uses away from destructive actions - I'm going to start a thread in the beam channel to get more confirmation we're super sure about this removal

stephenh: "Quaternary" :rofl:

PR #1155, opened 2025-09-16
feat: allow custom radio group name

allangaldinosilva:
This PR adds support to provide a `name` prop to the `RadioGroupField` component.

By default, the underlying <input type="radio"> element requires a name attribute in order to be logically grouped with other radios. Without name, each radio is treated as independent, meaning multiple radios can be selected at the same time instead of behaving as a mutually exclusive set.

This fixes some situations where the `RadioGroupField` is rendered individually for a parent group.

### DIFF
--- a/src/inputs/RadioGroupField.tsx
+++ b/src/inputs/RadioGroupField.tsx
@@ -38,8 +38,8 @@ export interface RadioGroupFieldProps<K extends string> extends Pick<Presentatio
   helperText?: string | ReactNode;
   onBlur?: () => void;
   onFocus?: () => void;
-  /** The group name for the radio group. */
-  name?: string;
+  /** The group name for the radio group. Only for legacy pages with custom layouts - avoid using this. */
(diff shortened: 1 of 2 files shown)


src/inputs/RadioGroupField.tsx
```diff
@@ -38,6 +38,8 @@ export interface RadioGroupFieldProps<K extends string> extends Pick<Presentatio
   helperText?: string | ReactNode;
   onBlur?: () => void;
   onFocus?: () => void;
+  /** The group name for the radio group. */
+  name?: string;
```
stephenh, line 42: This fix makes a lot of sense, my only hesitation is that this page/component really isn't supposed to be used this way -- i.e. with random "drop this radio better whereever I want" custom layouts 😅 

Can we name this something like `unsupportedNameHack` and mention that this is only to be used for legacy pages, until we can move them over to a standard layout?

PR #1175, opened 2026-01-08
feat: Gridtable layout right pane

apattersonATX-HB:
## GridTableLayout: Add Right Pane Support + ReactNode Page Title

We have to build the right pane into GridtableLayout so that it _only_ appears next to the table as opposed to filling the entire viewport - header

### Changes
- Add optional `rightPaneProps` to `GridTableLayout` for built-in right pane functionality
- Change `pageTitle` prop from `string` to `ReactNode` for more flexible headers
- Add `paddingXss` prop to `Pagination` component for custom padding styles
- Update `ScrollableContent` to use `omitBottomPadding` to prevent double padding with pagination

### DIFF
--- a/src/components/Layout/GridTableLayout/GridTableLayout.tsx
+++ b/src/components/Layout/GridTableLayout/GridTableLayout.tsx
@@ -67,7 +67,7 @@ export type GridTableLayoutProps<
   tertiaryAction?: ActionButtonProps;
   hideEditColumns?: boolean;
   totalCount?: number;
-  rightPaneProps?: RightPaneLayoutProps;
+  rightPaneProps?: Omit<RightPaneLayoutProps, "children">;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -55,7 +56,7 @@ export type GridTableLayoutProps<
   X extends Only<GridTableXss, X>,
   QData,
 > = {
-  pageTitle: string;
+  pageTitle: ReactNode;
```
apattersonATX-HB: Saw a use case for `<AutosaveIndicator />` but didnt want to make a decorator or autosaveIndicator prop

apattersonATX-HB:
Needed for [SC-78056](https://app.shortcut.com/homebound-team/story/78056/rollout-new-table-layout-to-change-logs-table)

apattersonATX-HB:
I'm pretty sure this design has died. 

CC @bdow if you're interested

PR #1229, opened 2026-04-08
fix: number field clipboard paste

allangaldinosilva:
## Summary

Fixes a regression from [#1226](https://github.com/homebound-team/beam/pull/1226) where pasting values into `NumberField` was silently discarded.

PR #1226 added an echo-detection guard (`if (formatted !== lastSentRef.current) return`) to prevent react-aria's stale blur-commits from overwriting user edits. However, when pasting, react-aria calls `commit()` directly via its `onPaste` handler *before* `TextFieldBase.onChange` can update `lastSentRef`, so the guard blocked the legitimate pasted value.

Storybook https://60466f7d43c37c0021788867-ltsykrbklt.chromatic.com/?path=/story/inputs-numberfield--number-field-styles

### DIFF
--- a/src/inputs/NumberField.test.tsx
+++ b/src/inputs/NumberField.test.tsx
@@ -292,6 +292,21 @@ describe("NumberFieldTest", () => {
     expect(lastSet).toEqual(7500);
   });
 
+  it("does not discard pasted values", async () => {
+    // Given a NumberField with a value
+    const r = await render(<TestNumberField label="Cost" type="cents" value={1200} />);
+    expect(r.cost).toHaveValue("$12.00");
+    // When the user focuses, selects all, and pastes a new value
+    const input = r.cost as HTMLInputElement;
+    focus(input);
+    input.setSelectionRange(0, input.value.length);
+    fireEvent.paste(input, { clipboardData: { getData: () => "75" } });
+    fireEvent.blur(input);
+    // Then the pasted value should be retained
+    expect(r.cost).toHaveValue("$75.00");
+    expect(lastSet).toEqual(7500);
+  });
+
   it("respects useGrouping as false", async () => {
     // Given a NumberField with `useGrouping` set to `false`
     const r = await render(<TestNumberField label="Code" useGrouping={false} value={123456} />);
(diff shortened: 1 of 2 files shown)


src/inputs/NumberField.tsx
```diff
@@ -142,6 +142,8 @@ export function NumberField(props: NumberFieldProps) {
   // Tracks the last value we sent to the parent via onChange, so the render-time sync
   // can distinguish "echo of user typing" from "genuinely external value change".
   const lastSentRef = useRef<number | undefined>(undefined);
+  // Tracks the value at focus time so we can identify stale blur-commits from react-aria.
+  const focusValueRef = useRef<number | undefined>(undefined);
```
allangaldinosilva, line 146: Track the value at focus time to improve the condition below so it suppresses commits that revert to the stale focus value, while allowing paste-originated commits through.

src/inputs/NumberField.test.tsx
```diff
@@ -292,6 +292,21 @@ describe("NumberFieldTest", () => {
...
 
+  it("does not discard pasted values", async () => {
+    // Given a NumberField with a value
+    const r = await render(<TestNumberField label="Cost" type="cents" value={1200} />);
+    expect(r.cost).toHaveValue("$12.00");
+    // When the user focuses, selects all, and pastes a new value
+    const input = r.cost as HTMLInputElement;
+    focus(input);
+    input.setSelectionRange(0, input.value.length);
+    fireEvent.paste(input, { clipboardData: { getData: () => "75" } });
```
apattersonATX-HB, line 303: thought: Do all of our fields have this sort of test? Seems valuable now that we've seen it break 😅

apattersonATX-HB (approved):
lgtm

src/inputs/NumberField.test.tsx
allangaldinosilva: That's a good call, I don't think we have that for other fields 🤔, the `NumberField` is a particular one that we've been struggling to handle some edge cases (https://github.com/homebound-team/beam/pull/1185, https://github.com/homebound-team/beam/pull/1226) 😅

PR #1230, opened 2026-04-08
fix: Don't highlight-on-row-hover readonly fields.

stephenh:


package.json
```diff
@@ -88,7 +88,7 @@
     "@homebound/eslint-config": "^1.10.2",
     "@homebound/rtl-react-router-utils": "1.3.0",
     "@homebound/rtl-utils": "^2.71.0",
-    "@homebound/truss": "^2.21.2",
+    "@homebound/truss": "^2.21.3",
```
stephenh, line 91: Just bumping truss for good measure

src/inputs/TextFieldBase.tsx
```diff
@@ -254,10 +254,9 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
               css={{
                 // Use input wrapper to get common styles, but then we need to override some
                 ...fieldStyles.inputWrapperReadOnly,
-                ...(multiline ? Css.fdc.aifs.gap2.$ : Css.if(wrap === false).truncate.$),
+                ...(multiline ? Css.fdc.aifs.gap2.$ : Css.if(!wrap).truncate.$),
                 ...xss,
               }}
-              className={BorderHoverChild}
```
stephenh, line 260: We're in the "this is read only" branch of the ternary on line 252

zovington (approved):

PR #1238, opened 2026-05-04
feat: add scrollable footer

allangaldinosilva:
### Description

Adds a `ScrollableFooter` layout primitive that portals content into a new sibling slot below ScrollableParent's scrollable region, so action bars (Save/Cancel, etc.) sit at the bottom of the page without overlapping the last row of long/virtualized tables.

### Screenshot

https://60466f7d43c37c0021788867-wqfjtbkqqy.chromatic.com/?path=/story/components-layout--with-footer

image

### DIFF
--- a/src/components/Layout/Layout.stories.tsx
+++ b/src/components/Layout/Layout.stories.tsx
@@ -5,12 +5,14 @@ import { TabsWithContent, TabWithContent } from "src/components/Tabs";
 import { Css, Palette } from "src/Css";
 import { FormLines } from "src/forms";
 import {
+  Button,
   FullBleed,
   GridColumn,
   GridDataRow,
   GridTable,
   PreventBrowserScroll,
   ScrollableContent,
+  ScrollableFooter,
   ScrollableParent,
   simpleHeader,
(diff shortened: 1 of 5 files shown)


src/components/Layout/Layout.stories.tsx
```diff
@@ -69,6 +71,23 @@ export function VirtualizedScrolling() {
...
 
+export function WithFooter() {
+  return (
+    <TestProjectLayout>
+      <TestHeader title="Enter Bid Information" />
+      <ScrollableContent virtualized>
+        <TableExample virtualized numCols={20} numRows={200} />
+      </ScrollableContent>
+      <ScrollableFooter>
+        <div css={Css.df.aic.jcfe.gap2.bgWhite.bt.bcGray200.px3.py2.bshHover.$}>
```
stephenh, line 82: @allangaldinosilva / @bdow I wonder if the `bgWhite` / `bcGray200` ... maybe the `bshHover`? should be something we should either bake directly into `ScrollableFooter` (probably not) or some other component in Beam 🤔

stephenh (approved):

PR #1239, opened 2026-05-07
feat: table review layout

apattersonATX-HB:
**TableReviewLayout**
- New full-screen modal layout for reviewing table data with a slide-in detail panel
- Supports both rows-based and query-based tables
- `emptyState` auto-detects empty rows for rows-based tables; for query-based tables, the caller controls when to show it
  - Ideally we could build this into gridTable to succeed `fallBackMessage`
- `SidePanel` is an internal component (not exported) with a sticky header/footer and scrollable body

**Other layout files**
- Extracted `ActionButtonProps` into `layoutTypes.ts`, eliminating duplicate definitions across `FormPageLayout`, `GridTableLayout`, and `SidePanel`
- Consolidated `BaseTableProps`, `GridTablePropsWithRows`, `BaseQueryTableProps`, and `isGridTableProps` into `layoutTypes.ts` so `TableReviewLayout` and `GridTableLayout` share the same table prop types

Screenshot 2026-05-07 at 12 04 32 PM

Screenshot 2026-05-07 at 12 04 47 PM


src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
```diff
@@ -0,0 +1,142 @@
...
+  GridTablePropsWithRows,
+  isGridTableProps,
+  BaseQueryTableProps as QueryTablePropsWithQuery,
+} from "../layoutTypes";
+import { HeaderBreadcrumb, PageHeaderBreadcrumbs } from "../PageHeaderBreadcrumbs";
+
+const defaultRightPaneWidth = 450;
+
+export type TableReviewLayoutProps<R extends Kinded, X extends Only<GridTableXss, X>, QData> = {
+  pageTitle: ReactNode;
```
apattersonATX-HB, line 24: made this and the other layouts also a ReactNode for the common use case of an icon/tag/autosave indicator

src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -17,47 +16,34 @@ import { useTestIds } from "src/utils";
...
-type GridTablePropsWithRows<R extends Kinded, X extends Only<GridTableXss, X>> = BaseTableProps<R, X> & {
-  rows: GridTableProps<R, X>["rows"];
-  query?: never;
-  createRows?: never;
-};
-type QueryTablePropsWithQuery<R extends Kinded, X extends Only<GridTableXss, X>, QData> = BaseTableProps<R, X> & {
-  query: QueryResult<QData>;
-  createRows: (data: QData | undefined) => GridDataRow<R>[];
+// GridTableLayout-specific query props extend the shared base with pagination/display extras.
+type QueryTablePropsWithQuery<R extends Kinded, X extends Only<GridTableXss, X>, QData> = BaseQueryTableProps<
```
stephenh, line 28: The `QueryTable...WithQuery` having query x2 is a little awkward -- do we need the `WithQuery` suffix?

src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -17,47 +16,34 @@ import { useTestIds } from "src/utils";
...
   F extends Record<string, unknown>,
   R extends Kinded,
   X extends Only<GridTableXss, X>,
   QData,
 > = {
-  pageTitle: string;
+  pageTitle: ReactNode;
   tableProps: GridTablePropsWithRows<R, X> | QueryTablePropsWithQuery<R, X, QData>;
-  breadcrumb?: HeaderBreadcrumb | HeaderBreadcrumb[];
+  breadCrumb?: HeaderBreadcrumb | HeaderBreadcrumb[];
```
stephenh, line 46: I don't mind the rename, but doing these as separate PRs can be nice

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
```diff
@@ -0,0 +1,142 @@
...
+  breadCrumb?: HeaderBreadcrumb | HeaderBreadcrumb[];
+  /** Instructional text rendered below the title, above the table. */
+  description: ReactNode;
+  closeAction: VoidFunction;
+  tableProps: GridTablePropsWithRows<R, X> | QueryTablePropsWithQuery<R, X, QData>;
+  /**
+   * Replaces the table region with centered content.
```
stephenh, line 31: "with centered content ... when empty" I assume? But then not following hte "shown whenever defined" comment for query-based tables.

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
```diff
@@ -0,0 +1,142 @@
...
+              <div css={Css.fg1.oh.mh0.$}>{panelContent}</div>
+            </motion.div>
+          )}
+        </AnimatePresence>
+      </div>
+    </div>
+  );
+}
+
+export const TableReviewLayout = React.memo(TableReviewLayoutComponent) as typeof TableReviewLayoutComponent;
```
stephenh: Kinda wonder if we need this? I know we have it in a few other places, but I feel like we've been copy/pasting it around without maybe really needing it 🤔

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
```diff
@@ -0,0 +1,142 @@
...
+    // that works as a successor to `fallbackMessage`.
+    tableContent = <div css={Css.h100.df.fdc.aic.jcc.$}>{emptyState}</div>;
+  } else {
+    tableContent = (
+      <QueryTable {...(tableProps as QueryTableProps<R, QData, X>)} style={{ allWhite: true }} stickyHeader />
+    );
+  }
+
+  return (
+    <div css={Css.fixed.top0.bottom0.left0.right0.z(1000).bgWhite.df.fdc.$} {...tid}>
```
stephenh, line 99: Oh, maybe this is a full-page-modal-ism? 😅 If so, would we remove this? And have the page that wants to use this just be a route?

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
apattersonATX-HB: Not particularly. Any query updates would just cause a re-render. I believe the moment we add in any kind of `formState` it'll need to come back but we can drop it for now.

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
apattersonATX-HB: I updated the comment to be clearer.

Basically if we have a query table that needs to show a prettier emptyState they'll need to update the emptyState prop on their own somehow. The DP created slot use case would be `!hasPendingSlotRequests`.
I assume we wouldn't want our `ReviewTableLayout` component peeking into queryTableProps to see if it has any rows. We could, but it'd be ugly. A better route is just expanding gridTable to use an `emptyState` ReactNode instead of only having a fallbackMessage prop.

Or we can toss it out and only show a fallbackMessage row for QueryTables. I don't have a preference. This never went through a design review process anyways.

src/components/Layout/GridTableLayout/GridTableLayout.tsx
apattersonATX-HB: 🤣 nah. I'll just use the normal type name.

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
apattersonATX-HB: That sounds like a bigger discussion with @bdow. 

It'd be worse if we made this one a full page while keeping FormPageLayout & GridTableLayout as modals.

apattersonATX-HB: Checked BP — 6/8 usages are already route components. The fixed positioning just handles the viewport fill; it works fine as a route. The side and headers are usually just taking away space that we can't fully get rid of without this modal style.

stephenh: "while keeping FormPageLayout & GridTableLayout as modals"

Our two existing page layouts have full-page-modal-isms built into them?

apattersonATX-HB: Actually only [FormPageLayout](https://github.com/homebound-team/beam/blob/285742e8d1df53f5cb16dfbb0e1510a40acba32e/src/components/Layout/FormPageLayout.tsx#L69) does. GridTableLayout doesn't. It's only got one usage in BP atm anyways.

To your point though, FormPage/ReviewTableLayout are supposed to take the entire screen. [GridTableLayout designs](https://www.figma.com/design/OfJJm1CXHRGkrHNcPu2grY/2025-UX-Standardization?node-id=542-14285&m=dev) also have it taking over full screens as well but when we were implementing we talked a lot about replacing current tables so it kind of does both probably closer to what you're envisioning the other two doing if I'm following right.

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
```diff
@@ -0,0 +1,146 @@
...
+   * Replaces the table region with centered content.
+   *
+   * For rows-based tables: shown automatically when `tableProps.rows` contains no data rows.
+   * For query-based tables: shown when prop is defined; TableReviewLayout does not peek into query results to determine if empty.
+   */
+  emptyState?: ReactNode;
+  /**
+   * When set, slides open the panel column and renders a `SidePanel` with the given props.
+   */
+  panelContent?: SidePanelProps;
```
apattersonATX-HB, line 40: Updated from a `ReactNode` to `SidePanelProps` so we're always using the same format.

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { SidePanel } from "./SidePanel";
+
+const defaultRightPaneWidth = 450;
+
+export type SidePanelProps = {
+  title: ReactNode;
+  content: ReactNode;
+  primaryAction?: ActionButtonProps;
+  secondaryAction?: ActionButtonProps;
+};
```
bdow, line 21: **thought/non-blocking**: Disclaimer; I'm still getting caught up with the newer layouts, so I might not know how we're expecting to make new layouts. 
Anyways, my take is that SidePanels would be a composable component that can be added to any existing layout. Making a new side panel per layout feels like another way for our UI's to become inconsistent quickly. 
I'm not going to block this on asking you to make an abstraction so we can have one SidePanel component that works across all layouts, because it feels like we're about to do a fairly big reset on UI/UX.

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
bdow: This makes me want to kick the can down the road given the soon to be discussed UI/UX discussions. I don't particularly like the full page modal sort of thing. To me it also feels like it should just be a new route. And then have it be more of a `<FocusedLayout />`(?) where we don't include the global nav or anything. Reasoning is that once we start stacking UI's on top of each other, we're eating up more resourcing on the client, susceptible to much large rerenders if the parent below this modal changes, it'd prevent sharing this page via a URL, and makes it super tricky to do things with query parameters. I'm not saying we'll run into these issues with this layout, but we're opening the door for them by beginning to support this paradigm.

bdow (approved):
Fine to move this through to unblock given the renewed push for a refreshed UX/UI.

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
stephenh: Yeah -- @apattersonATX-HB were you planning on this "review incoming slots" UX being it's own route, in the PP UI? If not, can we make it so?

Feel free to ping @bdow on the easiest way to get routes "without the navbar / header / etc" b/c I'm almost certain it hasn't changed since he setup it up. :sweat_smile: 

Even with the UX discussions happening this week, I can't see "we need more full page modals" being an outcome :sweat_smile: , so I'd prefer just not doing that here in `TableReviewLayout`, and then writing a ~tech debt ticket (probably for @bdow on the beam side of things :-) to remove it from `FormPageLayout` and update the internal-frontend pages either to a) be routes now, or b) still do the full page modal hack but only one-off in their current legacy places and not baked into the `FormPageLayout`.

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
apattersonATX-HB: So preference being something like `PanelTitle | PanelBody | PanelFooter` like we've done for Modals?

I can see that. I see this as more of a `FormPanelProps` since we expect to be loading forms inside of it more than anything. Maybe we'd want to turn this into something specific for forms like that and than also expose individual components for others. There were some conversations around updating our comment/messaging side panel too but those all exist inside of normal routes with a header/leftNav (with lots of other cruft around comm panels themselves).

Will leave this as is for now since we're investing time in updating all of Beam and BP and not just wishful thinking 🙇🏽‍♂️

src/components/Layout/TableReviewLayout/TableReviewLayout.tsx
apattersonATX-HB: @stephenh Yeah i can make it it's own route. 

@bdow What's pushing the actual usage of this as a modal in part is the [single entry point](https://www.figma.com/design/95RUxYOYOy4t7i9R2LJF7s/Q1-2026-%7C-Slot-creation?node-id=3960-24943&m=dev) that only shows when requests are active so it's not clear how much value their is in having a shareable link. **But** the point on re-renders happening under the hood would be an easy gotcha for callers of this component. Even within this use case if we attach the `refetchQueries: ["FooBarTakeoffQuery"]` to the mutation accepting these TLIs instead of `closeAction` on this layout it'd be bad perf 😅

PR #1240, opened 2026-05-08
fix: pick the dominant axis on DnDGrid

roberthdg:
### [P-2211](https://linear.app/homebound-team/issue/P-2211/beam-fix-drag-and-drop-component)

The drag “passed target” check treated X and Y equally: if the cursor crossed either the bottom half or the right half of a target, it counted as “passed” and triggered “insert after”.

That worked in the existing implementations because items are narrow and dragging naturally moves across X. But in a single-column grid with wide rows and right-side drag handles, the cursor stays in the right half of every target the whole time. The check becomes permanently true, so only “insert after” ever fires. Dragging down works, but dragging up never lands.

The fix chooses a single axis per move based on the dominant displacement between the placeholder and target. Single-column grids use Y, single-row grids use X, and 2D grids use whichever axis changes more.

Before:

https://github.com/user-attachments/assets/aa36e755-8126-4a2d-b061-d0cc0b91776f

After:

https://github.com/user-attachments/assets/aece389a-d159-4436-a57a-b190fd62afd9

## Storybook

* [Before](https://60466f7d43c37c0021788867-kxzpcopueq.chromatic.com/?path=/story/components-table-gridtable--draggable-card-rows)

* [After](https://60466f7d43c37c0021788867-pdjhmwmsjh.chromatic.com/?path=/story/components-table-gridtable--draggable-card-rows)

### DIFF
--- a/src/components/DnDGrid/DnDGrid.tsx
+++ b/src/components/DnDGrid/DnDGrid.tsx
@@ -4,7 +4,7 @@ import { Css, Palette, Properties, useTestIds } from "src";
 import { clearInlineStyles, isDefined, setInlineStyles } from "src/utils";
 import { DnDGridContext } from "./DnDGridContext";
 
-export interface DnDGridProps {
+export type DnDGridProps = {
   children: ReactNode;
   /** CSS Grid styles for the grid container. */
   gridStyles?: GridStyles;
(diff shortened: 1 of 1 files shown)


bdow (approved):
Looks good.
**non-blocking:** Does make me wonder if an improvement would be to optionally only allow dragging in a certain direction. Could be a good follow up at somepoint.

PR #1243, opened 2026-05-14
feat: Introduction of design tokens

bdow:
Begins utilization of design tokens and introduction of semantic tokens for theming capabilities.

Design tokens are written in a the `tokens/tokens.json`, which follows the designtokens.org specification format: https://www.designtokens.org/tr/2025.10/format/. Beam uses generation scripts to take the JSON and create consumable CSS styles and tokens for Beam and Truss.

Token naming semantics are to follow TailwindCss theming variable name prefixes going forward

As a POC of token work, we have introduced `<ContrastScope />` in favor of the `contrast: boolean` property across multiple components.


.cursor/rules/general.mdc
```diff
@@ -13,6 +13,8 @@ alwaysApply: true
 - Styles are generated via the `@homebound/truss` library
 - Style properties are defined in the generated @Css.ts file
   - ALWAYS check [Css.ts](mdc:src/Css.ts) for the correct property names and available methods
+- **Design tokens:** Edit [`tokens/tokens.json`](mdc:tokens/tokens.json) per [`tokens/README.md`](mdc:tokens/README.md). After changes: `yarn generate:design-tokens`, `yarn validate:tokens`. CI runs validate then `check:token-drift` before build — see [`AGENTS.md`](mdc:AGENTS.md).
+- **Linting:** After editing files under `src/`, run `yarn lint:fix:files` with those paths (see [`AGENTS.md`](mdc:AGENTS.md)). Before a PR, run `yarn lint:ci` so CI’s ESLint job matches local.
```
stephenh, line 17: Ngl I didn't know about these `.cursor/*` files 😅

src/Css.ts
```diff
@@ -1747,9 +2040,241 @@ class CssBuilder<T extends Properties, S extends StyleKind = "buildtime"> {
...
+  /** Sets `color: "var(--b-field-border-hover, rgba(221, 220, 220, 1))"`. */
+  get fieldBorderHover() {
+    return this.add("color", "var(--b-field-border-hover, rgba(221, 220, 220, 1))");
+  }
+  /** Sets `color: "var(--b-field-text-disabled, rgba(141, 141, 141, 1))"`. */
+  get fieldTextDisabled() {
+    return this.add("color", "var(--b-field-text-disabled, rgba(141, 141, 141, 1))");
+  }
+  /** Sets `color: "var(--b-focus-ring-inset, rgba(29, 78, 216, 1))"`. */
+  get focusRingInset() {
```
stephenh: I think `color` having "no prefix" was fine when things where `Css.white.$` / `Css.black.$` but now that our "colors" are all these token values, it seems like the Css of `Css.focusRingInset.$` is not incredibly intuitive that "oh I'm setting the color to focusRingInset"...

Ofc it'd be a huge breaking change, so we probably can't do it, but I'd kinda like to see `cFocusRingInside` -- wonder if we could somehow have the `c` prefix added for any "not one of the core/obvious colors".

tokens/tokens.json
```diff
@@ -0,0 +1,1522 @@
+{
+  "$schema": "https://www.designtokens.org/schemas/2025.10/format.json",
+  "beam": {
+    "$description": "Beam design tokens \u2014 DTCG 2025.10-shaped source. Authoring and `beam.*` layout: see tokens/README.md (colors under beam.color.* today; other namespaces as implemented). Run yarn generate:design-tokens after edits.",
+    "color": {
+      "primitive": {
```
stephenh, line 6: Do you use an editor/IDE plugin to maintain this? Like if there is some standard tooling for it -- it's hard for me to get excited about reading/maintaining this file by hand / or just in general.

But if it's just coming from tools, it doesn't matter.

tokens/tokens.json
```diff
@@ -0,0 +1,1522 @@
...
+          "$type": "color",
+          "$value": "{beam.color.primitive.Blue800}",
+          "$extensions": {
+            "com.homebound.beam": {
+              "cssVar": "--b-primary-pressed",
+              "contrast": "{beam.color.primitive.Blue800}"
+            }
+          }
+        },
+        "Scrim": {
```
stephenh, line 1368: Are these `Scrim` / `Surface` names pretty standard in the token space? Or are these the tokens we imported from Tailwind?

tokens/tokens.json
henrysipp: Pretty solid support from Figma, et al. So our design team should be able to maintain this!

src/Css.ts
bdow: Yea,... I dunno. I'm kind of on the other side of that fence. I like the idea of doing `Css.primary.$` and having semantic color tokens for those reasons. I get what you're saying, though, with that example as `inset` is a css property as well. I definitely would like to not have ambiguity between these values, so maybe "Inset" is the wrong term here anyways.

tokens/tokens.json
bdow: Yes, they are. Not imported from Tailwind, but Surface and Scrim are token names used by Google's Material design. Happy to bikeshed on naming conventions - I'm not too tied to it. I did like "Surface" and "OnSurface" to denote like background and color on the background.

tokens/tokens.json
bdow: The goal will be to have Figma or another source of truth generate this json. To start, I think hand editing is what we're doing, but the idea will be that it'll be generated elsewhere and synced to Beam and other consumers

src/Css.ts
stephenh: "and having semantic color tokens for those reasons"

Not 100% following? I'm good with semantic color tokens, my point is just that their names are so "not colors" that the previous "cute" approach of letting `color` drop its names directly onto `Css.(blue100|red100|etc)` is maybe not a good idea anymore.

Like maybe we could tell truss, all `blue*|red*|black*` colors are fine to do `Css.blue100` abbreviations for, but all others should have a `c` prefix, similar to `bc`, to get `Css.cPrimary.cFocusRingInside`?

Or do you think the `Css.focusRingInside` will actually be just fine in practice?

bdow: Okay, now you've got me on the fence 😄. I like being able to do "Css.primary.$" with the expectation that is just the primary color. But the `Css.focusRingInside.$` is definitely more vague of what it actually results in - is it an `outline`, `box-shadow`, etc. Though, I do hope some of this is further abstracted and instead of having a `Css.addIn("box-shadow", "0 0 0 3px ${Palette.FocusRingInset}")` becomes more of a `Css.focusRing.$`.
I get this is tangential to your ask, so maybe something doing a `cWhite`, `cTransparent` - having some sort of color namespace would be good.

bdow: Granted, I know Tailwind does not namespace their colors like they do the other styles. E.g `class="primary bgPrimary"`

stephenh: "becomes more of a `Css.focusRing.$`" --> totally agree, although now I'm left wondering "...wait, is `focusRing` also a token and so this is just setting color as well?" :sweat_smile: 

It's unfortunate b/c I think if we were "starting fresh" and knew our "colors could be semantically-named tokens", we'd definitely have the `c...` prefix -- it's just only our backwards compatibility we have to worry about.

I dunno, maybe we can fork this conversation into a Truss PR/config flag that changes all color abbreviations to be `c...` prefixed, unless they match the ~7-8 "known actual color" setup in the `truss-config`?

I can take a crack at that if you like, or very happy to have you run with it as well -- doesn't have to block this PR.

tokens/tokens.json
stephenh: Cool -- I'm generally ambivalent, except that things like `OnSurface` imo should a little too much like Truss `if...` / `when...` things, so I think when I read `Css.onSurface` I would think "oh this is a media query something"?

...granted, the `c...` would help some of that like `Css.cOnSurface`. :thinking: 

I'm also just curious, what does "on surface" denote? The color ... on a background? Not immediately clicking for me.

src/css/generated/theme-scopes.css
```diff
@@ -0,0 +1,65 @@
+/*
+ * AUTO-GENERATED — do not edit. Run `yarn generate:design-tokens`, `yarn build`, or `yarn build:truss`.
+ *
+ * [data-theme] attribute values must match `contrastDataTheme` (etc.) in src/components/ContrastScope.tsx.
+ * Source: beam.color.semantic.* theme axes in tokens/tokens.json (excluding baseline default).
+ */
+
+[data-theme="contrast"] {
```
stephenh, line 69: Nice!

tokens/tokens.json
bdow: Yea, good point regarding the Truss-isms. And yes, Surface is a background color. It's almost like the "Primary Background" - typically white. And then "onSurface" would be the color typically on that background color. So its not just the text color, but could be an icon color. Some other options were to use like "Ink" vs "onSurface". I'm in no means tied to the nomenclature here, but did enough back and forth with the AI that I wanted to push forward and have these discussions here instead. So extremely welcome to coming up with better ways to semantically define a background color and the colors that should be on that background. Like for us would be white and gray900. But if we did contrast, then it'd be flipped.

src/Css.ts
```diff
@@ -41,7 +41,68 @@ export type RuntimeStyles = RawCssProperties & { readonly __kind: "runtime" };
 
 export type Typography = "xs2" | "xs2Sb" | "xs" | "xsSb" | "sm" | "smSb" | "md" | "mdSb" | "lg" | "xl" | "xl2";
 
-export type CssSetVarKeys = `--${string}`;
+export enum Tokens {
+  ButtonCautionDisabledBg = "--b-button-caution-disabled-bg",
+  ButtonCautionDisabledFg = "--b-button-caution-disabled-fg",
+  ButtonDangerDisabledBg = "--b-button-danger-disabled-bg",
+  ButtonDangerDisabledFg = "--b-button-danger-disabled-fg",
```
stephenh, line 48: @bdow / @henrysipp this is where the designtokens "graduating levels of semantic tokens" (i.e. sure `primary = blue` is a base token -- totally on board with that, but then there keep being more & more tokens layer on top) kinda looses me...

Like `ButtonCautionDisabledBg` is _such_ a specific token, isn't this a slippery slope to:

* Every component...
* ...for every variation (caution | danger | ...)
* ...for every attribute (fg | bg | ...)

Ends up being a token? How does this not end up with ~50 components x 2-3 variants x 2-3 attributes = hundreds/1000s of tokens?

Or is that the point, and it's fine? 🤔

stephenh: I guess specifically the component level of tokens: https://m3.material.io/foundations/design-tokens/overview

bdow: 100%. I think this is more of an exercise to expose our inconsistencies and waaaayyy too many variations. I am hoping we can slim things down. Like, I discovered we have a Quaternary and Caution variants of our buttons. This is too many.
So I think its actually like a 50/50 sort of split - Hopefully we can eliminate many of these, but there may be times where the tokens need to be a bit more specific to have a narrow target and not accidentally impact elements we shouldn't.

src/Css.ts
```diff
@@ -106,9 +167,241 @@ class CssBuilder<T extends Properties, S extends StyleKind = "buildtime"> {
   }
 
   // borderColor
-  /** Sets `borderColor: "rgba(255,255,255,1)"`. */
+  /** Sets `borderColor: "var(--b-button-caution-disabled-bg, rgba(254, 240, 138, 1))"`. */
+  get bcButtonCautionDisabledBg() {
```
stephenh: @bdow this is specifically where the "1000s of tokens" concerns me, is that afaiu we assume: a) all tokens should go in Palette (afaict?( and b) all palette values can be used for any css prop.

So we get output like this `bc...` of "border-color = ...a token of buttonCautionDisabledBg..." that is specifically meant for background colors, not border colors, so this abbreviation is like invalid to start with.

I saw that `tokens.json` has some `com.homebound.beam` metadata floating around -- could we add some hints to there that would tell us things like "don't make a `Palette....` for this"?

...maybe that is another solution/alternative to the `c...` prefix quandary. 🤔 

I.e. instead of "oh we need to prefix colors with `c...`" this real issue is that Palette just isn't the right spot for any "non-reference" tokens, and anyone using a system/component token would need to do:

```
Css.bgColor(Tokens.ButtonCautionDisabledBg);
```

stephenh: So like does `tokens.json` know about reference vs. system vs. component tokens, and can we put only reference tokens in Palette, and then have some other way for everyone else to get the "other tokens"?

...I also assume we'll have non-color tokens at somepoint as well? So moving away from "all tokens go in Palette" is good anyway?

src/Css.ts
stephenh: Tying the comment threads together, I think https://github.com/homebound-team/beam/pull/1243/changes#r3249350631 is an alternative/better solution maybe.

src/Css.ts
bdow: Yea, for sure will have tokens for box-shadows, fonts, etc. Using https://tailwindcss.com/docs/theme#theme-variable-namespaces as a guide to what could be themeable and might need to be a token. I don't think we'd do everything they have, but just a guide.
So yea, not every token is going to be a color. This is where the namespace in tokens.json comes in for `beam.colors.[semantic|primitives]`. Then we also would have `beam.shadows.*`, `beam.fonts`, etc. We'd need to update the generator to handle those as they come.

bdow: I think maybe component level semantic tokens would be good to separate in the json. That makes sense. This way we can still get things like `Css.primary` and `Css.bgPrimary`, etc. But we could build into the schema a way of know that this "ButtonCautionDisabledBg" is _just_ a background color. I like that nudge and will move towards this as part of this PR.

stephenh: Nice! Thank you!

Do you think we'll need a Truss change to make the `Css.bgColor(Tokens.ButtonCautionDisabledBg)` more ergonomic? Like I don't think that works today? Or I forget what it looks like anyway :sweat_smile: 

...naive question on `Css.bgPrimary`, that also strikes me as "...would that even make sense?" :sweat_smile: -- maybe I'm being overly protective of `Palette`, but just given how the palette values "end up everywhere" (color, border color, background color), I'm ~maybe overly protective of it and would nudge for it to be as small as possible.

...one thought, I kinda wonder how many folks will being `Css.primary` / `Css.bgPimary` anyway; like the goal with `Css.blue500` and `Css.bgBlue400` and friends is that "there will be a lot of places setting those colors/bgs/etc so let's make it as ergonomic as possible" -- with some of these tokens, maybe `primary` is mostly owned by the Beam internal buttons anyway, and having them do `Css.color(Token.Primary)` a few times is fine, if the upshot is that "all the code _outside_ of beam" no longer seems `Css.primary` as a thing they should even be doing? 

Like we should probably bias all `Css.foo/bar/zaz` abbreviations to "will be used frequently by internal-beam & external-application callers", to "earn their spot" in the `Css` namespace, and have internal-only/infrequent-usages be fine with `Css.color(...)` type things? :thinking:

bdow: I think `Css.bgColor(Tokens.ButtonCautionDisabledBg)` would work fine. Hopefully thats more of just a Beam thing that happens once when its this specific of a component token. 
Per `Css.bgPrimary`, I think it makes sense. I'd rather keep it. I think there are places where using the primary color for a foreground/text color and a background color makes sense, and we'd want to keep those in sync. Agreed on keeping it small, and will keep that in mind when we expand these semantic tokens. I think keeping component level tokens out of the palette will make a big impact on your concern.
I'm not going to die on that hill, so if you feel strongly that _all_ non-primitive colors should be kept out of the color generation process, thats fine by me.

stephenh: I'd personally stick to only non-primitive in Palette but agreed keeping component out will be the biggest win, so yeah that sgtm.

bdow: Okay. Palette is back to just the colors it was before. Semantic color tokens do not generate class names in Css.ts. Instead, they must be used as `Css.color(Tokens.Primary).$`.

apattersonATX-HB (approved):

src/Css.ts
```diff
@@ -396,7 +457,7 @@ class CssBuilder<T extends Properties, S extends StyleKind = "buildtime"> {
   }
   /** Sets `borderColor: value`. */
   bc(value: Properties["borderColor"]) {
-    return this.add("borderColor", value);
+    return this.add("borderColor", maybeCssVar(value));
```
stephenh: Iirc this moved into the `add` method in the last iteration of the Truss PR? Just checking my understanding...

stephenh (approved):

src/Css.ts
bdow: Yes, you're correct. I hadn't run `yarn build:truss` yet. Update coming 👍

src/Css.ts
stephenh: Just catching up, but the Palette looks great -- ty!

PR #1244, opened 2026-05-14
feat: [P-2225] Add horizontal layout option to RadioGroupField

allangaldinosilva:
### [P-2225](https://linear.app/homebound-team/issue/P-2225)

Adds a `layout` prop to `RadioGroupField` so options can render side-by-side instead of stacked. Defaults to `"vertical"` to preserve existing behavior.

### DIFF
--- a/src/inputs/RadioGroupField.stories.tsx
+++ b/src/inputs/RadioGroupField.stories.tsx
@@ -294,3 +294,23 @@ export function HelperText() {
     />
   );
 }
+
+export function HorizontalLayout() {
+  const [state, setState] = useState<string | undefined>("a");
+  return (
+    <FormLines width="md">
+      <RadioGroupField
+        label="Estimate Type"
+        layout="horizontal"
+        value={state}
+        onChange={setState}
+        options={[
+          { label: "Temporary", value: "a" },
+          { label: "Lot Determined", value: "b" },
+        ]}
+        onBlur={action("onBlur")}
+        onFocus={action("onFocus")}
+      />
+    </FormLines>
+  );
+}
--- a/src/inputs/RadioGroupField.stories.tsx
+++ b/src/inputs/RadioGroupField.stories.tsx
@@ -296,7 +296,7 @@ export function HelperText() {
 }
 
 export function HorizontalLayout() {
-  const [state, setState] = useState<string | undefined>("a");
+  const [state, setState] = useState<string | undefined>("option-1");
   return (
     <FormLines width="md">
       <RadioGroupField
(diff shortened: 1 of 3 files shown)


src/inputs/RadioGroupField.stories.tsx
```diff
@@ -294,3 +294,23 @@ export function HelperText() {
     />
   );
 }
+
+export function HorizontalLayout() {
+  const [state, setState] = useState<string | undefined>("a");
```
stephenh: @allangaldinosilva could we have another story that also shows what this looks like with ~10 options? Like how it would wrap basically -- I assume the intent is that "we'd never use the horizontal layout with 10 options" but 🤷 seems worth covering from an edge case side of things.

src/inputs/RadioGroupField.tsx
```diff
@@ -70,26 +85,29 @@ export function RadioGroupField<K extends string>(props: RadioGroupFieldProps<K>
...
-                    isOptionDisabled={!!option.disabled}
-                    {...otherProps}
-                    {...tid[option.value]}
-                  />
-                ),
-              })}
-            </Fragment>
-          );
-        })}
+        <div css={Css.df.if(layout === "horizontal").fdr.gap3.else.fdc.$}>
```
bdow: **suggestion:** Might want to do a flex wrap here

src/inputs/RadioGroupField.stories.tsx
allangaldinosilva: Good call @stephenh, this is the result with 25 options:
Screenshot 2026-05-14 at 6 27 30 PM

To avoid this shrink I added `flex-wrap: wrap` then they are wrapped:
Screenshot 2026-05-14 at 6 27 39 PM

src/inputs/RadioGroupField.tsx
```diff
@@ -132,7 +152,17 @@ function Radio<K extends string>(props: {
   const { hoverProps, isHovered } = useHover({ isDisabled: disabled });
 
   return (
-    <label css={Css.df.cursorPointer.mb1.if(disabled).add("cursor", "initial").$} {...hoverProps}>
+    // In vertical layout, each option uses `mb1` for stacking spacing.
+    // In horizontal layout, spacing comes from the parent flex `gap`, so we skip `mb1`.
```
bdow: **question:** Could we use the flex gap property for the vertical layout and get rid of `mb1` all together here?

src/inputs/RadioGroupField.tsx
allangaldinosilva: Done 👍

src/inputs/RadioGroupField.tsx
allangaldinosilva: Yeah, it works fine, I was able to stop passing the `layout` down to the `Radio` with this change, also got ride of the `mb1` 👍 thanks

bdow (approved):

PR #1246, opened 2026-05-22
feat: Add SideNav layout and components

bdow:
# Changes include: 

### SideNav & SideNavLayout components
New layout components `SideNavLayout` and `SideNav` will be a replacement for the same components in `internal-frontend`.
This also standardizes the layout of the links and groups of links via accordions in the SideNav.

### Motion tokens    
Adds motion definitions for the tokens.json, following motion patterns from designtokens.org specification (https://www.designtokens.org/tr/2025.10/format/#cubic-bezier).
Introduces codegen to add these token definitions into the Truss config

### Z-Index organization
Adds a zIndices.tsx file to define all layering across the app. This is helpful to have in one place to quickly understand how elements stack and interact with eachother.
Updates all uses of zindex across the app to utilize the util file


.vscode/settings.json
apattersonATX-HB, line 1: question: Did you update this file intentionally?

bdow: yes. It wasn't doing anything for me using eslint on save vs using prettier formatting.

scripts/dtcg-shared.ts
```diff
@@ -58,6 +58,33 @@ export type DtcgSrgbColorValue = {
   hex?: string;
 };
 
+/** DTCG 2025.10 — `$type: "duration"` (value + unit `ms` | `s`). */
+export type DtcgDurationValue = {
+  value: number;
+  unit: "ms" | "s";
+};
+
+export function isDtcgDurationValue(v: unknown): v is DtcgDurationValue {
```
stephenh, line 67: Kinda curious how much of this design token support could eventually upstream into truss itself -- a lot easier to prototype here so 👍 to that / can keep doing that for awhile.

src/components/DnDGrid/DnDGrid.tsx
```diff
@@ -2,6 +2,7 @@ import equal from "fast-deep-equal";
 import React, { KeyboardEvent, ReactNode, useCallback, useRef } from "react";
 import { Css, Palette, Properties, useTestIds } from "src";
 import { clearInlineStyles, isDefined, setInlineStyles } from "src/utils";
+import { zIndices } from "src/utils/zIndices";
```
stephenh, line 5: Maybe zIndices could be a design token similar to motions that are not variable based, but just consts?

src/components/NavLinks/NavLink.tsx
```diff
@@ -22,14 +24,21 @@ export type NavLinkProps = {
   /** HTML attributes to apply to the button element when it is being used to trigger a menu. */
   menuTriggerProps?: AriaButtonProps;
   buttonRef?: RefObject<HTMLElement>;
+  /** Press handler for button-mode NavLinks (no `href`). */
+  onPress?: () => void;
```
stephenh: Could we make this like `Button` which has a single `onClick` that is "either a string or a callback" since we expect the user to only set one at a time:


```
  onClick: ((e: PressEvent) => void) | ((e: PressEvent) => Promise<void>) | string;

```

src/components/SideNav/useSideNavLinkGroupExpanded.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+import {
+  SIDE_NAV_EXPANDED_LINK_GROUPS_STORAGE_KEY,
+  useSideNavLinkGroupExpanded,
+} from "src/components/SideNav/useSideNavLinkGroupExpanded";
+import { useTestIds } from "src/utils";
+import { click, render } from "src/utils/rtl";
+
+describe("useSideNavLinkGroupExpanded", () => {
+  afterEach(() => {
+    window.localStorage.removeItem(SIDE_NAV_EXPANDED_LINK_GROUPS_STORAGE_KEY);
```
stephenh: In theory we have a suite global localStorage reset? Or if not, we should...

src/components/DnDGrid/DnDGrid.tsx
bdow: I could definitely zIndices into the truss-config so we have helpers like `zModal` or similarly named. But I don't think they belong in `tokens.json`. My initial reason for thinking this way is that the design token goal is for shared properties across platforms, and these z-indices seem pretty specific to just the web implementation of our design system.
Happy to chat through this. Are you suggesting they go in `tokens.json` or just added to `truss-config` - or both in the sense the `tokens.json` zindices feed into the `truss-config` as just consts?

src/components/NavLinks/NavLink.tsx
bdow: Yup, 100%. Good call out.

src/components/SideNav/useSideNavLinkGroupExpanded.test.tsx
bdow: Yup, good call. Will update

src/components/NavLinks/NavLink.tsx
bdow: Caused a bit more churn in this PR with updating all the `href="..."` to `onClick="..."`, but hopefully for the better.

src/components/DnDGrid/DnDGrid.tsx
stephenh: I was initially thinking tokens.json , but true they are probably web-only :thinking: -- I suppose, kind of talking myself out of it, if they were in the truss-config -- would anything else really reference them? Probably not?

So now it seems pretty fine as-is :thinking:

src/components/SideNav/SideNav.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+        label: "Budgets",
+        links: [
+          { label: "Budget", icon: "search", onClick: "/budget" },
+          { label: "POs", icon: "search", onClick: "/pos" },
+        ],
+      },
+    ];
+  }
+
+  function createUnlabeledSectionItems(): SideNavItem[] {
```
stephenh: If these fixtures are used only once, we should just inline them to help test locality/readability -- and ideally create as much data inline in the test as possible -- granted, no factories here

src/components/SideNav/SideNavItems.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  it("renders nested sections inside a section scope", async () => {
+    const r = await render(<SideNavItems items={[createNestedSection()]} panelCollapsed={false} />, withRouter());
+
+    expect(r.sideNav_section_label).toHaveTextContent("Outer");
+    expect(r.sideNav_section_section_label).toHaveTextContent("Inner");
+    expect(r.getByText("Deep")).toBeInTheDocument();
+  });
+
+  function createNestedSection(): SideNavItem {
```
stephenh, line 23: Same nudge on inlining the fixtures

src/components/SideNav/SideNavLinkGroup.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export function SideNavLinkGroupView(props: SideNavLinkGroupViewProps) {
+  const { linkGroup, panelCollapsed } = props;
+  const tid = useTestIds(props, "linkGroup");
+
+  if (panelCollapsed) {
+    return (
+      <>
+        {linkGroup.links.map((link) => (
+          <NavLink key={kebabCase(link.label)} variant="side" {...link} iconOnly={!!link.icon} {...tid.link} />
```
stephenh: I don't think the `key` value is every used anywhere, so we could skip the `kebabCase`?

src/components/SideNav/SideNavLinkGroup.tsx
```diff
@@ -0,0 +1,35 @@
...
+        ))}
+      </>
+    );
+  }
+
+  return <SideNavLinkGroupDisclosure {...props} />;
+}
+
+function SideNavLinkGroupDisclosure(props: SideNavLinkGroupViewProps) {
```
stephenh, line 28: Just curious, what is the "...Disclosure" terminology mean/for?

src/components/SideNav/SideNavSectionView.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    expect(r.sideNav_section_label).toHaveTextContent("Workspace");
+    expect(r.sideNav_section_link).toBeInTheDocument();
+    expect(r.sideNav_section_linkGroup_trigger).toHaveTextContent("Settings");
+  });
+
+  it("omits the section label when the section is unlabeled", async () => {
+    const r = await render(<Harness section={{ items: [{ label: "Help", onClick: "/help" }] }} />, withRouter());
+
+    expect(r.queryByTestId("sideNav_section_label")).toBeNull();
```
stephenh: We can do:

```
expect(r.query.sideNav_section_label).toBeNull();
```

now :-)

src/components/SideNav/sideNavUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function isSideNavSection(item: SideNavItem): item is SideNavSection {
+  return "items" in item && Array.isArray(item.items);
+}
+
+export function sideNavItemKey(item: SideNavItem | SideNavSectionItem): string {
+  if (isSideNavLink(item)) return kebabCase(item.label);
+  if (isSideNavLinkGroup(item)) return kebabCase(item.label);
+  if (item.label) return kebabCase(item.label);
```
stephenh: Same thinking that React `key`s don't need formatted, afaiu... (disclaimer I didn't re-open all the files in this PR to search for non `key=...` callers 😅 )

src/components/SideNav/useSideNavLinkGroupExpanded.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    window.localStorage.setItem(SIDE_NAV_EXPANDED_LINK_GROUPS_STORAGE_KEY, JSON.stringify({ Budgets: false }));
+
+    const r = await render(
+      <Harness linkGroup={createLinkGroup({ links: [{ label: "Budget", onClick: "/budget", active: true }] })} />,
+    );
+
+    expect(r.linkGroupExpanded_expanded).toHaveTextContent("false");
+  });
+
+  function createLinkGroup(overrides: Partial<SideNavLinkGroup> = {}): SideNavLinkGroup {
```
stephenh: These could probably be top-level instead of in the `describe`

stephenh (approved):

src/components/SideNav/SideNavLinkGroup.tsx
bdow: A disclosure is the industry term for content hidden with an accordion and the like: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details. 
React-aria also has a 'useDisclosure' hook as well. I tried to use it, but it doesn't animate well.

src/components/SideNav/SideNavLinkGroup.tsx
bdow: Yea, sure. I always kind of viewed "keys" as `id`'s in HTML, and had preferred them formatted more that way. But I guess they really don't need it.

PR #1247, opened 2026-05-22
feat: Update support to use React Router v6

bdow:


src/components/ButtonMenu.test.tsx
```diff
@@ -3,12 +3,12 @@ import { Button } from "src/components/Button";
...
 import { Css } from "src/Css";
 import { noop } from "src/utils";
-import { click, render, type } from "src/utils/rtl";
+import { click, render, type, withRouter } from "src/utils/rtl";
 
 describe("ButtonMenu", () => {
   it("can update menu items", async () => {
     // Given a Button Menu
-    const r = await render(<TestButtonMenu />);
+    const r = await render(<TestButtonMenu />, withRouter());
```
stephenh, line 11: We could potentially add `withRouter` to our `src/utils/rtl.rb` by default but np either way

stephenh (approved):

PR #1249, opened 2026-06-01
feat: Page Header

abadami:
## Page Header

A new `PageHeader` component added. Currently only supporting a `title` and `rightSlot` property to get what's needed for Product Offerings. Made quick notes and preparation for the breadcrumbs and tabs portions of the component once we get to that stage so it should (hopefully) be an pretty easy add once we get there.

### DIFF
--- a/src/components/PageHeader.stories.tsx
+++ b/src/components/PageHeader.stories.tsx
@@ -1,7 +1,6 @@
 import { Meta } from "@storybook/react-vite";
 import { Button } from "src/components/Button";
 import { PageHeader } from "src/components/PageHeader";
-import { Css } from "src/Css";
 import { action } from "storybook/actions";
 
 export default {
@@ -13,20 +12,14 @@ export default {
 } as Meta;
 
 export function NoRightSlot() {
-  return (
-    <div css={Css.bgWhite.$}>
-      <PageHeader title="Test Title" />
-    </div>
-  );
+  return <PageHeader title="Test Title" />;
 }
 
 export function WithRightSlot() {
   return (
-    <div css={Css.bgWhite.$}>
-      <PageHeader
-        title="Test Title"
-        rightSlot={<Button label="Test Action" variant="primary" onClick={action("clicked")} />}
-      />
-    </div>
+    <PageHeader
+      title="Test Title"
+      rightSlot={<Button label="Test Action" variant="primary" onClick={action("clicked")} />}
+    />
   );
 }
--- a/src/components/PageHeader.stories.tsx
+++ b/src/components/PageHeader.stories.tsx
@@ -1,6 +1,8 @@
 import { Meta } from "@storybook/react-vite";
+import { Button } from "src/components/Button";
 import { PageHeader } from "src/components/PageHeader";
 import { Css } from "src/Css";
+import { action } from "storybook/actions";
 
 export default {
   component: PageHeader,
@@ -21,7 +23,10 @@ export function NoRightSlot() {
 export function WithRightSlot() {
   return (
     <div css={Css.bgWhite.$}>
-      <PageHeader title="Test Title" rightSlot={<span>Right Slot</span>} />
+      <PageHeader
+        title="Test Title"
+        rightSlot={<Button label="Test Action" variant="primary" onClick={action("clicked")} />}
+      />
     </div>
   );
 }
--- a/src/components/PageHeader.stories.tsx
+++ b/src/components/PageHeader.stories.tsx
@@ -1,0 +1,27 @@
+import { Meta } from "@storybook/react-vite";
+import { PageHeader } from "src/components/PageHeader";
+import { Css } from "src/Css";
+
+export default {
+  component: PageHeader,
+  parameters: {
+    type: "figma",
+    url: "https://www.figma.com/design/yJcdAYy1rqkxTTNhUQnjjU/H2-2026-Beam-Work?m=dev",
+  },
+} as Meta;
+
+export function NoRightSlot() {
+  return (
+    <div css={Css.bgWhite.$}>
+      <PageHeader title="Test Title" />
+    </div>
+  );
+}
+
+export function WithRightSlot() {
+  return (
+    <div css={Css.bgWhite.$}>
+      <PageHeader title="Test Title" rightSlot={<span>Right Slot</span>} />
+    </div>
+  );
+}
--- a/src/components/PageHeader.test.tsx
+++ b/src/components/PageHeader.test.tsx
@@ -5,18 +5,13 @@ import { noop } from "src/utils";
 
 describe("PageHeader", () => {
   it("renders", async () => {
-    const r = await render(<PageHeader title="Test Title" data-testid={"pageHeader"} />);
+    const r = await render(<PageHeader title="Test Title" />);
     expect(r.pageHeader).toBeInTheDocument();
-    expect(r.pageHeader.textContent).toEqual("Test Title");
+    expect(r.pageHeader_title.textContent).toEqual("Test Title");
   });
 
   it("renders with right slot", async () => {
-    const r = await render(
-      <PageHeader
-        title="Test Title"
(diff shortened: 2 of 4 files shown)


src/components/PageHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+export function PageHeader<X extends Only<Xss<PageHeaderXss>, X>>(props: PageHeaderProps<X>) {
+  const { title, rightSlot, xss, ...otherProps } = props;
+  const tid = useTestIds(otherProps);
+
+  return (
+    <header {...tid} css={{ ...Css.df.fdc.pt3.pr3.pl3.$, ...xss }}>
+      <div css={Css.df.jcsb.mb2.w100.aic.$}>
+        <div>
+          {/* Breadcrumbs here */}
+          <h2 css={Css.xl.$}>{title}</h2>
```
abadami: `h2` element over `h1` felt right since it wasn't using the max `xl2` size, but can be easily convinced since h1 might make more sense being a page header and all.

src/components/PageHeader.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+    type: "figma",
+    url: "https://www.figma.com/design/yJcdAYy1rqkxTTNhUQnjjU/H2-2026-Beam-Work?m=dev",
+  },
+} as Meta;
+
+export function NoRightSlot() {
+  return (
+    <div css={Css.bgWhite.$}>
+      <PageHeader title="Test Title" />
+    </div>
```
bdow: **nit:** Would prefer to not have these wrapping divs -  unless they are really needed?

src/components/PageHeader.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+  });
+
+  it("renders with right slot", async () => {
+    const r = await render(
+      <PageHeader
+        title="Test Title"
+        rightSlot={<Button data-testid={"exampleButton"} label="Test Button" onClick={noop} />}
+      />,
+    );
+    expect(r.exampleButton).toBeInTheDocument();
```
bdow: **suggestion:** The `<Button />` component will automatically generate a `data-testid` based on the label. So this one should result in `data-testid="testButton"`.
Then we can update the assertion to be `r.testButton.toBeInTheDocument()`.

src/components/PageHeader.tsx
```diff
@@ -0,0 +1,28 @@
+import { ReactNode } from "react";
+import { Css, Margin, Only, Xss } from "src/Css";
+import { useTestIds } from "src/utils";
+
+type PageHeaderXss = Margin;
+export interface PageHeaderProps<X> {
+  title: ReactNode;
+  rightSlot?: ReactNode;
+  xss?: X;
```
bdow: **suggestion:** I wouldn't add in xss unless we have an existing need for it. `xss` properties are typically to allow users to override specific css defaults. I'm hoping we won't need to do that here.

src/components/PageHeader.tsx
bdow: This will most likely be used at the title of the page, so would prefer to use `h1`.

One interesting thing to note is the current `PageHeader` component in BP uses a [`useDocumentTitle`](https://github.com/homebound-team/internal-frontend/blob/main/src/hooks/useDocumentTitle.tsx) hook which also updates the `document.title`/`<title>` tag based on this header's title. 🤔 I wonder if we should do a similar thing here. Not something to do here now, but just something to think about.

src/components/PageHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export function PageHeader<X extends Only<Xss<PageHeaderXss>, X>>(props: PageHeaderProps<X>) {
+  const { title, rightSlot, xss, ...otherProps } = props;
+  const tid = useTestIds(otherProps);
+
+  return (
+    <header {...tid} css={{ ...Css.df.fdc.pt3.pr3.pl3.$, ...xss }}>
+      <div css={Css.df.jcsb.mb2.w100.aic.$}>
```
bdow: **suggestion:** Probably worth putting a `gap` value here to avoid the two elements colliding on smaller screens.

src/components/PageHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+type PageHeaderXss = Margin;
+export interface PageHeaderProps<X> {
+  title: ReactNode;
+  rightSlot?: ReactNode;
+  xss?: X;
+}
+
+export function PageHeader<X extends Only<Xss<PageHeaderXss>, X>>(props: PageHeaderProps<X>) {
+  const { title, rightSlot, xss, ...otherProps } = props;
+  const tid = useTestIds(otherProps);
```
bdow: **suggestion:** Kind of our best practices with the `useTestIds` hook is to provide a default value. E.g. `const tid = useTestIds(otherProps, "pageHeader");`
Then 
```
<header {...tid}>
  <h2 {...tid.title}>{title}</h2>
</header>
``` 
becomes 
```
<header data-testid="pageHeader">
  <h2 data-testid="pageHeader_title">Foo</h2>
</header>  
```

Then in tests, we can simply reference those values and not pass in our own `data-testid="foo"` in the test render method.

src/components/PageHeader.tsx
bdow: **nit:** Small thing, but I _think_ the Figma shows the vertically alignment as the top. So instead of `.aic.$`, we'd probably want to leave it as is (defaulting to `align-items: stretch`), or set to `.ais.$`.

src/components/PageHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    <header {...tid} css={{ ...Css.df.fdc.pt3.pr3.pl3.$, ...xss }}>
```
bdow: **suggestion:** Per the design, I think we'll want a border-bottom set, using `Gray200`. Or better yet, we could use our design tokens and set it via `bc(Tokens.SurfaceSeparator)`.

src/components/PageHeader.stories.tsx
abadami: Yeah that's no problem, I'll remove it. I was mostly using it as an easy visual test to confirm the margins but I could've just used the dev console for that in hindsight.

src/components/PageHeader.tsx
abadami: You're correct, I missed that. Thanks for catching it!

src/components/PageHeader.tsx
abadami: Oh that's really slick! Will do

src/components/PageHeader.tsx
abadami: I just checked this myself and you are correct! I feel like it may make more sense for that piece to be centered if there are no breadcrumbs present, but that's more of a design question and looking at the Product Offering page in Figma I didn't even notice it wasn't centered to begin with.

abadami: Good call, I forgot to check out smaller screens.

src/components/PageHeader.tsx
abadami: Yeah I was on the fence of adding it myself because theoretically we'd never need to override these styles, but felt like I should since most all other components do. I think keeping it off until we find that use case is a good option here.

bdow (approved):

src/components/PageHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+  title: ReactNode;
+  rightSlot?: ReactNode;
+}
+
+export function PageHeader(props: PageHeaderProps) {
+  const { title, rightSlot, ...otherProps } = props;
+  const tid = useTestIds(otherProps, "pageHeader");
+
+  return (
+    <header {...tid} css={Css.df.fdc.pt3.pr3.pl3.bb.bc(Tokens.SurfaceSeparator).$}>
```
bdow: **suggestion:** One more thing - We probably want to use `bgColor(Tokens.Surface)` on this. Giving it a background color would prevent content below it (if this stays sticky while scrolling) from showing.

PR #1251, opened 2026-06-02
feat: add preventTooltip opt-out to IconButton and Avatar

bdow:
IconButton and AvatarButton surface their `label`/`name` as a hover tooltip by default. Add a `preventTooltip` prop to keep the `aria-label` for screen readers without the tooltip (an explicit `tooltip` or disabled reason still shows). 
Renames Avatar/AvatarButton's existing `disableTooltip` to `preventTooltip` for consistency, and to disambiguate from the existing tooltips that show when an element is considered "disabled"


allangaldinosilva (approved):
Looks good to me!

PR #1255, opened 2026-06-02
feat: Page Header Tabs

abadami:
## Page Header Tabs

Adding support for Tabs in the page header! Set via a `tabs` prop that accepts the same props as the `Tabs` component.


src/components/PageHeader.tsx
```diff
@@ -1,14 +1,17 @@
...
+import { RouteTabsProps, Tabs, TabsProps } from "src/components/Tabs";
 import { Css, Tokens } from "src/Css";
+import { AnyObject } from "src/types";
 import { useTestIds } from "src/utils";
 
-export interface PageHeaderProps {
+export interface PageHeaderProps<V extends string> {
   title: ReactNode;
   rightSlot?: ReactNode;
+  tabs?: TabsProps<V, AnyObject> | RouteTabsProps<V, AnyObject>;
```
bdow: **request:** Let's get rid of `AnyObject` here.
This second generic is meant to restrict what styles a caller can apply to the component. It looks like the current Tabs component doesn't handle this 100%, and probably what led us to here.
But what we should do is:
1. Update the `Tabs.tsx` file, so that the `TabContent` and `Tabs` components no longer use `AnyObject` and instead define the generic properly, similar to how `TabsWithContent` already does it.
Example:
```
-export function TabContent<V extends string>(
-  props: Omit<RequiredRenderTabs<V, AnyObject>, "onChange"> | RequiredRenderRouteTabs<V, AnyObject>,
+export function TabContent<V extends string, X extends Only<TabsContentXss, X>>(
+  props: Omit<RequiredRenderTabs<V, X>, "onChange"> | RequiredRenderRouteTabs<V, X>,
```
2. Export `type TabsContentXss` from `Tabs.tsx` and utilize that in the `PageHeader` component
```
export interface PageHeaderProps<V extends string, X> {
   title: ReactNode;
   rightSlot?: ReactNode;
   tabs?: TabsProps<V, X> | RouteTabsProps<V, X>;
 }
 
export function PageHeader<V extends string, X extends Only<TabsContentXss, X>>(props: PageHeaderProps<V, X>) {
```

This then should ensure we're enforcing that rule properly across the Tabs integration.

bdow: Granted, I don't think anyone would be passing `contentXss` through these props as we're only ever displaying the tab bar here and not the `TabContent` in which the `contentXss` would apply. Either way, I think adding the restriction in properly is still a good practice and we should move forward doing so.

src/components/PageHeader.stories.tsx
```diff
@@ -23,3 +27,34 @@ export function WithRightSlot() {
     />
   );
 }
+
+export function WithTabs() {
```
bdow, line 33: **suggestion (non-blocking):** For these `WithTabs` and `WithRightSlotAndTabs` stories, can please include the `<TabContent />` component under the page header as well? I think that'd help just see the expected set up.

src/components/PageHeader.stories.tsx
```diff
@@ -9,6 +12,7 @@ export default {
     type: "figma",
     url: "https://www.figma.com/design/yJcdAYy1rqkxTTNhUQnjjU/H2-2026-Beam-Work?m=dev",
   },
+  decorators: [withRouter("/ce:2", "/:ceId/*"), withBeamDecorator],
```
bdow: **nit:** I don't think we need the route definitions in here as we're not using Route tabs in these examples - probably could get away with just `withRouter()`

src/components/PageHeader.stories.tsx
```diff
@@ -9,6 +12,7 @@ export default {
     type: "figma",
     url: "https://www.figma.com/design/yJcdAYy1rqkxTTNhUQnjjU/H2-2026-Beam-Work?m=dev",
```
bdow, line 14: **suggestion:** I think this story might look better if we adding the `layout: "fullscreen"` parameter in this object.

src/components/PageHeader.test.tsx
```diff
@@ -14,4 +15,35 @@ describe("PageHeader", () => {
...
+      <PageHeader
+        title="Test Title"
+        tabs={{
+          tabs,
+          selected: "tabA",
+          onChange: noop,
+        }}
+      />,
+      withRouter(),
+    );
```
bdow, line 35: **suggestion:** Something we try to do here at HB is to try and reduce the number of lines of code when possible. This is one good example where we could improve, and hopefully helps readability by reducing need to scroll. This could be reduced from 25 loc to 9:
```
    const tabs: Tab[] = [
      { name: "Tab A", value: "tabA" },
      { name: "Tab B", value: "tabB" },
      { name: "Tab C", value: "tabC" },
    ];
    const r = await render(
      <PageHeader title="Test Title" tabs={{ tabs, selected: "tabA", onChange: noop }} />,
      withRouter(),
    );
```

bdow: Keep a look out for other opportunities to reduce LOC! I'm sure there are existing areas that we could improve. But just something to keep in mind going forward.

src/components/PageHeader.stories.tsx
abadami: Interestingly I was getting `useLocation` errors without it. I might check really quick and see why that's happening and prevent it from occurring if it makes sense to do so.

bdow: I'm guessing you'll get those without a `withRouter()` at all. But you shouldn't need to specify paths/uris in the `withRouter` call unless you're matching against them.

src/components/PageHeader.test.tsx
abadami: Sounds good! I'll try to keep that in mind.

src/components/PageHeader.stories.tsx
abadami: Yeah that makes a lot of sense to me.

src/components/PageHeader.stories.tsx
abadami: Yeah that's fair. Plus I don't think I'll be able to easily prevent the need for a router anyway due to the conditional hook restriction ofc. I'll just remove the routes for now.

src/components/PageHeader.tsx
abadami: Yeah, I was wondering while I was doing this if I should omit the `contentXss` prop here due to not actually using content for this. I admittedly kinda forgot to go back and look into that.

Either way though I do agree that Tabs should be using the appropriate practice instead of `AnyObject` so I went ahead and did it.

src/components/PageHeader.tsx
```diff
@@ -1,14 +1,18 @@
...
+import { Css, Only, Tokens } from "src/Css";
 import { useTestIds } from "src/utils";
 
-export interface PageHeaderProps {
+export interface PageHeaderProps<V extends string, X> {
   title: ReactNode;
   rightSlot?: ReactNode;
+  tabs?:
+    | Omit<TabsProps<V, X>, "contentXss" | "omitFullBleedPadding" | "includeBottomBorder">
+    | Omit<RouteTabsProps<V, X>, "contentXss" | "omitFullBleedPadding" | "includeBottomBorder">;
```
abadami, line 11: I did omit `contentXss`, `omitFullBleedPadding`, and `includeBottomBorder` props from the Page Header tabs prop options, since the first two won't apply and the bottom border is already provided by `PageHeader`

bdow (approved):

PR #1257, opened 2026-06-03
feat: Navbar component

bdow:
Adds the Navbar component (top global nav) built on the shared AppNav model, with a mobile/overflow collapse to a hamburger menu via the useContentOverflow hook. Includes HomeboundLogo, viewportModes story helper, and Storybook docs.


src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,92 @@
...
+          <div css={Css.df.aic.fs0.gap2.$}>
+            {showMobile && <NavbarMobileMenu items={items} {...tid} />}
+            <div css={Css.fs0.$} {...tid.brand}>
+              {brand}
+            </div>
+          </div>
+          {!sm && (
+            // Stays mounted while overflowing (hidden) so the items remain measurable and the bar
+            // can expand again as space frees up.
+            <div ref={containerRef} css={Css.df.aic.fg1.mw0.oh.if(overflows).visibility("hidden").$} {...tid.items}>
```
abadami, line 60: **note:** I did pull this down to mess around with the storybook a little and I did find a potential scenario in the "With Open Nav" story if that setting is set to be open by default for whatever reason and we're in about a tablet sized view, then the drop down menu will always be visible even though we're in mobile mode. This is likely due to that overlay being an absolutely position element under the body element, so it doesn't get captured by the visibility rule here.

I don't think this scenario is likely to come up as I'm not sure if that menu setting is used in this way at all, but just wanted to note that it is something that could happen.

image

src/components/Navbar/NavbarMobileMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const { pathname, search } = useLocation();
+
+  useEffect(() => {
+    if (isOpen) {
+      document.body.style.overflow = "hidden";
+      return () => {
+        document.body.style.overflow = "";
+      };
+    }
+  }, [isOpen]);
```
apattersonATX-HB: thought: could `usePreventScroll` work just as well here?

src/components/index.ts
apattersonATX-HB, line 1: nit: could also export `HomeboundLogo` from here

src/hooks/useContentOverflow.ts
```diff
@@ -0,0 +1,47 @@
...
+ * static after mount.
+ *
+ * @param enabled When `false`, measurement is skipped and `overflows` is forced to `false`.
+ */
+export function useContentOverflow(enabled: boolean = true) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const contentRef = useRef<HTMLDivElement>(null);
+  const [overflows, setOverflows] = useState(false);
+
+  const check = useCallback(() => {
```
apattersonATX-HB: nit: 👺 


```suggestion
  const checkContentOverflow = useCallback(() => {
```

src/components/Navbar/NavbarMobileMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+          {/* AnimatePresence keeps the drawer mounted through its slide/fade-out before unmounting. */}
+          <AnimatePresence>{isOpen && <NavbarMobileDrawer items={items} onClose={close} tid={tid} />}</AnimatePresence>
+        </ContrastScope>,
+        document.body,
+      )}
+    </>
+  );
+}
+
+function NavbarMobileDrawer({
```
apattersonATX-HB, line 56: question: Are we testing on **real** devices 👀

src/components/HomeboundLogo.tsx
```diff
@@ -0,0 +1,17 @@
+import { BeamColor } from "src/colors";
+import { Css } from "src/Css";
+
+type HomeboundLogoProps = {
+  fill?: BeamColor;
+  width?: number | "auto";
+  height?: number | "auto";
+};
+
+export function HomeboundLogo(props: HomeboundLogoProps) {
```
abadami, line 10: **suggestion (non-blocking):** I wonder if it'd be worth it to go ahead and offer the `HomeboundLogo` as a component provided by Beam. That being said, that's something we can do whenever and I doubt there is much use case for it now. I could maybe see there being a use case in the future, but probably not something to worry about now

src/components/Navbar/NavbarMobileMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+        transition={{ ease: "linear", duration: 0.2 }}
+        onClick={onClose}
+        {...tid.mobileMenuScrim}
+      />
+      <FocusScope contain restoreFocus>
+        <motion.aside
+          key="navbarMobileMenuDrawer"
+          role="dialog"
+          aria-modal="true"
+          aria-label="Navigation"
```
apattersonATX-HB: nit: If we're calling this `aria-modal="true"` we also need `<FocusMode .. autoFocus>`

src/components/Navbar/NavbarMobileMenu.tsx
bdow: Will look. Forgot about that. Not sure if that's conditional and would release scrolling. But will confirm, thanks!

src/components/Navbar/NavbarMobileMenu.tsx
bdow: Is that right? I assumed `contain` would auto focus here to start out with. Will double check, thanks!

src/components/Navbar/NavbarMobileMenu.tsx
bdow: Yes, and taking chromatic snapshots for multiple viewports.

src/hooks/useContentOverflow.ts
```diff
@@ -0,0 +1,47 @@
...
+ * observe `containerRef` or re-measure on every commit: if reacting to `overflows` reflows the
+ * container (changing its `clientWidth`), observing it would re-trigger the measurement, flip the
+ * result, and loop ("Maximum update depth exceeded"). The content's natural width is treated as
+ * static after mount.
+ *
+ * @param enabled When `false`, measurement is skipped and `overflows` is forced to `false`.
+ */
+export function useContentOverflow(enabled: boolean = true) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const contentRef = useRef<HTMLDivElement>(null);
```
abadami: **suggestion:** Is there any scenario where these ref elements could be something other than divs? I could maybe see it at least for the container ref with some elements like `section` for example. Granted, `section` is the only other example I can think of on the top of my head.

src/components/Navbar/Navbar.tsx
bdow: Gotcha. I think this story is also a little odd as it is forcing this menu open. You're right it shouldn't happen. The only way to open this menu when showing its "mobile" size is by programmatically doing so. Good edge case to call out and track.

src/hooks/useContentOverflow.ts
bdow: Yup, I don't see why not. Can update.

src/components/index.ts
bdow: Sure, can do!

src/components/Navbar/NavbarMobileMenu.tsx
bdow: fwiw, if you ever want to try this out, you can do it via:
```
yarn start --host [your computer IP] -p 9001
```
Then, as long as you're on the same network, your phone can open up the local address
Screenshot 2026-06-04 at 9 06 37 PM

src/components/Navbar/NavbarMobileMenu.tsx
bdow: Works well here. Thanks!

src/components/Navbar/NavbarMobileMenu.tsx
bdow: Looks like `contain` doesn't initially contain it. You do also need `autoFocus`. And that's what I get for assuming 😄. Thanks!

src/components/Navbar/Navbar.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+    <Navbar
+      brand={createBrand()}
+      items={createGlobalNavItems({ activeLabel: "Finances", librariesMenuOpen: true })}
+      rightSlot={createDefaultRightSlot()}
+      user={createDefaultUser()}
+    />
+  );
+}
+
+WithOpenNav.play = async ({ canvasElement }: { canvasElement: HTMLElement }) => {
```
stephenh: Small nudge on the `setStory` / `play` DSL ...

Similarly curious if the `fullscreen` + `chromatic / modes`, from the top of the file, could be `setStory` DSL-d as well -- I know I am probably too paranoid about it, although I assert it is a legitimate paranoia 😅 , that the storybook / chromatic / etc DSL always seems to change every ~year or so.

I.e. our `plan-configuration-frontend` storybook upgrade was kind of a mess b/c we had ~2-3 ways of "setting the wait storybook viewport, wait chromatic mode, wait what now?" that just got confusing/ugly over time.

Granted, the `setStory` would mean repeating some of the props for each story (i.e. yes every story here uses the same viewport/etc), but kinda feel like that is okay, b/c ideally the `setStory` API/DSL is really declarative / ergonomic.

And so it's worth the upshot / win, of having just 1 spot in the repo that translates are "I want viewport a/b/c" to "whatever today's storybook API is".

bdow: Yup, can do.

bdow: Alrighty - updated.

abadami (approved):

PR #1258, opened 2026-06-03
fix: add toast to table review layout;

apattersonATX-HB:
- Both `TableReviewLayout` and `FormPageLayout` used an arbitrary `z(1000)` index. Replaced it with `pageOverlay: 1000` added to `zIndices`. Sits above BP navbar.
- `TableReviewLayout` now owns its own `<Toast />` inside the header, matching the pattern `FormPageLayout` already had
- Added some `WithToastVisible` stories to both components

Screenshot 2026-06-03 at 3 24 52 PM

----

This fixes an issue called out in BP accept plan package incoming slot. [APP-1489](https://linear.app/homebound-team/issue/APP-1489/bug-fe-validation-error-banners-do-not-show-in-the-slot-review-ux)

### DIFF
--- a/src/utils/zIndices.ts
+++ b/src/utils/zIndices.ts
@@ -23,7 +23,10 @@ export const zIndices = {
   // snackbar so toasts still land on top.
   sideNav: 100,
 
-  // Top of the stack.
+  // Full-page overlays sit above side-nav and use their own toast internally
+  pageOverlay: 110,
+
+  // Top of stack
   snackbar: 120,
 } as const;
 
--- a/src/utils/zIndices.ts
+++ b/src/utils/zIndices.ts
@@ -23,11 +23,12 @@ export const zIndices = {
   // snackbar so toasts still land on top.
   sideNav: 100,
 
-  // Full-page overlays sit above side-nav and use their own toast internally
-  pageOverlay: 110,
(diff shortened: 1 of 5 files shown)


src/utils/zIndices.ts
```diff
@@ -23,8 +23,12 @@ export const zIndices = {
   // snackbar so toasts still land on top.
   sideNav: 100,
 
-  // Top of the stack.
-  snackbar: 120,
+  // Full-page overlays — high enough to clear consuming-app nav bars (~999).
+  // Both layouts own their Toast internally so it renders inside the overlay header.
+  pageOverlay: 1000,
```
apattersonATX-HB, line 28: @bdow Originally i had this at a cool 110 but blueprint's navbar being set to z-999 broke it. 

Image

 Not sure if you had planned to touch zIndices in your [navbar](https://github.com/homebound-team/beam/pull/1257) work.

bdow: Yes, as part of the related layout work.

bdow (approved):

src/utils/zIndices.ts
apattersonATX-HB: awesome

PR #1261, opened 2026-06-05
issue: reproduce conditional action column rendering issue

allangaldinosilva:
**Problem**
GridTable's action column (e.g., trash icons) intermittently disappears after toggling between read and edit modes when the column is conditionally rendered via spread operator (`...(isEdit ? [actionColumn] : [])`).

https://github.com/user-attachments/assets/27230055-9082-4749-8957-ab0546dd59b3

cc @bdow

PR #1263, opened 2026-06-08
fix: keep conditional action column visible across remounts

allangaldinosilva:
### Summary

A conditionally-rendered action column (e.g. a trash icon shown only in edit mode) would sometimes disappear in edit mode after navigating away and back. This fixes the leaked MobX watcher that was corrupting the column's `sessionStorage` visibility state.

The issue was caught by [#1261](https://github.com/homebound-team/beam/pull/1261), which provides a Storybook repro and a screen recording demonstrating the column disappearing.

### Root cause

Visible columns are persisted to `sessionStorage` under a key auto-derived from the column ids, so the key differs per mode:
- read mode → `nameValue`
- edit mode → `nameValueAction`

Each `columns` change wired up a new `autorun` to write the visible set to that mode's key, but **the previous watcher was never disposed**. After entering edit mode, the `nameValueAction` watcher kept firing forever. Back in read mode (action column gone), it wrote `["name","value"]` into the `nameValueAction` key — recording the action column as user-hidden. On the next fresh mount, the table read that key, concluded the column had been hidden, and dropped it.

### Fix

`ColumnStorage` now holds the disposer for each watcher and disposes the previous one before creating a new one (`loadVisible` and `loadExpanded`). Each storage key only ever has one live watcher, so a mode's key is only updated while that mode is active.


bdow (approved):
This is good. I do wonder if we should allow for opting out of column storage. If I recall correctly, this is a much smaller table. Makes me wonder if we should have a slimmed down version of our table component for in-line sort of tables vs full page tables. Something to think about 🤔

PR #1264, opened 2026-06-08
feat: More mobile-friendly GridTableLayout & optional page header

abadami:
## GridTableLayout: Mobile-friendly improvements and optional page header

Right now Page headers are required for the layout, but as we move towards an active `PageHeader` component and preferring declaring headers that way, we no longer need it to be required. Also, made some mobile friendly improvements, especially to the search and filters, so that less space is taken by them unless necessary

### Changes

- `pageTitle` is now an optional prop and will hide the page header if not provided
- The search bar will now be hidden in a little search button that works similarly to the `FilterDropdownMenu` in mobile screens only, otherwise it operates the same
- The search bar was added as an option in the `FilterDropdownMenu` component, that way when having both the search and filter options open in a small screen, the layout of the search and filters are more user friendly.


src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -220,6 +220,22 @@ export function WithPagination() {
...
+    <TestProjectLayout>
+      <GridTableLayoutComponent
+        tableProps={{
+          columns,
+          rows: [simpleHeader, ...makeNestedRows(3)],
+          sorting: { on: "client", initial: [columns[1].id!, "ASC"] },
+        }}
+      />
+    </TestProjectLayout>
+  );
```
bdow, line 236: **thought/suggestion:** I think this `TestProjectLayout` isn't necessary for us to wrap the GridTableLayout with - at least not this story. I also think we could benefit from the mobile chromatic snapshot on this story as well - wdyt?

src/components/Filters/FilterDropdownMenu.test.tsx
```diff
@@ -89,12 +126,21 @@ function TestFilterDropdownMenu<F extends Record<string, unknown>>(props: {
...
+  searchProps?: {
+    onSearch: (value: string) => void;
+  };
 }) {
   const [filter, setFilter] = useState<F>(props.initialFilter || ({} as F));
   return (
-    <div>
-      <FilterDropdownMenu filterDefs={props.filterDefs} filter={filter} onChange={setFilter} groupBy={props.groupBy} />
+    <QueryParamProvider adapter={ReactRouter6Adapter}>
+      <FilterDropdownMenu
```
bdow: **suggestion:** I believe the `withRouter` method already wraps components in the `QueryParamProvider` - so we shouldn't need to wrap again here.
See: https://github.com/homebound-team/rtl-react-router-utils/blob/main/src/index.tsx#L25

If thats not working for some reason, let's discuss.

src/components/Filters/FilterDropdownMenu.tsx
```diff
@@ -61,26 +94,62 @@ function FilterDropdownMenu<F extends Record<string, unknown>, G extends Value =
...
-        active={isOpen}
-        {...testId.button}
-      />
+      {/* Large screen: search field visible inline before filter button */}
+      {!sm && hasSearch && <div css={Css.wPx(244).$}>{searchTextField}</div>}
+
+      {/* Small screen: search icon button inline with filter button */}
+      {sm && hasSearch && (
+        <Button
+          label=""
```
bdow, line 117: **request:** Buttons should always have a label for accessibility reasons. If you're trying to do an icon only button, then use `IconButton`. `label` is optional for that one, but please do include one. It'll be added via `aria-label`.

src/components/Filters/FilterDropdownMenu.tsx
```diff
@@ -61,26 +94,62 @@ function FilterDropdownMenu<F extends Record<string, unknown>, G extends Value =
...
+          label={sm ? "" : "Filter"}
+          icon="filter"
+          size="md"
+          endAdornment={
+            !sm && (
+              <div css={Css.df.aic.gap1.$}>
+                {activeFilterCount > 0 && <CountBadge count={activeFilterCount} />}
+                <Icon icon={isOpen ? "chevronUp" : "chevronDown"} />
+              </div>
+            )
```
bdow: **thought**: I see here that we use `sm` again. I'm kind of surprised we'd want to hide the active filter count on mobile - is that part of the design?
Could also avoid a re-render here by using css to hide. wdyt?

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
bdow: Probably fine to keep as is. I see we use the breakpoint helper again later in `hideEndAdornment={sm}`

src/components/Filters/FilterDropdownMenu.tsx
```diff
@@ -61,26 +94,62 @@ function FilterDropdownMenu<F extends Record<string, unknown>, G extends Value =
...
-        variant="secondaryBlack"
-        onClick={() => setIsOpen(!isOpen)}
-        active={isOpen}
-        {...testId.button}
-      />
+      {/* Large screen: search field visible inline before filter button */}
+      {!sm && hasSearch && <div css={Css.wPx(244).$}>{searchTextField}</div>}
+
+      {/* Small screen: search icon button inline with filter button */}
+      {sm && hasSearch && (
```
bdow, line 115: **thought (non-blocking):** I'm always looking for ways to not live in React's lifecycle. This could be one of those where we don't need the `useBreakpoint` hook here. We could just use css to hide/show the correct element here. Granted, we'd be always putting both in the DOM, but they are relatively small components, that I don't think it'd be a big deal. You can do this in truss via `Css.if("sm").flex.$`

src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -153,13 +151,13 @@ function GridTableLayoutComponent<
...
-          {layoutState?.search && <SearchBox onSearch={layoutState.setSearchString} />}
-          {layoutState?.filterDefs && (
+          {layoutState && (layoutState.filterDefs || layoutState.search) && (
             <FilterDropdownMenu
               filterDefs={layoutState.filterDefs}
               filter={layoutState.filter}
               onChange={layoutState.setFilter}
               groupBy={layoutState.groupBy}
+              searchProps={layoutState.search ? { onSearch: layoutState.setSearchString } : undefined}
             />
```
bdow, line 161: **thought/question:** I'm tempted to want to keep search and filters as separate. I get they are kind of the same thing -> searching is filtering your results list. I'm guessing this is done in order to put the search on the next line, similar to the filters? Might be worth getting design's eyes on this before merging and have feedback from them on the layout for mobile

src/components/Filters/FilterDropdownMenu.tsx
```diff
@@ -32,24 +46,43 @@ interface FilterDropdownMenuProps<F extends Record<string, unknown>, G extends V
     /** The list of group by options. */
     options: Array<{ id: G; name: string }>;
   };
+  /** When provided, renders a search box before the filter controls. */
+  searchProps?: SearchBoxProps;
```
bdow, line 50: **thought:** I think I get why we did this 🤔. Was it in order to get the mobile search bar below this line? 
I wonder if that is what design is requesting (I didn't see a mock for this). 
It does feel like we're kind of retrofitting the Search box in here, which is fine. There's just something about this that I can't put my finger on yet. My mind goes to the "single responsibility principle". And now mixing the search in with the filters (which again, a search is a type of filter, so that's fine). I'm arguing with myself to rationalize it. But I think the first step might be to ask design for guidance on how this is suppose to work on mobile - unless there is a mock and I just missed it.

src/components/Filters/FilterDropdownMenu.tsx
abadami: Ahhh, good call out. Will take care of it.

src/components/Filters/FilterDropdownMenu.tsx
abadami: Oh this is what I get for not looking into Truss too deeply yet. Yeah I think I'd prefer to do it that way as well. I highly doubt we'll get performance issues with it but we can always switch it if we somehow do.

src/components/Filters/FilterDropdownMenu.tsx
abadami: Yep, this was purely to get the mobile search bar in the proper position. Most importantly, it puts it below the two buttons. Otherwise if we had it setup the previous way, it would unfortunately push the filters button down below as well.

FWIW, I did try to figure out a way to keep them separate, but I couldn't find a solution quickly and I started justifying it as "it's also a filter" like you were in the other comments. As I'm typing this though, I'm feeling there may be a solution. I'll explore it, but I want to get confirmation from the design team first that this is the direction we're going

src/components/Filters/FilterDropdownMenu.tsx
abadami: That's a good point, both on the CSS and the active filter count. We don't really have the design for those states in place so worth asking the design team in general about it

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
abadami: Oh right, the mobile chromatic snapshot is a good point. Will for sure add.

src/components/Filters/FilterDropdownMenu.tsx
bdow: We can create a follow up ticket for this after checking with design on expectations for mobile behaviors

src/components/Filters/FilterDropdownMenu.tsx
bdow: If we see flickering or odd rerenders, then lets move towards a css solution

src/components/Filters/FilterDropdownMenu.tsx
bdow: Let's follow up to add IconButton outline variant once that PR lands. No need to block here.

bdow (approved):

PR #1266, opened 2026-06-10
feat: Adding an outline variant to the IconButton component

abadami:
## IconButton Outline Variant

Similar to circle. Will likely be used for buttons on mobile screens quite a bit.

### DIFF
--- a/src/components/IconButton.tsx
+++ b/src/components/IconButton.tsx
@@ -35,6 +35,8 @@ export type IconButtonProps = {
    * for screen readers without showing the tooltip. An explicit `tooltip` or disabled reason still shows.
    */
   preventTooltip?: boolean;
+  /** Whether to display the square variant */
+  outline?: boolean;
 } & BeamButtonProps &
   BeamFocusableProps;
 
@@ -58,6 +60,7 @@ export function IconButton(props: IconButtonProps) {
     forceFocusStyles = false,
     label,
     preventTooltip = false,
+    outline = false,
   } = props;
   const isDisabled = !!disabled;
   const ariaProps = { onPress, isDisabled, autoFocus, ...menuTriggerProps };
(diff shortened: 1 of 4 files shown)


src/components/IconButton.tsx
```diff
@@ -77,14 +80,14 @@ export function IconButton(props: IconButtonProps) {
   const styles = useMemo(
     () => ({
       ...iconButtonStylesReset,
-      ...(circle ? iconButtonCircle : compact ? iconButtonCompact : iconButtonNormal),
-      ...(isHovered && (circle ? iconButtonCircleStylesHover : iconButtonTokenHover)),
+      ...(circle ? iconButtonCircle : compact ? iconButtonCompact : outline ? iconButtonOutline : iconButtonNormal),
+      ...(isHovered && (circle || outline ? iconButtonCircleStylesHover : iconButtonTokenHover)),
       ...(isFocusVisible || forceFocusStyles ? (circle ? iconButtonCircleStylesFocus : iconButtonStylesFocus) : {}),
-      ...(active && (circle ? activeStylesCircle : iconButtonTokenHover)),
+      ...(active && (circle || outline ? activeStylesCircle : iconButtonTokenHover)),
```
abadami: As of right now outline mimics the circle styling quite a bit when it comes to these various states. I could see this changing if we want to divert from circle's setup at all.

src/components/IconButton.tsx
```diff
@@ -77,14 +80,14 @@ export function IconButton(props: IconButtonProps) {
...
       ...(isDisabled && iconButtonStylesDisabled),
       ...(bgColor && Css.bgColor(bgColor).$),
```
abadami, line 89: I wonder if it would be worth moving this closer to the top. As of right now if a `bgColor` then the hover and active state background color changes don't activate since this override is at the bottom. Granted, we might need an override on the `bgColors` for those states as well.

I personally think it might be worth incorporating the usual `xss` pattern to handle these situations instead since you can handle the hover and active states through it easily enough, but we can save that for a little later. If we were to set this up we could keep `bgColors` as a legacy option to not negatively affect the current usage of it.

I will note if we do want to just move `bgColors` up on the list here my only worry is it affecting buttons that use the `bgColors` that aren't currently expecting a hover or active state color change, especially one they can't control.

Curious what others think on this.

src/components/IconButton.tsx
```diff
@@ -129,6 +132,7 @@ const iconButtonStylesReset = Css.bcTransparent.bss.bgTransparent.cursorPointer.
 const iconButtonNormal = Css.hPx(28).wPx(28).br8.bw2.$;
 const iconButtonCompact = Css.hPx(18).wPx(18).br4.bw1.$;
 const iconButtonCircle = Css.br100.wPx(48).hPx(48).bcGray300.ba.bw1.df.jcc.aic.$;
+const iconButtonOutline = Css.br8.wPx(48).hPx(40).pxPx(12).py1.bcGray300.ba.bw1.df.jcc.aic.$;
```
bdow: **nit:** I don't think we need the padding declarations here if we're specifying height, width, and to center the content.

src/components/IconButton.tsx
```diff
@@ -58,6 +60,7 @@ export function IconButton(props: IconButtonProps) {
     forceFocusStyles = false,
     label,
     preventTooltip = false,
+    outline = false,
```
bdow: **suggestion:** I'd like to move toward an `IconButtonVariant` type that includes `Default`, `Circle`, and `Outline`. Right now we have a weird case where a consumer could specify `<IconButton circle={true} outline={true} />`, but I think the user should really just be choosing one, right?
I get this will be a bit of a breaking API change, but that's fine. Shoudl be easy to change `<IconButton circle />` to `<IconButton variant={IconButtonVariant.Circle} />`

bdow (changes requested):

src/components/IconButton.tsx
bdow: I would say leave as is. Existing issue that we should address separately.

bdow (approved):

PR #1268, opened 2026-06-11
feat: Add View Toggle support to GridTableLayout

abadami:
## Add View Toggle Support to GridTableLayout

This PR adds View Toggle Support to Grid Table Layout. Currently stylistically acts very similar to the `EditColumnsButton`, both in mobile view and desktop view. As of right now, `GridTableLayout` supports this through a `renderContent` prop that we intend to be temporary. Ideally, the layout would support the card tile view internally rather than needing an external definition. You're also able to change the default view via a prop as well.


src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -48,6 +49,9 @@ export type GridTableLayoutProps<
   tertiaryAction?: ActionButtonProps;
   hideEditColumns?: boolean;
   totalCount?: number;
+  /** Temporary prop for card views. When provided, shows a view toggle button. Rendered in place of the table when in tile mode. */
+  renderContent?: ReactNode;
```
bdow: **thought/suggestion:** Could this be renamed to be more specific to what we're doing, and potentially use for later - like `withCardView`.
For now `withCardView: ReactNode` and later `withCardView: boolean;` once we have the card view build in. Comments can help explain the future state we're moving towards.
And let's stick with one term - either "tile" or "card". I think "card" is what I've heard most people already say, so probably would prefer that.

src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -111,8 +117,9 @@ function GridTableLayoutComponent<
     () => (tableProps.api as GridTableApiImpl<R>) ?? new GridTableApiImpl(),
     [tableProps.api],
   );
+  const [view, setView] = useState<TableView>(defaultView);
   const clientSearch = layoutState?.search === "client" ? layoutState.searchString : undefined;
-  const showTableActions = layoutState?.filterDefs || layoutState?.search || hasHideableColumns;
+  const showTableActions = layoutState?.filterDefs || layoutState?.search || hasHideableColumns || !!renderContent;
```
bdow: **nit (non-blocking):** I wonder when we should change this to a `shouldShowTableActions(props)` helper. Not a big deal, though.

src/components/Table/components/ViewToggleButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+type ViewToggleButtonProps = {
+  view: TableView;
+  onChange: (view: TableView) => void;
+  defaultOpen?: boolean;
+};
+
+const menuItems: MenuItem[] = [
+  { label: "List", icon: "projects", onClick: () => {} },
+  { label: "Tile", icon: "tile", onClick: () => {} },
+];
```
bdow: **request:** Would prefer these to be inline set their onClick values to be explicit about what they are choosing. Right now, we assume the "key" for these menu items will match `TableView` and we cast it as such. But if we change a lable to "Card" vs "List", then we'll need to also update the TableView type. 
As a general rule, I would prefer changes to the presentation layer does not dictate or require model changes. Then we'd be able to change the label as needed, and it will not require updating the data models that support it.

src/components/Table/components/ViewToggleButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+      {...tid}
+    >
+      <Menu
+        ariaMenuProps={menuProps}
+        onClose={() => state.close()}
+        items={menuItems}
+        selectedItem={view === "list" ? "List" : "Tile"}
+        onChange={(key) => onChange(key.toLowerCase() as TableView)}
+      />
+    </OverlayTrigger>
```
bdow: **suggestion:** I think we'd want to use `<ButtonMenu />` which does all of this OverlayTrigger and Menu for us and the ref connection between them. If there is something missing from ButtonMenu to the desired outcome, let's add it (looking at maybe `hideEndAdornment`?). 
btw, the `onChange` as part of the Menu or ButtonMenu was created as a way to (optionally) denote which item is currently selected.

bdow (changes requested):

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -6,14 +6,14 @@ import { collapseColumn, column, numericColumn, selectColumn } from "src/compone
...
-import { withBeamDecorator, withRouter, zeroTo } from "src/utils/sb";
+import { viewportModes, withBeamDecorator, withRouter, zeroTo } from "src/utils/sb";
 import { TestProjectLayout } from "../Layout.stories";
 import { GridTableLayout as GridTableLayoutComponent, useGridTableLayoutState } from "./GridTableLayout";
 
 export default {
   component: GridTableLayoutComponent,
   decorators: [withBeamDecorator, withRouter()],
-  parameters: { layout: "fullscreen" },
+  parameters: { layout: "fullscreen", chromatic: { modes: viewportModes("desktop", "mobile1") } },
```
bdow: Let's not add these modes at this level. We'll be creating a lot of snapshots and not always will they look good in mobile, so maybe not worth adding them yet.

src/components/Table/components/ViewToggleButton.tsx
abadami: Yeah, it was the `hideEndAdornment`. In hindsight, yeah I should've went ahead and added it. Will do so as I agree that's the better move.

src/components/Table/components/ViewToggleButton.tsx
abadami: That's a good call out. I'll absolutely make sure to get this one switched over.

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
abadami: Yeah I added these now since I _think_ we talked about it in the last PR, but yeah it might be a better plan to wait until all the layouts are in place and good to go since it's still a WIP as a whole.

bdow: Yea - if we do it, let's do it per story, not per file of stories.

src/components/Layout/GridTableLayout/GridTableLayout.tsx
abadami: Ah yeah I forgot we went with that direction prop name. Easy enough change so no issue. Also, I'll comb through to make sure we're using consistent terminology with that.

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -323,4 +323,45 @@ describe("GridTableLayout", () => {
...
+          rows: [simpleHeader, ...rows],
+        }}
+        renderContent={<Content />}
+      />,
+      withRouter(),
+    );
+
+    // Then ViewToggleButton is not rendered
+    expect(r.viewToggleButton).toBeInTheDocument();
+  });
```
bdow, line 372: can we update this test to click the toggle and assert the `Content` is displayed?

src/components/Layout/GridTableLayout/GridTableLayout.tsx
abadami: Yeah it might be over the line a touch. I'll say let's hold off on it purely since `hasHideableColumns` is memoized at the moment.

bdow (approved):

PR #1269, opened 2026-06-11
feat: Teach GridTableLayout and TableActions to be layout aware

bdow:
GridTableLayout and TableActions can now smartly be sticky when in NavbarLayout and within PageHeaderLayout. Required also teaching GridTable to adjust some things when in a layout with document level scrolling. Also needed to adjust some ways we calculate columns to avoid unnecessary scroll bars

### DIFF
--- a/src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
+++ b/src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
@@ -1,39 +1,16 @@
 import { checkboxFilter } from "src/components/Filters";
 import { actionColumn, column, numericColumn } from "src/components/Table/utils/columns";
 import { simpleHeader } from "src/components/Table/utils/simpleHelpers";
+import { DocumentScrollLayoutProvider } from "src/layouts/DocumentScrollLayoutContext";
+import { beamTableActionsHeightVar } from "src/layouts/layoutVars";
 import { noop } from "src/utils";
 import { click, render, tableSnapshot, withRouter } from "src/utils/rtl";
+import { vi } from "vitest";
 import {
   GridTableLayout as GridTableLayoutComponent,
   GridTableLayoutProps,
   useGridTableLayoutState,
 } from "./GridTableLayout";
-
-type Data = { name: string | undefined; value: number | undefined };
-type HeaderRow = { kind: "header"; id: string; data: undefined };
-type DataRow = { kind: "data"; id: string; data: Data };
-type Row = HeaderRow | DataRow;
-
-// Because we also want to test the use of the useGridTableLayoutState hook, we need create a wrapper component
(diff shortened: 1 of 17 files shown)


src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -342,32 +319,133 @@ describe("GridTableLayout", () => {
...
+// Because we also want to test the use of the useGridTableLayoutState hook, we need create a wrapper component
+type TestWrapperProps = Omit<GridTableLayoutProps<any, Row, any, any>, "layoutState"> & {
+  layoutStateProps: Parameters<typeof useGridTableLayoutState>[0];
+};
+function TestWrapper(props: TestWrapperProps) {
+  const layoutState = useGridTableLayoutState(props.layoutStateProps);
+  return <GridTableLayoutComponent {...props} layoutState={layoutState} />;
+}
+
+function getColumns() {
```
abadami, line 437: **question**: Two questions, wanted to ask for confirmation that we move helper variables and functions to the bottom stylistically for all Homebound files?

Lastly wanted to ask as to what was the reasoning for converting to `getColumns` and `getRows`?

abadami: Answered in pair review: Yes to the first one, and they were converted for handling a quirk with the vitest runner where if they are just variables they are evaluated immediately (aka, the object or array is eager created), but as a function they are not until they are called. Just helps with performance.

src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -307,13 +357,47 @@ type HeaderProps = {
...
+      tableWrapper.style.removeProperty(beamTableActionsHeightVar);
+    }
+  }, [enabled, tableActionsRef, tableWrapperRef]);
+
+  useResizeObserver({ ref: tableActionsRef, onResize: enabled ? syncHeightVar : noop });
+  useLayoutEffect(() => {
+    syncHeightVar();
+    const tableWrapper = tableWrapperRef.current;
+    return () => {
+      tableWrapper?.style.removeProperty(beamTableActionsHeightVar);
```
abadami, line 389: **question**: Any particular reason we're opting to remove the css property instead of setting it to 0?

abadami: Pair review answer: Basically there wasn't too much difference, just a choice.

In hindsight afterwards, I think it's better to remove the property to allow any consumer to have its own fallback value. That being said, if something doesn't have a fallback value it'll be an issue, but due to how `layoutVars` works it should be fine.

abadami (approved):

PR #1270, opened 2026-06-12
fix: editColumnsButton scroll and minimal style update;

apattersonATX-HB:
- New figma designs on our GridTable use this style for our `EditColumnsButton` so I made it the default. It doesn't match perfectly, still too much padding, but I believe that's a broader design system fix that needs to happen.
- Fixes popover growing too long. It triggered a page overflow and made edit columns unusable. I noticed this when implementing the giant personalization option group table [here](https://www.figma.com/design/z8KoZW6FQoSUkE6arpgMcQ/Personalization-Group-Management-Improvements--BP-?node-id=1-14&p=f&m=dev)

Screenshot 2026-06-11 at 7 22 20 PM


src/components/Table/components/EditColumnsButton.tsx
```diff
@@ -79,31 +87,36 @@ export function EditColumnsButton<R extends Kinded>(props: EditColumnsButtonProp
...
-                )
-              }
-              labelStyle="hidden"
-              label={option.label}
-              {...tid[`option${option.value}`]}
-            />
-          </Fragment>
-        ))}
-        <div css={Css.gc("1 / -1").df.jcc.$}>
+      <div css={Css.df.fdc.bgWhite.maxwPx(326).maxhPx(512).onHover.bshHover.$}>
```
bdow: **nit/suggestion:** Would love to start moving towards using tokens when possible. In her we could do:
```
bgColor(Tokens.Surface)
```

src/components/Table/components/EditColumnsButton.tsx
```diff
@@ -79,31 +87,36 @@ export function EditColumnsButton<R extends Kinded>(props: EditColumnsButtonProp
...
+                }
+                labelStyle="hidden"
+                label={option.label}
+                {...tid[`option${option.value}`]}
+              />
+            </Fragment>
+          ))}
+        </div>
+        {/* Pinned footer */}
+        <div css={Css.df.jcc.p2.bt.bcGray200.$}>
```
bdow: Same regarding tokens, this would be `bc(Tokens.onSurfaceMuted)`

src/components/Table/components/EditColumnsButton.tsx
```diff
@@ -23,10 +23,18 @@ type EditColumnsButtonProps<R extends Kinded> = {
...
 export function EditColumnsButton<R extends Kinded>(props: EditColumnsButtonProps<R>) {
-  const { defaultOpen, disabled, columns, trigger, api } = props;
+  const { defaultOpen, disabled, columns, api } = props;
+  // Defaults to a compact icon-only trigger; consumers can override it by passing `trigger`.
+  const trigger: OverlayTriggerProps["trigger"] = props.trigger ?? {
+    icon: "kanban",
+    size: "md",
+    label: "",
+    variant: "secondaryBlack",
+  };
```
bdow: **thought:** Any reason to allow changing these defaults? I think I'd push that its always the same as this is a very bespoke component for one use case.

src/components/Table/components/EditColumnsButton.tsx
apattersonATX-HB: Thanks! I forgot all about Tokens 😅

src/components/Table/components/EditColumnsButton.tsx
apattersonATX-HB: 🤷🏽‍♂️ None other than it supported `trigger` first. I'll update

bdow: Fair - I saw the only consumer of it in Beam is GridTableLayout, which hard codes the trigger. 

Though, just did another search and the Schedules page in `internal-frontend` also uses this, and just sets a `{ label: "Columns" }`. So that's my bad. - maybe keep for now?

apattersonATX-HB: Naw I think its fine. It's in the same location and limited to one page. Users will pick up the change quick.

All the latest designs i've seen involving gridTable have this icon so i'll call schedules an early adopter 😄

bdow (approved):

PR #1271, opened 2026-06-12
feat: Persist selected TableView in GridTableLayout to localstorage

bdow:


src/components/Layout/GridTableLayout/usePersistedTableView.ts
```diff
@@ -0,0 +1,43 @@
...
+  return [view, setViewAndPersist];
+}
+
+/** localStorage key for list/card view, scoped to pathname. */
+export function getGridTableViewStorageKey(pathname: string): string {
+  return `beam.gridTableLayout.view.${pathname}`;
+}
+
+function parseStoredTableView(raw: string | null): TableView | undefined {
+  return raw === "list" || raw === "card" ? raw : undefined;
```
abadami, line 42: **suggestion (non-blocking)**: I wonder if there to change this in a way to check if it satisfies the `TableView` type more dynamically. Otherwise if we ever introduce a new view type into the `TableView` a location in the code that'd we have to update manually that we might miss.

Granted I vaguely remember wanting to do this in Typescript in the past and it ending up feeling janky. What do you think though?

abadami (approved):

src/components/Layout/GridTableLayout/usePersistedTableView.ts
bdow: Yea, definitely. I haven't used `satisfies` much yet, so happy to explore it.

PR #1272, opened 2026-06-12
feat: Using Outline variant in FilterDropdown Menu & Adding endAdornment to IconButton

abadami:
## `IconButton` outline variant in `FilterDropdownMenu` and adding `endAdornment` to `IconButton`

This PR has the `GridTableLayout` actually use the outline variant `IconButton` by updating `FilterDropdownMenu` to use them. This does also update the `IconButton` by adding `endAdornment` to `IconButton` for outline and circle variants only, as the filter button really felt like they needed them for showing currently active filters for consistency with the normal button

Note: This did not add the end adornment to the IconButton. Will incorporate that either through [this pr](https://github.com/homebound-team/beam/pull/1273) or a future one

### DIFF
--- a/src/components/Filters/FilterDropdownMenu.tsx
+++ b/src/components/Filters/FilterDropdownMenu.tsx
@@ -5,7 +5,7 @@ import { Filter, FilterDefs, FilterImpls, filterTestIdPrefix, updateFilter } fro
 import { Icon } from "src/components/Icon";
 import { IconButton } from "src/components/IconButton";
 import { ToggleChip } from "src/components/ToggleChip";
-import { Css, Palette, Tokens } from "src/Css";
+import { Css, Palette } from "src/Css";
 import { useBreakpoint } from "src/hooks";
 import { SelectField } from "src/inputs/SelectField";
 import { TextField } from "src/inputs/TextField";
@@ -120,7 +120,6 @@ function FilterDropdownMenu<F extends Record<string, unknown>, G extends Value =
           label="Search"
           onClick={() => setSearchIsOpen(!searchIsOpen)}
           active={searchIsOpen}
-          bgColor={Tokens.Surface}
           {...testId.searchButton}
         />
       )}
@@ -133,7 +132,6 @@ function FilterDropdownMenu<F extends Record<string, unknown>, G extends Value =
           label="Filter"
           active={isOpen}
           onClick={() => setIsOpen(!isOpen)}
-          bgColor={Tokens.Surface}
           {...testId.button}
         />
       )}
--- a/src/components/Filters/FilterDropdownMenu.tsx
+++ b/src/components/Filters/FilterDropdownMenu.tsx
@@ -133,7 +133,6 @@ function FilterDropdownMenu<F extends Record<string, unknown>, G extends Value =
           label="Filter"
           active={isOpen}
           onClick={() => setIsOpen(!isOpen)}
-          endAdornment={activeFilterCount > 0 ? <CountBadge count={activeFilterCount} /> : undefined}
           bgColor={Tokens.Surface}
           {...testId.button}
         />
--- a/src/components/Filters/FilterDropdownMenu.tsx
+++ b/src/components/Filters/FilterDropdownMenu.tsx
@@ -3,8 +3,9 @@ import { Button } from "src/components/Button";
 import { CountBadge } from "src/components/CountBadge";
 import { Filter, FilterDefs, FilterImpls, filterTestIdPrefix, updateFilter } from "src/components/Filters";
 import { Icon } from "src/components/Icon";
+import { IconButton } from "src/components/IconButton";
 import { ToggleChip } from "src/components/ToggleChip";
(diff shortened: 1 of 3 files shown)


src/components/IconButton.tsx
```diff
@@ -10,7 +10,9 @@ import { noop } from "src/utils";
 import { getButtonOrLink } from "src/utils/getInteractiveElement";
 import { useTestIds } from "src/utils/useTestIds";
 
-export type IconButtonVariant = "default" | "circle" | "outline";
+type IconButtonVariantProps =
+  | { variant?: "default"; endAdornment?: never }
+  | { variant: "circle" | "outline"; endAdornment?: ReactNode };
```
abadami: I think this may be an interesting method of defining the specific variant props, as to me we shouldn't be using the `endAdornment` prop ever with the default variant, and this felt like a cleaner way of ensuring that doesn't happen via Typescript rather than using React conditional rendering, at the cost of readability for the variant type specifically.

Open to adjusting how this works and using the more React way of enforcing that rule with `endAdornment`

src/components/IconButton.tsx
```diff
@@ -103,19 +124,22 @@ export function IconButton(props: IconButtonProps) {
...
+        icon={icon}
+        color={
+          color ||
+          (isDisabled
+            ? Tokens.TextDisabled
+            : isCircle && (isHovered || active || isFocusVisible)
+              ? defaultIconColor
+              : iconColor)
+        }
+        bgColor={bgColor}
```
bdow: **suggestion:** I would prefer the bgColor to be driven by the variant. "Outline". We don't want people just adding whatever bgcolor they want to these. Our components very opiniated.

src/components/IconButton.tsx
```diff
@@ -81,14 +83,33 @@ export function IconButton(props: IconButtonProps) {
   const styles = useMemo(
     () => ({
       ...iconButtonStylesReset,
-      ...(isCircle ? iconButtonCircle : isOutline ? iconButtonOutline : compact ? iconButtonCompact : iconButtonNormal),
+      ...(isCircle
+        ? iconButtonCircle
+        : isOutline
+          ? endAdornment
+            ? iconButtonOutlineWithAdornment
```
abadami: Having a specific style with Outline makes this ternary even more messy, which might be worth cleaning up a bit for readability, but I do think the extra gap styling makes it nicer to read.

src/components/IconButton.stories.tsx
```diff
@@ -152,7 +160,8 @@ function HoveredIconButton(args: IconButtonStoryArgs) {
   const hoverBlock = (
     <div className="hovered-icon-button">
       <style>{`.hovered-icon-button button { background-color: ${bg};${borderColor ? ` border-color: ${borderColor};` : ""} }`}</style>
-      <IconButton {...iconArgs} variant={variant} active={isCircle} />
+      {/* eslint-disable-next-line @typescript-eslint/no-explicit-any */}
+      <IconButton {...(iconArgs as any)} variant={variant} active={isCircle} />
```
abadami: This was added due to Typescript being angry about this due to the variant prop definitions with `endAdornment`. Should be possible to cleanup though, missed this in my initial self review.

src/components/Filters/FilterDropdownMenu.tsx
```diff
@@ -113,28 +114,41 @@ function FilterDropdownMenu<F extends Record<string, unknown>, G extends Value =
...
-      {hasFilters && (
+      {/* Small screen: filter icon button with optional active count badge */}
+      {sm && hasFilters && (
+        <IconButton
+          variant="outline"
+          icon="filter"
+          label="Filter"
+          active={isOpen}
+          onClick={() => setIsOpen(!isOpen)}
+          endAdornment={activeFilterCount > 0 ? <CountBadge count={activeFilterCount} /> : undefined}
```
bdow: @apattersonATX-HB is working on a similar concept to build a badge into IconButton. Let's have you two connect here.

abadami: I think with Arthur's PR and design taking another look at it I'll go ahead and revert the `endAdornment` changes for the purposes of this PR and keep this under the scope of just using outline for the `IconButtons` here.

src/components/IconButton.tsx
abadami: I do agree, especially if we want to keep the opinionated design principle. My only concern with that at the moment is that I believe the `bgColor` prop here is being actively used elsewhere in Blueprint. I'd say we can still have outline use the surface color by default so that overriding it isn't necessary.

I think that's something to work towards, which I'll do with changing the outline styling to reflect it, but not something I think we can get rid of immediately.

Although I did check it again and we're only using it in ~3 places, but I do think we'll want to be intentional about it still.

bdow: Gotcha - I didn't realize it was an existing prop that's being used. Agreed to move forward by not using it ourselves and having the `variant` drive the look & feel of the component.

bdow (approved):

PR #1273, opened 2026-06-12
feat: icon badge;

apattersonATX-HB:
**Badge support on `Icon`, surfaced on `EditColumnsButton`**

- `Icon` gains an optional `badge` (`{ color; count? }`) overlaying in top-right corner.
- `Button` and `IconButton` forward `badge` to their `Icon`, and `OverlayTrigger` trigger configs accept it — needed because `EditColumnsButton`'s trigger is a `Button` (icon + chevron), not an actual `IconButton`.
- `EditColumnsButton` shows a built-in badge of the **visible** hideable-column count, only while some are hidden — absent when all are shown, `0` when all are hidden.

Screenshot 2026-06-12 at 3 32 11 PM
Screenshot 2026-06-12 at 3 32 00 PM

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -15,6 +15,8 @@ export type ButtonProps = {
   variant?: ButtonVariant;
   size?: ButtonSize;
   icon?: IconProps["icon"] | null;
+  /** Shows a small badge overlapping the button's icon (top-right). Omit `count` for a plain dot. */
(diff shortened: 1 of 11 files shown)


bdow:
FYI, this might get placed on hold. I spoke with the rest of the design team regarding this as we want to do something similar for filters and they want to chat through it as a team before we implement anything.

PR #1274, opened 2026-06-15
fix: Filter chips to display label not value

bdow:


src/components/Filters/FilterDropdownMenu.stories.tsx
```diff
@@ -0,0 +1,66 @@
+import { Meta } from "@storybook/react-vite";
```
abadami, line 1: **suggestion (non-blocking)**: It might be worth going ahead and adding stories for the filters that don't have stories here, namely BooleanFilter, CheckboxFilter, and NumberRangeFilter. I think if our eventual goal is to have the storybook be our documentation resource, then it'd likely benefit adding them in.

Non-blocking because it might be best to look at all that holistically when it comes to what our storybook is going to look like in the future.

src/components/Filters/selectedFilterLabelUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Value } from "src/inputs/Value";
+
+/** Resolves a display label for a selected option value (static or lazy-loaded options). */
+export function resolveOptionSelectedFilterLabel<O, V extends Value>(
+  optionsOrLoad: OptionsOrLoad<O>,
+  getOptionValue: (o: O) => V,
+  getOptionLabel: (o: O) => string,
+  value: V,
+): string {
+  const options = initializeOptions(optionsOrLoad, getOptionValue, getOptionLabel, undefined, false, false);
```
abadami, line 12: **question**: Just a clarity question to make sure I understand this correct.

"OptionsOrLoad<O>" as a type represents a list of options that need to be loaded from a separate source (presumably the server). The type may have some options pre-loaded, but it won't load them in until interacted with, which I guess actually as the comment suggests makes this more for lazy-loading.

I believe the initializeOptions function here invokes that loading process so we can fetch the option for the chip we need?

abadami (approved):

src/components/Filters/selectedFilterLabelUtils.ts
bdow: OptionsOrLoad represents a set of Options that are already there and can be used without lazy loading, or has a lazy loaded feature where some options are already available via `current` and then a method to call for loading.
`initializeOptions` does not trigger the loading. Instead it translates the `OptionsOrLoad` to `Options[]` so we can stand up the component properly.

bdow: The loading only happens when the user expands the select field's list.

src/components/Filters/FilterDropdownMenu.stories.tsx
bdow: Agreed. Could be a nice add to put those in as well. Agreed to do this in the future when doing a new pass out our storybook set up.

PR #1275, opened 2026-06-15
feat: add onEnter callback param for SelectField

Jacob-Diebold:
Added an optional onEnter callback to SelectField (via ComboBoxBase / ComboBoxInput) so consumers can handle Enter in the filter input when no dropdown row is keyboard-focused.

React Aria’s combobox reverts typed text on Enter when the menu is open but nothing in the list has focus (aria-activedescendant is unset). That breaks flows like type-then-Enter to create or commit a custom value. This wires Enter through Beam instead of requiring app-level onKeyDownCapture workarounds.

When **onEnter** is set:

- Enter with non-empty trimmed input and no keyboard-focused list item → calls onEnter(inputValue) and blurs the field
- Enter when the user has arrowed to a list row (including “Add New”) → normal selection; onEnter is not called
- Disabled/read-only fields are ignored
- Single-select only (MultiSelectField is unchanged)

### DIFF
--- a/src/inputs/internal/ComboBoxBase.tsx
+++ b/src/inputs/internal/ComboBoxBase.tsx
@@ -66,6 +66,11 @@ export type ComboBoxBaseProps<O, V extends Value> = {
   /* Only supported on single Select fields */
   onAddNew?: (v: string) => void;
   /**
+   * Called when Enter is pressed in the filter input with typed text and no keyboard-focused list item.
+   * Only supported on single Select fields.
+   */
+  onEnter?: (inputValue: string) => void;
+  /**
    * When true (default), options are sorted alphabetically by their label.
    * Set to false to maintain the original order of options.
    */
@@ -102,6 +107,7 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
     fullWidth = fieldProps?.fullWidth ?? false,
     onSearch,
     onAddNew,
+    onEnter,
     autoSort = true,
     ...otherProps
   } = props;
@@ -387,6 +393,7 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
         borderless={borderless}
         tooltip={resolveTooltip(disabled, undefined, readOnly)}
         resetField={resetField}
+        onEnter={multiselect ? undefined : onEnter}
       />
       {state.isOpen && (
         <Popover
--- a/src/inputs/internal/ComboBoxInput.tsx
+++ b/src/inputs/internal/ComboBoxInput.tsx
@@ -47,6 +47,8 @@ type ComboBoxInputProps<O, V extends Value> = {
   isTree?: boolean;
   /* Allows input to wrap to multiple lines */
   multiline?: boolean;
+  /** Only supported on single Select fields. See ComboBoxBaseProps.onEnter. */
+  onEnter?: (inputValue: string) => void;
 } & PresentationFieldProps;
 
 export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V>) {
@@ -71,6 +73,7 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
     inputRef,
     inputWrapRef,
     multiline = false,
+    onEnter,
     ...otherProps
   } = props;
 
@@ -103,7 +106,27 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
     value: inputProps.value,
   });
 
-  return (
+  function onEnterKeyDownCapture(e: React.KeyboardEvent) {
+    if (!onEnter) return;
+    if (e.key !== "Enter") return;
+    if (inputProps.readOnly || inputProps.disabled) return;
+    if (!(e.target instanceof HTMLInputElement)) return;
+
+    const input = e.target;
+
+    // User arrowed to a dropdown row — let react-aria select that option (including "Add New").
+    if (input.getAttribute("aria-activedescendant")) return;
+
+    const typedValue = input.value.trim();
+    if (!typedValue) return;
+
+    e.preventDefault();
+    e.stopPropagation();
+    onEnter(typedValue);
+    input.blur();
+  }
+
+  const field = (
     <TextFieldBase
       {...otherProps}
       {...multilineProps}
@@ -246,6 +269,8 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
       }}
     />
   );
+
+  return onEnter ? <div onKeyDownCapture={onEnterKeyDownCapture}>{field}</div> : field;
 }
 
 function SelectedOptionBullets({ labels = [] }: { labels: string[] | undefined }) {
--- a/src/inputs/SelectField.test.tsx
+++ b/src/inputs/SelectField.test.tsx
@@ -493,6 +493,51 @@ describe("SelectFieldTest", () => {
     expect(onSelect.mock.calls[2][0]).toBe(undefined);
   });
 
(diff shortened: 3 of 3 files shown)


Jacob-Diebold:
First beam PR, let me know what you guys think!

src/inputs/internal/ComboBoxInput.tsx
```diff
@@ -246,6 +269,8 @@ export function ComboBoxInput<O, V extends Value>(props: ComboBoxInputProps<O, V
       }}
     />
   );
+
+  return onEnter ? <div onKeyDownCapture={onEnterKeyDownCapture}>{field}</div> : field;
```
bdow, line 273: **nit:** This extra wrapping element may cause unintended style issues. For example, if the `field` was expected to be the immediate descendent of a flex wrapper. Could we have wired this handler into an existing element? If not, then I would suggest setting `display: contents` on this element so that it's a behavior only wrapper, and styles pass right through it.

bdow (commented):
Curious on @stephenh 's thoughts too. 
I'm hesitant to provide a "choose your own adventure" sort of escape hatch for a very specific "Add new option" workflow. If this is blocking an urgent user issue, then I'd rather provide quick fix locally in `internal-frontend`, then add a ticket/issue to address the tech debt and build out the functionality we need in a more structured and reusable way.

Jacob-Diebold:
@bdow Yeah, that makes sense. I don't think its super urgent anymore, I was able to give them a workaround to unblock them.

I just had a different thought after coming back to this with fresh eyes this morning. 

What would you think if this component treated the 'Enter' key as 1. selecting an option if the currently typed option existed and 2. if the user typed an option that doesn't exist AND there is an 'onAddNew' callback defined then it would call the onAddNew callback on the current value.

The UX would be pretty much the same as it currently is, right now users just have to type and click an option; this would just give users the option of pressing the 'Enter' key as well. And its similar to my approach without having to add on a new callback param.

I think that would remove most of the code in the <SelectMaybeNewField> component in internal-frontend. It wouldn't need to swap between the SelectField and the TextField anymore which is where I think users are facing the issue.

Thanks for the feedback, still trying to figure out whats appropriate to do where within the stack and that all makes sense!

Jacob-Diebold:
Or if we don't like the 'Enter' key press being a first class citizen in this case. Maybe we just remove the whole SelectField to TextField swap in SelectMaybeNewField to make that component more beam native. Users could still add new fields via the 'onAddNew' functionality, they just won't be able to press enter to submit it; but the component wouldn't be swapping on them midway through so the experience would be more uniform.

bdow:
@Jacob-Diebold - That's be a pretty seamless UX, for sure. 
A quick **opinion** piece 😄:
My only push back is that in some of these instances, we may want more friction. I often think about the single responsibility principle in these flows. My personal preference is if we're adding something new, that there is a different mechanism for that. It doesn't have be a lot of friction, but I'd prefer it be a different component - like a regular text input that is meant to hold a freeform value. 
As fun for us as it can be to make components do many things, in my experience it makes components more brittle and harder to maintain.

Where we found a good middle ground, imo, was the [ChipSelectField](https://main--60466f7d43c37c0021788867.chromatic.com/?path=/story/inputs-chipselectfield--example)'s "Create New" flow. from the dropdown, you select "Create New". That changes the field from a select to an text input. Then there is an 'onEnter' from there which triggers a callback and appends the new option to the select field. I'm not saying that is perfectly correct either, as there is not feedback for the user that the only way to submit your new option is by hitting the Enter key. 
I'd personally rather go back to the drawing board with the design team and figure out the best way to handle the flows where a value doesn't exist as an option, and how we can improve that.

Jacob-Diebold:
Yeah, that makes sense. Maybe the component is already in a good middle ground between usability and friction. Users can still type a new option, and if onAddNew is provided, an “Add New” option appears in the dropdown. From there, they can press Arrow Down to focus the dropdown and Enter to trigger the onAddNew callback when “Add New” is selected.

Maybe the answer is simply to remove the SelectField → TextField swap in internal-frontend's SelectMaybeNewField. That’s the part that’s currently behaving inconsistently, and removing it would make the implementation more uniform with the rest of the app.

@stephenh thoughts?

src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -65,6 +65,11 @@ export type ComboBoxBaseProps<O, V extends Value> = {
   onSearch?: (search: string) => void;
   /* Only supported on single Select fields */
   onAddNew?: (v: string) => void;
+  /**
+   * Called when Enter is pressed in the filter input with typed text and no keyboard-focused list item.
+   * Only supported on single Select fields.
+   */
+  onEnter?: (inputValue: string) => void;
```
stephenh, line 72: @Jacob-Diebold curious if we need the new `onEnter` prop? Like afaiu the use case in Blueprint is already passing `onAddNew` which expects to be told "here is the new value", and we're somehow just missing/dropped the new value?

Jacob-Diebold: @stephenh Yeah I don't think we do need it, referring to this previous message:

> I just had a different thought after coming back to this with fresh eyes this morning.
> 
> What would you think if this component treated the 'Enter' key as 1. selecting an option if the currently typed option existed and 2. if the user typed an option that doesn't exist AND there is an 'onAddNew' callback 
> defined then it would call the onAddNew callback on the current value.
> 
> The UX would be pretty much the same as it currently is, right now users just have to type and click an 
> option; this would just give users the option of pressing the 'Enter' key as well. And its similar to my approach without having to add on a new callback param.
> 
> I think that would remove most of the code in the component in internal-frontend. It wouldn't need to swap 
> between the SelectField and the TextField anymore which is where I think users are facing the issue.

The current issue in that internal frontend component is that a user can hit Enter to create the new option after the component has swapped over to the text input component (debounced with a 300ms delay); but if it hasn't swapped over yet, Enter will just revert the input to the previously selected input. So if a user types something and then hits Enter right away it just deletes whatever they typed. I think we just need it to canonically do the same thing i.e. Enter works for both components, or it doesn't work for either. And really if we don't want to support 'Enter' here to put up a small barrier to entry of creating new items; then the front end component doesn't need to swap over to a text input field.

Jacob-Diebold: Or I guess a third option, which I think would be the most similar to "[ChipSelectField](https://main--60466f7d43c37c0021788867.chromatic.com/?path=/story/inputs-chipselectfield--example)'s "Create New" flow" that @bdow mentioned. We remove the auto-switch functionality from the SelectMaybeNewField component so it doesn't auto-switch from SelectField -> TextInput component. It only switches over if the user clicks the 'Add New' dropdown option, then they could type the new value in & hit 'Enter' to save it. It would probably make sense to bake this into Beam since it would be inline with ChipSelectField & then we can still remove the SelectMaybeNewField.

stephenh: @Jacob-Diebold lazy ask, is this "swap over to the text component" something that is happening one-off in internal-frontend?

stephenh: I'm looking at `SelectMaybeNewField` -- I'm forgetting why it works this way -- ngl I had thought `onAddNew` made `SelectField` provide "everything the end page needs".

I guess imo I'd love to see internal-frontends `SelectMaybeNewField` go away and/or get folded into Beam proper so that it can "just work", i.e. maybe we just remove this "300ms delay between switching over between fields".

Jacob-Diebold: Yeah, I agree I don't think we need SelectMaybeNewField after looking at all of this; SelectField should have the ability to do everything we need it to. Unless I'm missing something that I'll run into while removing it. 

I'm going to consolidate it on the frontend to just use SelectField where ever its in use (only 5 places). And then if we ever want to add handling for pressing 'Enter' we just need to natively add it into beam.

PR #1276, opened 2026-06-15
feat: Query Table Infinite Scroll Support

abadami:
## Query Table Infinite Scroll Support

For GridTableLayout, Infinite Scrolling was already supported via the `infiniteScrolling` prop for the normal grid table usage, but not for the QueryTable usage. This adds the ability to use infinite scrolling with the query table. This also adds stories for both GridTable and QueryTable infinite scrolling.

Lastly, this removes "normal" pagination from GridTableLayout as we want to enforce infinite scrolling in our table layouts.

### DIFF
--- a/src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
+++ b/src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
@@ -173,43 +173,6 @@ export function QueryTableLayout() {
   );
 }
 
-export function WithPagination() {
-  const filterDefs = useMemo(() => getFilterDefs(), []);
-  const columns = useMemo(() => getColumns(), []);
-
-  const layoutState = useGridTableLayoutState({
-    persistedFilter: {
-      filterDefs,
-      storageKey: "grid-table-layout",
-    },
-    search: "client",
-    pagination: {
-      pageSizes: [25, 50, 100, 500],
-      storageKey: "grid-table-pagination",
-    },
-  });
-
-  return (
-    <TestProjectLayout>
-      <GridTableLayoutComponent
-        pageTitle="Grid Table With Pagination"
-        breadCrumb={[
-          { href: "/", label: "Home" },
-          { href: "/", label: "Product Offerings" },
-        ]}
-        layoutState={layoutState}
-        totalCount={543}
-        tableProps={{
-          columns,
-          rows: [simpleHeader, ...makeNestedRows(10)],
-          sorting: { on: "client", initial: [columns[0].id!, "ASC"] },
-        }}
-        primaryAction={{ label: "Add Product", onClick: noop }}
-      />
-    </TestProjectLayout>
-  );
-}
-
 export function WithoutHeader() {
   const columns = useMemo(() => getColumns(false), []);
 
@@ -349,8 +312,18 @@ export function WithInfiniteScroll() {
 }
 
 export function WithQueryTableInfiniteScroll() {
-  const query = useExamplePaginatedQuery({ filter: {} });
   const columns = useMemo(() => getColumns(false), []);
+  const filterDefs = useMemo(() => getFilterDefs(), []);
+
+  const layoutState = useGridTableLayoutState({
+    persistedFilter: {
+      filterDefs,
+      storageKey: "grid-table-layout",
+    },
+    search: "client",
+  });
+
+  const query = useExamplePaginatedQuery({ filter: layoutState.filter });
 
   return (
     <TestProjectLayout>
@@ -366,6 +339,7 @@ export function WithQueryTableInfiniteScroll() {
           ],
           infiniteScroll: { pageSize: 50 },
         }}
+        layoutState={layoutState}
       />
     </TestProjectLayout>
   );
@@ -398,11 +372,11 @@ function useExampleQuery({ filter }: { filter: Record<string, unknown> }) {
   };
 }
 
-function useExamplePaginatedQuery({ filter }: { filter: Record<string, unknown> }) {
+function useExamplePaginatedQuery({ filter }: { filter: { status: string[] } }) {
   const filterString = JSON.stringify(filter);
   const pageSize = 50;
 
-  const fakedb = useMemo(
+  const fakeDatabase = useMemo(
     () =>
       zeroTo(200).map((index) => ({
         id: String(index),
@@ -414,6 +388,11 @@ function useExamplePaginatedQuery({ filter }: { filter: Record<string, unknown> 
     [],
   );
 
+  const filteredDatabase = useMemo(
+    () => (filter.status ? fakeDatabase.filter((row) => filter.status.includes(row.status)) : fakeDatabase),
+    [fakeDatabase, filter],
+  );
+
   const [loading, setLoading] = useState(true);
   const [data, setData] =
     useState<Array<{ id: string; name: string; value: number; status: string; priority: number }>>();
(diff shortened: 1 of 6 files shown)


src/components/Layout/GridTableLayout/QueryTable.tsx
```diff
@@ -9,17 +9,35 @@ export type QueryResult<QData> = {
   loading: boolean;
   error?: { message: string };
   data?: QData;
+  /**
+   * Optional. When provided alongside `infiniteScroll`, QueryTable will call this as the user
+   * scrolls to the bottom to load the next page. Requires `as="virtual"` on the table.
+   */
+  fetchMore?: (variables: { offset: number; limit: number }) => Promise<{ data?: QData }>;
```
abadami: I need to revisit how this particular piece works as it is not playing nice with Apollo in Internal Frontend at the moment.

bdow: I would try to remove this and not have QueryTable be aware of "fetchMore". My thought, and maybe wrong, would be that we're just passing the `infiniteScroll` property through, and that'd be enough. From the consumer perspective, they'd do something like:
```
// internal-frontend
function ProjectsPage() {
  // Not: this is an existing callback from `ProjectsPage.tsx in internal-frontend
  const maybeFetchNextPage = useCallback(async () => {
    if (query.data?.projectsPage.pageInfo.hasNextPage) {
      await query.fetchMore({ variables: { page: { offset: query.data?.projectsPage.entities?.length, limit: 100 } } });
    }
  }, [query]);

  const tableProps = useMemo(() => {
    return { 
      query: query,
      infiniteScroll: { onEndReached: maybeFetchNextPage },
    };
  }, [maybeFetchNextPage, query]);

  return (
    <GridTableLayout tableProps={tableProps} />
} 
```
I haven't completely thought through this, but my initial assumption is that this is just a pass-through sort of behavior. Hopefully we don't need to teach QueryTable how/when to fetch data.
Take a look at `ProjectsPage.tsx` and how that does the infinite scrolling with the local `QueryTable.tsx` in `internal-frontend`. I'd suggest modeling that pattern

abadami: This has taught me the very important lesson of checking to make sure `internal-frontend` didn't already have it's own implementation up and running 😆 

I think with this in mind, we really don't have too many actual changes needed to the `QueryTable` and layout itself. All we're doing in removing support for the old pagination and adding stories for infinite scrolling

abadami: My only thought that I'll add really quick is that it seems like the purpose of QueryTable was to handle the the Apollo Query under the hood, and I would argue that that would include handling "fetchMore". 

That all being said, with the ProjectsPage checking if there is a next page there is already precedent that passing in our own version could be helpful and likely common. I'm down with supporting it like how internal-frontend is at the moment to ease transition and we can perhaps re-evaluate later.

bdow: Fair. Though, it doesn't trigger the queries at the moment. It just accepts an Apollo query, displays the proper state (loading, error, or data result), and calls the `createRows` method internally. I agree, we could teach QueryTable to do more, but I don't think its necessary to do right now. If we see a lot of boiler plate happening on the consumer side of this, and there's a good opportunity to abstract, then I think we'd do it. I'm not 100% sure I see the right abstraction point yet.

abadami: Yeah, good point on waiting to abstract until we have a real need to do so, good thing to get a reminder on every so often. I made the revisions!

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -347,6 +381,77 @@ function useExampleQuery({ filter }: { filter: Record<string, unknown> }) {
...
+    [fakeDatabase, filter],
+  );
+
+  const [loading, setLoading] = useState(true);
+  const [data, setData] = useState<ExampleQueryData>();
+
+  useEffect(() => {
+    setLoading(true);
+    const timer = setTimeout(() => {
+      console.log(filteredDatabase);
```
bdow: **nit:** Not sure if we need the log. But if its helpful for the story, that's fine.

abadami: Forgot to remove that as I needed a quick debug on something. Thanks for catching it

src/components/Layout/GridTableLayout/QueryTable.tsx
```diff
@@ -40,10 +37,6 @@ export function QueryTable<R extends Kinded, QData, X extends Only<GridTableXss,
   const data = query.loading || query.error ? undefined : query.data;
   const rows = useMemo(() => createRows(data), [createRows, data]);
 
-  // Detect our `pageInfo` response pattern
-  const hasNextPage = data && getPageInfo && getPageInfo(data).hasNextPage;
-  const infoMessage = hasNextPage ? "Too many rows" : undefined;
-
```
bdow, line 46: **thought (non-blocking):** I don't really recall why this was here, but I'm not sure if it was _only_ for the pagination UX style we're removing. This could've been necessary for any pagination approach, including infinite scroll. I'm find removing it if its not being used and just moving on. Just calling it out as something for us to watch for. 
We have displayed a "Too many rows" in GridTable when the amount of data coming back was just too big and it'd cause issues rendering. That'd be a hint to users to use some filters, or to engineers that the page needed pagination and virtualization. But this doesn't appear to be the same thing... 🤔

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
bdow, line 1: **question:** Would it be worth adding a test here, or are the infinite scrolling behaviors already tested by GridTable?

bdow (approved):

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
abadami: I had just had that thought myself just now because I assumed GridTable was testing it, but it appears that it isn't! I'll look into this.

src/components/Layout/GridTableLayout/QueryTable.tsx
abadami: Dug into this a bit deeper. It seems like this is being used as an info message on the top row to signify that there is more to this query or that it has pagination support but the table isn't displaying everything at the moment.

This requires the `getPageInfo` prop to be passed into QueryTable, which is already an optional prop in GridTableLayout to begin with. Plus, in internal-frontend, this prop isn't being used in any of the usages of QueryTable.

I'd say we keep it removed and add it back in if asked for or we notice it's useful, perhaps in a way where we don't need an optional prop to nudge the users / engineers about it and it just does it. Any objections?

src/components/Table/GridTable.test.tsx
```diff
@@ -2538,6 +2538,192 @@ describe("GridTable", () => {
...
+        "
+        | Name  | Value |
+        | ----- | ----- |
+        | row 1 | 1     |
+        | row 2 | 2     |
+        "
+      `);
+    });
+
+    it("accepts endOffsetPx without error", async () => {
```
abadami, line 2705: This test might not be too necessary looking at it again but probably doesn't hurt to just keep it.

PR #1278, opened 2026-06-17
feat: improve SelectField add new UX

Jacob-Diebold:
## Summary

Improves `SelectField`'s `onAddNew` UX by replacing the static **"Add New"** row with a contextual **Add `"typed value"`** pseudo-option that appears only while the user is typing a value that doesn't already exist.

## Motivation

The old behavior always showed a generic **"Add New"** row at the bottom of the options list, even when nothing was typed. That made the affordance unclear and encouraged awkward workarounds (e.g. swapping to a `TextField`, or capturing Enter to commit typed text). This change makes add-new feel like a normal combobox flow: type → see a filtered list → explicitly choose the add-new row.

Pairs with Blueprint work to remove the `SelectMaybeNewField` wrapper and use Beam's native `onAddNew` directly.

## What changed

- **Dynamic add-new row** — While filtering, shows `Add "Kiwi"` at the bottom when the trimmed input doesn't exactly match an existing option label (case-insensitive). Hidden when input is empty or matches an existing label (e.g. typing `banana` when `Banana` exists).
- **Default Deduplication** — The `onAddNew` callback can only be triggered from a new pseudo-option; the option is hidden for matching values so it provides deduplication out of the box.
- **Visual treatment** — Add-new row label is italicized in the menu to show a visual distinction from actual options.
- **Docs & examples** — Expanded JSDoc on `onAddNew`; new `OnAddNew` Storybook story; expanded test coverage.

## Breaking changes

Consumers that selected the add-new option by the literal label **"Add New"** need to update:

- Type a value first, then select the dynamic row (e.g. `Add "Kiwi"`).
- Tests using `select(field, "Add New")` should type the value and select `Add "..."` instead.

Screenshot 2026-06-17 at 9 12 51 AM
Screenshot 2026-06-17 at 9 13 05 AM

https://github.com/user-attachments/assets/09f5f7d9-3cf0-484a-b9d4-8245ef057a96


src/inputs/SelectField.test.tsx
```diff
@@ -493,23 +493,155 @@ describe("SelectFieldTest", () => {
...
+  describe("onAddNew", () => {
+    const fruitOptions: HasIdAndName[] = [
+      { id: "1", name: "Apple" },
+      { id: "2", name: "Banana" },
+      { id: "3", name: "White Oak" },
+    ];
+
+    function typeToFilter(input: HTMLElement, value: string) {
+      click(input);
+      fireEvent.input(input, { target: { value } });
```
stephenh, line 505: Does the `type(input, value)` util not work for this? I kinda thought it did `fireEvent.input` for us.

stephenh (approved):
Looks great!

src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -511,7 +533,30 @@ function sortOptions<O>(options: O[], getOptionLabel: (opt: O) => string): O[] {
 
 /** A marker option to automatically add an "Unset" option to the start of options. */
 export const unsetOption = {};
-export const addNewOption = { id: "new", name: "Add New" };
+const addNewOptionId = "new";
```
bdow: **nit:** We might want a special `beam_` prefix or something here to help prevent potential collisions of actual `id` properties. Probably would be fine as is, as most ids are our tagged ids. But in the case options are set up using a union type or enums, then I think we'd want to supply a bit more of a special case here.

Jacob-Diebold: Makes sense! We might as well just to add a bit of extra safety!

src/inputs/internal/ComboBoxBase.tsx
```diff
@@ -110,45 +125,47 @@ export function ComboBoxBase<O, V extends Value>(props: ComboBoxBaseProps<O, V>)
...
-      propOptionMenuLabel
-        ? propOptionMenuLabel(o, Boolean(unsetLabel) && o === unsetOption, Boolean(onAddNew) && o === addNewOption)
-        : getOptionLabel(o),
+    (o: O) => {
+      const isAddNew = Boolean(onAddNew) && isAddNewOption(o);
+      if (propOptionMenuLabel) {
+        return propOptionMenuLabel(o, Boolean(unsetLabel) && o === unsetOption, isAddNew);
+      }
+      const label = getOptionLabel(o);
+      return isAddNew ? <span css={Css.add("fontStyle", "italic").$}>{label}</span> : label;
```
bdow: **thought (non-blocking)** I'm surprised we don't have an "italic" style built into Truss! Will note that and add it in soon. If that's something you're interested in trying, I'd encourage you to give it a shot. More exposure to more of our stack is always better! Let me know, and happy to chat through it.

bdow (approved):
👏 Nice work. I appreciate the tests and comments throughout. Thanks!

src/inputs/SelectField.test.tsx
Jacob-Diebold: Good call out, I hadn't seen that function. Unfortunately it looks like `type(input, value)` immediately blurs right after it calls `fireEvent.input` so it would reset the component state and wouldn't work here to be able to observe the dropdown state.

src/inputs/internal/ComboBoxBase.tsx
Jacob-Diebold: Thats a great idea, I want as much exposure as possible. I'll give it a shot and let you know if I have any questions while I'm working through it!

bdow:
One more thing - this isn't really a `chore:` PR. This would probably be more of a `feat:` or `fix:`. 
A chore is typically when just updating a test, or something that wouldn't impact a downstream consumer. Also when doing `chore` you also will not get a version of Beam generated. `fix:` bumps a patch version, and `feat:` bumps a minor version. Not a big deal on which to use, but this could be looked at as more of a `fix` than a `feat` maybe? 
I believe you'll just need to update the title with the right prefix. As this is doing a Squash & Merge, you'll just want to make sure the first line of that squashed commit message has the proper prefix.

Jacob-Diebold:
> One more thing - this isn't really a `chore:` PR. This would probably be more of a `feat:` or `fix:`. A chore is typically when just updating a test, or something that wouldn't impact a downstream consumer. Also when doing `chore` you also will not get a version of Beam generated. `fix:` bumps a patch version, and `feat:` bumps a minor version. Not a big deal on which to use, but this could be looked at as more of a `fix` than a `feat` maybe? I believe you'll just need to update the title with the right prefix. As this is doing a Squash & Merge, you'll just want to make sure the first line of that squashed commit message has the proper prefix.

@bdow  Ahh ok cool, thanks for the clarification!

PR #1280, opened 2026-06-18
feat: GridTable -> Card View mapping integration

abadami:
## Card View for GridTable

This PR adds in a new render variant for GridTable, simply called CardView, that reads GridTable data and outputs our Card View. We now have a nice mapping of card properties that we can add to column data to map out what a row will look like as a card instead. Currently proposed API will accept a shorthand value for everything except data, progress, and status. Those require a `{ kind: CardProperty, getValue?: (data) => {specific type for property} }`, to specify their more specific needs.


src/components/Table/types.ts
```diff
@@ -32,6 +33,22 @@ export type GridTableScrollOptions =
...
+  Status = "status",
+  Eyebrow = "eyebrow",
+  Badge = "badge",
+  Title = "title",
+  DataBlock = "data",
+  Progress = "progress",
+  Image = "image",
+}
+
+type CardPropertyDef<R extends Kinded> =
```
abadami: @bdow I mentioned I had a minor problem, but I realized a bit late that this method doesn't enforce "value" being set in the `GridContentCell` type, which I'm not a fan of since this method requires it. I did add a console warning for non-production environments, but I think that enforcement is less than idea since it's easily missable.

Thinking about it further what might make the most sense would be to just have this type enforce the `{ kind: CardProperty; getValue: (data) => {whatever is needed for the specific card property} }` for each CardProperty type. Biggest tradeoff compared to the current version is that we lose the nice shorthand way of handling the less complicated card properties (just putting `CardProperty.title`, for instance), but I think having the type enforcement here is worth it.

I do also recognize this is basically a variation of the function idea you had, but I think this method is a little more clear on what property you're declaring for instead of just a function, which I think is helpful for readability. That being said, what are your thoughts on these ideas for this particular issue?

src/components/Table/GridTable.tsx
```diff
@@ -598,34 +615,53 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
     return { ...style, minWidthPx };
   }, [contentWidth, inDocumentScrollLayout, style, tableWidth]);
 
+  console.log(visibleDataRows);
```
bdow: **request:** please remove

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,198 @@
...
+};
+
+export function TableCardView<X extends Only<Xss<TagXss>, X>>(props: TableCardViewProps<X>) {
+  const { title, imgSrc, eyebrow, badge, data, status, progress } = props;
+  const tid = useTestIds(props, "tableCardView");
+
+  return (
+    <div css={Css.p3.w("330px").h("100%").bshBasic.bgColor(Tokens.Surface).df.fdc.gap2.$} {...tid}>
+      <div css={Css.relative.$}>
+        <img css={Css.h("184px").w("100%").objectFit("cover").$} src={imgSrc} alt={String(title)} {...tid.image} />
```
bdow: **nit/suggestion:** Truss has some shorthand helpers for these, like `.h100` `.w100` `.hPx(184)` `.wPx(330)`
Would prefer to use those when possible - not a big deal.

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,198 @@
...
+    );
+  }
+  return card;
+}
+
+function toDisplay(value: GridCellValue): ReactNode {
+  if (value == null) return null;
+  if (typeof value === "string" || typeof value === "number" || typeof value === "boolean") return value;
+  return String(value);
+}
```
bdow: **request:** Can we reorganize this file so that helper methods are at the bottom. Typically we do:
```
// imports

// type ComponentProps / related Types

// function Component

// Helper methods & types unrelated to Component Props

File: src/components/Table/components/TableCard.tsx [hunk 5]
```diff
@@ -0,0 +1,198 @@
...
+
+export type TableCardViewProps<X> = {
+  imgSrc: string;
+  eyebrow?: GridCellValue;
+  title: GridCellValue;
+  badge?: GridCellValue;
+  data: CardData[];
+  status?: TagProps<X>;
+  progress?: AriaProgressBarProps;
```
bdow: **question:** I think we spoke about this, but since we aren't using the AriaProgressBar component, maybe we can simplify this type? Maybe call it something like `progressPercentage` and we can just accept a number between 0-100

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,198 @@
...
+              {badge && (
+                <span css={Css.sm.wsnw.$} {...tid.badge}>
+                  {toDisplay(badge)}
+                </span>
+              )}
+            </div>
+          )}
+        </div>
+        {data && data?.length > 0 && (
+          <div css={Css.dg.gtc("1fr 1fr").sm.$}>
```
bdow: **nit/suggestion:** grid template columns of `1fr 1fr` may not actually result in two equally sized columns. If that is what we want, then we should use `gtc("repeat(2, minmax(0, 1fr))")`.
More details on this here: https://css-tricks.com/equal-width-columns-in-css-grid-are-kinda-weird/

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,198 @@
...
+      <Link to={to} css={Css.tdn.color("unset").$} className={navLink}>
+        {card}
+      </Link>
+    );
+  }
+  if (rowStyle?.onClick) {
+    return (
+      <div onClick={() => rowStyle.onClick!(rs.row as any, api)} css={Css.cursorPointer.$}>
+        {card}
+      </div>
```
bdow: **nit/suggestion:** This might be something we're doing already in other places, but I'd like us to avoid using `div` elements for something clickable. Ideally, this is a `<button>` element. We get a lot of accessibility benefits out of the box by using the right elements. Our default button element should be unstyled - if its not, then thats a bigger problem, and would prefer a follow up ticket to address.

src/components/Table/GridTable.test.tsx
```diff
@@ -4500,3 +4500,155 @@ function TestFilterAndSelect(props: {
...
+    data: { address: "123 Main St", city: "Austin", beds: "3", bidOut: 65, status: "Active" },
+  },
+  {
+    kind: "data",
+    id: "row2",
+    data: { address: "456 Oak Ave", city: "Dallas", beds: "4", bidOut: 30, status: "Pending" },
+  },
+];
+
+describe("card view", () => {
```
bdow, line 4516: **request:** For all the tests here, can we use the `// Given, When, Then` comment pattern, please?

src/components/Table/types.ts
bdow: I'm not sure I'm 100% follow, so let's chat about it.

src/components/Table/components/TableCard.test.tsx
```diff
@@ -0,0 +1,57 @@
+import { TableCardView } from "src/components/Table/components/TableCard";
+import { render } from "src/utils/rtl";
+
+describe("TableCardView", () => {
```
bdow, line 4: **request:** Can we use the `//Given, when then` comment structure here, please? And for any other new tests created. Thanks!

src/components/Table/components/TableCard.tsx
abadami: I went with changing progress to just a number and then clamping the value between 0 - 100. I don't think there is an idiomatic way in Typescript to enforce a numeric range like that but if there is do let me know.

src/components/Table/components/TableCard.tsx
abadami: Oh interesting, nice article find on this. I'll keep this in mind with grid in the future and do the switch here since I'd say we'd want them equal

src/components/Table/components/TableCard.tsx
abadami: I glossed over this when I was reviewing. Nice shout out and I agree. Looks like button shouldn't cause any styling issues here so I've got it.

src/components/Table/components/TableCard.tsx
bdow: I think type enforcement is good enough. Clamping is fine, but hopefully its clear enough that progress is a percentage value, so it should just be clear that only a number between 0 and 100 would make sense.

src/components/Table/components/TableCard.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <TableCardView imgSrc={imgSrc} title="123 Main St" data={[]} status={{ text: "Active", type: "success" }} />,
+    );
+    // Then the status Tag shows the text
+    expect(r.tableCardView_status).toHaveTextContent("Active");
+  });
+
+  it("renders data blocks as 'header: value' pairs", async () => {
+    // Given a card with data blocks
+    // When rendered
+    const r = await render(<TableCardView imgSrc={imgSrc} title="123 Main St" data={data} />);
```
bdow, line 50: **request:** This looks like the only place we use the `const data` defined at the top of the tests. Can we inline it here instead? 
This is something we'll typically look for in code reviews - where is something declared more globally than really needed in tests and stories.  There's a concept out there you may have heard: tests should be DAMP, not DRY ([example article](https://enterprisecraftsmanship.com/posts/dry-damp-unit-tests/)).
The idea is when reading through tests, everything you need in order to understand the test should be right there with it. You should still share something code where it makes sense, but it's all about finding a balance for readability and maintainability.
GridTable tests are a great example of this. We share the types for `Row` and `Data`, but most of the tests define their own `const columns` and `const rows`. They just reuse the shapes/types.

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+  columns: GridColumnWithId<R>[];
+  rowStyle: RowStyle<any> | undefined;
+  api: GridTableApi<R>;
+};
+
+export function TableCard<R extends Kinded>({ rs, columns, rowStyle, api }: TableCardProps<R>) {
+  let title: string | undefined;
+  let eyebrow: string | undefined;
+  let badge: string | undefined;
+  let status: CardTag | undefined;
```
bdow, line 34: **nit (non-blocking):** I had never seen `CardTag` before. Looks like it's a more narrow type for supporting Tags. I don't know much about this `Card` component - so I'm hesitant to depend on it. Nothing to change here, but just something to keep an eye on. Just feels a little funny to be using a type for the Card component in order to render the Tag component. I don't mind it, but I don't think we'll want to be coupled to the Card component for too long. Or if anything, I could see us replacing the card component with `TableCard` and making it more generic.

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+        break;
+    }
+  }
+
+  if (!title) return null;
+
+  const card = (
+    <TableCardView
+      data-testid={`card_${rs.row.id}`}
+      imgSrc={(rs.row as any).imgSrc ?? ""}
```
bdow: **question:** Lazy ask, but why do we need the `as any` here?

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+      title={title}
+      eyebrow={eyebrow}
+      badge={badge}
+      status={status}
+      data={dataBlocks}
+      progress={progress}
+    />
+  );
+
+  const to = rowStyle?.rowLink?.(rs.row as any);
```
bdow: Looks like we're doing `as any` in a couple more places. Curious why, and if this is a precedent we already have. I'm hoping we can get the typing right on these

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { RowState } from "src/components/Table/utils/RowState";
+import { applyRowFn, isGridCellContent } from "src/components/Table/utils/utils";
+import { Css, Tokens } from "src/Css";
+import { navLink } from "src/css/CssReset";
+import { useTestIds } from "src/utils";
+import { defaultTestId } from "src/utils/defaultTestId";
+
+export type CardData = {
+  header: string;
+  value: ReactNode | string | number;
```
bdow: **nit:** Maybe `label` instead of `header`? Would be cool to try and use the `header` value if it is a `string` and omit the `label` here. But not a request to do that now. Maybe naming it `label` will also be clear in the API that this can deviate from the header's value if needed. Plus, its not really being displayed as a header, so I think semantically header could be good to avoid here. Granted, its' not a `<label>` either, but then we get to splitting hairs 😄 
And fwiw, now that I'm going, these would be a good fit for the definition list element [[mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl)].
>Common uses for this element are to implement a glossary or to display metadata (a list of key-value pairs). 
```
<dl>
  <dt>Sqft:</dt><dd>2,400</dd>
  <dt>Bedrooms:<dt><dd>2.5</dd>
  ...
</dl>
```

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -241,35 +249,317 @@ export function GridTableLayoutWithColor() {
...
+  baths: string;
+  elevations: string;
+  width: string;
+  depth: string;
+  bidOut: number;
+  status: string;
+};
+type PlanRow = SimpleHeaderAndData<PlanData>;
+
+const planColumns = [
```
bdow: **request:** Can we make `planColumns` and `planRows` to be functions that return the columns and rows. This is that eagerly vs lazily evaluating thing we talked about. I'm pretty sure when storybook starts up, its going to go through and evaluate everything to build out the list of stories.

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -241,35 +249,317 @@ export function GridTableLayoutWithColor() {
...
-        sorting: { on: "client", initial: [columns[2].id!, "ASC"] },
-      }}
-      withCardView={tileContent}
-      defaultView="card"
-    />
+    <TestProjectLayout pageTitle="With View Toggle">
+      <GridTableLayoutComponent
+        tableProps={{
+          columns: planColumns,
+          rows: planRows,
```
bdow: **thought:** Looks like these are only used here, maybe we can just inline them?

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -248,33 +247,30 @@ describe("GridTableLayout", () => {
...
+          withCardView
         />,
         withRouter(),
       );
 
       expect(r.viewToggleButton).toBeInTheDocument();
-      expect(r.query.cardContent).not.toBeInTheDocument();
 
       click(r.viewToggleButton);
       click(r.viewToggleButton_card);
```
bdow: **suggestion:** If we want this test to assert the view toggle, then I don't think we need to check local storage for this. We can probably also remove the clicking of the toggle button. It looks like testing the localstorage is done below in the test `persists view selection to localStorage when toggled`

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -552,3 +598,11 @@ function getRows(): Row[] {
     { kind: "data", id: "3", data: { name: "Gamma", value: 30 } },
   ];
 }
+
+function getHideableColumns() {
```
bdow: **question:** Why did we need to add this 'getHideableColumns'?

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+            ))}
+          </div>
+        )}
+        {progress !== undefined && (
+          <div css={Css.df.fdc.gap1.$}>
+            <div css={Css.df.aic.gap1.fs("10px").lh("14px").$}>
+              <div css={Css.w25.hPx(8).br4.bgGray200.$}>
+                <div css={Css.h100.br4.bgBlue500.w(`${clampProgress(progress)}%`).$} />
+              </div>
+              <span {...tid.progressValue}>{clampProgress(progress)}%</span>
```
bdow: **suggestion:** Maybe call `clampProgress` just once here and store the clamped value. This would avoid duplicate warnings for the same entry.

src/components/Table/components/TableCard.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+  component: TableCardView,
+} as Meta;
+
+const imgSrc = "plan-exterior.png";
+const data = [
+  { header: "Sq Ft", value: "2,400" },
+  { header: "Beds", value: "4" },
+  { header: "Baths", value: "3" },
+  { header: "Close Date", value: "Aug 2026" },
+];
```
bdow: **nit:** These are so simple I'm not concerned about them being module level (e.g. no need to wrap in a function), but I would prefer them to be at the bottom of the file.

src/components/Table/components/TableCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+        progress = slot.value;
+        break;
+    }
+  }
+
+  if (!title) return null;
+
+  const card = (
+    <TableCardView
+      data-testid={`card_${rs.row.id}`}
```
bdow: **suggestion:** Would prefer to the `useTestIds` hook. 

**thought (non-blocking):** Fwiw, my opinion (and it doesn't match everyones) is that `data-testid`s should avoid using entity ids in them. I personally prefer that a card has the same test id as every other card. If I want to test a specific card, then I just test that one card independently. Or if I am rendering a list of cards, then I can `queryAllByTestId` and assert the proper count, and then use array notation to pick a specific one. I believe our test helpers have this notion in them as well to do something like `expect(r.card_title_0).toHaveTextContent(...)`/`expect(r.card_title_1)...`

src/components/Table/cardSlots.ts
bdow, line 1: **nit (non-blocking):** This looks great. Only nit is that there are any comments, which could be helpful as new people start using these `cardEyebrowSlot`/etc functions. Doesn't need to be here, but could be in a readme or other markdown file that helps describe how the card layout for tables work.

src/components/Table/GridTable.stories.tsx
```diff
@@ -2419,3 +2420,209 @@ export function MultipleStickyColumns() {
...
+  return (
+    <div css={Css.df.fdc.vh100.$}>
+      <div css={Css.fg1.$}>
+        <GridTable as="card" columns={columns} rows={rows} />
+      </div>
+    </div>
+  );
+}
+
+export function CardViewInfiniteScroll() {
```
bdow, line 2542: **praise:** Nice! 🎉

src/components/Table/GridTable.test.tsx
```diff
@@ -4500,3 +4507,187 @@ function TestFilterAndSelect(props: {
...
+  beforeEach(() => {
+    setRunningInJest();
+  });
```
bdow, line 4519: **thought (non-blocking):** I wonder if we should do this always in setupTests.ts 🤔. Wouldn't hurt anything, and I think it'd be necessary to ensure all tests select the correct `const _as`.

src/components/Table/GridTable.test.tsx
```diff
@@ -4500,3 +4507,187 @@ function TestFilterAndSelect(props: {
...
+    expect(r.card_row2_title).toHaveTextContent("456 Oak Ave");
+  });
+
+  it("does not render the header row as a card", async () => {
+    // Given rows that include a header row and two data rows
+    // When rendered as card
+    const r = await render(<GridTable as="card" columns={cardColumns} rows={cardRows} />);
+    // Then only the data rows produce cards
+    expect(r.card_row1_title).toBeInTheDocument();
+    expect(r.card_row2_title).toBeInTheDocument();
```
bdow: **suggestion:** This would be a good place to count the number of card rows that actually render vs just checking that these two are in the document. If the header actually did render in some fashion, this test would still pass.

src/components/Table/GridTable.test.tsx
```diff
@@ -4500,3 +4507,187 @@ function TestFilterAndSelect(props: {
...
+      <GridTable
+        as="card"
+        columns={cardColumns}
+        rows={cardRows}
+        rowStyles={{ data: { rowLink: () => "/detail/1" } }}
+      />,
+      withRouter(),
+    );
+    // Then each card is wrapped in an anchor tag
+    expect(r.container.querySelectorAll("a").length).toBeGreaterThan(0);
```
bdow: **suggestion:** This assertion feels a little too broad and could easily be a false positive. 
I'd suggest we test the tag name and href value:
 `expect(r.card-testid.parentElement.tagName).toBe("a")` and `expect(r.card-testid.parentElement).toHaveAttribute('href', '/details/1')`

src/components/Table/GridTable.tsx
```diff
@@ -602,30 +618,47 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
...
-          infiniteScroll,
-          tableContainerRef,
-          persistScrollPosition,
+        {as === "card" ? (
+          // VirtuosoGrid requires layout measurement; in jsdom fall back to a plain div so tests can render cards.
+          runningInJest ? (
+            <div css={Css.py2.$}>
+              <div css={Css.dg.gtc("repeat(auto-fill, 330px)").jcc.gap3.p3.$}>{visibleDataRows}</div>
+            </div>
+          ) : (
```
bdow: **suggestion/thought:** what do you think about moving this `runningInJest` check down into `CardView`? Feels a little more appropriate there, imo. It's basically the "how does the card view render in jest" handling.

src/components/Table/components/TableCard.test.tsx
abadami: This all makes sense to me. I'm actually surprised I haven't heard of the DAMP acronym before. I've heard the concept of course, just not that it has an acronym for it like DRY does. But yes I'll do my best to internalize this going forward and move data inline.

src/components/Table/components/TableCard.tsx
abadami: Yeah, that's all completely fair. I'd say we can wait to see if we have a need of a generic card first before we make a decision on this since yes, I can also see us converting `TableCard` to be the new `Card` component. I'm not sure if we have any non-GridTable type use cases for card currently at the moment...

src/components/Table/components/TableCard.tsx
abadami: It's a good ask. We don't. I think I falsely assumed it had to do with the variable nature of the data in rows and didn't think much of it, but looking deeper it's just a `GridDataRow` as it should be. I wonder on a previous iteration I had it actually had some reason to be there, but in this current iteration it didn't but I didn't revert it back.

src/components/Table/components/TableCard.tsx
abadami: See other comment on this, but will remove.

src/components/Table/components/TableCard.tsx
abadami: I'll go ahead and switch it to label since I think it makes more sense. Going along with the idea to try and use the header value, I think that'd it might be cool to see if we can allow the data content value to be a function and it's able to interpret it as the values for card slot, content, and value in the GridCellContent. I think in the current version it's able to do that for just content and value. 

I guess though that it won't know which card slot to use so maybe that's wishful thinking...but maybe there's a way. Something worth exploring later perhaps.

abadami: Oh also I didn't know about the definition list element and yeah, this is a fantastic use case for it. I'll switch it as well.

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
abadami: Yep, just here. I'll inline them to this story.

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
abadami: Since they're only used in a specific story, I'll inline them

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
abadami: To differentiate between the hide column set and not, but yeah looking at it again we don't have an actual reason to do that. I'll just set the corresponding columns in `getColumns` to `canHide` and remove this.

src/components/Table/components/TableCard.tsx
abadami: Good call on this one.

src/components/Table/components/TableCard.tsx
abadami: Ah forgot to convert this to useTestId. On it. Also, I wasn't aware that the test helpers might have something build in already to handle lists like this (without having to specify our own), I'll try it out and see if it just works.

bdow: Yea, these are our own internal test helpers. Just some fancy stuff we've tried to add to reduce the amount we need to write ourselves in tests. Code reviews are most likely where you'll learn about homebound specific tooling.

src/components/Table/cardSlots.ts
abadami: I was gonna go ahead and do this one, but I think it'd warrant this being a part of a documentation discussion as a whole. I think it could be useful to have a MD file for each component (or even multiple for select components like, for example, I think the card view would be appropriate for GridTable).

src/components/Table/GridTable.tsx
abadami: Oh yeah for sure. Shoot I feel like I was thinking about this too and just forgot to go back. I should really add TODOs for my "I should come back to this" stuff lol

src/components/Table/GridTable.test.tsx
abadami: That's not a bad idea. Granted only GridTableLayout and GridTable tests seem to have a need to actively use this set function at the moment. I guess if we see opportunity for it later down the line elsewhere we should go ahead and do it.

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -242,34 +250,321 @@ export function GridTableLayoutWithColor() {
...
+                cardSlot: cardDataBlockSlot({ label: "Sq ft", value: sqft }),
+              }),
+            }),
+            column<PlanRow>({
+              id: "beds",
+              name: "Bed",
+              header: "Bed",
+              data: ({ beds }) => ({
+                content: beds,
+                value: beds,
```
bdow, line 311: **nit (non-blocking):** if `content` and `value` are the same, then `value` is not necessary to set.

bdow (approved):

PR #1281, opened 2026-06-22
feat: Environment Banners

bdow:
Adds new layout wrapper to support environment banners that serve two purposes:
1. Inform the user of which environment they are in.
2. Inform the user if they are impersonating another user.

These banners stay stuck at the top of the screen at all times.

Includes new semantic colors specific for environments. Adds new method for setting environment favicons, using the same AppEnvironment type used by the banner component.

Updates all other layouts to be aware of the new banner and its presence to properly align when it is there.


src/components/Tooltip.tsx
```diff
@@ -35,7 +35,7 @@ type TooltipProps = {
 const gapCrossingDelay = 100;
 
 export function Tooltip(props: TooltipProps) {
-  const { placement, children, title, disabled, delay = 0, bgColor, xss } = props;
+  const { placement, children, title, disabled, delay = 150, bgColor, xss } = props;
```
abadami, line 38: **question**: Out of curiosity, what is the reason for the change in the delay default here? Is this a situation of we usually want a delay more often than not?

I guess as a note to that, this would affect every tool tip without a delay set to have one now. I doubt it's a big deal but good to note.

abadami (approved):

src/components/Tooltip.tsx
bdow: It was proposed by Julia in one of our channels. I agree it makes sense to not have it be so eager to display. Current behavior has the tooltip show immediately on mouse over, which can trigger it to flicker when just moving your cursor over a tooltip element.

PR #1282, opened 2026-06-22
feat: grid table row pinning

apattersonATX-HB:
[app-1523](https://linear.app/homebound-team/issue/APP-1523/gridtable-row-pinning)
----
Adds **row pinning** to `GridTable` — pin a row and it sticks to the top, staying visible while the rest of the table scrolls. 

- `useGridTableApi`: `pinRow` / `unpinRow` / `togglePinnedRow` / `isPinnedRow` / `getPinnedRowIds`
- `pinColumn` (with a `PinToggle` control), plus `initPinned` to start a row pinned
- Pinned rows get a green highlight and stay visible even when a filter would otherwise hide them
- Reuses the existing sticky-header block across all three render modes (div / table / virtual) instead of inventing a new sticky layer
- Kept deliberately separate from the existing declarative `pin` prop (that one only reorders within a group); this is the interactive, sticky version layered on top

Screenshot 2026-06-23 at 10 57 35 AM


src/components/Table/utils/RowState.ts
```diff
@@ -320,7 +337,18 @@ export class RowState<R extends Kinded> {
...
+   *
+   * NOTE: declarative `row.pin: "first" | "last"` predates the interactive row-pinning added
+   * and is a *different* feature — it reorders a row within its sibling group and (here)
+   * exempts it from the filter; it is NOT the sticky "pin to top". The shared "pin" vocabulary is
+   * confusing (`row.pin` / this getter vs the runtime `pinned` / `pinRow` / `pinColumn`). Renaming
+   * this getter from `isPinned` -> `isAlwaysVisible` avoids the `isPinned` vs `pinned` clash without a
+   * breaking change. A larger follow-up could rename the declarative prop itself (e.g. `pin` ->
+   * `anchor`) to free up "pin" for the sticky feature — see the PR description.
+   */
+  private get isAlwaysVisible(): boolean {
```
apattersonATX-HB, line 346: @blambillotte @bdow 
I renamed this internal prop from `isPinned` to `isAlwaysVisible` to _hopefully_ reduce some confusion between the two features. This PR adds **runtime row pinning**. The old pre-existing declarative `pin` prop (`GridDataRow.pin: "first" | "last"`) is a *different* feature. That reorders a row **within its sibling group** and exempts it from client-side filtering;

So now we have confusing syntax floating around. "Pin" now means two things:
- `row.pin` (old) → in-group ordering + filter-exemption
- runtime pinning (new) → sticky hoist to the top of the table

So what i'm wondering is do we want to rename the consumer API to from `pin` to something like `isAlwaysVisible` (or any other name!), to keep the two distinct 🤔

stephenh: @apattersonATX-HB naive question, but does this _have_ to be a totally separate feature? Like would a new `pin: "top"` be basically what we're being asked to do?

stephenh: Clarifying, we really should not have "two pins" floating around, even if we slightly rename one -- we need to either:

a) Keep using the `row.pin` API but with a new `pin: "top"` value, or
b) Remove `row.pin` and use this new `initPinned` API-based approach that is more similar to how selected/collapsed works.

stephenh: ...apologies for the stream-of-conciousness thinking, but maybe these really are two separate features:

* This "new pinning" is the "true pinning" feature, and
* We rename the "old pinning" / `pin` attribute to something like `sort: "first" | "last"` because it really is more limited "sort within your group" than true "pin to the top of the table"?

apattersonATX-HB: Exactly they're two separate features.

Maybe `groupSort`? `anchor` or `pinInGroup`? I'm terrible with names so i don't have a strong opinion just looking for consensus on the rename before i go ahead and do it since I imagine it'll be a lot of updates around BP. Whether we decide to do that as part of this PR or a follow up

stephenh: @bdow how do you feel about renaming the current `pin: first | last` to `sort: first | last`? Feels like that is a pretty apt/succinct name for that it's doing -- just overriding sort order within groups of regular rows.

bdow: 🤔 Yea, `pin` was a misleading name in hindsight. 
I'm fine with `sort` as the new property. Still doesn't feel great, but I don't have any better ideas. Throwing some out though:
```
fixed
fixedSort
pinInPlace
```
But either way, I don't think any are better than the other.

src/components/Table/components/PinToggle.tsx
```diff
@@ -0,0 +1,30 @@
+import { useContext } from "react";
+import { IconButton } from "src/components/IconButton";
+import { TableStateContext } from "src/components/Table/utils/TableState";
+import { Palette } from "src/Css";
+import { useComputed } from "src/hooks";
+import { useTestIds } from "src/utils";
+
+type PinToggleProps = {
+  id: string;
```
stephenh: Maybe call this `rowId`? A little ambiguous being "just id"

bdow (approved):
Looks great, performs great. Great tests. Great comments. Great great great. 😄 
I'll be very interested to see how this works with the new document level scrolling, but since you append these pinned rows to the header row wrapper, then I think we're in good shape and get the sticky behavior properly.

stephenh (approved):
I'm good with this shipping, but @apattersonATX-HB would like "renaming the old pin to `sort`" to be considered part of this work / a next-PR follow-up instead of a later-ish thing.

src/components/Table/components/PinToggle.tsx
apattersonATX-HB: ahh good call

PR #1283, opened 2026-06-22
feat: Extend SelectAndWait util

Jacob-Diebold:
## Summary

Extends `@homebound/beam/rtl` `selectAndWait` so tests can work with async search-driven and creatable Beam comboboxes without manual open → type → wait → select boilerplate.

### Why

Beam `SelectField` comboboxes often don't show options until the user types (async `onSearch`) or expose an `Add "…"` row for creatable values (`onAddNew`). Tests were repeating the same low-level steps:

    click(field);
    fireEvent.input(field, { target: { value: "…" } });
    await wait();
    select(field, "Option label");

The goal is a higher-level helper: pass the field and the outcome you want, and let Beam's RTL utilities handle combobox mechanics so it works out of the box for free.

### What changed

**`src/utils/rtlUtils.tsx`**

1. **Auto-search when the listbox is empty**  
   For single-string values, if opening the listbox yields zero `[role=option]` elements:
   - Wait briefly (lazy-load / debounced search)
   - If still empty, type the target option label via `fireEvent.input`
   - Wait, reopen, then select as before  

   Uses `fireEvent.input` directly rather than rtl-utils `input`, which blurs and can close the dropdown.

       await selectAndWait(r.food, "Apple Pie");

2. **`{ addNew: true }` for creatable fields**  
   Types the value and selects the `Add "<value>"` row rendered by `ComboBoxBase`:

       await selectAndWait(r.brand, "GE", { addNew: true });


src/utils/rtlUtils.tsx
```diff
@@ -164,16 +164,75 @@ export function select(select: HTMLElement, value: string | string[]) {
...
 }
 
-export async function selectAndWait(select: HTMLElement, value: string | string[]): Promise<void> {
+/**
+ * Selects an option from the Beam SelectField, MultiSelectField, and TreeSelectField components.
+ *
+ * When `value` is a single string and the listbox opens empty, types the value to trigger async
+ * search before selecting. Multi-select (`value` array) assumes options are already loaded.
+ *
+ * When `{ addNew: true }`, types the value and selects the `Add "<value>"` creatable row.
```
stephenh, line 175: Great in-source comment, maybe move it to the `selectAndWait` directly instead of the ops.

src/utils/rtlUtils.tsx
```diff
@@ -164,16 +164,75 @@ export function select(select: HTMLElement, value: string | string[]) {
...
+    await maybeAutoSearch(select, value);
+  }
+
   return allowAndWaitForAsyncBehavior(() => {
-    const optionValues = Array.isArray(value) ? value : [value];
     optionValues.forEach((optionValue) => selectOption(select, optionValue));
   });
 }
 
+function addNewOptionLabel(value: string): string {
```
stephenh: Personally would be fine just inlining this "only invoked once" function, particularly since we're not sharing the logic with the real impl (which is fine, it's trivial)

stephenh (approved):
Nice!

src/utils/rtlUtils.tsx
Jacob-Diebold: Thanks for catching that. I didn't catch that I had accidentally added the opt type between the js doc and the actual function I wrote it on. 🤦‍♂️ It does not make sense to be on the option type at all lol.

PR #1284, opened 2026-06-23
fix: Fix weird style issues when env banner overflows

bdow:
Before | After
-- | -- 
Screenshot 2026-06-23 at 8 49 09 AM | Screenshot 2026-06-23 at 8 48 49 AM


allangaldinosilva (approved):
nice!! 👏

PR #1285, opened 2026-06-23
feat: Support setting document.title from PageHeader

bdow:
Whenever using the `PageHeader` or `PageHeaderLayout` components, then the document title would get set for free via the `PageHeader` component.
Examples:

Results in: **[QA] Product Offerings | Blueprint | Homebound**
```
<BeamContext 
  documentTitleConfig={{
    env: "qa", // Sets environment prefix
    suffix: "Blueprint | Homebound",
  }}
>
  <PageHeader title="Product Offerings" />
</BeamContext>
```

Results in: **Schedule | 123 Sesame St | Blueprint | Homebound**
```
<BeamContext 
  documentTitleConfig={{
    env: "prod", // Prod env does not apply a prefix
    suffix: "Blueprint | Homebound",
  }}
>
  <PageHeader title="Schedule" documentTitleSuffix="123 Sesame St" />
</BeamContext>
```

### DIFF
--- a/src/components/DocumentTitle/DocumentTitleContext.tsx
+++ b/src/components/DocumentTitle/DocumentTitleContext.tsx
@@ -1,0 +1,21 @@
+import { createContext, PropsWithChildren, useContext, useMemo } from "react";
+import type { AppEnvironment } from "src/components/EnvironmentBanner/EnvironmentBanner";
+
+export type DocumentTitleConfig = {
+  env: AppEnvironment;
+  suffix?: string;
+};
+
+const DocumentTitleContext = createContext<DocumentTitleConfig | undefined>(undefined);
+
+export function DocumentTitleProvider(props: PropsWithChildren<DocumentTitleConfig>) {
+  const { children, env, suffix } = props;
+  const value = useMemo(() => ({ env, suffix }), [env, suffix]);
+
+  return <DocumentTitleContext.Provider value={value}>{children}</DocumentTitleContext.Provider>;
+}
+
+/** Used by {@link useDocumentTitle}. */
+export function useDocumentTitleConfig(): DocumentTitleConfig | undefined {
+  return useContext(DocumentTitleContext);
+}
--- a/src/components/PageHeader.test.tsx
+++ b/src/components/PageHeader.test.tsx
@@ -1,3 +1,4 @@
+import { BeamProvider } from "src/components/BeamContext";
 import { Button } from "src/components/Button";
 import { PageHeader } from "src/components/PageHeader";
 import { Tab } from "src/components/Tabs";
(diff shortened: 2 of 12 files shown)


src/components/PageHeader.tsx
```diff
@@ -1,19 +1,23 @@
 import { ReactNode } from "react";
 import { RouteTabsProps, Tabs, TabsContentXss, TabsProps } from "src/components/Tabs";
 import { Css, Only, Tokens } from "src/Css";
+import { useDocumentTitle } from "src/hooks/useDocumentTitle";
 import { useTestIds } from "src/utils";
 
-export interface PageHeaderProps<V extends string, X> {
-  title: ReactNode;
+export type PageHeaderProps<V extends string, X> = {
+  title: string;
```
bdow, line 8: Purposefully restricting this to just a `string` for two reasons:
1. Keeps title consistent in terms of styling.
2. Allows to reuse this title in the `document.title`.

If there is need for additional elements around the title string, then those should be separate props with clear intent.

src/components/PageHeader.test.tsx
```diff
@@ -1,16 +1,49 @@
...
   });
 
+  it("sets document.title when BeamProvider documentTitleConfig is configured", async () => {
+    // Given BeamProvider with documentTitleConfig and a PageHeader
+    // When rendered
+    await render(
+      <BeamProvider documentTitleConfig={{ env: "local", suffix: "Blueprint | Homebound" }}>
+        <PageHeader title="Projects" />
+      </BeamProvider>,
+      { omitBeamContext: true },
```
abadami, line 26: **question**: I find it interesting that we use this `omitBeamContext` flag despite using the Beam Provider...oh, is it to omit the default beam context so that we can pass in our own provider values?

src/components/DocumentTitle/DocumentTitleContext.tsx
```diff
@@ -0,0 +1,21 @@
...
+import type { AppEnvironment } from "src/components/EnvironmentBanner/EnvironmentBanner";
+
+export type DocumentTitleConfig = {
+  env: AppEnvironment;
+  suffix?: string;
+};
+
+const DocumentTitleContext = createContext<DocumentTitleConfig | undefined>(undefined);
+
+export function DocumentTitleProvider(props: PropsWithChildren<DocumentTitleConfig>) {
```
abadami, line 11: **question**: I'm a little curious on the what your reasoning was for using a Context for setting the env and suffix values for the document title. I rationalize it myself as I state below here and I think it makes sense, but I was curious if there was any other reasoning I could be missing.

My initial instinct would be baking it into the `useDocumentTitle` hook directly. But I suppose for the suffix, that'll get repetitive, and for the env, that could be different from application from application and thus not reasonable to hardcode in there.(maybe less likely for Homebound, but possible for anyone choosing to use Beam).

abadami (approved):

src/components/PageHeader.test.tsx
bdow: That's correct. Just to provide our own values into the BeamProvider.

src/components/DocumentTitle/DocumentTitleContext.tsx
bdow: The idea is that a user will set this on BeamProvider and have a consistent suffix and environment throughout the app, and not feel so boiler-platey when engineers are just want to set the page title of "Schedules". That's all they need to do.
One of our goals with Beam and all this layout work is to make it so most everything for the consuming apps is just a "plug-n-play" sort of experience. The less they need to do and remember to set up, the better.

Potentially consuming apps don't need this and we avoid wrapping in the extra provider.

PR #1287, opened 2026-06-24
feat: Update design of Outline Icon Button

abadami:
## Outline Icon Button Design Update

This PR updates the icon button to match the outline styling shown here: https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=58-3735&t=pQoJQEzLTmVDP4LR-4

### DIFF
--- a/src/components/IconButton.tsx
+++ b/src/components/IconButton.tsx
@@ -78,25 +78,19 @@ export function IconButton(props: IconButtonProps) {
 
   const isCircle = variant === "circle";
   const isOutline = variant === "outline";
-  const styles = useMemo(
-    () => ({
+  const styles = useMemo(() => {
+    const variantKey = isCircle ? "circle" : isOutline ? "outline" : compact ? "compact" : "default";
+    const vs = variantStyles[variantKey];
+    return {
       ...iconButtonStylesReset,
-      ...(isCircle ? iconButtonCircle : isOutline ? iconButtonOutline : compact ? iconButtonCompact : iconButtonNormal),
-      ...(isHovered &&
-        (isCircle ? iconButtonCircleStylesHover : isOutline ? iconButtonOutlineStylesHover : iconButtonTokenHover)),
-      ...(isFocusVisible || forceFocusStyles
-        ? isCircle
-          ? iconButtonCircleStylesFocus
-          : isOutline
-            ? iconButtonOutlineStylesFocus
-            : iconButtonStylesFocus
-        : {}),
-      ...(active && (isCircle ? activeStylesCircle : isOutline ? activeStylesOutline : iconButtonTokenHover)),
+      ...vs.base,
+      ...(isHovered && vs.hover),
+      ...((isFocusVisible || forceFocusStyles) && vs.focus),
+      ...(active && vs.pressed),
       ...(isDisabled && iconButtonStylesDisabled),
       ...(bgColor && Css.bgColor(bgColor).$),
-    }),
-    [isHovered, isFocusVisible, isDisabled, compact, isCircle, isOutline, active, bgColor, forceFocusStyles],
-  );
+    };
+  }, [isHovered, isFocusVisible, isDisabled, compact, isCircle, isOutline, active, bgColor, forceFocusStyles]);
   const iconColor = isCircle ? circleIconColor : defaultIconColor;
 
   const buttonAttrs = {
@@ -153,3 +147,29 @@ const iconButtonCircleStylesFocus = Css.bgBlue100.bcBlue700.$;
 const iconButtonStylesDisabled = Css.cursorNotAllowed.$;
 const activeStylesCircle = Css.bgGray200.bcGray200.$;
 const activeStylesOutline = Css.bgGray200.$;
+const variantStyles = {
+  default: {
+    base: iconButtonNormal,
+    hover: iconButtonTokenHover,
+    focus: iconButtonStylesFocus,
+    pressed: iconButtonTokenHover,
+  },
+  compact: {
+    base: iconButtonCompact,
+    hover: iconButtonTokenHover,
+    focus: iconButtonStylesFocus,
+    pressed: iconButtonTokenHover,
+  },
+  circle: {
+    base: iconButtonCircle,
+    hover: iconButtonCircleStylesHover,
+    focus: iconButtonCircleStylesFocus,
+    pressed: activeStylesCircle,
+  },
+  outline: {
+    base: iconButtonOutline,
+    hover: iconButtonOutlineStylesHover,
+    focus: iconButtonOutlineStylesFocus,
+    pressed: activeStylesOutline,
+  },
+} as const;
--- a/src/components/IconButton.tsx
+++ b/src/components/IconButton.tsx
@@ -80,13 +80,13 @@ export function IconButton(props: IconButtonProps) {
   const isOutline = variant === "outline";
   const styles = useMemo(() => {
     const variantKey = isCircle ? "circle" : isOutline ? "outline" : compact ? "compact" : "default";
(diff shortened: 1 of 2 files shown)


src/components/IconButton.tsx
```diff
@@ -133,14 +134,42 @@ const iconButtonStylesReset = Css.bcTransparent.bss.bgTransparent.cursorPointer.
...
-const iconButtonOutline = Css.br8.wPx(48).hPx(40).bgColor(Tokens.Surface).bc(Tokens.SurfaceSeparator).ba.bw1.df.jcc
-  .aic.$;
+const iconButtonOutline = Css.br8.wPx(42).hPx(40).bgColor(Tokens.Surface).bcGray300.ba.bw1.df.jcc.aic.$;
 /** Semantic hover fill; contrast is driven by `--b-*` when inside {@link ContrastScope}. */
 const iconButtonTokenHover = Css.bgColor(Tokens.NeutralFillHoverStrong).$;
 export const iconButtonStylesHover = Css.bgGray200.$;
 export const iconButtonContrastStylesHover = iconButtonTokenHover;
 export const iconButtonCircleStylesHover = Css.bgBlue100.bcBlue200.$;
+const iconButtonOutlineStylesHover = Css.bgGray100.$;
+const iconButtonOutlineStylesFocus = Css.boxShadow(`0 0 0 2px #ffffff, 0 0 0 5px var(${Tokens.Primary})`).$;
```
bdow: **question:** Is this different from the `bshFocus`?

src/components/IconButton.tsx
```diff
@@ -133,14 +134,42 @@ const iconButtonStylesReset = Css.bcTransparent.bss.bgTransparent.cursorPointer.
...
 const iconButtonStylesDisabled = Css.cursorNotAllowed.$;
 const activeStylesCircle = Css.bgGray200.bcGray200.$;
+const activeStylesOutline = Css.bgGray200.$;
+const variantStyles = {
+  default: {
+    base: iconButtonNormal,
+    hover: iconButtonTokenHover,
+    focus: iconButtonStylesFocus,
+    pressed: iconButtonTokenHover,
+  },
```
bdow, line 141: **question/suggestion:** What do you think about inlining the styles here? I'm okay with duplication as it looks like we're doing the same style for focus and hover (I'd suspect these will deviate at some point soon).
I think it'd just be easier to read vs jumping from const to const.

src/components/IconButton.tsx
```diff
@@ -133,14 +134,42 @@ const iconButtonStylesReset = Css.bcTransparent.bss.bgTransparent.cursorPointer.
...
 export const iconButtonContrastStylesHover = iconButtonTokenHover;
 export const iconButtonCircleStylesHover = Css.bgBlue100.bcBlue200.$;
+const iconButtonOutlineStylesHover = Css.bgGray100.$;
+const iconButtonOutlineStylesFocus = Css.boxShadow(`0 0 0 2px #ffffff, 0 0 0 5px var(${Tokens.Primary})`).$;
 const iconButtonStylesFocus = Css.bcBlue700.$;
 const iconButtonCircleStylesFocus = Css.bgBlue100.bcBlue700.$;
 const iconButtonStylesDisabled = Css.cursorNotAllowed.$;
 const activeStylesCircle = Css.bgGray200.bcGray200.$;
+const activeStylesOutline = Css.bgGray200.$;
+const variantStyles = {
```
abadami, line 135: I added this to clean up the styles memo a little bit, help make it a little more readable. Looking at the overall design for IconButton though, there's a decent chance we may not need this depending on what will be supported and locking in on consistency for the different states across the variants.

src/components/IconButton.tsx
```diff
@@ -78,18 +78,19 @@ export function IconButton(props: IconButtonProps) {
 
   const isCircle = variant === "circle";
   const isOutline = variant === "outline";
-  const styles = useMemo(
-    () => ({
+  const styles = useMemo(() => {
+    const variantKey = isCircle ? "circle" : isOutline ? "outline" : compact ? "compact" : "default";
+    const vs = variantStyles[variantKey];
```
bdow: **suggestion:** Might look nicer to destructure here:
```
const { base, hover, focus, pressed } = variantStyles[variantKey];
```

src/components/IconButton.tsx
```diff
@@ -133,14 +134,42 @@ const iconButtonStylesReset = Css.bcTransparent.bss.bgTransparent.cursorPointer.
...
 const iconButtonCircle = Css.br100.wPx(48).hPx(48).bcGray300.ba.bw1.df.jcc.aic.$;
-const iconButtonOutline = Css.br8.wPx(48).hPx(40).bgColor(Tokens.Surface).bc(Tokens.SurfaceSeparator).ba.bw1.df.jcc
-  .aic.$;
+const iconButtonOutline = Css.br8.wPx(42).hPx(40).bgColor(Tokens.Surface).bcGray300.ba.bw1.df.jcc.aic.$;
 /** Semantic hover fill; contrast is driven by `--b-*` when inside {@link ContrastScope}. */
 const iconButtonTokenHover = Css.bgColor(Tokens.NeutralFillHoverStrong).$;
 export const iconButtonStylesHover = Css.bgGray200.$;
 export const iconButtonContrastStylesHover = iconButtonTokenHover;
 export const iconButtonCircleStylesHover = Css.bgBlue100.bcBlue200.$;
+const iconButtonOutlineStylesHover = Css.bgGray100.$;
```
bdow: **nit (non-blocking):** There's a new variant of this outline button coming which is basically the "dark" mode ([Figma](https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=58-3735&t=MFozTapIf0qxTRMM-4)). Would be nice to use the semantic tokens here when we get to building that instance out and we can just wrap that in the ContrastScope component when we need the dark mode.

src/components/IconButton.tsx
abadami: There actually is a very slight difference as `bshFocus` has 4px of spread on the primary color compared to this using 5px, which the figma has set. That being said, I forgot about the box shadow shorthand. I personally think it may be worth bringing this up to design to see if we want to change the `bshFocus` value itself to 5px.

I do think this should be using the `bshFocus` value regardless though. I don't see a reason for a deviation at the moment...

bdow: 100% bring this up to design. Ideally this focus style is consistent.

src/components/IconButton.tsx
abadami: That's a good improvement to this, as I agree jumping between consts would get annoying.

src/components/IconButton.tsx
abadami: I'll went ahead and switched it

bdow (approved):

src/components/IconButton.tsx
```diff
@@ -130,17 +131,30 @@ export function IconButton(props: IconButtonProps) {
...
+    pressed: Css.bgColor(Tokens.NeutralFillHoverStrong).$,
+  },
+  compact: {
+    base: Css.hPx(18).wPx(18).br4.bw1.$,
+    hover: Css.bgColor(Tokens.NeutralFillHoverStrong).$,
+    focus: Css.bcBlue700.$,
+    pressed: Css.bgColor(Tokens.NeutralFillHoverStrong).$,
+  },
+  circle: {
+    base: Css.br100.wPx(48).hPx(48).bcGray300.ba.bw1.df.jcc.aic.$,
```
stephenh, line 149: Just lurking through all PR reviews, but anytime we do w/h=48, there is a `sqPx(48)` as well 😅 (fyi, no follow up necessary)

PR #1288, opened 2026-06-24
feat: Add scroll to top button for document scrolling layouts

bdow:


src/layouts/DocumentScrollToTopButton.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    expect(r.getAllByTestId("documentScrollToTop")).toHaveLength(1);
+  });
+});
+
+function mockDocumentViewport(width: number, height: number): void {
+  Object.defineProperty(document.documentElement, "clientWidth", { configurable: true, get: () => width });
+  Object.defineProperty(document.documentElement, "clientHeight", { configurable: true, get: () => height });
+}
+
+function scrollWindow(y: number): void {
```
abadami: **suggestion**: We do have a lot of scroll related logic now, I wonder if this might be more globally useful within Beam for testing. Do you think it might be worth being exported in a testing utility file?

src/layouts/DocumentScrollToTopButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const scrollToTop = useCallback(() => {
+    window.scrollTo({ top: 0, behavior: "smooth" });
+  }, []);
+
+  return (
+    <div
+      {...tid.wrapper}
+      aria-hidden={!visible}
+      {...(!visible ? { inert: "" } : {})}
```
abadami, line 37: **question**: Never knew about the inert attribute until today! That being said, I'm curious why you used it instead of using something like disabled on the IconButton?

bdow: I'm doing both right now. I initially didn't disable the button because it makes it flicker to the disabled state once we hide the arrow, but decided to still disable it anyways.

src/layouts/DocumentScrollToTopButton.test.tsx
bdow: Good idea. Will do

src/layouts/DocumentScrollToTopButton.tsx
abadami: Oh you did! I can't believe I missed that reading here.

After commenting I did have the thought that you might've been doing it if the disabled state was showing for a split second on the transform transition...that would make sense.

My personal thought that it dropping out of visibility should be enough of an indication that it's "disabled". Though if somehow the button is still visible in some way, then the disabled state is going to clear up potential confusion. I don't see that happening currently.

abadami (approved):

PR #1289, opened 2026-06-25
feat: Update beam select component for "Option Type" input

kbesingeryeh:
* Update `IconCardGroup` to use new name `SelectCard`
* Add new icons

Screenshot 2026-06-25 at 11 53 37 AM


src/forms/BoundSelectCardGroupField.test.tsx
```diff
@@ -25,21 +25,26 @@ const categories: IconCardGroupItemOption<Category>[] = [
...
 
-describe("BoundIconCardGroupField", () => {
+describe("BoundSelectCardGroupField", () => {
   it("shows the label", async () => {
     const author = createObjectState(formConfig, { favoriteGenres: [Category.Math] });
-    const r = await render(<BoundIconCardGroupField field={author.favoriteGenres} options={categories} />);
+    const r = await render(<BoundSelectCardGroupField field={author.favoriteGenres} options={categories} />);
     expect(r.favoriteGenres_label).toHaveTextContent("Favorite Genres");
   });
+  it("renders option descriptions", async () => {
```
stephenh, line 35: Should have newlines between `it` tests

src/inputs/SelectCard.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+      {children}
+    </div>
+  );
+}
+
+/** Hover styled version of the SelectCard — uses a scoped stylesheet to force hover styles for visual testing. */
+function HoveredSelectCard(args: SelectCardProps) {
+  return (
+    <div className="hovered-select-card">
+      <style>{`.hovered-select-card button { border-width: 2px; border-color: ${Palette.Blue600}; }`}</style>
```
stephenh: @kbesingeryeh cc @bdow honestly I forget the best Truss-v2 way of forcing styles, but it is generally not copy/pasting the definitions like this.

Brandon, what's your preference? A `forceHover` prop like we've done before?

stephenh (commented):
Generally lgtm, will defer to Brandon for 👍

src/inputs/SelectCard.stories.tsx
blambillotte: Since this is a storybook-only ask, what if we did `HoveredSelectCard.play = async ({ canvasElement }) => ...` to just interact with a `userEvent.hover` rather than trying to modify the styles?

src/inputs/SelectCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const toggleState = useToggleState(ariaProps);
+  const { inputProps } = useCheckbox(checkboxProps, toggleState, ref);
+
+  const styles = useMemo(
+    () => ({
+      ...baseStyles,
+      // The two layouts are meaningfully different: a label-only card is a fixed, compact,
+      // center-aligned box (the original IconCard sizing), while a card with a description
+      // fills the row evenly, stretches to a shared height, and top-aligns its content.
+      ...(description ? withDescriptionStyles : withoutDescriptionStyles),
```
blambillotte: question: since the size style is set at the card level, what happens when you have a mix of cards with or without descriptions? Does the `CardGroup` look okay? Or do we need some styles at the group level to also check "do any of my cards have a description"?

src/inputs/SelectCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+      ...(isHovered && !isDisabled && hoverStyles),
```
blambillotte, line 56: Also noticed the hover border styles cause some layout jank where not-hovered elements resize when viewed as a group - looks like we need to account for the border width always (or tweak the styles such that the layout is immune to small box size changes)

Image

src/inputs/SelectCard.stories.tsx
stephenh: Oh right :thinking: yeah that'd be nice if we can get it triggered that way

blambillotte (approved):

PR #1290, opened 2026-06-25
fix: Remove smooth scrolling behavior that doesn't place nicely with virtuoso

bdow:


abadami (approved):
Interesting, I wonder why the smooth scrolling doesn't play nicely...

PR #1293, opened 2026-06-26
feat: add showProdWarning to EnvironmentBanner

chr1sjf0x:
## Description

Adds an optional `showProdWarning?: boolean` prop to `EnvironmentBanner` so consuming apps can opt a **non-impersonating** production session into the environment banner (e.g. a flag-gated developer warning).

- `shouldShowEnvironmentBanner` now returns `true` for `prod` when impersonating **or** when `showProdWarning` is set.
- Prod copy: impersonation message still wins; otherwise shows "You are in the Production Environment" with a matching tooltip.
- The developer prod warning uses the red `EnvBrandLocalProd` fill to convey the same "live prod data" urgency as local-prod; the impersonation banner keeps its existing `EnvBrandProd` (turquoise) styling.
- Added tests + a `ProdWarning` story.

Consumed by internal-frontend HACK-24 (prod warning banner) once published.

## Screenshots

`ProdWarning` story added to `EnvironmentBanner.stories.tsx` — red banner with PROD badge and the production warning message.

### DIFF
--- a/src/components/EnvironmentBanner/EnvironmentBanner.test.tsx
+++ b/src/components/EnvironmentBanner/EnvironmentBanner.test.tsx
@@ -77,24 +77,25 @@ describe("EnvironmentBanner", () => {
     expect(r.query.environmentBanner).toBeNull();
   });
 
-  it("renders the prod warning banner for developers when showProdWarning is set", async () => {
-    // Given a prod environment without impersonation but with the developer warning opted in
+  it("renders the prod warning banner when showProdWarning is set", async () => {
+    // Given a prod environment without impersonation but with the prod warning set
     const r = await render(<EnvironmentBanner env="prod" showProdWarning />);
 
-    // Then the prod banner renders the developer warning copy
+    // Then the prod banner renders the warning copy with the red local-prod fill
     expect(r.environmentBanner_badge).toHaveTextContent("PROD");
     expect(r.environmentBanner_message).toHaveTextContent("You are in the Production Environment");
     expect(r.query.environmentBanner_impersonating).toBeNull();
-    // Red, matching local-prod urgency rather than the turquoise impersonation fill.
     expect(r.environmentBanner).toHaveStyle({ backgroundColor: Tokens.EnvBrandLocalProd });
   });
 
-  it("prefers impersonation copy over the prod warning when both apply", async () => {
+  it("lets the prod warning trump impersonation when both apply", async () => {
     // Given prod while impersonating and with showProdWarning set
     const r = await render(<EnvironmentBanner env="prod" impersonating={{ name: "Andrea Eppy" }} showProdWarning />);
 
-    // Then the impersonation message wins
-    expect(r.environmentBanner_message).toHaveTextContent("You are impersonating Andrea Eppy");
+    // Then the prod warning copy and red fill win, with impersonation still shown on the right (like local-prod)
+    expect(r.environmentBanner_message).toHaveTextContent("You are in the Production Environment");
+    expect(r.environmentBanner).toHaveStyle({ backgroundColor: Tokens.EnvBrandLocalProd });
+    expect(r.environmentBanner_impersonating).toHaveTextContent("Impersonating Andrea Eppy");
     expect(r.environmentBanner_impersonatingIcon).toBeInTheDocument();
   });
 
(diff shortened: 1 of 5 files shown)


src/components/EnvironmentBanner/EnvironmentBanner.test.tsx
```diff
@@ -77,6 +77,27 @@ describe("EnvironmentBanner", () => {
     expect(r.query.environmentBanner).toBeNull();
   });
 
+  it("renders the prod warning banner for developers when showProdWarning is set", async () => {
```
bdow: **nit:** There's no concept of a "developer" in the code here, so the test name is a bit misleading. Basically, its just `renders the prod warning banner when showProdWarning is set`

src/components/EnvironmentBanner/EnvironmentBanner.test.tsx
```diff
@@ -77,6 +77,27 @@ describe("EnvironmentBanner", () => {
...
+    // Given a prod environment without impersonation but with the developer warning opted in
+    const r = await render(<EnvironmentBanner env="prod" showProdWarning />);
+
+    // Then the prod banner renders the developer warning copy
+    expect(r.environmentBanner_badge).toHaveTextContent("PROD");
+    expect(r.environmentBanner_message).toHaveTextContent("You are in the Production Environment");
+    expect(r.query.environmentBanner_impersonating).toBeNull();
+    // Red, matching local-prod urgency rather than the turquoise impersonation fill.
```
bdow: **nit:** This comment would be better suited in the code than in the test.

src/components/EnvironmentBanner/EnvironmentBanner.test.tsx
```diff
@@ -77,6 +77,27 @@ describe("EnvironmentBanner", () => {
...
+    expect(r.environmentBanner).toHaveStyle({ backgroundColor: Tokens.EnvBrandLocalProd });
+  });
+
+  it("prefers impersonation copy over the prod warning when both apply", async () => {
+    // Given prod while impersonating and with showProdWarning set
+    const r = await render(<EnvironmentBanner env="prod" impersonating={{ name: "Andrea Eppy" }} showProdWarning />);
+
+    // Then the impersonation message wins
+    expect(r.environmentBanner_message).toHaveTextContent("You are impersonating Andrea Eppy");
+    expect(r.environmentBanner_impersonatingIcon).toBeInTheDocument();
```
bdow: What color is it in this instance? I would assume if we're doing the "showProdWarning", then it'd act more like `local-prod`, where we display both the "You're in the prod env" and the copy on the right of "You're impersonating...".
I wouldn't fall back to "impersonation only" style if `showProdWarning` is true.

src/components/EnvironmentBanner/EnvironmentBanner.tsx
```diff
@@ -13,14 +13,16 @@ export type ImpersonatedUser = { name: string };
...
   env: AppEnvironment;
   impersonating?: ImpersonatedUser;
+  /** When in prod without impersonation, show a developer "you are in production" warning banner. */
+  showProdWarning?: boolean;
 };
 
 /** Environment banner; horizontal/vertical pinning is owned by {@link EnvironmentBannerLayout}. */
 export function EnvironmentBanner(props: EnvironmentBannerProps) {
-  const { env, impersonating, ...others } = props;
+  const { env, impersonating, showProdWarning, ...others } = props;
```
bdow: suggestion: Would prefer to default this value to `showProdWarning: false`. And then in the `shouldShowEnvironmentBanner`, expect that the value is passed in.

src/components/EnvironmentBanner/EnvironmentBanner.tsx
```diff
@@ -70,9 +72,21 @@ export function EnvironmentBanner(props: EnvironmentBannerProps) {
   );
 }
 
-/** True when {@link EnvironmentBanner} should display: `dev`, `qa`, `local-prod`, or `prod` while impersonating. */
-export function shouldShowEnvironmentBanner(env: AppEnvironment, impersonating: ImpersonatedUser | undefined): boolean {
-  return env === "dev" || env === "qa" || env === "local-prod" || (env === "prod" && isDefined(impersonating));
+/**
+ * True when {@link EnvironmentBanner} should display: `dev`, `qa`, `local-prod`, or `prod` while impersonating
+ * or when `showProdWarning` opts a developer session into the prod warning banner.
```
bdow: **nit:** I don't love these comments. I think it makes up this idea of opting into a developer session. No one opts into a developer session. It'd make more sense to say something along the lines of: `showProdWarning is typically set to true only for developers in the consuming apps`.

src/components/EnvironmentBanner/EnvironmentBanner.tsx
```diff
@@ -101,14 +115,20 @@ function getEnvironmentBannerConfig(
...
       };
-    // Prod env banner should only be displayed when impersonating
+    // Prod env banner displays when impersonating or when a developer opts in via `showProdWarning`;
+    // impersonation copy takes priority since it is the more specific warning.
     case "prod":
       return {
-        bgColor: Tokens.EnvBrandProd,
+        // Developer prod warning reuses the red local-prod fill to convey the same "live prod data" urgency;
+        // impersonation keeps its own prod fill.
+        bgColor: isDefined(impersonating) ? Tokens.EnvBrandProd : Tokens.EnvBrandLocalProd,
```
bdow: I think if we're impersonating and we have showProdWarning, then we should stick with the red color. Feels like `showProdWarning` styles should trump others, imo. What do you think?

chr1sjf0x: Agreed. TY

chr1sjf0x:
Thanks for the review — addressed all of it in be7b0a73:

- **`showProdWarning` now trumps impersonation** (per the thread on the bgColor): when set, the banner uses the red `EnvBrandLocalProd` fill + "You are in the Production Environment" copy, and impersonation (if any) now shows on the right rail like local-prod — no more falling back to impersonation-only styling.
- **Default `showProdWarning: false`** in the component; `shouldShowEnvironmentBanner` now takes it as a required arg. Threaded through `EnvironmentBannerLayout` and `docs/layouts.md`.
- **Reworded** the prop/function comments to drop the "developer session" framing; moved the color rationale into the component.
- Renamed the test to `renders the prod warning banner when showProdWarning is set` and rewrote the both-apply test to assert the warning trumps (red + warning copy + impersonation on the right).

bdow (approved):

PR #1296, opened 2026-06-26
fix: add unlisted react-aria packages as dependencies

Jacob-Diebold:
## Summary

Beam's build externalizes `@react-aria/*` imports in `dist/index.js` (e.g. `@react-aria/textfield`, `@react-aria/interactions`), but those packages were not declared in Beam's `dependencies`. Consumers only get them today via fragile transitive hoisting.

This PR declares the runtime packages Beam's bundle actually imports:

- `@react-aria/interactions`
- `@react-aria/textfield`
- `@react-aria/numberfield`

(`@react-aria/utils` was already declared.)


bdow (commented):
This is fine to do. I would be curious if we could've just imported directly from `react-aria`, which I think bundles these already - maybe?
At least looking at this, we could've changed:
```
import type { NumberFieldAria } from "@react-aria/numberfield";
```
to 
```
import type { NumberFieldAria } from "react-aria";
```
And 
```
import { useTextField } from "@react-aria/textfield";
```
to 
```
import { useTextField } from "react-aria";
```

I think the root of this might be that we're importing incorrectly to begin with.

Jacob-Diebold:
> This is fine to do. I would be curious if we could've just imported directly from `react-aria`, which I think bundles these already - maybe? At least looking at this, we could've changed:
> 
> ```
> import type { NumberFieldAria } from "@react-aria/numberfield";
> ```
> 
> to
> 
> ```
> import type { NumberFieldAria } from "react-aria";
> ```
> 
> And
> 
> ```
> import { useTextField } from "@react-aria/textfield";
> ```
> 
> to
> 
> ```
> import { useTextField } from "react-aria";
> ```
> 
> I think the root of this might be that we're importing incorrectly to begin with.

I'll give it a shot to see if it works

bdow (approved):

stephenh (approved):

Jacob-Diebold:
@bdow you were right! react-aria exports `useTextField` & `NumberFieldAria`. It doesn't include `getInteractionModality` so we still need `@react-aria/interactions` added; but that still cuts down on the packages. Thanks!

PR #1298, opened 2026-06-26
feat: add BeamOverlays for modal and drawer app context

RoseAndresHB:
## Summary
- Adds `BeamOverlays`, an opt-in mount point that apps can place inside `BeamProvider` below app-wide context providers so `openModal` / `openInDrawer` content can use those hooks.
- `BeamProvider` keeps backwards-compatible fallback overlay rendering when `BeamOverlays` is not mounted.
- Documents setup in `docs/overlays.md` and ships it in the npm package.

## Test plan
- [x] `npx vitest run src/components/BeamOverlays.test.tsx`
- [x] `npx vitest run src/components/SuperDrawer/SuperDrawer.test.tsx`
- [x] `npx vitest run src/components/Modal/useModal.test.tsx`
- [ ] Blueprint follow-up: mount `<BeamOverlays />` in `AppProviders` after Beam release

### DIFF
--- a/docs/overlays.md
+++ b/docs/overlays.md
@@ -1,0 +1,51 @@
+# Beam global overlays (Modal & SuperDrawer)
+
+This document is the **canonical contract** for mounting Beam's global overlays so modal and super-drawer content can use app React context. It ships with the `@homebound/beam` package for consuming apps.
+
+## Rules (normative)
+
+1. **Wrap the app in `BeamProvider`** — Required for `useModal`, `useSuperDrawer`, snackbars, and other Beam globals.
+2. **Mount `<BeamOverlays />` once per `BeamProvider`** — Place it **inside** `BeamProvider`, **below** app-wide providers whose context overlay content should read.
+3. **Do not mount multiple `BeamOverlays`** — A second instance logs a dev warning and can render duplicate modals.
+4. **Route-scoped context still requires props** — Providers mounted only on a page or route do not wrap `BeamOverlays` unless you restructure the tree. Pass that data into `openModal` / `openInDrawer` content explicitly.
+
+## Why this exists
+
+`useModal` and `useSuperDrawer` render a **single global** Modal and SuperDrawer. Without `BeamOverlays`, Beam mounts them as **React siblings** of your app `children`. Content opened via `openModal({ content })` is then **outside** providers nested under `children`, so app context hooks in overlay content see default values.
+
+This is a **React tree** issue, not a DOM portal issue. Portals change where nodes paint in the DOM; they do not change which providers wrap overlay content in React.
+
+## Setup
+
+```tsx
+import { BeamOverlays, BeamProvider } from "@homebound/beam";
+
+<BeamProvider>
+  <AppWideProviders>
+    {children}
+    <BeamOverlays />
+  </AppWideProviders>
+</BeamProvider>;
+```
+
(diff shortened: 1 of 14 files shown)


bdow (commented):
disclaimer: I haven't really read through the whole thing, but I have thought about this in the past a little bit. My first thought was to try and utilize portals.
1. `openModal` would take the same props, just call a `createPortal` and point to an overlay container (where ever that is).
2. We'd need to keep track of modals being opened, and enforce another closed.

And not really sure how else it'd work but that was my first thougth/idea. ANyways, gotta split. Will read through this later

docs/overlays.md
```diff
@@ -0,0 +1,51 @@
...
+`useModal` and `useSuperDrawer` render a **single global** Modal and SuperDrawer. Without `BeamOverlays`, Beam mounts them as **React siblings** of your app `children`. Content opened via `openModal({ content })` is then **outside** providers nested under `children`, so app context hooks in overlay content see default values.
+
+This is a **React tree** issue, not a DOM portal issue. Portals change where nodes paint in the DOM; they do not change which providers wrap overlay content in React.
+
+## Setup
+
+```tsx
+import { BeamOverlays, BeamProvider } from "@homebound/beam";
+
+<BeamProvider>
```
stephenh, line 23: @RoseAndresHB cc @bdow not against this approach, but I'd like this doc to outline why `BeamProvider` must be outside the `AppWideProviders`.

I.e. I naively thought just putting `BeamProvider` itself/as-is as a child of `AppWideProviders` would fix our context issues.

src/components/BeamOverlaysContent.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { SuperDrawer } from "src/components/SuperDrawer/SuperDrawer";
+
+/** Renders the global Modal and SuperDrawer; used by {@link BeamOverlays} and BeamProvider fallback. */
+export function BeamOverlaysContent() {
+  const { modalState } = useBeamContext();
+  return (
+    <>
+      {modalState.current && <Modal {...modalState.current} />}
+      <SuperDrawer />
+    </>
```
bdow, line 12: **question:** We are moving where it renders to have a _more_ context, but still not the full available context as the caller. Is that right?
For example, in Takeoffs we have a TakeoffManagerContext(? i think thats the name...), and we're in a nested set of Routes, so there's two additional contexts that we wouldn't have access to. I could be wrong, but my understanding here is that `BeamOverlays` still wouldn't have full access to that context, right?
In that sense, I wonder if its really worth doing something like this if it only gets us halfway there.

RoseAndresHB: That is correct, we will not solve the issue of contexts that are added at the route-specific level, i.e. providers nested deeper within the tree than the `<BeamOverlays>` component.

src/components/BeamContext.tsx
```diff
@@ -88,23 +100,45 @@ export function BeamProvider({ children, documentTitleConfig, ...presentationPro
...
+    <OverlayHostContext.Provider value={overlayHostContext}>
+      <PresentationProvider {...presentationProps}>
+        <RightPaneProvider>
+          <AutoSaveStatusProvider>
+            <SnackbarProvider>
+              {/* OverlayProvider is required for Modals generated via React-Aria */}
+              <ToastProvider>
+                <OverlayProvider>
+                  {children}
+                  {!hasOverlayHost && <BeamOverlaysContent />}
```
RoseAndresHB:
```suggestion
                  {!hasOverlayHost && <BeamOverlaysContent /> // fallback for Overlays when opted-out of <BeamOverlays>}
```

PR #1299, opened 2026-06-29
feat: GridTable empty state

bdow:


src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -530,6 +530,101 @@ describe("GridTableLayout", () => {
...
+
+      await typeAndWait(r.search, "zzz-no-match");
+      expect(r.clearFilters).toBeInTheDocument();
+
+      // When the clear-filters action is clicked
+      click(r.clearFilters);
+      expect(r.gridTable).toBeInTheDocument();
+
+      // Then the table shows the data rows again
+      expect(tableSnapshot(r)).toMatchInlineSnapshot(`
```
abadami, line 637: **thought (non-blocking)**: This is a cool way to check data rows. I wonder if it'd be worth making a util function someday to build the snapshot string for us based on provided rows and columns. Could perhaps make dealing with tests that need a lot of data easier? Probably pretty niche at the moment though.

src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -295,6 +319,41 @@ export function useGridTableLayoutState<F extends Record<string, unknown>>({
...
+function composeEmptyState<F extends Record<string, unknown>, R extends Kinded, X extends Only<GridTableXss, X>, QData>(
+  tableProps: GridTablePropsWithRows<R, X> | QueryTablePropsWithQuery<R, X, QData>,
+  layoutState: ReturnType<typeof useGridTableLayoutState<F>> | undefined,
+  layoutEmptyFallback: string | undefined,
+): GridTableEmptyStateProps | undefined {
+  const tableEmptyState = "emptyState" in tableProps ? tableProps.emptyState : undefined;
+  const tableEmptyFallback = "emptyFallback" in tableProps ? tableProps.emptyFallback : undefined;
+  const baseTitle = tableEmptyState?.title ?? tableEmptyFallback ?? layoutEmptyFallback;
+
+  if (!baseTitle && !tableEmptyState) return undefined;
```
abadami: **question**: Is there ever a situation where we'd want to show nothing if the table is empty, even if a table empty state / title isn't provided?

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -197,6 +197,42 @@ export function WithoutHeader() {
...
+    const layoutState = useGridTableLayoutState({
+      persistedFilter: {
+        filterDefs,
+        storageKey: "grid-table-layout-empty-state",
+      },
+      search: "client",
+    });
+
+    useEffect(() => {
+      layoutState.setSearchString("no-match");
```
abadami, line 235: **question**: Just a sanity check, but does setting the search string here with the layout state as a dependency not trigger this effect infinitely? 

I imagine it might not if the layout state doesn't update when the search string is matching.

src/components/Layout/GridTableLayout/QueryTable.tsx
```diff
@@ -37,8 +46,14 @@ export function QueryTable<R extends Kinded, QData, X extends Only<GridTableXss,
...
 
-  // loading/error/empty can all use the one fallback prop.
-  const fallbackMessage = query.loading ? "Loading…" : query.error ? `Error: ${query.error.message}` : emptyFallback;
+  const fallbackMessage = query.loading ? "Loading…" : query.error ? `Error: ${query.error.message}` : undefined;
+
+  const emptyState: GridTableEmptyStateProps | undefined = useMemo(() => {
+    if (query.loading || query.error) return undefined;
+    if (emptyStateProp) return emptyStateProp;
+    if (emptyFallback) return { title: emptyFallback };
+    return undefined;
```
abadami: **question**: Similar to the other comment about having no empty state, but is there merit to having a default value here?

src/hooks/useBodyBackgroundColor.tsx
```diff
@@ -0,0 +1,20 @@
+import { useLayoutEffect } from "react";
+import { BeamColor } from "src/colors";
+import { maybeCssVar } from "src/Css";
+
+/** Sets `document.body` background color and restores the prior inline value on unmount. */
+export function useBodyBackgroundColor(color: BeamColor): void {
```
abadami, line 6: **question**: I'm curious as to why you'd want to dynamically change the body background color like this? To me it seems simpler to just have the appropriate child containers (like the child of whatever is using document scroll in this case) dictate their backgrounds rather than changing the entire document's body. I suppose this does make it less work to make sure all document scrolls have the appropriate background though, so that's nice ergonomically 😄.

bdow: internal-frontend denotes a gray background color across the app. Design is moving towards a white background. This allows us to set the full background. Given this empty state only takes up ~100-200px of height, then we can't just set a background color there. Otherwise, we'll still see the gray background elsewhere.

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
bdow: Yea, this was basically done to just initially render the empty state. Will fix to make the story more usable.

src/components/Layout/GridTableLayout/GridTableLayout.tsx
bdow: Great point. Does seem like we should have a default empty state provided. Will update

src/components/Layout/GridTableLayout/QueryTable.tsx
bdow: Thanks. Updated to always have a default empty state now.

abadami (approved):

PR #1300, opened 2026-06-29
feat: Improve Table Actions Styling

abadami:
### Improve Table Actions Styling

This is accomplished by creating a new `GridTableLayoutActions` component, which will allow us to more fine tune the experience when the search / filters are open. `FilterDropdownMenu` might no longer be needed due to this as it is only used for the Grid Table Layout.


src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,292 @@
...
+      )}
+    </div>
+  );
+}
+
+// memo doesn't support generic parameters, so cast the result to the correct type
+const _GridTableLayoutActions = memo(GridTableLayoutActionsComponent) as typeof GridTableLayoutActionsComponent;
+export { _GridTableLayoutActions as GridTableLayoutActions };
+
+// --- FilterPanel ---
```
stephenh: Once we hit `// --- worth a comment` we might as well break out `FilterPanel` into a separate file, even if it's not exported / only used by `GridTableLayout`?

I'm generally +1 on deleting the other filter impl, the only caveat is that we have 5+ filter impls over the years (basically 1 per designer 🫤 ) and if we ever have a UX that needs "the new filters ... in a slightly different place", I'd rather use "the new filters" than have a 6th filter impl 😅 

But hopefully we'd catch that sort of thing at Figma review time anyway. 🤞

src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -142,29 +141,23 @@ function GridTableLayoutComponent<
   const visibleColumnsStorageKey = layoutState?.persistedColumnsStorageKey;
 
   const tableActionsEl = (
-    <TableActions
+    <GridTableLayoutActions
+      filterDefs={layoutState?.filterDefs}
+      filter={layoutState?.filter}
+      onChange={layoutState?.setFilter}
```
bdow: **suggestion:** Maybe keep calling this `setFilter` vs `onChange`. `onChange` is a bit ambiguous to what change we're paying attention to here.

src/components/Layout/GridTableLayout/GridTableLayout.tsx
```diff
@@ -142,29 +141,23 @@ function GridTableLayoutComponent<
...
+      groupBy={layoutState?.groupBy}
+      searchProps={layoutState?.search ? { onSearch: layoutState.setSearchString } : undefined}
       right={
-        (hasHideableColumns || withCardView) && (
+        hasHideableColumns || withCardView ? (
```
bdow: **thought:** Now that we're being a bit more bespoke and strict for "this component is only for the Edit Columns Button and ViewToggleButton" if we can just pass the props we need through and have `GridTableLayoutActions` be more in control of rendering these. This might give us a bit better of a re-rendering as we are never rebuilding the whole `right` side when the view toggle changes, but hopefully the internal React rebuilding can notice that the only thing that really changed was the ViewToggleButton's state, or visibility of the EditColumnsButton.
My rationale is that this component is now specific for this case only. Before it was "whatever you want to put on the right side". But now we're enforcing specific UI elements, so I think we can make a bit more of a strict API. Wdyt?

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,292 @@
...
+      onChange={(v) => setSearchValue(v ?? "")}
+      placeholder="Search"
+      clearable
+      fullWidth
+      startAdornment={<Icon icon="search" color={Palette.Gray700} />}
+    />
+  );
+
+  return (
+    <div css={Css.df.fdc.gap1.pb1.$}>
```
bdow: **suggestion (non-blocking):** I see this has a `padding-bottom: 8px` - What do you think about have the `GridTableLayout` doing a `gap1` between the Actions and the table? 
I usually lean towards having the spacing between components being handled by the parent component that is composing them together. 
I think this one is fine here, but just a mentality for the future.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,292 @@
...
+      <div css={Css.df.gap1.aic.jcsb.aifs.pt3.if(inDocumentScrollLayout).pl3.pr3.$}>
```
bdow: **nit:** `pl3.pr3` can be `px3`.

Another thought - not something we need to do here, but would be nice to create a spacing variable for our horizontal padding i the different viewports. Like, I think in mobile we might only want a `px1` here? At least save on some horizontal spacing. Maybe we can create a follow up ticket for "Create horizontal layout spacing css variable for consistent padding across layout components in DocumentScroll layouts".

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,292 @@
...
+        <div css={Css.df.gap1.aic.fww.$}>
```
bdow: **question:** Do we really want this to wrap (`fww`)? I assumed these elements would always stay on one line.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+          {/* Small screen: filter icon toggle */}
+          {sm && hasFilters && (
+            <IconButton
+              variant="outline"
+              icon="filter"
+              label="Filter"
+              onClick={() => setShowFilters(!showFilters)}
+              active={showFilters}
+              {...testId.button}
```
bdow: **suggestion:** Might be worth having a more specific testid name here.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,292 @@
...
+            />
+          )}
+        </div>
+        {right}
+      </div>
+
+      {/* Search row — spans full width below TableActions (including under right-side buttons) */}
+      {sm && showSearch && hasSearch && <div css={Css.pl3.pr3.$}>{searchTextField}</div>}
```
bdow: **question (non-blocking):** Would it be possible for `showSearch` to be true if `hasSearch` is false? I'm wondering if this conditional can be trimmed down to just `showSearch`? This is fine, but just a thought.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
bdow: +1 on this. Prefer more component files with their own dedicated test files and stories where applicable.

src/components/Table/components/EditColumnsButton.tsx
```diff
@@ -64,16 +64,14 @@ export function EditColumnsButton<R extends Kinded>(props: EditColumnsButtonProp
...
-        icon: "kanban",
-        size: "md",
-        label: "",
-        variant: "secondaryBlack",
-      }}
+      trigger={
+        sm
+          ? { icon: "kanban", variant: "outline" }
+          : { icon: "kanban", label: "", size: "md", variant: "secondaryBlack" }
+      }
```
bdow, line 76: **question/thought (non-blocking):** I think the result of these are going to be super similar, except for the down caret icon on large screens. I wonder how meaningful that extra icon really is. I'd maybe ask Design do you think we really need to have that down caret icon on desktop?
One nice thing about designing for mobile is that it forces recognize and reduce the amount extra UI elements that might not be providing the biggest benefit. If we can deem something isn't necessary for mobile, then it's always good to question if its also really necessary for desktop too.

src/components/Table/components/ViewToggleButton.tsx
```diff
@@ -19,11 +19,12 @@ export function ViewToggleButton({ view, onChange, defaultOpen }: ViewToggleButt
     { label: "Card", icon: "tile", onClick: () => onChange("card") },
   ];
 
+  const icon = view === "list" ? "projects" : "tile";
+
   return (
     <ButtonMenu
-      trigger={{ icon: view === "list" ? "projects" : "tile", label: "", size: "md", variant: "secondaryBlack" }}
+      trigger={sm ? { icon, variant: "outline" } : { icon, label: "", size: "md", variant: "secondaryBlack" }}
```
bdow, line 26: **nit (non-blocking):** This is an existing problem, but maybe something we can try to fix later on. We have to set `label: ""` to get the right button we want here. But that's kind of an antipattern. I think we should _always_ pass a label, but just have an `iconOnly` option here. But again, then why can't we just use the icon button per the last comment. 
but the problem ends up being we recognize which trigger component to render based on the existence of the `label` property. This is something I'd like us to create a ticket for an fix in the future.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
abadami: Oh yeah, very fair point. Absolutely will do this. I'll also go ahead and drop `FilterDropdownMenu` with Stephen's comment in mind.

src/components/Layout/GridTableLayout/GridTableLayout.tsx
abadami: Yeah, I'd say we would want to make the API strict on this for sure. I don't really foresee any scenario where this actions layout would be worth abstracting to be more generic with a custom right slot at this point so I think it'd be good to keep it strict.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
abadami: Ooo, I like that css variable for consistent padding in layouts. I'll make a ticket for that in that Core Layout/Component improvements project.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
abadami: Yeah that's a good mentality, and seeing that throughout Beam's code base I do agree it's better.

I think in this case this particular element is being wrapped by a `display: contents` div. I don't want to mess with the entire layout here so I'd say we leave this one as an outlier for now.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
abadami: Yeah that's a good point. Besides, I don't think there is a scenario where they could wrap to another line.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
abadami: I believe you're right! Good to trim it down.

src/components/Table/components/EditColumnsButton.tsx
abadami: Yeah, I'm down with bringing it up with design, but will likely want to save that for a different PR

src/components/Table/components/ViewToggleButton.tsx
abadami: Yeah, I think this also leads to a greater discussion regarding buttons and "do we even need `IconButton`, or should it be wrapped into the `Button` component" because to me, I could see IconButton just becoming an "IconOnly" prop on Button at some point. That being said, I think that's worthy of a discussion later on and we can start with updating this API here. I'll create a ticket in the core improvement project for now.

src/components/Filters/FilterDropdownMenu.test.tsx
bdow, line 1: 🎉 !!

src/components/Layout/GridTableLayout/FilterPanel.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+const groupByOptions: GroupOption[] = [
+  { id: "none", name: "None" },
+  { id: "status", name: "Status" },
+];
+
+export default {
+  component: FilterPanel,
+  decorators: [withBeamDecorator, withRouter()],
+  parameters: { layout: "padded" },
+} satisfies Meta;
```
bdow, line 13: **request:** Would prefer this at the top of the file. I think for stories, the order of content in the code would be:
```
// Export for Storybook's purposes
export default { ... }

// Exported Stories

// Local types needed for stories

// Local data functions needed for stories

File: src/components/Layout/GridTableLayout/FilterPanel.stories.tsx [hunk 12]
```diff
@@ -0,0 +1,96 @@
...
+
+export function OpenWithControls() {
+  const [filter, setFilter] = useState<TestFilter>({});
+  return (
+    <FilterPanel
+      isOpen={true}
+      filterImpls={filterImpls}
+      filter={filter}
+      setFilter={setFilter}
+      activeFilterCount={0}
```
bdow: **question:** What do you think of getting rid of the prop `activeFilterCount` and just count the filters in the component? There's a function that'll already do that for us at `getActiveFilterCount`.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,182 @@
...
+      {/* Combined filter panel — controls when open, chips when closed */}
+      {hasFilters && (
+        <FilterPanel
+          isOpen={showFilters}
+          groupBy={groupBy}
+          filterImpls={filterImpls}
+          filter={filter}
+          setFilter={setFilter}
+          activeFilterCount={activeFilterCount}
+          onClear={() => setFilter?.({} as F)}
```
bdow, line 174: **thought/suggestion:** If all this is doing is calling `setFilter` to empty, then we can avoid this prop all together as we're already passing in `setFilter`. 
Though, I wonder if we should also be clearing out the `search` text as well when the user clicks clear 🤔 Might be worth asking design - I would tend to yes. And in that vein, I'd probably keep this `onClear` and just have it reset both the search & filters.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,182 @@
...
+            {hasHideableColumns && view === "list" && columns && api && (
+              <EditColumnsButton columns={columns} api={api} tooltip="Display columns" />
+            )}
+            {withCardView && view !== undefined && setView && <ViewToggleButton view={view} onChange={setView} />}
+          </div>
+        )}
+      </div>
+
+      {/* Search row — spans full width below TableActions (including under right-side buttons) */}
+      {sm && showSearch && <div css={Css.pl3.pr3.$}>{searchTextField}</div>}
```
bdow: **nit:** super minor, but could do `px(3)` here.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,182 @@
...
+      placeholder="Search"
+      clearable
+      fullWidth
+      startAdornment={<Icon icon="search" color={Palette.Gray700} />}
+    />
+  );
+
+  return (
+    <div css={Css.df.fdc.gap1.pb1.$}>
+      <div css={Css.df.gap1.aic.jcsb.aifs.pt3.if(inDocumentScrollLayout).px3.$}>
```
bdow: **request:** This looks like we have conflicting `align-items` definitions with `.aic.` and `.aifs.`. I'm guessing that `aifs` is winning out - but I woul initially think we'd want `aic`. Granted, probably all of these elements in the row are of equal height by design so whatever this value is set at probably doesn't matter - could be omitted and probably fine.

src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -0,0 +1,149 @@
...
+            onSelect={(g) => g && groupBy.setValue(g)}
+          />
+        )}
+        {filterControls}
+        {activeFilterCount > 0 && <Button label="Clear" variant="tertiary" onClick={onClear} {...testId.clearBtn} />}
+      </div>
+    );
+  }
+
+  if (!filter || !setFilter) return null;
```
bdow, line 86: **suggestion:** Can we enforce that both `filter` and `setFilter` are required props for this?

bdow: Looking over this - appears we're doing this to support a `groupBy` element on its own... that isn't something that we've ever done before, but I don't see why we wouldn't introduce that... 🤔

src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -0,0 +1,149 @@
...
+    options: Array<{ id: G; name: string }>;
+  };
+  filterImpls: FilterImpls<F>;
+  filter?: F;
+  setFilter?: (filter: F) => void;
+  activeFilterCount: number;
+  onClear: () => void;
+};
+
+export function FilterPanel<F extends Record<string, unknown>, G extends Value = string>({
```
bdow: **thought** The split early-return structure in here was hard for me to follow — it reads like two components in one function, especially with the IIFE for filterControls. What do you think of moving the open/closed views into their own smaller components? FilterDropdownMenu did this with FilterChips extracted. We could abstract both.

src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -0,0 +1,149 @@
...
+          })()
+        : null;
+
+    return (
+      <div
+        style={{ scrollbarWidth: "none" }}
+        css={
+          sm
+            ? Css.df.gap1.aic.oxa.mw0.if(inDocumentScrollLayout).px3.$
+            : Css.df.fww.gap1.aic.if(inDocumentScrollLayout).px3.$
```
bdow: **thought:** I wonder if we could reduce the need for JS to provide the media query styles here. There is a `.ifSm.` in truss. I remember having problems mixing multiple 'if' blocks in truss, but worth seeing if that has been fixed.
Maybe something like:
```
Css.df.aic.gap1.ifSm.oxa.mw0.ifMdAndUp.fww.if(inDocumentScrollLayout).px3.$
```

src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  if (!filter || !setFilter) return null;
+
+  const chips = safeEntries(filterImpls).flatMap(([key, f]) => chipsForFilterKey(key, f, filter, setFilter, testId));
+
+  if (chips.length === 0) return null;
+
+  return (
+    <div css={Css.df.gap1.aic.oxa.mw0.fww.if(inDocumentScrollLayout).pl3.$}>
```
bdow: **question:** Should chips also only `overflow-x: auto` when on the smaller breakpoint? Otherwise, people without a trackpad or side scrolling mouse won't be able to see the overflow without some actual `<`/`>` buttons, I think.

bdow (changes requested):

src/components/Layout/GridTableLayout/FilterPanel.stories.tsx
abadami: Yep for sure! Sorry, it's taking me a bit to internalize this one.

src/components/Layout/GridTableLayout/FilterPanel.tsx
abadami: Actually I missed removing that. I talked to design regarding the clear button yesterday and they determined that we'd have chips wrap vertical always now so that the clear button is always discoverable.

src/components/Layout/GridTableLayout/FilterPanel.tsx
abadami: Yes absolutely I must've glossed over that being an IIFE I am really not a fan of that, yeesh.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
abadami: Yeah, looking at this again I do believe we won't need the `aic` at all. Should be able to omit it, but I'll make sure it doesn't change the look at all.

src/components/Layout/GridTableLayout/FilterPanel.tsx
abadami: Yeah for sure, let's see if I can figure that out here...hopefully the multiple if block stuff works.

src/components/Layout/GridTableLayout/FilterPanel.stories.tsx
abadami: Good idea, I'm always down to remove unnecessary props.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
abadami: Hmm, I think UX wise if we have the clear button also clear search, then there will need to be a chip representation for the search value...otherwise I feel like it isn't clear that the clear button will also clear search at the moment. 

So yeah probably worth asking design about for sure 😄

src/components/Layout/GridTableLayout/FilterPanel.tsx
abadami: Yeah, I do feel like that situation is rare, but at the same time, I think it's good to leave the option open. We can always enforce it later if it literally never happens.

I don't see too many tables that have literally 0 filters that are provided by the GridTableLayout though, so maybe we can make that assumption now...

src/components/Layout/GridTableLayout/FilterPanel.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { checkboxFilter } from "src/components/Filters";
+import { click, render, withRouter } from "src/utils/rtl";
+import { vi } from "vitest";
+import { buildFilterImpls, FilterPanel } from "./FilterPanel";
+
+type TestFilter = { needsRevision?: boolean };
+const filterDefs: FilterDefs<TestFilter> = {
+  needsRevision: checkboxFilter({ label: "Needs Revision" }),
+};
+const filterImpls = buildFilterImpls(filterDefs);
```
bdow: **request:** Can these be moved to the bottom of the file, similar set up to all other stories and components.

src/components/Layout/GridTableLayout/FilterPanel.test.tsx
bdow, line 1: Can we do the `// Given/When/Then` comments in here for consistency?

src/components/Layout/GridTableLayout/FilterPanel.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+          setFilter={setFilter}
+          onClear={() => setFilter({})}
+        />
+      );
+    }
+    const r = await render(<Wrapper />, withRouter());
+
+    expect(r.filter_chip_needsRevision).toBeInTheDocument();
+    click(r.filter_chip_needsRevision);
+    expect(r.query.filter_chip_needsRevision).not.toBeInTheDocument();
```
bdow: **question/thought:** This test seems like its validating behavior of the Chip component rather than this FilterPanel. Maybe we remove this and can put it elsewhere if this isn't tested yet?
My understanding is clicking a chip will call the setFilter to remove itself from the list of filters, and doesn't actually even touch this `useState` filter defined in the Wrapper component.

src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -0,0 +1,164 @@
...
+  onClear,
+}: Omit<FilterPanelProps<F, G>, "isOpen">) {
+  const tid = useTestIds({}, filterTestIdPrefix);
+  const inDocumentScrollLayout = useDocumentScrollLayout();
+  const activeFilterCount = getActiveFilterCount(filter ?? {});
+  const filterControls = filter && setFilter ? buildFilterControls(filterImpls, filter, setFilter, tid) : null;
+
+  return (
+    <div
+      style={{ scrollbarWidth: "none" }}
```
bdow: **suggestion:** We can update truss and use `Css.sbwn.$`

src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -0,0 +1,164 @@
...
+  filter?: F;
+  setFilter?: (filter: F) => void;
+  onClear: () => void;
+};
+
+export function FilterPanel<F extends Record<string, unknown>, G extends Value = string>(
+  props: FilterPanelProps<F, G>,
+) {
+  if (props.isOpen) return <FilterPanelOpen {...props} />;
+  return <FilterPanelClosed {...props} />;
```
bdow: **nit:** with a single conditional, this could be nicer to write as one return:

```suggestion
  return isOpen ? <FilterPanelOpen {...props} /> : <FilterPanelClosed {...props} />;
```

src/components/Layout/GridTableLayout/GridTableLayoutActions.test.tsx
bdow, line 1: **nit/request;** Can we use the `// Given / When / Then` comments here too?

src/components/Layout/GridTableLayout/GridTableLayoutActions.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+      const onSearch = vi.fn();
+      await render(<GridTableLayoutActions searchProps={{ onSearch }} />, withRouter());
+      expect(onSearch).toHaveBeenCalledWith("");
+    });
+
+    it("calls onSearch after typing a value", async () => {
+      const onSearch = vi.fn();
+      const r = await render(<GridTableLayoutActions searchProps={{ onSearch }} />, withRouter());
+      change(r.search, "alpha");
+      act(() => vi.advanceTimersByTime(300));
```
bdow: **nit:** I think we have a `typeAndWait` helper that can do this for us: 
```
await typeAndWait(r.filter, "bar");
```

abadami: Ah yeah I saw that in one of your latest PRs. Will switch.

src/components/Layout/GridTableLayout/FilterPanel.test.tsx
abadami: Hmm, that's a good point. I think I'll move this over to the GridTableLayout test file then. I should be able to test out the layout state setting the filters correctly with that. We'll see.

src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -0,0 +1,163 @@
...
+  onClear,
+}: Omit<FilterPanelProps<F, G>, "isOpen">) {
+  const tid = useTestIds({}, filterTestIdPrefix);
+  const inDocumentScrollLayout = useDocumentScrollLayout();
+  const activeFilterCount = getActiveFilterCount(filter ?? {});
+  const filterControls = filter && setFilter ? buildFilterControls(filterImpls, filter, setFilter, tid) : null;
+
+  return (
+    <div
+      style={{ scrollbarWidth: "none" }}
```
bdow: **request:** Can we remove this line now?

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -530,6 +530,41 @@ describe("GridTableLayout", () => {
...
+
+      function FilterChipWrapper() {
+        const layoutState = useGridTableLayoutState<ChipFilter>({
+          persistedFilter: { filterDefs: chipFilterDefs, storageKey: "chip-click-test" },
+        });
+        capturedFilter = layoutState.filter;
+        return (
+          <GridTableLayoutComponent
+            layoutState={layoutState}
+            hideEditColumns
```
bdow: **nit:** I don't think setting `hideEditColumns` is necessary for this test.

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -530,6 +530,41 @@ describe("GridTableLayout", () => {
     }
   });
 
+  describe("filters", () => {
+    it("removes the filter value when a chip is clicked", async () => {
+      // Given the panel is closed with an active filter
+      sessionStorage.setItem("chip-click-test", JSON.stringify({ needsRevision: true }));
```
bdow: **nit (non-blocking):** Personal preference would've been to use query params via the `withRouter` call. Would've been able to validate changes to the filters via changes to query params as well. But this works too, doesn't hurt to test this way. 
Maybe worth making a `const storageKey = "chip-click-test"` just to ensure they stay in sync?

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -578,6 +613,10 @@ function TestWrapper(props: TestWrapperProps) {
   return <GridTableLayoutComponent {...props} layoutState={layoutState} />;
 }
 
+const chipFilterDefs = {
+  needsRevision: checkboxFilter({ label: "Needs Revision" }),
+};
```
bdow: **request:** Can this just be inlined? Appears to only be used once.

src/components/Layout/GridTableLayout/GridTableLayoutActions.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+      expect(r.query.search).not.toBeInTheDocument();
+    });
+
+    it("calls onSearch on mount with the initial search value", async () => {
+      // Given searchProps is provided
+      const onSearch = vi.fn();
+      // When the component mounts
+      await render(<GridTableLayoutActions searchProps={{ onSearch }} />, withRouter());
+      // Then onSearch is called with an empty string
+      expect(onSearch).toHaveBeenCalledWith("");
```
bdow: **question:** This test kind of sticks out as maybe an accidental re-render on load? Why do we expect `onSearch` to initially be fired - what are we doing that value that we can't already do on initial render?

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const [{ search: initialValue }, setQueryParams] = useQueryParams({ search: StringParam });
+  const [searchValue, setSearchValue] = useState<string>(initialValue || "");
+  const [debouncedSearch] = useDebounce(searchValue, 300);
+
+  useEffect(() => {
+    if (searchProps) {
+      searchProps.onSearch(debouncedSearch);
+      setQueryParams({ search: debouncedSearch || undefined }, "replaceIn");
+    }
+  }, [debouncedSearch, searchProps, setQueryParams]);
```
bdow: I see where this initial `onSearch` call is happening now. Still not sure I see if its necessary. Maybe this could be a `useDebouncedCallback` instead?

bdow (approved):
Would love to see about removing the unnecessary (_I think unnecessary_) initial `onSearch` call. Otherwise, happy to approve and move forward

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
abadami: Oh yeah, I don't think it's necessary either looking at it. Should be able to change it and I think that'll remove the need for that test as well.

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
abadami: Ah, I didn't realize it set query params automatically. I think I'll leave it as is for now, could be worth having a few tests using the session storage method. I did set the storage key to a const though.

bdow (approved):

PR #1303, opened 2026-06-30
feat: Adding "with dot" icons for search, filter, and columns, and rename "kanban" icon to "columns"

abadami:
### Adding "with dot" icons

Adding the new "filters with dot", "columns with dot", and "search with dot" icons. These have their own specific fill applied to their dots, which are using our tokens. 

This also renames our "kanban" icon to "columns" as per design.


stephenh (approved):
Neat

src/components/Icon.stories.tsx
```diff
@@ -14,6 +15,7 @@ export default {
     },
   },
   parameters: { controls: { exclude: ["xss", "icon", "inc"] } },
+  decorators: [withBeamDecorator],
```
bdow: **question:** Why was this decorator needed now?

src/components/Icon.tsx
```diff
@@ -56,11 +56,20 @@ export const Icon = React.memo((props: IconProps) => {
...
+  columns: (
     <>
       <path d="M19.893 3.001H4C2.897 3.001 2 3.898 2 5.001V19.001C2 20.104 2.897 21.001 4 21.001H19.893C20.996 21.001 21.893 20.104 21.893 19.001V8V7V6V5.001V5C21.892 3.897 20.994 3.001 19.893 3.001ZM8 19.001H4V8H8V19.001ZM14 19.001H10V8H14V19.001ZM16 19.001V8H19.893L19.894 19.001H16Z" />
     </>
   ),
+  columnsWithDot: (
+    <>
+      <path
+        d="M15.6006 3.00098C15.2173 3.75094 15 4.59994 15 5.5C15 6.67655 15.3708 7.76585 16 8.66016V19.001H19.8936L19.8926 10.9658C20.0921 10.9877 20.2947 11 20.5 11C20.9812 11 21.4477 10.9374 21.8926 10.8213V19.001C21.8926 20.104 20.9956 21.001 19.8926 21.001H4C2.897 21.001 2 20.104 2 19.001V5.00098C2 3.89798 2.897 3.00098 4 3.00098H15.6006ZM4 8V19.001H8V8H4ZM10 8V19.001H14V8H10Z"
+        fill="#242424"
```
bdow: **request:** This `fill` should be inherited by whatever color is passed to the component and set on the outer svg element.

src/components/Icon.tsx
```diff
@@ -56,11 +56,20 @@ export const Icon = React.memo((props: IconProps) => {
...
+      />
+      <circle cx="20.5" cy="5.5" r="3.5" fill={`var(${Tokens.Primary})`} />
```
bdow, line 67: **nit/suggestion:** Truss exports a `maybeCssVar` that will conditionally check if the string starts with `--` and wrap it in a `var()`. Not necessary here, as we know that the Token color will return a `--` var name. But for times where we might use a dynamic color that could either be a variable or static color, than we should look to use the helper.

src/components/Icon.tsx
```diff
@@ -225,6 +240,15 @@ export const Icons = {
...
   ),
+  filterWithDot: (
+    <>
+      <path
+        d="M7.625 4C8.10817 4 8.49988 4.39786 8.5 4.88867V5.92969C9.51948 6.29576 10.25 7.28332 10.25 8.44434C10.25 9.60537 9.51948 10.5929 8.5 10.959V19.1113C8.49988 19.6022 8.10817 20 7.625 20C7.14183 20 6.75012 19.6022 6.75 19.1113V10.959C5.73052 10.5929 5 9.60537 5 8.44434C5.00005 7.28332 5.73052 6.29576 6.75 5.92969V4.88867C6.75012 4.39786 7.14183 4 7.625 4ZM15.5 8.1084C16.0277 8.45265 16.6177 8.70846 17.25 8.85547V12.1514C18.2696 12.5174 19 13.5059 19 14.667C18.9999 15.8279 18.2695 16.8156 17.25 17.1816V19.1113C17.2499 19.6022 16.8582 20 16.375 20C15.8918 20 15.5001 19.6022 15.5 19.1113V17.1816C14.4805 16.8156 13.7501 15.8279 13.75 14.667C13.75 13.5059 14.4804 12.5174 15.5 12.1514V8.1084ZM16.375 13.7773C15.8917 13.7773 15.5 14.1761 15.5 14.667C15.5002 15.1578 15.8918 15.5557 16.375 15.5557C16.8582 15.5557 17.2498 15.1578 17.25 14.667C17.25 14.1761 16.8583 13.7773 16.375 13.7773ZM7.625 7.55566C7.14179 7.55566 6.75006 7.95347 6.75 8.44434C6.75 8.93526 7.14176 9.33301 7.625 9.33301C8.10824 9.33301 8.5 8.93526 8.5 8.44434C8.49994 7.95347 8.10821 7.55566 7.625 7.55566Z"
+        fill="#242424"
```
bdow: **request:** Same here regarding removal of the fill and just let it be inherited.

src/components/Icon.stories.tsx
```diff
@@ -59,6 +61,9 @@ export const Icon = (props: IconProps) => {
     "remove",
     "pin",
     "impersonate",
+    "columnsWithDot",
+    "searchWithDot",
+    "filterWithDot",
```
bdow: **suggestion:** I'd like to see these `WithDot` icons either:
1. Next to their normal, not with dot, icons.
2. Have their own "Badged" group.

src/components/Icon.tsx
```diff
@@ -56,11 +56,20 @@ export const Icon = React.memo((props: IconProps) => {
  */
 export const Icons = {
   // Actions
-  kanban: (
+  columns: (
     <>
       <path d="M19.893 3.001H4C2.897 3.001 2 3.898 2 5.001V19.001C2 20.104 2.897 21.001 4 21.001H19.893C20.996 21.001 21.893 20.104 21.893 19.001V8V7V6V5.001V5C21.892 3.897 20.994 3.001 19.893 3.001ZM8 19.001H4V8H8V19.001ZM14 19.001H10V8H14V19.001ZM16 19.001V8H19.893L19.894 19.001H16Z" />
     </>
   ),
+  columnsWithDot: (
```
bdow: **thought:** I'm leaning towards a more semantic naming for these, like "columnsActive", or "columnsBadged".
"Active" might imply too much, where I think "Badge" is a widely used term for this sort of style that something of active.

bdow (changes requested):

src/components/Icon.stories.tsx
abadami: I prefer next to their normal icons for now. If we do end up getting more "with dot" type icons, then I'd say we make the "Badged" group.

src/components/Icon.tsx
abadami: Good to know!

src/components/Icon.tsx
abadami: :+1:

src/components/Icon.tsx
abadami: I'm finding I agree with updating to `{iconName}Badged`, but is this something we'd want to talk with design about first to maintain consistency in naming?

src/components/Icon.tsx
abadami: Yep, missed that with these last two I did.

src/components/Icon.stories.tsx
abadami: It's not, I glossed over this one. I had an idea regarding how to do this that I scrapped. Forgot to remove this.

src/components/Icon.tsx
bdow: Yup. Let's chat with Design and suggest naming it to `___Badged`

bdow (approved):

src/components/Icon.tsx
abadami: Got Design approval on this one! Will be changed for the release.

PR #1304, opened 2026-06-30
feat: Flexible min/max cards

abadami:
### Flexible min/max cards

Apply a min value and a `1fr` max value. This should allow it to more dynamically fill the space as needed.

compact-card-view


src/components/Table/GridTable.tsx
```diff
@@ -57,6 +57,9 @@ export function setRunningInJest() {
   runningInJest = true;
 }
 
+const CARD_MIN_WIDTH_PX = 280;
+const CARD_MAX_WIDTH_PX = 330;
```
bdow: **thought:** I think the idea was to try and allow them to grow as wide as needed. They just can't shrink below a certain size. Then we'd show this to design and see if they are okay with the cards growing that wide. 
Try putting the max to `1fr` and then we'll see these cards actually be a bit more flexible.

abadami: Yeah, I drafted this as I wasn't 100% certain on the ask, but this makes sense to me since I thought the max part was weird the way I did it (as it'd always fill it to max anyway if it can).

FWIW, I messed around with the min value a little bit to see what works. I feel like 280 is the bare minimum we should support.

Anywho, yeah `1fr` was the piece I was missing for this one. This is doing what we need now.

bdow (approved):

PR #1305, opened 2026-07-01
fix: GridTable virtual render function to be a component

bdow:
As we can toggle between the 'as' properties (Card <-> List in GridTableLayout), then we need to have the virtual render method to be a component, as it has state. Toggle between the 'card' and 'virtual' currently causes errors of "changes hook count on the same component". This fixes those errors.

### DIFF
--- a/src/components/Table/GridTable.tsx
+++ b/src/components/Table/GridTable.tsx
@@ -648,8 +648,26 @@ export function GridTable<R extends Kinded, X extends Only<GridTableXss, X> = an
             infiniteScroll={infiniteScroll}
             persistScrollPosition={persistScrollPosition}
           />
+        ) : _as === "virtual" ? (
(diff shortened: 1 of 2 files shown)


src/components/Table/GridTable.tsx
abadami, line 703: **thought**: While it's immediately unnecessary, I wonder if we should just go ahead and convert both the `renderDiv` and `renderTable` to be their own components as well?

abadami (approved):

src/components/Table/GridTable.tsx
bdow: Yea, not a bad idea. 
My thought there is that they have no state, and do not need to be elevated to be React Components. They just return JSX, which I think is good to keep them as-is. We can keep the react component tree size down (not by much granted).

PR #1307, opened 2026-07-01
feat: Support showing Tags in TableCard badge slot

bdow:


src/components/Table/components/TableCard.tsx
```diff
@@ -135,10 +140,13 @@ export function TableCardView(props: TableCardViewProps) {
               <h4 css={Css.xl.fwb.$} {...tid.title}>
                 {title}{" "}
               </h4>
-              {badge && (
-                <span css={Css.sm.wsnw.$} {...tid.badge}>
-                  {badge}
-                </span>
+              {(badge || badgeTags?.length) && (
```
abadami, line 143: **question**: I think I remember this discussion with design, but just to be sure. It was decided that these tags are always tied to another piece of column data? Basically, could we ever want a badge tag but without the regular badge text? I could maybe see it on cards that have the tag tied to the title... 

I don't think we have to worry about it now regardless, we can change it if it comes up.

abadami (approved):
It could be worth having one of GridTable or GridTableLayout card view stories also implement badge tags, but I don't think it needs to hold off releasing the PR.

src/components/Table/components/TableCard.tsx
bdow: Correct. I think we build it as needed, rather than just saying a tag can go into any of these right now. Let's wait for a design to show how it should look, then we'll add it in.

PR #1308, opened 2026-07-02
feat: TagGroup component

bdow:


src/components/TagGroup.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+  component: TagGroup,
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=238-2165",
+    },
+  },
+} as Meta<TagGroupProps>;
+
+export const List = newStory(Template.bind({}), {});
```
stephenh: Fwiw this "not with the edit button" story still shows the edit button, not sure if that really matters, just mentioning it.

src/components/TagGroup.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const tid = useTestIds(otherProps, "tagGroup");
+
+  return (
+    <div {...tid} css={Css.df.aic.fww.gap1.$}>
+      {tags.map((tag) => (
+        <Tag key={tag.text} {...tag} variant="secondary" />
+      ))}
+      {onEdit && (
+        <span css={Css.ml1.sm.$}>
+          <Button label="Edit" variant="text" onClick={onEdit} {...tid.edit} />
```
stephenh, line 30: Totally get we're implementing what came over from design/UX, so this isn't a critique of the code, but imo this Edit feels like a weird UX, even where it's used in the product figmas.

I.e. normally I expect a component that has "edit" CTAs to flip over to like inline editing, i.e. a textbox that I click into, a selectfield I click in to.

But for this TagGroup, the edit experience in the product page that happens to want this is "...go back a step and change the list on the previous step".

Which, dunno, just feels weird in general, and particularly to bake into the component as a standard pattern 🤷 

Like maybe it's actually better for TagGroup to never have edit, like it's really read-only (b/c it basically is), and just let the product page do this weird "...oh yeah and if you want to edit the tags you here see, here's the edit CTA that goes back on step in the stepper flow".

stephenh (approved):

src/components/TagGroup.stories.tsx
bdow: Shucks - I changed to using the template.bind paradigm and must've accidentally let that one in. Not intended and would prefer to not have it there. Thanks for double checking that.

src/components/TagGroup.tsx
bdow: Given its placement in the component, it's either open up a `endNode: ReactNode` sort of prop to get the "Edit" button in the right place per design, or we bake it in like this. 
Per how it'll be used - I agree its an odd placement for a "go back" sort of button. I had some blinders on here and only saw the component implementation - will try to get a better read on how it'll be used in the future.
I'll make a note in Figma asking for a little more clarity on this and see if that goes anywhere. Can do follow up changes if necessary.

src/components/TagGroup.stories.tsx
stephenh: Oh yeah ... fwiw I basically "design ergonomic programming DSLs for a living" 😅 and cannot understand what the `Template.bind` API is supposed to be doing.

Like I get the goal is "a set of defaults to use across the stories that then we override" ... but isn't that the same thing as:

```
function TestTagGroup(props: Partial<TagGroupProps>) {
  const tags = [...];
  return ... <TagGroup {...props} tags={tags} />;
}
```

...I feel like the answer is probably that it drives this Controls tab:

Image

And for some reason, storybook needs this `Template.bind` incantation to do that? 

...actually the `Template` is just a function, and so we're calling the JS native `FUnction.bind`, like usually you see that for `onClick.bind(this)` or what not, so now I am even more confused 😅

bdow: haha. Yea. I don't think its anything special. It's just a recommended structure from Storybook.

Well, after I wrote that above line, I went to find the page where they suggest the pattern. Turns out, I'm old and this is their v2 recommended pattern, which is no longer recommended 😄. Will remove.

stephenh: "no longer recommended" -> very on brand of storybook to have "slightly different but still kinda wtf" APIs from release to release. :sweat_smile: 

(I know I'm being kinda a jerk about their APIs, to be clear I love storybook, and would never build a frontend without it, I'm just kinda OCD about these things, and they're a super-popular / widely-used project, so I feel it's fair to apply a high bar to their DX. :sweat_smile: )

src/components/TagGroup.tsx
stephenh: Cool, yeah that sg!

PR #1310, opened 2026-07-02
feat: Table Card Style Update

abadami:
### TableCard Styling Update

Making some style adjustments to better match the Figma designs. Also, fixes the card height to fix scroll jittering.

Figma design for reference: https://www.figma.com/design/yJcdAYy1rqkxTTNhUQnjjU/H2-2026-Beam-Work?node-id=376-36664&t=mXEL9NgEWQro4CW3-4


src/components/Table/components/TableCard.tsx
```diff
@@ -137,40 +135,49 @@ export function TableCardView(props: TableCardViewProps) {
...
-          <div css={Css.df.fdc.gap1.$}>
-            <div css={Css.df.aic.gap1.fs("10px").lh("14px").$}>
-              <div css={Css.w25.hPx(8).br4.bgGray200.$}>
-                <div css={Css.h100.br4.bgBlue500.w(`${progressValue}%`).$} />
+              <div css={Css.df.fdc.fg1.$}>
+                {col2.map((d) => (
+                  <dl key={d.label} css={Css.df.gapPx(4).$} {...tid[defaultTestId(d.label)]}>
+                    <dt>{d.label}:</dt>
+                    <dd>{d.value}</dd>
+                  </dl>
```
bdow: **suggestion:** Could both of these be wrapped in a single `<dl>`? It's really all one list, just spread across two columns, which should be fine.

abadami: Right, forgot to switch that in. Good call.

bdow (approved):

src/components/Table/components/TableCard.tsx
```diff
@@ -118,8 +118,11 @@ export function TableCardView(props: TableCardViewProps) {
 
   const progressValue = useMemo(() => (progress !== undefined ? clampProgress(progress) : 0), [progress]);
 
+  const col1 = data.slice(0, Math.ceil(data.length / 2));
+  const col2 = data.slice(Math.ceil(data.length / 2));
+
   return (
-    <div css={Css.p3.w100.h100.bshBasic.bgColor(Tokens.Surface).df.fdc.gap2.$} {...tid}>
+    <div css={Css.p3.w100.hPx(430).bshBasic.bgColor(Tokens.Surface).df.fdc.gap2.br8.$} {...tid}>
```
bdow, line 125: Since we have only one place using this now, this hardcoded height is fine. Would love to make this a property we can pass in at somepoint, though. Maybe worth creating a ticket for this.

abadami: I went with a hard coded height for now because I wasn't sure if we wanted to chat about that a little. We could maybe have presets in a "sm", "md", or "lg" fashion that way we can enforce sizes a little bit better, but I think it's worth chatting about w/ design at some point.

PR #1315, opened 2026-07-06
feat: IconButton contrast mode update

bdow:
Includes new semantic css tokens for proper styling.


src/components/IconButton.tsx
```diff
@@ -37,6 +37,12 @@ export type IconButtonProps = {
    * for screen readers without showing the tooltip. An explicit `tooltip` or disabled reason still shows.
    */
   preventTooltip?: boolean;
+  /** Storybook-only visual state overrides for snapshotting pseudo-interactions. */
+  __storyState?: {
```
abadami, line 41: **thought**: I think this is a neat idea to help enforce the styles for storybook since the previous method wasn't consistent. My only worry is introducing this pattern of more "internal / dev" type props might clutter prop definitions if we overuse them. Probably not an immediate issue, nor do I think it'll actually become a problem, but something to keep in mind.

src/components/IconButton.tsx
```diff
@@ -109,10 +120,14 @@ export function IconButton(props: IconButtonProps) {
...
-          : isCircle && (isHovered || active || isFocusVisible)
-            ? defaultIconColor
-            : iconColor)
+          ? Tokens.OnSurfaceDisabled
+          : isOutline && isPressing
+            ? Tokens.OnSurfaceRaisedPressed
+            : isOutline && isHovered
+              ? Tokens.OnSurfaceRaisedHover
+              : isCircle && (isHovered || active || isFocusVisible)
+                ? defaultIconColor
```
abadami, line 129: **nit/question (non-blocking)**: Do we have any best practices for how deep nested ternary operators should go before we reconsider it syntactically? Personally I think this one is on the border of being too deep, but I don't think it's a big enough deal to block the PR (and admittedly I don't have a significantly better idea for readability tbh).

abadami (approved):
Looks good! Left a couple comments but nothing that I think should block the PR going in.

PR #1316, opened 2026-07-06
feat: Search is now cleared when clicking on any "clear filters" button in GridTableLayout

abadami:
### Clear search when clearing filters

This PR will clear the search textbox and state whenever the filters are cleared in `GridTableLayout`. This also adds a `clearFiltersToken` that helps denote when the filters are cleared, to react to as needed (like with the debounced search value in `GridTableLayoutActions`).


src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -40,7 +40,7 @@ function FilterPanelOpen<F extends Record<string, unknown>, G extends Value = st
   filterImpls,
   filter,
   setFilter,
-  onClear,
+  onClear = () => {},
```
bdow: **question:** Do you think it'd make more sense for us to just allow it to be `undefined`? We have a util method called `maybeCall` that we can use to conditionally call this.

src/components/Layout/GridTableLayout/FilterPanel.tsx
```diff
@@ -78,7 +78,7 @@ function FilterPanelClosed<F extends Record<string, unknown>, G extends Value =
...
```
bdow: **suggestion:** Same here - I think it'd make more sense to just allow it to be undefined.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -100,6 +104,16 @@ function GridTableLayoutActionsComponent<
...
 
+  // Resync the local search input whenever `clearFilters` runs elsewhere (e.g. the empty state's
+  // "Clear Filters" button), since that can reset `layoutState.searchString` without this component's
+  // knowledge. `clearFiltersToken` starts at 0 and only ever increases, so 0 means "no reset yet".
+  useEffect(() => {
+    if (!clearFiltersToken) return;
+    handleSearchDebounced.cancel();
+    setSearchValue("");
+    setQueryParams({ search: undefined }, "replaceIn");
+  }, [clearFiltersToken, handleSearchDebounced, setQueryParams]);
```
bdow: **thought:** `useEffect`'s alway make me question them. There's always the "run immediately" and "run whenever it changes", which you're handling here via the special `0` sentinel token - but it does make me wonder if there's another way of handling this. 
I'm wondering if theres a more imperative api (similar to `TextField.api`), or a sort of "clear search ref" approach that would simplify this and make it more explicit of a call, and less of a reaction (moving away from useEffect).

abadami: Completely fair. The other idea I was considering was having `useGridTableLayoutState` own everything and having `GridTableLayoutActions` be completely controlled, but I believe `useGridTableLayoutState` is being more widely used compared to the actual layout in Blueprint currently and I didn't want to mess with that API too much. Also, I felt that was a heavier change and this, while using a `useEffect`, felt cleaner.

I'll explore other explicit call methods real quick. I'm pretty intrigued by the `TextField.api` implementation, so I'll check that out and see if it makes sense to do something similar here.

bdow (approved):

PR #1317, opened 2026-07-07
feat: Refactor of SelectCardGroup to support list and grid views

bdow:
SelectCard is no longer supported as an individual component - only available through the SelectCardGroup.

SelectCardGroup API and implementation changed quite a bit in order to properly handle multi (checkbox group) vs single (radio group) selection behaviors. This allows for proper keyboard and other accessibility features to work based on the underlying component.

Adds a "list" vs "grid" view, which dictates the layout and the card view.

Other changes include adjusting the Bound* (formstate aware) versions of the components to fit the new API structure. Included in the BoundForm stories for testing.


src/forms/BoundSelectCardGroupField.tsx
```diff
@@ -1,38 +1,74 @@
...
 import { defaultLabel } from "src/utils/defaultLabel";
 
-export type BoundSelectCardGroupFieldProps<V extends Value> = Omit<
-  SelectCardGroupProps<V>,
+type BoundSelectCardCheckboxGroupFieldProps<V extends Value> = DistributiveOmit<
+  SelectCardCheckboxGroupProps<V>,
   "label" | "values" | "onChange"
 > & {
   field: FieldState<V[] | null | undefined>;
-  /** Make optional so that callers can override if they want to. */
```
bdow, line 13: Comment worth keeping

src/inputs/internal/radioStyles.ts
bdow, line 1: Consider putting in the CSS reset

src/inputs/SelectCard/GridSelectCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+      label={label}
+      selected={isSelected}
+      disabled={isDisabled}
+      tooltip={tooltip}
+      view="grid"
+      __storyState={__storyState}
+      {...others}
+    >
+      <VisuallyHidden>
+        <input {...mergeProps(inputProps, focusProps)} {...tid.value} />
```
bdow: **thought:** this focusProp might be unnecessary.

src/inputs/SelectCard/SelectCardGroup.tsx
```diff
@@ -0,0 +1,23 @@
...
+  SelectCardRadioGroupProps,
+  SelectCardView,
+} from "src/inputs/SelectCard/types";
+
+export function SelectCardGroup<V extends Value>(props: SelectCardGroupProps<V>) {
+  if (isSelectCardRadioGroupProps(props)) {
+    return <SelectCardRadioGroup {...props} />;
+  }
+
+  const { multiselect: _multiselect, ...checkboxProps } = props;
```
bdow: **nit:** probably unnecessary

src/inputs/SelectCard/SelectCardCheckboxGroup.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <SelectCardCheckboxGroupItem
+          key={String(option.value)}
+          option={option}
+          groupState={state}
+          isSelected={values.includes(option.value)}
+          view="grid"
+          {...tid[defaultTestId(option.label)]}
+        />
+      ));
+
```
bdow: **question:** Can we simplify this?

src/inputs/SelectCard/SelectCardCheckboxGroupItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+type SelectCardCheckboxGroupItemProps<V extends Value> = SelectCardGroupItemProps<V> & {
+  groupState: CheckboxGroupState;
+};
+
+export function SelectCardCheckboxGroupItem<V extends Value>(props: SelectCardCheckboxGroupItemProps<V>) {
+  const { option, groupState, isSelected, view, ...others } = props;
+  const { label, description, disabled, tooltip, value } = option;
+  const ref = useRef<HTMLInputElement>(null);
+  const { inputProps, isDisabled: isOptionDisabled } = useCheckboxGroupItem(
+    { value: value as string, "aria-label": label, isDisabled: disabled ?? false },
```
bdow: **question:** Do we need to keep doing this `value as string` and `value as unknown as string`?

src/inputs/SelectCard/SelectCardGroup.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+  ];
+}
+
+function createListCategoryOptions(): SelectCardListGroupItemOption<Category>[] {
+  return [
+    { label: "Math", description: "Numbers and equations", value: Category.Math },
+    { label: "History", description: "Past events", value: Category.History },
+    { label: "Not Applicable", description: "None apply", value: Category.Na, selectionBehavior: "exclusive" },
+  ];
+}
```
bdow, line 27: move to bottom

src/inputs/SelectCard/SelectCardGroup.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+        values={[Category.Math]}
+        onChange={() => {}}
+      />,
+    );
+    // Then list rows with descriptions are rendered
+    expect(r.math).toHaveTextContent("Numbers and equations");
+    expect(r.math_value).toBeInTheDocument();
+  });
+
+  it("ignores option icons in list view", async () => {
```
bdow: these 3 tests can be removed - validated via storybook/chromatic

src/inputs/SelectCard/SelectCardRadioGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+    : (props as SelectCardGridGroupLayoutProps<V>).options.map((option) => (
+        <SelectCardRadioGroupItem
+          key={String(option.value)}
+          option={option}
+          groupState={state}
+          isSelected={value === option.value}
+          view="grid"
+          {...tid[defaultTestId(option.label)]}
+        />
+      ));
```
bdow: same with checkboxes - can we simplify?

src/inputs/SelectCard/StyledRadio.tsx
```diff
@@ -0,0 +1,51 @@
...
+      css={{
+        ...radioReset,
+        ...radioDefault,
+        ...(isDisabled && isSelected
+          ? radioDisabledSelected
+          : isDisabled
+            ? radioDisabled
+            : isSelected
+              ? radioChecked
+              : radioUnchecked),
```
bdow: suggestion: hoist values up to const's instead of so many inline ternarys

src/inputs/SelectCard/StyledRadio.tsx
```diff
@@ -0,0 +1,51 @@
...
+        ...(isHovered && !isDisabled ? radioHover : {}),
+        ...Css.mtPx(2).fs0.$,
```
bdow: mtPx isn't needed

src/inputs/SelectCard/types.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export function isSelectCardRadioGroupProps<V extends Value>(
+  props: SelectCardGroupProps<V>,
+): props is SelectCardRadioGroupProps<V> {
+  return props.multiselect === false;
+}
+
+export function isSelectCardListLayoutProps(props: { view?: SelectCardView }): props is { view: "list" } {
+  return props.view === "list";
+}
```
bdow: Move to `utils`

src/inputs/SelectCard/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  return [...currentValues.filter((v) => !exclusiveValues.has(v)), clickedValue];
+}
+
+/** Returns the value that was toggled between two selection arrays. */
+export function findToggledSelectCardGroupValue<V extends Value>(prev: V[], next: V[]): V | undefined {
+  const added = next.find((v) => !prev.includes(v));
+  if (added !== undefined) return added;
+  return prev.find((v) => !next.includes(v));
+}
```
bdow, line 43: can we simplify the selection behavior for checkboxes?

src/inputs/SelectCard/ListSelectCard.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+        label="Pressed selected"
+        description={description}
+        selected
+        __storyState={{ pressed: true }}
+      />
+    </div>
+  );
+}
+
+export function RadioStates() {
```
allangaldinosilva, line 44: chore: because this is using the `ListSelectCardWrapper`, it's currently rendering `checkbox`es insead of `radio`:

Image

Maybe just add a `type` prop to the wrapper?

src/inputs/SelectCard/SelectCardGroup.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+    expect(r.math_value).not.toBeChecked();
+  });
+
+  it("supports single-select in list view", async () => {
+    const onChange = vi.fn();
+    const r = await render(
+      <SelectCardGroup
+        label="Categories"
+        view="list"
+        options={createListCategoryOptions()}
```
allangaldinosilva, line 90: thought/suggestion: I noticed this component does not expose a `data-testid={label}` like others do, for example if we run `expect(r.categories)...` in a test, it fails since `r.categories` is not found:

Image

I think it would make writing tests easier, for example I used to use `getOptions` very often to validate the options were properly rendered.

allangaldinosilva: This is also happening with `MultiSelectCardGroup` so maybe ensure all of them are aligned.

src/inputs/RadioGroupField.tsx
```diff
@@ -192,61 +193,3 @@ function Radio<K extends string>(props: {
     </label>
   );
 }
-
```
allangaldinosilva, line 195: praise: nice clean up 🔪

allangaldinosilva (approved):
This is looking great to me!

src/inputs/SelectCard/ListSelectCard.stories.tsx
bdow: Good catch - I initially had the opposite issue. Fixing now!

src/inputs/SelectCard/SelectCardGroup.test.tsx
bdow: We can add a label for the group, but I don't think `getOptions` would work here still. That looks specifically for elements with a `role="option"` within a `listbox`, which works properly for SelectField, but radio and checkbox inputs do not have that role.

allangaldinosilva: gotcha, makes sense 👍

bdow: I think that'd be a nice improvement though. Will look to add in.

src/inputs/SelectCard/types.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export type SelectCardView = "grid" | "list";
+
+type SelectCardGroupItemOptionBase<V extends Value> = {
+  label: string;
+  /** Optional secondary copy shown beneath the label. */
+  description?: string;
+  disabled?: boolean;
+  /** Tooltip shown on hover, i.e. to explain why the option is disabled. */
+  tooltip?: string;
```
stephenh: Feels like `description` / `tooltip` will eventually end up as ReactNodes? 

I get keeping `label` as a string for accessibility, but these others seems like invariably want "some bold" / "some formatting" in them.

src/inputs/SelectCard/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+export function getNextSelectCardGroupValues<V extends Value>({
+  currentValues,
+  clickedValue,
+  options,
+}: GetNextSelectCardGroupValuesArgs<V>): V[] {
+  const option = options.find((o) => o.value === clickedValue);
+  if (!option) return currentValues;
+
+  const isExclusive = option.selectionBehavior === "exclusive";
+  const exclusiveValues = new Set(options.filter((o) => o.selectionBehavior === "exclusive").map((o) => o.value));
```
stephenh: Wouldn't mind these `const` being defined directly before their respective usage.

stephenh (commented):
Really like the `SelectCardGroup` / `MultiSelectCardGroup` split!

src/inputs/SelectCard/types.ts
bdow: That's fair. I'll update to support it now.

src/inputs/SelectCard/utils.ts
bdow: I noticed this too during my self review, but forgot to make note of it. Will update. Thanks for flagging!

src/forms/BoundForm.tsx
```diff
@@ -232,16 +233,20 @@ export function boundCheckboxGroupField(props: Omit<BoundCheckboxGroupFieldProps
   });
 }
 
-export function boundSelectCardField(props: Omit<BoundSelectCardFieldProps, KeysToOmit>) {
+export function boundSelectCardGroupField<V extends Value>(
+  props: DistributiveOmit<BoundSelectCardGroupFieldProps<V>, KeysToOmit>,
```
stephenh, line 238: I wouldn't mind a `// Use DistributeOmit to ...the rationale...` just as a reminder for why it's here

bdow: Sure, can do. Will also add a comment at the definition site.

src/inputs/SelectCard/types.ts
```diff
@@ -0,0 +1,93 @@
...
+  label: string;
+  /** Optional secondary copy shown beneath the label. */
+  description?: ReactNode;
+  disabled?: boolean;
+  /** Tooltip shown on hover, i.e. to explain why the option is disabled. */
+  tooltip?: ReactNode;
+  /** The value of the SelectCardGroup item. */
+  value: V;
+  /** For checkbox groups, selecting this option clears other options and cannot be combined. */
+  selectionBehavior?: "exclusive";
```
stephenh, line 18: Ah neat -- seems like a potentially annoying ask (i.e. to have pages remember "oh you selected n/a, unselect all the others, etc"), but this is a great declarative API for it 👍

src/utils/rtl.test.tsx
```diff
@@ -437,4 +447,91 @@ describe("rtl", () => {
...
+          label="Categories"
+          options={createSelectCardOptions()}
+          values={["math"]}
+          onChange={() => {}}
+          data-testid="custom"
+        />,
+      );
+      // When calling getOptions with the component test-id root
+      // Then it finds the group and returns option labels
+      expect(getOptions(r.custom)).toEqual(["Math", "History", "Finance"]);
```
stephenh, line 526: Nice!

src/inputs/SelectCard/SelectCardShell.tsx
```diff
@@ -0,0 +1,67 @@
...
+  selected?: boolean;
+  disabled?: boolean;
+  tooltip?: ReactNode;
+  __storyState?: SelectCardStoryState;
+  children: ReactNode;
+  view: SelectCardView;
+};
+
+/** Tooltip, hover/focus shell, and shared selection-state borders shared by both card variants. */
+export function SelectCardShell(props: SelectCardShellProps) {
```
stephenh, line 20: Nice wrapper around both variants 👍

stephenh (approved):
🎉

PR #1322, opened 2026-07-08
feat: Page Header Breadcrumbs

abadami:
### PageHeader Breadcrumbs

Adds Breadcrumb support to `PageHeader`.


src/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,36 @@
...
+export type BreadcrumbsProps = {
+  breadcrumbs: Breadcrumb | Breadcrumb[];
+};
+
+export function Breadcrumbs({ breadcrumbs }: BreadcrumbsProps) {
+  const tid = useTestIds({}, "breadcrumb");
+  const items = Array.isArray(breadcrumbs) ? breadcrumbs : [breadcrumbs];
+
+  return (
+    <div css={Css.df.aic.gapPx(4).$}>
```
bdow, line 43: **thought/suggestion:** I don't know if design has determined this yet, but on mobile we could potentially wind up fairly deep in the app. We might want to ensure we support this scrolling off the page horizontally. Maybe similar to the active filters with the hidden scrollbar.
I'd also suggest that this has a `useLayoutEffect` that scrolls the container to the end by default.

src/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,36 @@
...
+      {items.map((bc, i) => (
+        <Fragment key={bc.label}>
+          {i > 0 && <span css={Css.fs0.xs.$}>/</span>}
+          <Link
+            {...tid.link}
+            to={bc.href}
+            title={bc.label}
+            css={Css.xs.gray900.truncate.mw0.maxwPx(120).onHover.gray800.$}
```
bdow: I think if we do the scroll behavior, then we don't need to do this truncation & max-width here.

bdow: Maybe I'm getting ahead of myself, but are these styles for hover and such already defined by design? I can't find it in Figma if it is.

src/components/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    // Given a breadcrumb with a very long label
+    const longLabel = "A Really Long Project Name That Should Truncate Instead Of Wrapping";
+
+    // When rendered
+    const r = await render(<Breadcrumbs breadcrumbs={{ label: longLabel, href: "/" }} />, {});
+
+    // Then the link is capped at a max width, with the full label available via a title attribute
+    expect(r.breadcrumb_link).toHaveStyle({ maxWidth: "120px" });
+    expect(r.breadcrumb_link).toHaveAttribute("title", longLabel);
+  });
```
bdow: **thought:** I'm not sure if this test is worth it. I think its more visual than anything and can just be validated via Storybook.

src/components/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  it("does not clip short labels", async () => {
```
bdow: Same here - this doesn't really test that its not clipping anything. It's basically the same exact test as above, and wouldn't fail if you made "Home" a super long title.

src/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,36 @@
...
+  href: string;
+};
+
+export type BreadcrumbsProps = {
+  breadcrumbs: Breadcrumb | Breadcrumb[];
+};
+
+export function Breadcrumbs({ breadcrumbs }: BreadcrumbsProps) {
+  const tid = useTestIds({}, "breadcrumb");
+  const items = Array.isArray(breadcrumbs) ? breadcrumbs : [breadcrumbs];
```
bdow, line 18: **praise:** I like this normalization front and center

src/components/Breadcrumbs.tsx
bdow: I saw what you're doing with the truncation after writing this comment. I've tried that sort of set up before on apps, and I think truncation can end up making things look funky. If we can avoid truncation, I think we should. I'd also be curious to know what design thinks.

src/components/PageHeader.tsx
```diff
@@ -12,23 +13,29 @@ export type PageHeaderProps<V extends string, X> = {
...
   return (
-    <header {...tid} css={Css.df.fdc.pt3.pr3.pl3.bb.bc(Tokens.SurfaceSeparator).bgColor(Tokens.Surface).$}>
-      <div css={Css.df.jcsb.mb2.w100.gap1.$}>
-        <div>
-          {/* Breadcrumbs here */}
+    <header {...tid} css={Css.df.fdc.pt3.px3.bb.gap2.bc(Tokens.SurfaceSeparator).bgColor(Tokens.Surface).$}>
+      <div
+        css={{
+          ...Css.df.jcsb.w100.gap1.$,
+          ...Css.if(!tabs).mb2.$,
```
bdow, line 29: **question:** Was this conditional of `mb2` only when not showing tabs a miss before?

src/components/PageHeader.tsx
```diff
@@ -12,23 +13,29 @@ export type PageHeaderProps<V extends string, X> = {
...
+        }}
+      >
+        <div css={Css.mw0.$}>
+          {breadcrumbs && <Breadcrumbs {...breadcrumbs} />}
```
bdow, line 33: **thought (non-blocking):** I wonder if breadcrumbs will be better off being on its own horizontal space depending on how we approach the scrolling behavior with it.

src/components/Breadcrumbs.tsx
abadami: Honestly I'm not a big fan of the truncation either, but I did think it was better than doing nothing. In hindsight, I probably should've asked design about it so I will do so now.

src/components/Breadcrumbs.tsx
abadami: I could not find it if they were either. I'll include it in my question regarding long / deep breadcrumbs

src/components/PageHeader.tsx
abadami: Yeah, essentially. Previously the mb2 was always present, which was great normally if we didn't have tabs. If we did have tabs though, the tabs were clearly a bit more spaced from the rest of the header elements than they should be, which I didn't really notice until working on this.

src/components/Breadcrumbs.test.tsx
abadami: Hm, yeah that's fair. I think there's an argument for the title attribute showing, but that's not a big deal to test and there's a chance we're going to be ditching that anyway.

src/components/Breadcrumbs.tsx
abadami: Looks like for deep breadcrumbs, we're going with a collapsible state. As of right now assumed that deep breadcrumbs are >= 3, but waiting for official word from design.

This also doesn't technically do anything for long breadcrumb labels yet, and I'm also waiting for design on that, but that likely won't be an issue until later.

src/components/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+      { label: "Projects", href: "/projects" },
+    ];
+
+    // When rendered
+    const r = await render(<Breadcrumbs breadcrumbs={breadcrumbs} />, {});
+
+    // Then they render as links in order with the correct hrefs
+    const links = r.getAllByTestId("breadcrumb_link");
+    expect(links.map((l) => l.textContent)).toEqual(["Home", "Projects"]);
+    expect(links.map((l) => l.getAttribute("href"))).toEqual(["/", "/projects"]);
```
stephenh: I think the `_n` syntax of rtl-utils would be more idiomatic:

```
expect(r.breadbrumb_link_0).toHaveTextContext("Home");
expect(r.breadbrumb_link_0).toHaveAttributeValue("href", "/");
expect(r.breadbrumb_link_1).toHaveTextContext("Projects");
expect(r.breadbrumb_link_1)....("/projects");
```

abadami: Yeah, I think it's more explicit too. Will switch it over.

src/components/Breadcrumbs.stories.tsx
bdow, line 1: **thought (non-blocking):** For something that is fairly simple as breadcrumbs, there's a decent amount of rendering combinations and truncation logic that we deal with here. If it were up to me, we'd have rendered all the breadcrumbs, never truncating anything, and just let that container scroll horizontally. And on load, it'd immediately scroll itself to the end so the user see's the immediate parent page, allowing them to scroll back to see the rest of the breadcrumbs.
I'm not knocking the design and implementation, it's a cool pattern to do the truncation and ability to load the missing breadcrumbs on click. But even with truncation that means we need to properly support the horizontal overflow when expanded.
To sum it up: Let's look for ways to keep our UI patterns as simple as possible. Save the complex stuff for the really complex stuff.

src/components/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+    { label: "Projects", href: "" },
+    { label: "Project 123", href: "" },
+  ];
+
+  return <Breadcrumbs breadcrumbs={breadcrumbs} />;
+}
+
+// On desktop, collapses to the first two crumbs + "..." + last. Resize the
+// preview below 600px to see it drop to just the first crumb + "..." + last.
+export function CollapsesWithManyLongBreadcrumbs() {
```
bdow, line 39: **suggestion:** When I expand this on mobile it looks like: Image.
I think we should make the container holding the breadcrumbs stay as one line of text and scroll horizontally

bdow (approved):
Giving the approval, but would also like to see the text wrapping issue fixed (see screenshot) either in this PR or a follow up.

src/components/Breadcrumbs.stories.tsx
abadami: I think that's fair. My mindset with the implementation was that we were designing to hide the breadcrumbs in the mobile view entirely, so I figured the general issues with truncation wouldn't appear too often, if at all.

I think the scroll is the better long term solution, but I'll handle it in a follow-up since we are not implementing any breadcrumb'd pages yet

PR #1323, opened 2026-07-10
feat: Stepper Tabs

abadami:
### Stepper Tabs

Adds in the new stepper tabs. Does leave the legacy stepper currently as I don't think we want these to override current styling at the moment. Open to more clean ways to transition


src/components/StepperTabs/StepperTab.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+} as Meta;
+
+const states: { state: StepperTabState; label: string }[] = [
+  { state: "completed", label: "Step Label" },
+  { state: "active", label: "Step Label" },
+  { state: "activeCompleted", label: "Step Label" },
+  { state: "notVisited", label: "Details & Slots" },
+];
+
+export function States() {
```
abadami, line 10: These stories could be cleaned up a little bit...they are current kinda messy.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,105 @@
...
+      {...tid[defaultTestId(value)]}
+    >
+      {!collapsed && (
+        <>
+          <span css={Css.lineClamp1.$}>{label}</span>
+          {showCheck && (
+            <span css={Css.fs0.ml1.$}>
+              <Icon icon="check" inc={2.5} {...tid.check} />
+            </span>
+          )}
```
bdow: **suggestion:** It's not really a concern for us at Homebound, but best practice would be to supply something for the screenreader here along with the check mark. Like:
```<VisuallyHidden>Complete/Not Complete</VIsuallyHidden>```
I'd like to do these sorts of things whenever possible, even if its not within out specific needs. But maybe/hopefully one day we'll be using this for our external customers and we'll be ready to go for any type of user.

File: src/components/StepperTabs/StepperTab.tsx [hunk 3]
```diff
@@ -0,0 +1,105 @@
...
+      )}
+    </button>
+  );
+}
+
+const borderBottomWidthPx = 6;
+
+function getStepperTabStyles() {
+  const withBorderBottom = (color: Properties) => ({
+    ...Css.bb.add("borderBottomWidth", `${borderBottomWidthPx}px`).$,
```
bdow: **suggestion:** Maybe just inline `borderBottomWidthPx`. It's just used in this one place afaict

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,105 @@
...
+  };
+
+  const hoverStyles: Record<StepperTabState, Properties> = {
+    notVisited: Css.bgGray100.$,
+    completed: Css.bgGray100.$,
+    active: Css.bgGray100.$,
+    activeCompleted: Css.bgGray100.$,
+  };
+
+  // When collapsed, every state's border goes gray except "completed" (visited, inactive, completed), which stays blue.
```
bdow: **thought (non-blocking):** What this comment calls out is kind of an inconsistency in behavior. In desktop we treat this component as steps, allowing the user to back and forth, while in mobile we change the behavior to be just a progress bar. Just feels a little odd, imo.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const collapsedStyles: Record<StepperTabState, Properties> = {
+    notVisited: Css.bcGray300.cursor("default").hPx(0).py0.$,
+    completed: Css.bcBlue600.cursor("default").hPx(0).py0.$,
+    active: Css.bcGray300.cursor("default").hPx(0).py0.$,
+    activeCompleted: Css.bcGray300.cursor("default").hPx(0).py0.$,
+  };
```
bdow: **thought (non-blocking):** A lot of duplicated styles here. Could we consolidate these somehow? Everything is nearly `Css.bcGray300.cursor("default").hPx(0).py0.$`.
I'm actually a little surprised that the "active" is also gray. Feels like a miss from the handoff or design side..

src/components/StepperTabs/StepperTab.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export default {
+  component: StepperTab,
+} as Meta;
+
+const states: { state: StepperTabState; label: string }[] = [
+  { state: "completed", label: "Step Label" },
+  { state: "active", label: "Step Label" },
+  { state: "activeCompleted", label: "Step Label" },
+  { state: "notVisited", label: "Details & Slots" },
```
bdow: **thought:** I wonder if it'd be easier or more difficult to handle the active state separately from the completed state. They seem a bit orthogonal to each other. Maybe they overlap more than I think?

src/components/StepperTabs/StepperTab.tsx
abadami: I agree, that'd be really good to plan for ahead of time. I'll look out for other opportunities.

src/components/StepperTabs/StepperTab.tsx
abadami: I don't suppose we don't need to call it out per se, though I do think it's good to note...I could go either way on removing it or not. Though we could also be more direct and acknowledge the progress bar like behavior, but again that could be unnecessary.

src/components/StepperTabs/StepperTab.stories.tsx
abadami: I took a look into this, and yeah just keeping it split seems cleaner. I don't think we'll add any other additional state to this so I think keeping it separated makes sense.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -70,24 +69,19 @@ function getStepperTabStyles() {
...
   });
 
-  const stateStyles: Record<StepperTabState, Properties> = {
-    notVisited: { ...Css.gray400.$, ...withBorderBottom(Css.bcGray300.$) },
-    completed: { ...Css.blue700.$, ...withBorderBottom(Css.bcBlue600.$) },
-    active: { ...Css.blue700.smSb.$, ...withBorderBottom(Css.bcBlue600.$) },
-    activeCompleted: { ...Css.blue700.smSb.$, ...withBorderBottom(Css.bcBlue600.$) },
-  };
+  // Blue once either completed or active; bold only while active.
+  const stateStyles = (active: boolean, completed: boolean): Properties => ({
```
bdow: **request:** Our typical practice has been to write these as `function stateStyles() {` rather than `const stateStyles = () => {...}`
This is just a opinion we hold here, and would like to stay consistent. "If its a function, then declare it as a function" is typically the mindset.

abadami: Got it! Will switch it over.

Out of curiosity, I believe eslint has a rule for this. Is there a reason we aren't enforcing it with eslint? Or is that ruling too strict? I think it enforces all functions by default and I could see there being exceptions...

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  const hoverStyles: Properties = Css.bgGray100.$;
+
+  function getCollapsedStyles(active: boolean, completed: boolean): Properties {
+    return {
+      ...Css.cursor("default").hPx(0).py0.$,
+      ...(completed && !active ? Css.bcBlue600.$ : Css.bcGray300.$),
+    };
+  }
```
bdow: **nit/suggestion:** Could we just place this function at the module level? Doesn't feel like it needs to be part of this `getStepperTabStyles` function.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const ref = useRef(null);
+  const { buttonProps } = useButton(ariaProps, ref);
+  const { isFocusVisible, focusProps } = useFocusRing();
+  const { hoverProps, isHovered } = useHover(ariaProps);
+  const tid = useTestIds(props, "stepperTab");
+
+  const { baseStyles, getStateStyles, hoverStyles, focusRingStyles, disabledStyles, getCollapsedStyles } = useMemo(
+    () => getStepperTabStyles(),
+    [],
+  );
```
bdow: **question:** Given this has not dependencies, it makes me think that this is always a stable value returned and maybe we can just define these as stable values to start out with. 
So like, rather than calling `getStepperTabStyles`, maybe there is a module level const that returns what we need here that can return `baseStyles`, `hoverStyles`, etc. 
But the functions like `getStateStyles` and `getCollapsedStyles` should just be module level functions as well. Specifically those functions feel a little unnecessary to be memo'd. The values returned from those functions are re-evaluated on every render anyways, so there's not really a difference between making them module level functions vs what we're doing here in terms of performance - just a difference in how we allocate memory (though, very minor). I think I'd prefer we simplify the set up.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,99 @@
...
+  return (
+    <button
+      ref={ref}
+      {...mergeProps(buttonProps, focusProps, hoverProps)}
+      aria-label={label}
+      css={{
+        ...baseStyles,
+        ...getStateStyles(active, completed),
+        ...(isHovered && !disabled ? hoverStyles : {}),
+        ...(collapsed ? getCollapsedStyles(active, completed) : {}),
```
bdow, line 40: **thought:** The `getCollapsedStyles` set a height of 0px. Do we still need to do `{!collapsed && (` below? Seems like we're trying to solve the hiding of the label in two ways, and would prefer just to do it once. If we did the `h0` style, then we could also remove the `aria-label` defined on this element, as we wouldn't end up removing the label. Or vice-versa - keep the aria-label, remove the `h0` definition, and just rely on the `{!collapsed && (` to hide the content.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <>
+          <span css={Css.lineClamp1.$}>{label}</span>
+          {completed && (
+            <span css={Css.fs0.ml1.$}>
+              <Icon icon="check" inc={2.5} {...tid.check} />
+            </span>
+          )}
+          <VisuallyHidden>{completed ? "Complete" : "Not Complete"}</VisuallyHidden>
+        </>
+      )}
```
bdow: **thought (non-blocking):** Correct me if I'm wrong - but on desktop we treat this as a stepper sort of behavior, and on mobile, its a progress bar. Because those are different semantic components, I wonder if we should be treating them more different here. Especially from the accessibility perspective. Just a passing thought. More something I think for the design team and us to chat through than something for us to just determine. IMO, changing this behavior to be strictly linear on mobile (vs ability to click any step that is enabled) is a bit odd. Granted, we're just "mobile friendly" and probably it won't really impact many users.

src/components/StepperTabs/StepperTab.tsx
abadami: Oh yeah, good point! We do want the 0px height because otherwise when collapsed it still has a height that it had originally, and both I felt like it'd be better to have less height on the header when in mobile and design had no height assigned to its collapsed state anyway, so we'll stick with the height method for this.

src/components/StepperTabs/StepperTab.tsx
abadami: I'm always down to simplify the setup. We do use this pattern in other places (granted, they usually have dependencies, but they also don't have smaller functions constructing the styles like this does), I wonder if we should perhaps consider simplifying this pattern / have those components go towards this direction rather than a memo'd value.

src/components/StepperTabs/StepperTab.tsx
abadami: You're correct on how it's behavior works. I think if steps have a back button then it might technically be behaving the same because once a step is complete you can go next, then go back and remove a value from a required field, then you'd have differing a broken progress bar technically...

Yeah, I think it's worth validating with design in general on that.

src/components/StepperTabs/StepperTab.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+export default {
+  component: StepperTab,
+} as Meta;
+
+const combos: { active: boolean; completed: boolean; label: string }[] = [
+  { active: false, completed: true, label: "Step Label" },
+  { active: true, completed: false, label: "Step Label" },
+  { active: true, completed: true, label: "Step Label" },
+  { active: false, completed: false, label: "Details & Slots" },
+];
```
bdow: **nit:** please place combos inside of the States function

src/components/StepperTabs/StepperTab.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+    [true, false],
+    [true, true],
+  ])("renders the label (active=%s, completed=%s)", async (active, completed) => {
+    // Given a tab with the given active/completed combination
+    // When rendered
+    const r = await render(
+      <StepperTab label="Step Label" value="step" active={active} completed={completed} onClick={vi.fn()} />,
+    );
+    // Then the label is displayed
+    expect(r.stepperTab_step).toHaveTextContent("Step Label");
```
bdow: **question:** Is this really testing anything? Feels like we don't ever have a case where the label shouldn't be displayed, so why go through every combination?

src/components/StepperTabs/StepperTab.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+  ])("shows the check icon only when completed (active=%s, completed=%s)", async (active, completed) => {
+    // Given a tab with the given active/completed combination
+    // When rendered
+    const r = await render(
+      <StepperTab label="Step Label" value="step" active={active} completed={completed} onClick={vi.fn()} />,
+    );
+    // Then the check icon is only shown when completed, regardless of active-ness
+    const hasIcon = !!r.query.stepperTab_check;
+    expect(hasIcon).toBe(completed);
+  });
```
bdow: **nit/suggestion:** This testing every combination doesn't really seem necessary - I think you only need to do it twice - once with completed as false, and once with completed as true.
I think the assumption that "active-ness" could potentially influence the complete check is maybe unfounded?
Fwiw, my own personal opinion (that I'm not saying is right or wrong), is to avoid these `it.each` as I find them more difficult to follow what the test is doing. Like, I think this could be two separate tests of "shows completed check when step is completed" "does not show check when step is not completed". I think it goes back to that "DRY" vs "DAMP" mindset.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+  label: string;
+  value: string;
+  /** Whether this is the currently-selected step. */
+  active: boolean;
+  /** Whether this step's content has been completed. */
+  completed: boolean;
+  onClick: (value: string) => void;
+  disabled?: boolean;
+  /** Collapses the tab down to its colored bottom border only, hiding the label — for narrow/many-tab layouts. */
+  collapsed?: boolean;
```
bdow: **nit:** I don't think this comment `...for narrow/many-tab layouts.` is accurate. Isn't it just for the mobile view?

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+        </span>
+      )}
+      <VisuallyHidden>{completed ? "Complete" : "Not Complete"}</VisuallyHidden>
+    </button>
+  );
+}
+
+function withBorderBottom(color: Properties) {
+  return {
+    ...Css.bb.add("borderBottomWidth", `6px`).$,
```
bdow: **thought:** I don't think this border bottom width is really 6px. That seems huge. I think it's 3px.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+function getCollapsedStyles(active: boolean, completed: boolean): Properties {
+  return {
+    ...Css.cursor("default").hPx(0).py0.$,
+    ...(completed && !active ? Css.bcBlue600.$ : Css.bcGray300.$),
+  };
+}
+
+const stepperTabStyles = {
+  baseStyles: Css.df.aic.fg1.py1.prPx(12).plPx(24).sm.br0.oh.$,
```
bdow: **question:** Why `br0` and `oh`? I would assume we don't have a border-radius to begin with do we?

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+  hoverStyles: Css.bgGray100.$,
+  focusRingStyles: Css.bshFocus.$,
+  // Disabled always wins over both the state's and the collapsed border color.
+  disabledStyles: { ...Css.gray400.cursorNotAllowed.$, ...Css.bcGray300.$ },
```
bdow: **suggestion:** can we just combine these two truss definitions vs spreading them together?

src/components/StepperTabs/StepperTabs.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useState } from "react";
+import { Button } from "src/components/Button";
+import { StepperTabsStep } from "src/components/StepperTabs/StepperTabs";
+import { Css } from "src/Css";
+import { noop } from "src/utils";
+import { StepperTabs } from "./StepperTabs";
+
+export default {
+  component: StepperTabs,
+} as Meta;
```
bdow, line 14: **suggestion:** These examples are probably worth getting the mobile version of the screenshots too - wdyt?

src/components/StepperTabs/StepperTabs.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    // Then the first step shows as visited, inactive, and completed
+    expect(r.stepperTabs_tab_trade).toHaveTextContent("Trade Partners");
+    expect(r.stepperTabs_tab_check_0).toBeInTheDocument();
+    // And the second step shows as the active, not-yet-completed step
+    expect(r.stepperTabs_step_0).toHaveAttribute("aria-current", "false");
+    expect(r.stepperTabs_step_1).toHaveAttribute("aria-current", "true");
+    expect(r.query.stepperTabs_tab_check_1).not.toBeInTheDocument();
+    // And the third step shows as not visited
+    expect(r.stepperTabs_tab_send).toHaveTextContent("Send Email");
+    expect(r.query.stepperTabs_tab_check_2).not.toBeInTheDocument();
```
bdow: **suggestion:** I think the way we're testing the existence of the checkmark could trigger a false-positive. Like, if only the last "inactive" element had a check mark, all of these still would pass. We're not really testing where the checkmark is, but just that only one exists on the page. Other ideas for testing could be:
1. query the first stepper tab and look for the completed check within it. 
2. denote the tab as complete in some other way like via `data-completed=true`. 
FWIW, the existence of the check is also being validated via the StepperTab itself, so not sure if this check is even needed - maybe just ensuring its being passed through correctly

src/components/StepperTabs/StepperTabs.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  it("shows the check icon for a step that is both active and completed", async () => {
+    // Given the first step is both completed and current
+    const r = await render(<StepperTabs steps={makeSteps()} currentStep="trade" onChange={vi.fn()} />);
+    // Then it's marked current and still shows its check icon
+    expect(r.stepperTabs_step_0).toHaveAttribute("aria-current", "true");
+    expect(r.stepperTabs_tab_check_0).toBeInTheDocument();
```
bdow: This is the same thing as above - the check could be anywhere, not specifically in this tab.

src/components/StepperTabs/StepperTab.stories.tsx
abadami: :+1:

src/components/StepperTabs/StepperTab.tsx
abadami: I must've glossed over that, but yes it's mobile view only. I don wonder if there is consideration for having too many tabs but I think based on the style guidelines for the component that shouldn't ever happen.

src/components/StepperTabs/StepperTab.tsx
abadami: I'm going to confirm this with design really quick. I'm seeing 6px in Figma, but that does seem inconsistent.

src/components/StepperTabs/StepperTab.tsx
abadami: br0 was a leftover from a previous styling that had button styles applied to this, which can be removed now. oh is to make sure the text doesn't show when collapsed.

src/components/StepperTabs/StepperTab.test.tsx
abadami: I personally like the it.each approach since it feels succinct and I don't think it's that much more difficult to follow, but I will acknowledge it is still more difficult. I think for code consistency, I'm down to drop it.

src/components/StepperTabs/StepperTabs.stories.tsx
abadami: Good call, let's do it.

src/components/StepperTabs/StepperTab.test.tsx
abadami: Good point, I think this test was originally different and I forgot to go back through these tests. Will switch it up.

src/components/StepperTabs/StepperTabs.test.tsx
abadami: It may be a little redundant, but I think there's some value for making sure the three stepper tabs here are represented correctly, especially since that's one of the main functions StepperTabs has. I'll go with number 1

src/components/StepperTabs/StepperTab.tsx
bdow: I see where it says 6px border, but I'm nearly certain its really a 3px border in here - not 6. I think this is doing something odd.

bdow: I think its because its a "Vector" line vs being rendered as a border. I believe the vector lines render at twice the pixel density so it can look more crisp when designers are drawing in Figma. But if I had a real border to the component, then a 3px border covers the same space.
No 3px border:
Screenshot 2026-07-16 at 11 31 18 PM


With 3px border 
Screenshot 2026-07-16 at 11 30 30 PM

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+function getCollapsedStyles(active: boolean, completed: boolean): Properties {
+  return {
+    ...Css.cursor("default").hPx(0).py0.$,
+    ...(completed && !active ? Css.bcBlue600.$ : Css.bcGray300.$),
+  };
+}
+
+const stepperTabStyles = {
+  baseStyles: Css.df.aic.fg1.py1.prPx(12).plPx(24).sm.oh.$,
```
bdow: I'm not sure if I suggested removing this in this PR or another one, but looks like the `button` element will center text by default, so we do need to add a `text-align: left`. Sorry if I had asked to remove!

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return {
+    ...Css.bb.add("borderBottomWidth", `6px`).$,
+    ...color,
+  };
+}
+
+function getStateStyles(active: boolean, completed: boolean): Properties {
+  return {
+    ...Css.gray400.if(active || completed).blue700.if(active).smSb.$,
+    ...withBorderBottom(active || completed ? Css.bcBlue600.$ : Css.bcGray300.$),
```
bdow, line 67: **suggestion (non-blocking):** I think it could be really nice to add a transition of when the step completes and have the bar fill in the blue color from left to right.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return {
+    ...Css.cursor("default").hPx(0).py0.$,
+    ...(completed && !active ? Css.bcBlue600.$ : Css.bcGray300.$),
+  };
+}
+
+const stepperTabStyles = {
+  baseStyles: Css.df.aic.fg1.py1.prPx(12).plPx(24).sm.oh.$,
+  hoverStyles: Css.bgGray100.$,
+  focusRingStyles: Css.bshFocus.$,
```
bdow: **suggestion:** I think we need to add an `outline: none` in here or the base styles to avoid the default focus styles that come with the browser. Right now if you tab through the steps it'll look like this:

Image

src/components/StepperTabs/StepperTabs.tsx
```diff
@@ -0,0 +1,57 @@
...
+export type StepperTabsStep = {
+  label: string;
+  value: string;
+  /** Whether this step's required fields are filled out. Combined with `currentStep` to derive each tab's visual state. */
+  completed: boolean;
+  disabled?: boolean;
+};
+
+export type StepperTabsProps = {
+  steps: StepperTabsStep[];
```
bdow, line 9: **suggestion** Maybe instead of recreating a `StepperTabsStep` type here, we could just do a `Pick<StepperTabProps, "active" | "completed" | etc>[]`

bdow (approved):
Again giving the approval, but I think still a couple things in here to still address and would also be good to pass along the Storybook link to design so they can AT this while in PR. Otherwise, happy to have follow up issues/PRs just to keep this moving.

src/components/StepperTabs/StepperTab.tsx
abadami: Actually I think we still don't need text-align left. I think since this is using flex, it's overriding the normal button text alignment. I believe that's how that works, anyway, as it looks fine in storybook.

src/components/StepperTabs/StepperTabs.tsx
abadami: I think it may still benefit to have a type alias but rather than doing the recreation, we can just have `StepperTabsStep` be the `Pick<StepperTabProps, "active" | "complete" | ect>`. I think having to type out the Pick syntax every time could be a bit much for consumers. I'm open to renaming it to Step though, might make sense to do so.

bdow: Yea, that's fine. I just want to keep the underlying types in sync more easily.

src/components/StepperTabs/StepperTab.tsx
bdow: Screenshot 2026-07-17 at 12 31 09 PM
This is what I'm seeing. You can see it is inheriting the `text-align: center` from the button, and it causes some extra space there.

src/components/StepperTabs/StepperTab.tsx
abadami: I agree, that could be neat! I'll leave that as a follow-up.

src/components/StepperTabs/StepperTab.tsx
abadami: Oh, huh, right. Weird that it seems to only do that at a specific width it feels. Well, either way, I'll add it back in

PR #1325, opened 2026-07-13
fix: Update Mobile Nav and Side Nav per design

bdow:


src/components/NavLinks/NavLink.tsx
```diff
@@ -44,14 +51,16 @@ export function NavLink(props: NavLinkProps) {
...
     variant,
     icon,
+    nested = false,
+    __storyState,
     ...otherProps
   } = props;
   const asLink = typeof onClick === "string";
   const isIconOnly = !!iconOnly && !!icon;
-  const labelContent = isIconOnly ? <span css={Css.visuallyHidden.$}>{label}</span> : label;
+  const labelContent = isIconOnly ? <VisuallyHidden>{label}</VisuallyHidden> : label;
```
abadami, line 60: **thought**: I wonder if there would be a way to standardize this kind of `VisuallyHidden` labeling for icons, like maybe including it as an optional prop for icons itself, making a small component for "labeled icons", or something like that. wdyt?

src/components/NavLinks/NavLink.stories.tsx
abadami, line 1: **praise**: Love how much cleaner this story file is now.

src/components/NavLinks/NavLink.tsx
bdow: Good question - I only changed this because TS was complaining about something. We already support passing a `label` prop to the IconButton component - but maybe we're not using that here because these are all "NavLinks" under the hood. We could add this into NavLink instead.

abadami: If you're talking about using IconButton to handle this situation instead, that could make sense. Though, I think considering NavLink uses `getButtonOrLink`, it might be weird in the link situations. 

I was talking more about the link content (in hindsight, should've put the comment there instead) where we have a pattern emerging of having an Icon with some sort of label content, but in smaller screens it becomes an icon only with the label wrapped in `VisuallyHidden`, so I was thinking we could maybe just wrap up that behavior in its own component for reuse.

Regardless, probably doesn't have to block this PR, just something to think about really.

abadami (approved):

PR #1326, opened 2026-07-14
feat: Jump Link Component

abadami:
## Jump Link Component

Adds in the component for representing "scroll to" jump links.

### DIFF
--- a/src/components/JumpLink.stories.tsx
+++ b/src/components/JumpLink.stories.tsx
@@ -47,7 +47,7 @@ export function BaseStates() {
 /** Confirms wrapping to 2 lines with an unbroken bar, and clamping/ellipsis beyond that. */
 export function Wrapping() {
   return (
-    <div css={Css.df.fdc.gap2.$}>
+    <div css={Css.df.fdc.gap2.wPx(180).$}>
       <JumpLink active label="A jump link that wraps to two lines" onClick={() => {}} />
       <JumpLink active={false} label="Line Items Long Section Name That Should Truncate" onClick={() => {}} />
     </div>
@@ -57,7 +57,7 @@ export function Wrapping() {
 /** Plain vertical stack usage — no grouping/wrapper component needed. */
 export function ListUsage() {
   return (
-    <div css={Css.df.fdc.gap1.$}>
+    <div css={Css.df.fdc.gap1.wPx(180).$}>
       <JumpLink active label="Jump Link 1" onClick={() => {}} />
       <JumpLink active={false} label="Jump Link 2" onClick={() => {}} />
       <JumpLink active={false} label="Jump Link 3" onClick={() => {}} />
--- a/src/components/JumpLink.stories.tsx
+++ b/src/components/JumpLink.stories.tsx
@@ -47,7 +47,7 @@ export function BaseStates() {
 /** Confirms wrapping to 2 lines with an unbroken bar, and clamping/ellipsis beyond that. */
 export function Wrapping() {
   return (
-    <div css={Css.df.fdc.gap2.wPx(180).$}>
+    <div css={Css.df.fdc.gap2.wPx(200).$}>
       <JumpLink active label="A jump link that wraps to two lines" href="#jump-link" />
       <JumpLink active={false} label="Line Items Long Section Name That Should Truncate" href="#jump-link" />
     </div>
@@ -57,7 +57,7 @@ export function Wrapping() {
 /** Plain vertical stack usage — no grouping/wrapper component needed. Sections below demonstrate the smooth scroll-to behavior. */
 export function ListUsage() {
   return (
-    <div css={Css.df.gap2.$}>
+    <div css={Css.df.gap2.hPx(2000).$}>
       <div css={Css.df.fdc.gap1.wPx(180).$}>
         <JumpLink active label="Jump Link 1" href="#section-1" />
         <JumpLink active={false} label="Jump Link 2" href="#section-2" />
@@ -65,7 +65,7 @@ export function ListUsage() {
       </div>
       <div>
         {["section-1", "section-2", "section-3"].map((id) => (
-          <div key={id} id={id} css={Css.hPx(400).df.aic.jcc.bt.bcGray200.$}>
+          <div key={id} id={id} css={Css.hPx(400).df.bt.bcGray200.$}>
             <h2>{id}</h2>
           </div>
         ))}
--- a/src/components/JumpLink.stories.tsx
+++ b/src/components/JumpLink.stories.tsx
@@ -4,10 +4,6 @@ import { Css } from "src/Css";
 
 export default {
   component: JumpLink,
-  args: {
-    label: "Jump Link",
-    href: "#jump-link",
-  },
 } as Meta<typeof JumpLink>;
 
 /** Replicates the Figma "Jump-Link States" grid: Active/Inactive columns x Default/Hover/Pressed/Focus/Disabled rows. */
--- a/src/components/JumpLink.stories.tsx
+++ b/src/components/JumpLink.stories.tsx
@@ -1,0 +1,66 @@
+import { Meta } from "@storybook/react-vite";
+import { JumpLink, JumpLinkProps } from "src/components/JumpLink";
+import { Css } from "src/Css";
+
+export default {
+  component: JumpLink,
+  args: {
+    label: "Jump Link",
+    onClick: () => {},
+  },
+} as Meta<typeof JumpLink>;
+
+/** Replicates the Figma "Jump-Link States" grid: Active/Inactive columns x Default/Hover/Pressed/Focus/Disabled rows. */
+export function BaseStates() {
+  const rows: Array<{ name: string; storyState?: JumpLinkProps["__storyState"]; disabled?: boolean }> = [
+    { name: "Default" },
+    { name: "Hover", storyState: { hovered: true } },
+    { name: "Focus", storyState: { focusVisible: true } },
+    { name: "Disabled", disabled: true },
+  ];
+
+  return (
+    <div css={Css.df.gap4.$}>
+      <div css={Css.df.fdc.gap2.$}>
+        <h2>Inactive</h2>
+        {rows.map(({ name, storyState, disabled }) => (
+          <JumpLink
+            key={name}
+            active={false}
+            label={name}
+            onClick={() => {}}
+            __storyState={storyState}
+            disabled={disabled}
+          />
+        ))}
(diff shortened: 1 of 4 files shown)


src/components/JumpLink.tsx
```diff
@@ -0,0 +1,62 @@
+import { PressEvent } from "@react-types/shared";
+import { ReactNode, useMemo, useRef } from "react";
+import { mergeProps, useButton, useFocusRing, useHover } from "react-aria";
+import { Css, Properties, Tokens } from "src/Css";
+import { useTestIds } from "src/utils/useTestIds";
+
+export type JumpLinkProps = {
+  label: ReactNode;
+  active: boolean;
+  onClick: (e: PressEvent) => void;
```
bdow: **thought:** My assumption would be that JumpLinks would be targeting an anchor element on the page - For example
```
<a href="#details-section">Details</a>
```
We could set:
```
html {
  scroll-behavior: smooth;
}
```
if we don't already, and have that be a nice smooth scrolling from section to section.

src/components/JumpLink.tsx
```diff
@@ -0,0 +1,62 @@
...
+      {...tid}
+      css={{
+        ...baseStyles,
+        ...(active && activeStyles),
+        ...(isHovered && !isDisabled && hoverStyles),
+        ...(isFocusVisible && focusStyles),
+        ...(isDisabled && disabledStyles),
+      }}
+    >
+      <span css={Css.wPx(140).lineClamp2.$}>{label}</span>
```
bdow: **suggestion:** I think the width restriction should live as part of the layouts, not part of the component. The component should take up whatever width is available to it. Testing the proper truncation can be done by setting width restrictions in the storybook examples.

src/components/JumpLink.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export function getJumpLinkStyles(): {
+  baseStyles: Properties;
+  activeStyles: Properties;
+  hoverStyles: Properties;
+  focusStyles: Properties;
+  disabledStyles: Properties;
+} {
+  return {
+    baseStyles: Css.wPx(184).tal.sm.cursorPointer.px3.py1.color(Tokens.TextLinkDefault).lhPx(24).$,
```
bdow: **suggestion:** Same thing here - we shouldn't need to set a width restriction here. 
Also, I'd assume `text-align: left`, `cursor: pointer;` would be default, so probably can be removed.

This `lhPx(24)` also seemed odd, so I just looked at the designs - this should be using the `md` font styles, which result in a 16px font size with 24px line height. So just switch out `sm` to `md` and you should be good to go without needing the extra `lhPx` declaration.

src/components/JumpLink.tsx
```diff
@@ -0,0 +1,62 @@
...
+    activeStyles: Css.smSb.bl.add("borderLeftWidth", "3px").bcBlue600.lhPx(24).$,
```
bdow: **request:** Same here regarding `smSb` and `lhPx` - just use `mdSb`.

bdow: **thought/suggestion:** This border turning on and off is going to cause things to jump and the stacked Jump Links to look off a bit. I think you might want to set a transparent border on the baseSTyles and just change the border color on active.

src/components/JumpLink.tsx
bdow: **thought/question:** Could we have put the lineclamp on the button element?

abadami: I agree with that assessment. I don't think we need to force a specific width in hindsight.

src/components/JumpLink.tsx
abadami: Yeah that makes a lot more sense. For the scroll behavior, that's something that has to be changed on an app by app basis? I don't think beam has a universal css file for stylings on the html tag. I wonder if that's worth enforcing somehow.

src/components/JumpLink.tsx
abadami: Yeah I think I misread the design there. For some reason I thought the font size was smaller still, which in hindsight if I thought that I should've commented on it 🤷 .

src/components/JumpLink.tsx
abadami: Good call

src/components/JumpLink.tsx
bdow: We should be able to do that via beam's css reset. Or we could take the anchor - look for the id on the page and do a `window.scrollTo(document.getElementById(targeId), { behavior: "smooth" })`

abadami: I guess with that in mind do we have a reason for any scroll behavior to not be smooth? I think if we agree from a design perspective that it's always nicer for that smoother transition, we should just opt to make it universal via the css reset.

bdow: the smooth scrolling doesn't work well with virtualized content as things are being added/removed from screen on scroll. It can potentially short circuit the smooth scrolling behavior.

src/components/JumpLink.tsx
abadami: That's...a fair point. Makes more sense on the container too.

abadami: Ok when I switched things to use an anchor tag, putting the line clamp on that tag has weird behavior. I have no clue as to why, but for some reason it doesn't like the anchor tag.

src/components/JumpLink.stories.tsx
```diff
@@ -0,0 +1,75 @@
+import { Meta } from "@storybook/react-vite";
+import { JumpLink, JumpLinkProps } from "src/components/JumpLink";
+import { Css } from "src/Css";
+
+export default {
+  component: JumpLink,
+  args: {
+    label: "Jump Link",
+    href: "#jump-link",
+  },
```
bdow: **question:** I'm not noticing where these args are being used. If we're not using them, can we remove?

src/components/JumpLink.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  it("scrolls to the target element when clicked", async () => {
+    const r = await render(
+      <div>
+        <JumpLink active={false} label="Section One" href="#section-one" data-testid="jumpLink" />
+        <div id="section-one" />
+      </div>,
+    );
+    click(r.jumpLink);
+    expect(window.scrollTo).toHaveBeenCalledTimes(1);
+    expect(window.scrollTo).toHaveBeenCalledWith({ top: expect.any(Number), behavior: "smooth" });
```
bdow, line 30: **praise:** Nice! I've never used or seen `expect.any(Number)` that I can recall. Good use of it here.

src/components/JumpLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+  disabled?: boolean;
+  /** Storybook-only visual state overrides for snapshotting pseudo-interactions. */
+  __storyState?: { hovered?: boolean; focusVisible?: boolean };
+};
+
+/** A link that smooth-scrolls the page to a section, e.g. for "on this page" tables of contents. */
+export function JumpLink(props: JumpLinkProps) {
+  const { label, active, href, disabled, __storyState, ...otherProps } = props;
+  const ref = useRef<HTMLAnchorElement>(null);
+  const { linkProps } = useLink({ isDisabled: disabled, onClick: handleClick, href }, ref);
```
bdow, line 21: **praise:** Good use of `useLink`, I hadn't thought of why we'd ever need this as a hook before, but to be able to use the `disabled` via it is nice.

bdow, line 21: **thought:** I'm kind of surprised the `handleClick ` isn't throwing some sort of ReferenceError since its bing referenced before instantiation, but maybe fine because of hoisting 🤷?

src/components/JumpLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const isHovered = __storyState?.hovered ?? isHoveredFromEvents;
+  const isFocusVisible = __storyState?.focusVisible ?? isFocusVisibleFromEvents;
+  const tid = useTestIds(otherProps, "jumpLink");
+
+  function handleClick(e: MouseEvent<FocusableElement>) {
+    // Let modifier/non-primary clicks fall through to native anchor behavior (open in new tab, etc).
+    if (e.metaKey || e.ctrlKey || e.shiftKey || e.altKey || e.button !== 0) return;
+    e.preventDefault();
+    const target = document.getElementById(href.replace(/^#/, ""));
+    if (target) {
```
bdow, line 33: **thought:** When we build these jump links into the layouts, we'll want to ensure this is aware of any sticky elements (like the page header) and offset the scrollTo `top` property as necessary.

src/components/JumpLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+      ref={ref}
+      {...mergeProps(linkProps, focusProps, hoverProps)}
+      {...tid}
+      css={{
+        ...jumpLinkStyles.baseStyles,
+        ...(active && jumpLinkStyles.activeStyles),
+        ...(isHovered && !disabled && jumpLinkStyles.hoverStyles),
+        ...(isFocusVisible && jumpLinkStyles.focusStyles),
+        ...(disabled && jumpLinkStyles.disabledStyles),
+        ...(disabled && active && jumpLinkStyles.disabledActiveStyles),
```
bdow, line 49: **question:** Is there a real use case for this this state - of being active and disabled? If a step is disabled, I'd assume we'd never be able to get to it.

bdow (approved):
Looks good. Just a few questions and thoughts. Happy to include any changes in this PR or as a follow up.

src/components/JumpLink.tsx
abadami: Yeah it's hoisted. I wanted to keep the group of consts all together. I'm not too strict on that code style though so if you feel strongly about having things declared before used I'm down for adjusting that. I just like the loose organization of variables / hook usages -> functions -> useEffects (if we have them) -> render in a component

src/components/JumpLink.tsx
abadami: Good to keep in mind!

src/components/JumpLink.tsx
abadami: Design does have it spec'd out, but realistically probably not. I think it's good to keep just in case though. May as well have it covered.

abadami: Although actually now that I think about it if we want to keep a strict API we probably shouldn't. I'll say we keep it since design spec'd it in, but check to see if this will ever be represented anywhere and if not, we can just get rid of it.

PR #1331, opened 2026-07-20
feat: Support table styles as part of GridTableLayout

bdow:


abadami (approved):
Lookin' good. No real notes on this one.

PR #1332, opened 2026-07-20
feat: Update default text color and storybook background

bdow:
Defaults the body text color to our Gray900 in CSS Reset - this was a miss from the beginning.

Our primary consuming app (internal-frontend) is moving toward a white background by default, so updating our storybook default background to match

### DIFF
--- a/src/css/CssReset.css
+++ b/src/css/CssReset.css
@@ -387,6 +387,7 @@ Add the correct display in Chrome and Safari.
   body {
     font-family: inherit;
     line-height: inherit;
+    color: var(--b-on-surface);
   }
 
   /**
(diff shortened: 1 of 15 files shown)


src/css/CssReset.css
```diff
@@ -387,6 +387,7 @@ Add the correct display in Chrome and Safari.
   body {
     font-family: inherit;
     line-height: inherit;
+    color: var(--b-on-surface);
```
stephenh, line 390: Very nit but personally a file like `CssReset.css` seems like something 100% app agnostic that gets copy/pasted around with zero changes, and this is starting to have app-specific styling/theme in it -- which is fine, just fwiw.

stephenh (approved):

src/css/CssReset.css
bdow: I agree, but counterpoint:
The tokens used here will/can be changed to fit whatever theme properly.

PR #1333, opened 2026-07-20
feat: Remove built in Header component from GridTableLayout

bdow:
Moves the actionMenu to the GridTableLayoutActions area, where I believe it was originally intended to be. Updates stories and tests as needed

### DIFF
--- a/src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
+++ b/src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
@@ -1,6 +1,8 @@
 import { Meta } from "@storybook/react-vite";
 import { useCallback, useEffect, useMemo, useState } from "react";
+import { Button } from "src/components/Button";
 import { checkboxFilter, multiFilter } from "src/components/Filters";
+import { PageHeader } from "src/components/PageHeader";
 import { GridDataRow, SimpleHeaderAndData } from "src/components/Table";
 import {
   cardBadgeSlot,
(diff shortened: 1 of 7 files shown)


src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
```diff
@@ -43,26 +45,30 @@ export function GridTableLayout() {
   });
 
   return (
-    <TestProjectLayout pageTitle="Grid Table Layout">
```
abadami, line 46: **thought (non-blocking)**: I wonder if it's worth using the baked in Page Header Layout for this instead.

abadami: Actually since you add in specific code for using TestProjectLayout without the pageTitle here, I imagine there's a reason for it?

abadami (approved):
Just a non-blocking comment. Glad to remove the page header piece from this 🥳

src/components/Layout/GridTableLayout/GridTableLayout.stories.tsx
bdow: I had a similar thought. But these stories have the side nav and are not using the new layouts - so I didn't want to introduce those here. I'm actually hoping we can remove these stories at some point as they are more anti-pattern at the moment as we move towards everything is nested in the layout components

bdow: Actually, these are using the new layouts, how about that! I'm guessing you migrated them, because I don't recall this - unless Cursor/Claude went rouge in a good way here. But either way, I'll do update as a follow up. Thanks for calling out and having me do a double take here!

PR #1335, opened 2026-07-20
feat: Workflow Header

abadami:
### Workflow Header

Adds in the Workflow Header. This adjusts how headers work internally by providing a `BaseHeader` component to allow for different behaviors between headers while still following the base styles. If we only have these two headers, it might be a tad overkill but the abstraction is incredibly simple so I don't think it'll be too big of an issue down the road.

_note_: Workflow headers will have their CTAs appear on the bottom of the page in a sticky footer-like fashion in mobile. The plan is to have the future `WorkflowLayout` handle that piece so it can appropriately keep track of what it needs to keep track of w/ regards to heights and such.

### DIFF
--- a/src/components/Headers/PageHeader.tsx
+++ b/src/components/Headers/PageHeader.tsx
@@ -1,0 +1,16 @@
+import { BaseHeader, BaseHeaderProps } from "src/components/Headers/BaseHeader";
+import { RouteTabsProps, Tabs, TabsContentXss, TabsProps } from "src/components/Tabs";
+import { Only } from "src/Css";
+import { useTestIds } from "src/utils";
+
+export type PageHeaderProps<V extends string, X> = Omit<BaseHeaderProps, "bottomSlot"> & {
+  tabs?:
+    | Omit<TabsProps<V, X>, "contentXss" | "omitFullBleedPadding" | "includeBottomBorder">
+    | Omit<RouteTabsProps<V, X>, "contentXss" | "omitFullBleedPadding" | "includeBottomBorder">;
+};
+
+export function PageHeader<V extends string, X extends Only<TabsContentXss, X>>(props: PageHeaderProps<V, X>) {
+  const { tabs, ...otherProps } = props;
+  const tid = useTestIds(otherProps, "header");
+  return <BaseHeader {...otherProps} bottomSlot={tabs && <Tabs {...tabs} {...tid.tabs} />} />;
+}
--- a/src/components/Headers/PageHeader.tsx
+++ b/src/components/Headers/PageHeader.tsx
@@ -1,6 +1,6 @@
 import { BaseHeader, BaseHeaderProps } from "src/components/Headers/BaseHeader";
 import { RouteTabsProps, Tabs, TabsContentXss, TabsProps } from "src/components/Tabs";
-import { Only } from "src/Css";
+import { Css, Only } from "src/Css";
 import { useTestIds } from "src/utils";
 
 export type PageHeaderProps<V extends string, X> = Omit<BaseHeaderProps, "bottomSlot"> & {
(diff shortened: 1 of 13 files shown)


src/components/Headers/PageHeader.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const { tabs, ...otherProps } = props;
+  const tid = useTestIds(otherProps, "header");
+  return (
+    <BaseHeader
+      {...otherProps}
+      bottomSlot={
+        tabs && (
+          <div css={Css.px3.$}>
+            <Tabs {...tabs} {...tid.tabs} />
+          </div>
```
bdow, line 22: **thought/suggestion (non-blocking):** This is an existing behavior for the tabs area of the header, so more of a follow up I just remembered; I believe tabs include their own overflow scrolling left & right. When we add this `px3` around them, then it means they'll never fully bleed to the edge of the screen, and overflown tabs look like they could've utilized more space on small viewports. I'd like to do a follow up ticket/task once we get to a place with more tabs the improve the UX of that.

src/components/Headers/WorkflowHeader.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+function useSteps() {
+  const [currentStep, setCurrentStep] = useState(steps[0].value);
+  return { steps, currentStep, onChange: setCurrentStep };
+}
+
+export function Default() {
+  const stepperTabs = useSteps();
+  return <WorkflowHeader title="Test Title" stepperTabs={stepperTabs} />;
+}
```
bdow, line 31: **thought (non-blocking):** Seems like by default we'd be requiring some sort of actions, right? Something to think about as we continue to build this out with whatever CTA API we're leaning towards

src/components/Headers/WorkflowHeader.tsx
```diff
@@ -0,0 +1,30 @@
+import { Button, ButtonProps } from "src/components/Button";
+import { BaseHeader, BaseHeaderProps } from "src/components/Headers/BaseHeader";
+import { StepperTabs, StepperTabsProps } from "src/components/StepperTabs";
+import { Css } from "src/Css";
+import { useTestIds } from "src/utils";
+
+export type WorkflowHeaderProps = Omit<BaseHeaderProps, "rightSlot" | "bottomSlot"> & {
+  rightSlot?: ButtonProps[];
```
bdow, line 8: **thought (non-blocking):** This won't be needed - at least not in this way - as we continue through that CTA API process. Though, it does make me wonder if expositing the `rightSlot` to the consuming engineers is something we even want to do? We're going to be reserving that area for our CTAs....
makes me think the WorkflowLayout component will have a sort of interface where:
```
type workflowProps = {
  header: Omit<WorkFlowHeaderProps, "rightSlot">
```
I think the only time a WorkflowHeader is used is through that layout, so we might be able to accomplish the reserving of the "rightSlot" via that omission.

bdow (approved):
Looks great! I like the abstraction for consistency here 👏 👏 👏

src/components/Headers/WorkflowHeader.tsx
abadami: This I'm pretty torn on. Because on one hand, I think if we expose WorkflowHeader for consumers, then it should be able to be used independently and thus we should keep the `rightSlot`. On the other hand, the WorkflowHeader being able to be used independently can be seen as its own sort of escape hatch, which I think goes against the opinionated / stricter API design philosophy we're going for, so we should only have it accessible via a layout.

I think there's a good argument either way. It may make sense to limit it to being used only in a layout(s) currently. I think the only other layout this would be used for is the Focused Form layout purely for the mobile footer functionality, but we can talk about that when we get there. wdyt?

src/components/Headers/WorkflowHeader.stories.tsx
abadami: Yeah, initially the header essentially had the CTAs required, but now that layouts are more dictating when a header shows the actions or not (due to the footer), I opted to leave it optional for now. I think we can keep that cadence especially if we don't expose the WorkflowHeader publicly (see other comment on CTA stuff).

src/components/Headers/WorkflowHeader.tsx
bdow: I think we avoid the escape hatch until we know we need it. If someone used this workflow header independently, then they'll run into issues on mobile where the actions are crowding the header. I think it'd end up being more of an anti-pattern than an escape hatch to be able to use these headers outside of their intended layouts.

abadami: In that case, I will go ahead and remove these headers from being consumed outside of layouts entirely now. I think establishing that early on is good. If we change our mind, we can always reverse it. I don't think PageHeader is being used independently at all currently so shouldn't be a big deal (and hopefully if it is my Beam Bump action catches it.

PR #1336, opened 2026-07-21
fix: keep ListBox height when filtering options (ON-285)

RoseAndresHB:
## Summary
- When filtering shrinks the `ListBox` option list, keep the established popover height so MultiSelect/Select scroll boxes no longer collapse ([ON-285](https://linear.app/homebound-team/issue/ON-285/bug-fe-or-be-option-assignment-filter-shrinks-option-list-scroll-box))
- Avoids a “latch peak height” approach that could thrash with multi-select chip `maxh(50%)` under load
- Adds regression coverage via `MultiSelectField.filterHeight.test.tsx` and a Storybook `FilterHeightRepro` story for visual checks

## Test plan
- [ ] `yarn test src/inputs/MultiSelectField.filterHeight.test.tsx src/inputs/MultiSelectField.test.tsx src/components/Filters/TreeFilter.test.tsx`
- [ ] Storybook → MultiSelectField → FilterHeightRepro: open Options, type `ZZZ Unique`, confirm listbox stays tall (~512px)
- [ ] After Beam release + IFE bump: Takeoff Manager → edit option assignment → filter Options → list does not collapse


src/inputs/MultiSelectField.stories.tsx
```diff
@@ -219,6 +219,30 @@ export function Loading() {
...
+    id: String(i + 1),
+    name:
+      i === 99
+        ? "ZZZ Unique Option For Filter Repro"
+        : `Option ${String(i + 1).padStart(3, "0")} — Plan Package Selection`,
+  }));
+  return (
+    <div css={Css.df.fdc.gap2.p2.maxw("480px").$}>
+      <p css={Css.sm.gray700.$}>
+        Open Options, type <code>ZZZ Unique</code>, and confirm the listbox stays tall (ON-285).
```
RoseAndresHB, line 234: @bdow not sure your/our preference for adding regression stories and/or text tying them to specific tickets. Fine w/ ripping this out if it isn't necessary/desired

src/inputs/internal/ListBox.tsx
```diff
@@ -59,17 +59,29 @@ export function ListBox<O, V extends AriaKey>(props: ListBoxProps<O, V>) {
...
+    // it is the persistent section with a single item and account for that height.
+    const next = Math.min(
+      popoverMaxHeight,
+      hasSections ? height + persistentItemHeight + sectionSeparatorHeight : height,
     );
+
+    setPopoverHeight((prev) => {
+      // When filtering shrinks the option list, keep the established popover height
+      // instead of collapsing the scroll box. Avoids a chip maxh(50%) ↔ height feedback loop
+      // that a pure "latch peak" approach can cause under load.
```
stephenh: What does this last "latch peak" sentence really mean? I feel like it's the LLM saying something it understands that I don't 😅

src/inputs/internal/ListBox.tsx
```diff
@@ -59,17 +59,29 @@ export function ListBox<O, V extends AriaKey>(props: ListBoxProps<O, V>) {
   // Using a ref, this itself should not trigger a rerender, only `popoverHeight` changes will trigger a rerender.
   const virtuosoListHeight = useRef<number>(0);
   const onListHeightChange = (listHeight: number) => {
+    const prevListHeight = virtuosoListHeight.current;
     virtuosoListHeight.current = listHeight;
     // The "listHeight" is only the list of options.
     // For multiple selects we need to also account for the height of the list of currently selected elements when re-evaluating.
     // Using `offsetHeight` to account for borders
     const height = (selectedList.current?.offsetHeight || 0) + listHeight;
```
stephenh, line 67: I kinda feel like this `selectedList` height would be this yellow bar:

Image

But it seems like Tyler's screenshot that's coming back wrong:

Image

src/inputs/internal/ListBox.tsx
RoseAndresHB: a previous approach "latched" the greatest height used for the popover, which ran into problems with a TreeFilter test timing out and some autocalculating height stuff. I initially was going to rip it out, but ended up leaving the verbiage in there thinking that it might be helpful to explain why we went this direction instead of that one; seems demonstrably more confusing than helpful though, so maybe should just remove 😅

src/inputs/internal/ListBox.tsx
RoseAndresHB: @stephenh i agree with the first part of this comment but am not quite grokking the second half.

stephenh: In Tyler's screenshot, we see that he has a few chips selected the `CAB... something` -- afaiu those chips are the `selectedList.current.offsetHeight` is supposed to be calculating (what in my screenshot is the orange bar).

So my musing is that maybe the bug is that `selectedList.current.offsetHeight` is returning zero when we didn't think it would/should.

RoseAndresHB: The bug that we found is that we set an upper bound on our popover's height, but do not have a corresponding lower bound set. So when our search filter wouldn't return any results AND we already have some selected, our selections already take up most/all of the minimum size of the popover.

We're accounting for this by only re-sizing the popover when it would need to grow in height, and leaving it the same size when fewer options would be displayed as a result of filtering 🙂

stephenh (approved):

src/inputs/internal/ListBox.tsx
RoseAndresHB: simplified this comment

PR #1338, opened 2026-07-22
feat: Workflow Layout

abadami:


src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -20,14 +21,24 @@ import { WorkflowActions, WorkflowActionsProps } from "./WorkflowActions";
 
 export type WorkflowHeaderConfig = Pick<BaseHeaderProps, "title" | "documentTitleSuffix" | "breadcrumbs"> &
```
abadami: I wonder if we should actually just separate this out from being a "WorkflowHeader" prop, especially since we aren't exposing "WorkflowHeader" anymore

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -69,22 +80,31 @@ export function WorkflowLayout(props: WorkflowLayoutProps) {
   const cssVars: Record<string, string> | undefined =
     headerHeight > 0 ? { [beamPageHeaderLayoutHeightVar]: `${headerHeight}px` } : undefined;
 
-  const { steps, currentStep, onChange } = stepperTabs;
-  const currentIndex = steps.findIndex((step) => step.value === currentStep);
+  const { currentStep, onChange } = stepperTabs;
+
+  const tabSteps = steps.map((step) => ({
```
abadami: The one flaw with this is that if there is ever matching labels, this will likely cause some problems. I highly highly doubt that will ever happen, but good to note

src/utils/sbComponents.tsx
```diff
@@ -83,14 +83,8 @@ export function TestProjectLayout({ pageTitle, children }: { pageTitle?: string;
...
-export function TestWorkflowProjectLayout({
-  workflowHeader,
-  children,
-}: {
-  workflowHeader: WorkflowLayoutProps["workflowHeader"];
-  children: ReactNode;
-}) {
-  return <WorkflowLayout workflowHeader={workflowHeader}>{children}</WorkflowLayout>;
+export function TestWorkflowProjectLayout(props: WorkflowLayoutProps) {
+  return <WorkflowLayout {...props} />;
```
bdow: **thought/suggestion:** Doesn't look like this wrapper is providing anything. Maybe we can remove it?

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -20,14 +21,24 @@ import { WorkflowActions, WorkflowActionsProps } from "./WorkflowActions";
...
+  label: string;
+  /** Drives the tab's completed checkmark, and gates the Continue/Complete CTA when this is the active step. */
+  completed: boolean;
+  /** Whether this step's tab can be clicked/jumped to. Independent of `isValid`. */
+  disabled?: boolean;
+  /** Rendered as the page body while this is the active step. */
+  content: ReactNode;
+};
+
 export type WorkflowLayoutProps = {
```
bdow: **suggestion:** Per discussion, let's update WorkflowLayoutProps to be shaped like:
```
type WorkflowLayoutProps = 
  Pick<BaseHeaderProps, "title" | "documentTitleSuffix" | "breadcrumbs"> & 
  Pick<WorkflowActionsProps, "onCancel" | "completeLabel" | "onComplete" | "onSaveAndExit"> &
  steps: Omit<StepperTabsProps, "collapsed"> & { content: ReactNode }
}
```
Not sure if `steps` can be that way exactly, but would be nice to just expand upon steps here and not require the user to pass in two separate props for steps.

src/forms/WorkflowLayoutFormApp.tsx
```diff
@@ -0,0 +1,198 @@
...
+        const steps: WorkflowLayoutStep[] = [
+          {
+            label: "Author Details",
+            completed: step1Valid,
+            content: <AuthorDetails formState={formState} />,
+          },
+          {
+            label: "Books",
+            completed: step2Valid,
+            disabled: !step1Valid,
```
bdow, line 48: **suggestion:** Let's create a ticket to see about how we can better scope this to reduce re-renders on field changes

src/forms/WorkflowLayoutFormApp.tsx
```diff
@@ -0,0 +1,198 @@
...
+}
+
+type Row = SimpleHeaderAndData<FormValue["books"]["rows"][number]>;
+
+function createColumns(formState: FormValue): GridColumn<Row>[] {
+  return [
+    { header: "#", data: ({ id }) => <span>{id.value}</span> },
+    {
+      header: "Title",
+      data: ({ title }) => <BoundTextField label="" compact field={title} />,
```
bdow: **nit:** we should always supply a label for accessibility - we have a `labelStyle: hidden`

src/layouts/WorkflowLayout/WorkflowActions.tsx
```diff
@@ -38,9 +50,9 @@ export function WorkflowActions(props: WorkflowActionsProps) {
         <Button label="Cancel" variant="quaternary" onClick={onCancel} />
         {onSaveAndExit && <Button label="Save & Exit" variant="secondary" onClick={onSaveAndExit} />}
         {isLastStep ? (
-          <Button label={completeLabel} variant="primary" onClick={onComplete} />
+          <Button label={completeLabel} variant="primary" onClick={onComplete} disabled={primaryDisabled} />
```
bdow, line 53: **thought** Maybe we can somehow use the observer and know which field states trigger this disabled property.

src/layouts/WorkflowLayout/WorkflowLayout.stories.tsx
```diff
@@ -17,7 +16,7 @@ export default {
 } satisfies Meta;
 
 export function Default() {
-  const [currentStep, setCurrentStep] = useState("trade");
+  const [currentStep, setCurrentStep] = useState("tradePartners");
```
bdow: **suggestion:** Per our discussion, we are thinking about having this state be controlled by the workflow layout. And a user would just pass in their current step.

src/layouts/WorkflowLayout/WorkflowActions.tsx
abadami: Will be part of Observer exploration ticket follow-up ticket as discussed

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -18,18 +19,20 @@ import { useBannerAndNavbarHeight } from "../useBannerAndNavbarHeight";
...
+  content: ReactNode;
+};
+
+export type WorkflowLayoutProps = Pick<BaseHeaderProps, "title" | "documentTitleSuffix" | "breadcrumbs"> &
   Pick<WorkflowActionsProps, "onCancel" | "completeLabel" | "onComplete" | "onSaveAndExit"> & {
-    stepperTabs: StepperTabsProps;
+    /** The workflow's steps; the active step's `content` is the body, and it drives the header's tab strip. */
+    steps: WorkflowLayoutStep[];
+    /** The step shown initially (matched against `defaultTestId(step.label)`); defaults to the first step. Uncontrolled — the layout owns step navigation from here. */
+    defaultStep?: string;
```
bdow, line 34: **suggestion:** We could make this more strict to ensure that the `defaultStep` is actually a value of the steps being passed in, maybe? Would be a nice API improvement in t he event someone things they can do something like `defaultStep: "2"` or something. This would require a generic and happy to do it as a follow up to the API improvements.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -43,8 +46,9 @@ export type WorkflowLayoutProps = {
  * of the public API.
  */
 export function WorkflowLayout(props: WorkflowLayoutProps) {
-  const { children } = props;
-  const { stepperTabs, onCancel, completeLabel, onComplete, onSaveAndExit, ...headerProps } = props.workflowHeader;
+  const { steps, defaultStep, onCancel, completeLabel, onComplete, onSaveAndExit, ...headerProps } = props;
+  const tabSteps = steps.map((step) => ({ ...step, value: defaultTestId(step.label) }));
```
bdow, line 51: I guess since we don't require `value` then it'd be hard to make the `defaultStep` be a strict type.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -43,8 +46,9 @@ export type WorkflowLayoutProps = {
...
+  const [currentStep, setCurrentStep] = useState(defaultStep ?? tabSteps[0]?.value);
```
bdow: **suggestion:** if `defaultStep` isn't actually a step, then we'll run into a weird issue here. Probably worth validating this somehow via a `getInitialStep(steps, defaultStep)` which would ensure if the user passes a `defaultStep` then it actually exists, otherwise default to the first step.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -69,22 +73,23 @@ export function WorkflowLayout(props: WorkflowLayoutProps) {
   const cssVars: Record<string, string> | undefined =
     headerHeight > 0 ? { [beamPageHeaderLayoutHeightVar]: `${headerHeight}px` } : undefined;
 
-  const { steps, currentStep, onChange } = stepperTabs;
-  const currentIndex = steps.findIndex((step) => step.value === currentStep);
+  const currentIndex = tabSteps.findIndex((step) => step.value === currentStep);
```
bdow: Per earlier discussions, this would result in a `-1` if we don't guard against it.

bdow (approved):
Would like to add in the proper guards to ensure currentStep is a valid step, otherwise good to go and do any other changes as follow ups.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
abadami: Good call

src/layouts/WorkflowLayout/WorkflowLayout.tsx
abadami: Yeah that's what will make it tricky, but I think it's worth looking into. I made a ticket for it.

bdow: Only way to keep the API slim and enforce the `value` would be to do something like a factory that ensures all steps include a `value` prop ahead of time.

PR #1339, opened 2026-07-22
fix: Restore table row hover color

bdow:
The table row hover was accidentally moved in the truss migration. Adding this back


stephenh (approved):
Thanks!

PR #1340, opened 2026-07-22
feat: CTA API & WorkflowHeaderLayout

abadami:
### CTA API & WorkflowHeaderLayout

This PR introduces the CTA API, which rather than working that into the Header...since Headers won't be publically exposed, this was going to be attached to the Layout instead. Thus, this PR will be the PR to introduce `WorkflowHeaderLayout`. We are going to be strict on allowed CTAs to start for Workflows, and thus are enforcing "back", "continue", "save & exit", "create/save" and "cancel" CTAs.

This layout also introduces moving CTAs to a footer in mobile, which requires some CSS vars to make sure things like the "scroll to top" button off screen.


src/components/Headers/WorkflowHeader.stories.tsx
```diff
@@ -35,10 +26,12 @@ export function WithRightSlotButtons() {
...
       title="Test Title"
-      rightSlot={[
-        { label: "Save Draft", variant: "secondary", onClick: action("save clicked") },
-        { label: "Continue", variant: "primary", onClick: action("continue clicked") },
-      ]}
+      rightSlot={
+        <div css={Css.df.aic.gap1.$}>
+          <Button label="Save Draft" variant="secondary" onClick={action("save clicked")} />
+          <Button label="Continue" variant="primary" onClick={action("continue clicked")} />
+        </div>
```
bdow: **nit/suggestion:** I think it'd be nice if `<div css={Css.df.aic.gap1.$}>` was rendered as part of a helper method that is used between both the stories and actual component to prevent drift.

src/components/Headers/WorkflowHeader.stories.tsx
```diff
@@ -35,10 +26,12 @@ export function WithRightSlotButtons() {
...
     <WorkflowHeader
       title="Test Title"
-      rightSlot={[
-        { label: "Save Draft", variant: "secondary", onClick: action("save clicked") },
-        { label: "Continue", variant: "primary", onClick: action("continue clicked") },
-      ]}
+      rightSlot={
+        <div css={Css.df.aic.gap1.$}>
+          <Button label="Save Draft" variant="secondary" onClick={action("save clicked")} />
+          <Button label="Continue" variant="primary" onClick={action("continue clicked")} />
```
bdow: **suggestion:** Haven't read any further than this, but I'm guessing/assuming we have a strict API about how buttons are rendered. Can we utilize that rendering functionality here to also prevent drift between stories and the real results?

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -70,13 +70,17 @@ function getStateStyles(active: boolean, completed: boolean): Properties {
...
   };
 }
 
 const stepperTabStyles = {
-  baseStyles: Css.df.aic.fg1.py1.prPx(12).plPx(24).sm.oh.tal.$,
+  baseStyles: Css.df.aic.fg1.py1
+    .prPx(12)
+    .plPx(24)
+    .sm.oh.tal.maxhPx(48)
+    .add("transition", "max-height 200ms cubic-bezier(0.4, 0, 0.2, 1), padding 200ms cubic-bezier(0.4, 0, 0.2, 1)").$,
```
bdow: **suggestion:** I'd like to keep our transitions as similar as possible and utilize what is in the truss configuration that follows our design tokens.
Is padding necessary to transition here? I don't see a change in padding at all. collapsed styles have a `py0`, but doesn't appears we set a padding-top/bottom anywhere - just left & right. Could be missing something, though since I'm not seeing the full thing here.

src/utils/sbComponents.tsx
```diff
@@ -79,6 +85,24 @@ export function TestProjectLayout({ pageTitle, children }: { pageTitle?: string;
...
+}: {
+  workflowHeader: WorkflowHeaderLayoutProps["workflowHeader"];
+  children: ReactNode;
+}) {
+  return (
+    <NavbarLayout navbar={createNavbar()}>
+      <WorkflowHeaderLayout workflowHeader={workflowHeader}>{children}</WorkflowHeaderLayout>
+    </NavbarLayout>
+  );
+}
```
bdow: **request:** Workflow layouts are not expected to have the Navbar. That is intentional as we want a focused page where the user has limited options to exit the workflow/form pages. Can you please remove so stories are more inline with what we'll be building?

docs/layouts.md
```diff
@@ -4,23 +4,24 @@ This document is the **canonical contract** for structural page layouts in Beam.
...
+2. **Preserve nesting order** when all apply: **`EnvironmentBannerLayout` → `NavbarLayout` → `SideNavLayout` → `PageHeaderLayout`**. For step-based workflow pages, use **`WorkflowHeaderLayout`** as the innermost layout instead of `PageHeaderLayout` — the two are not nested together, since `WorkflowHeaderLayout` renders a different header component (`WorkflowHeader`) and owns its own chrome. Its body is arbitrary content, passed as `children`, the same way `GridTableLayout`/arbitrary content is `PageHeaderLayout`'s.
+3. **Layouts render Beam components, not arbitrary nodes** — Each layout owns its chrome and renders the real Beam component internally. Pass the component's props as a **nested object** (`environmentBanner`, `navbar`, `sideNav`, `pageHeader`/`workflowHeader`); pass page body content via **`children`**. The layouts handle the document-scroll coordination (sticky chrome, auto-hide, CSS-var offsets) for you.
```
bdow: This guidance is incorrect. Specifically:
> or step-based workflow pages, use **`WorkflowHeaderLayout`** as the innermost layout instead of `PageHeaderLayout`

My understanding is that WorkflowHeaderLayout is closer to a top level layout, more similar to NavbarLayout than PageHeaderLayout. It will get wrapped still within the EnvBannerLayout, but it should never be wrapped within a NavbarLayout or SideNavLayout.

And again here:
>(`environmentBanner`, `navbar`, `sideNav`, `pageHeader`/`workflowHeader`);

Insinuating that pageHeader and workflowHeader are interchangeable will provide incorrect guidance to the agents.

src/layouts/WorkflowHeaderLayout/WorkflowHeaderLayout.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+  trade: "Trade Partners",
+  draft: "Draft Email",
+  send: "Send Email",
+};
+
+function makeSteps(): StepperTabsStep[] {
+  return tabValues.map((value, i) => ({ value, label: tabLabels[value], completed: false, disabled: i > 0 }));
+}
+
+function StepContent({ title, heightPx = 0 }: { title: string; heightPx?: number }) {
```
bdow: **nit:** `heightPx` is kind of misleading here, or at least its not really setting the height that is passed in. Instead it just creates more height for the page somewhat arbitrarily. Like, passing in `2000` px for height, results in an actual height of this element to be **1,896px**. 
If the end result is really just number of rows, I'd lean towards passing `numRows` as the prop.

src/layouts/WorkflowHeaderLayout/WorkflowHeaderLayout.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { WorkflowHeaderLayout } from "./WorkflowHeaderLayout";
+
+describe("WorkflowHeaderLayout", () => {
+  it("renders the header and body children", async () => {
+    // Given a WorkflowHeaderLayout with body content
+    const r = await render(<TestWrapper>{<div data-testid="body">Body content</div>}</TestWrapper>, withRouter());
+
+    // Then the header and body both render
+    expect(r.workflowHeaderLayout_header).toHaveTextContent("Test Workflow");
+    expect(r.body).toBeInTheDocument();
```
bdow: **suggestion:** This test is asserting the header is "Test Workflow", but that is never explicitly set in this test. My opinion in tests is that we should be asserting against something that the test doesn't not explicitly define. 
While `TestWrapper` does help reduce the lines of code, I think it is doing us a disservice by abstracting too much and making this not as explicitly testable as it should be. Wrapping components in tests should only be necessary when we need to test against an internal state / use a `useState` in side of the `TestWrapper`, or something similar. 
This abstraction also prevents us from validating things like the `onCancel` is clicked, or `onComplete` is clicked, or the default `completeLabel` is "Save", etc. 
If you want to limit the lines of code in something like this, I'd prefer we have something like:
```
const baseProps: WorkflowLayoutProps = { 
  title: "Test Title",
  onComplete: () = {}, 
  stepperTabs: [...],
  ...
}
```
Then when you render, you can apply these base prop and override the specific props you are testing in the test via:
```
const r = await render(<WorkflowLayout {...baseProps} title="Custom Title">....</WorkflowLayout>);
------------------------------
const onComplete = vi.fn();
const r = await render(<WorkflowLayout {...baseProps} onComplete={onComplete}>....</WorkflowLayout>);
```
That'll give you more flexibility on being able to properly test each property

src/layouts/WorkflowHeaderLayout/WorkflowHeaderLayout.tsx
```diff
@@ -0,0 +1,214 @@
...
+      } else {
+        root.style.removeProperty(beamWorkflowLayoutFooterHeightVar);
+      }
+    };
+  }, [showFooter]);
+
+  const headerEl = (
+    <WorkflowHeader
+      {...headerProps}
+      rightSlot={isMobile ? undefined : <div css={Css.df.aic.gap1.$}>{buttons}</div>}
```
bdow: **thought:** To reduce the size of this component, what do you think of doing something like:
```
rightSlot={<WorkflowActions isFirstStep={} isLastStep={} onCancel={} />}
```
And passing any data it needs to properly build out the buttons. Then we can more easily reuse this area in stories and such.

src/layouts/WorkflowHeaderLayout/WorkflowHeaderLayout.tsx
```diff
@@ -0,0 +1,214 @@
...
+  );
+
+  return (
+    <DocumentScrollLayoutProvider>
+      <div css={Css.df.fdc.w100.$} style={cssVars} {...tid}>
+        {/* Spacer reserves height for the always-sticky header; also the geometry anchor for scroll-collapse. */}
+        <div ref={spacerRef} css={Css.fs0.w100.$} style={{ height: headerHeight }} {...tid.spacer}>
+          <div
+            ref={headerMetricsRef}
+            css={Css.fixed.left(headerLeft).w(headerWidth).z(zIndices.pageStickyHeader).$}
```
bdow: **thought:** At the moment, workflow layouts should always be `left0`. There'll never be nested within a SideNav. I don't think we need to do this `headerLeft`. Doesn't hurt, but also probably unnecessary.

src/layouts/WorkflowHeaderLayout/WorkflowHeaderLayout.tsx
```diff
@@ -0,0 +1,214 @@
...
+            style={{ top: outerTop } as CSSProperties}
```
bdow: **question:** Why do we need this casting? and could we have just adding this into the `css` definition above via `css=Css.top(outerTop).$`?

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -0,0 +1,214 @@
...
+    const root = document.documentElement;
+    const previous = root.style.getPropertyValue(beamWorkflowLayoutFooterHeightVar);
+    root.style.setProperty(beamWorkflowLayoutFooterHeightVar, showFooter ? `${mobileFooterHeightPx}px` : "0px");
+    return () => {
+      if (previous) {
+        root.style.setProperty(beamWorkflowLayoutFooterHeightVar, previous);
+      } else {
+        root.style.removeProperty(beamWorkflowLayoutFooterHeightVar);
+      }
+    };
```
bdow, line 105: **nit:** Let's see how other layouts are handling local variables and follow those.

src/layouts/WorkflowHeaderLayout/WorkflowHeaderLayout.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+        <div css={Css.df.fdc.fg1.mh0.w100.$} {...tid.body}>
+          {children}
+
+          {/* Spacer so body content isn't hidden behind the fixed mobile footer. */}
+          {showFooter && <div css={Css.fs0.w100.hPx(mobileFooterHeightPx).$} />}
+        </div>
+
+        {showFooter && (
+          <div
```
bdow: **suggestion:** Can we put the spacer as a sibling to the fixed element?

src/layouts/DocumentScrollToTopButton.tsx
```diff
@@ -36,8 +37,10 @@ export function DocumentScrollToTopButton({ viewportHeight }: DocumentScrollToTo
       aria-hidden={!visible}
       {...(!visible ? { inert: "" } : {})}
       css={{
-        ...Css.fixed.bottomPx(scrollToTopOffsetPx).rightPx(scrollToTopOffsetPx).z(zIndices.scrollToTop).df.jcfe
-          .transitionTransform.$,
+        ...Css.fixed
+          .bottom(`calc(${scrollToTopOffsetPx}px + var(${beamWorkflowLayoutFooterHeightVar}, 0px))`)
```
bdow: **suggestion:** Let's create a util method in `layoutVars` that does this calculation for `getFloatingBottomOffset(initialOffset)` returns basically this calculation.

src/components/StepperTabs/StepperTab.tsx
abadami: We do in fact need the padding transition, otherwise when it transitions, since it does change the padding, it will slightly snap before it does the transition.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
abadami: You made the right call with this being done because this particular instance doesn't follow the usual instance of descendants using the variable, but in this instance the scrollToTopButton is a sibling. We might be able to properly scope it, but I wonder if it's more of a hassle than it's worth.

src/layouts/WorkflowLayout/WorkflowActions.tsx
```diff
@@ -0,0 +1,65 @@
...
+      </div>
+      <div css={Css.df.aic.gap1.$}>
+        <Button label="Cancel" variant="quaternary" onClick={onCancel} {...tid.cancel} />
+        {canExitEarly && onSaveAndExit && (
+          <Button label="Save & Exit" variant="secondary" onClick={onSaveAndExit} {...tid.saveAndExit} />
+        )}
+        {isLastStep ? (
+          <Button label={completeLabel} variant="primary" onClick={onComplete} {...tid.complete} />
+        ) : (
+          <Button label="Continue" variant="primary" onClick={onContinue} {...tid.continue} />
```
bdow: **suggestion:** We don't need this `{...tid.continue}`

src/layouts/WorkflowLayout/WorkflowActions.tsx
```diff
@@ -0,0 +1,65 @@
...
+  isFirstStep: boolean;
+  isLastStep: boolean;
+  isMobile: boolean;
+  onBack: () => void;
+  /** Leaves the workflow without saving. Always shown. */
+  onCancel: (e: PressEvent) => void;
+  /** Whether Save & Exit is available on the current step. Default false. */
+  canExitEarly?: boolean;
+  /** Saves partial progress and exits. Used whenever canExitEarly is true. */
+  onSaveAndExit?: (e: PressEvent) => void | Promise<void>;
```
bdow: **suggestion:** Let's assume if `onSaveAndExit` exists, then `canExitEarly` is true. So we don't need a `canExitEarly` prop then.

src/layouts/WorkflowLayout/WorkflowLayout.test.tsx
bdow, line 1: Let's split out ownership of the tests between WorkflowActions and WorkflowLayout

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -0,0 +1,177 @@
...
+  /** Leaves the workflow without saving. Always shown. */
+  onCancel: (e: PressEvent) => void;
+  /** Label for the completion button shown on the last step. */
+  completeLabel: "Create" | "Save";
+  /** Called when the completion button is clicked. Only shown on the last step. */
+  onComplete: (e: PressEvent) => void | Promise<void>;
+  /** Whether Save & Exit is available on the current step. Default false. */
+  canExitEarly?: boolean;
+  /** Saves partial progress and exits. Used whenever canExitEarly is true. */
+  onSaveAndExit?: (e: PressEvent) => void | Promise<void>;
```
bdow: **suggestion:** Let's `Pick<>` these props from WorkflowActionProps

bdow (approved):

PR #1342, opened 2026-07-24
feat: Show single filter inline rather than under filter toggle

bdow:
In GridTableLayoutActions we were hiding all filters under a filters toggle. However, when there is only one action, it seemed unnecessary to hide on desktop views. This change allows for displaying the single fitler, or group by, control in place of the toggle button when in the desktop view. In mobile, the toggle button still exists and will display the single filter or group by option, which is the same behavior as it is today.

### DIFF
--- a/src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
+++ b/src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
@@ -95,10 +95,9 @@ function GridTableLayoutActionsComponent<
   const hasSearch = !!searchProps;
   const filterCount = Object.keys(filterDefs ?? {}).length;
   const hasGroupBy = !!groupBy;
-  const hasFilterControls = filterCount > 0 || hasGroupBy;
   const controlCount = filterCount + (hasGroupBy ? 1 : 0);
+  const hasFilterControls = controlCount > 0;
   // One control only — nothing to nest behind a toggle; show it inline in the toolbar on desktop.
-  // Fall back to the Filter button if a lone filter is missing filter/setFilter.
   const showInlineControl = !sm && controlCount === 1 && (hasGroupBy || !!(filter && setFilter));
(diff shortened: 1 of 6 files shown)


src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -94,7 +93,13 @@ function GridTableLayoutActionsComponent<
   }, 300);
 
   const hasSearch = !!searchProps;
-  const hasFilters = !!filterDefs && Object.keys(filterDefs ?? {}).length > 0;
+  const filterCount = Object.keys(filterDefs ?? {}).length;
+  const hasGroupBy = !!groupBy;
+  const hasFilterControls = filterCount > 0 || hasGroupBy;
```
stephenh: Would `hasFilterControls` be simpler by moving it after `controlCount` and being:

```
const hasControls = controlCount > 0;
```

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
```diff
@@ -94,7 +93,13 @@ function GridTableLayoutActionsComponent<
...
+  const controlCount = filterCount + (hasGroupBy ? 1 : 0);
+  // One control only — nothing to nest behind a toggle; show it inline in the toolbar on desktop.
+  // Fall back to the Filter button if a lone filter is missing filter/setFilter.
```
stephenh: ...why would a lone filter by missing a filter/setFilter? Maybe include an `i.e. this happens when...`

stephenh (approved):
Nice! Great rule-of-thumb to build into the layout to have it "just know" what to do.

src/components/Layout/GridTableLayout/GridTableLayoutActions.tsx
bdow: Yea, this would most likely be a problem at the call site that didn't include these. Not a real issue. Will clean up this comment.

PR #1344, opened 2026-07-27
feat: Increase & improve design token usage across system

bdow:
Consolidates tokens that were similar.
Updates many more components to use design tokens for defining colors, allowing to support a dark mode. 
Adds a story for all semantic tokens, and updates the colors story location.


src/components/internal/DatePicker/DatePickerOverlay.tsx
```diff
@@ -1,14 +1,14 @@
...
+type DatePickerOverlayProps = {
   overlayProps: HTMLAttributes<HTMLElement>;
-}
+};
 
 // Small wrapper around DatePicker to provide necessary styling and state handling when displayed as an overlay.
 export function DatePickerOverlay({ overlayProps, children }: PropsWithChildren<DatePickerOverlayProps>) {
   return (
-    <div css={Css.br4.bshModal.$} {...overlayProps}>
+    <div css={Css.br4.bshModal.bgColor(Tokens.SurfaceRaised).color(Tokens.OnSurface).oh.$} {...overlayProps}>
```
stephenh: Just curious why these are having to set bcColor/color and they used to not have that?

scripts/generate-token-usages.ts
```diff
@@ -0,0 +1,220 @@
+/**
+ * Scans `src/` for `Tokens.*` references and maps each token to Storybook stories
+ * that cover those call sites. Output: `src/foundations/generated/tokenUsages.json`.
```
stephenh: Is this trying to say `Token.OnSurface` is used in `TextField.tsx` and `SelectField.tsx`, so if you change `OnSurface`, look in the `TextField.stories.tsx` and `SelectField.stories.tsx`?

Maybe include that in the jsdoc.

How does this does things like internal components where the file name doesn't directly have a file name == public component name == story name?

Is this going to be a long-lived file, or just something for this transient migration? Maybe include what/why users would use the token usages for.

Fwiw instead of file name mapping, I would probably do something runtime-y, i.e.:

* Monkey patch the truss `Tokens` keys to record their usage to a global array
* Have each story clear the array, run its story, then write `story: [OnSurface, OnPrimary]` etc to a json file
* After all stories run, we know definitely which tokens were accessed in which story

Granted, it requires actually running all the stories, and not doing static analysis -- not sure how exact we need this to be.

package.json
```diff
@@ -37,16 +37,17 @@
...
-    "build-storybook": "NODE_OPTIONS=--openssl-legacy-provider storybook build",
+    "build-storybook": "yarn generate:token-usages && NODE_OPTIONS=--openssl-legacy-provider storybook build",
     "test": "vitest run",
     "test:watch": "vitest",
     "lint": "eslint --ext js,ts,tsx src",
     "lint:ci": "eslint --ext js,ts,tsx src --cache --cache-location .eslintcache --quiet",
     "lint:fix": "eslint --ext js,ts,tsx --fix --cache --cache-location .eslintcache src",
     "lint:fix:files": "eslint --ext js,ts,tsx --fix --cache --cache-location .eslintcache",
-    "storybook": "NODE_OPTIONS=--openssl-legacy-provider storybook dev -p 9000",
+    "storybook": "yarn generate:token-usages && NODE_OPTIONS=--openssl-legacy-provider storybook dev -p 9000",
```
stephenh: Do `storybook` / `build-storybook` need to run the token-usages?

src/components/internal/DatePicker/DatePickerOverlay.tsx
bdow: This one wasn't needed. Will update and look for other similar areas. I think the agent went a bit too heavy handed. Thanks for catching

package.json
bdow: No, this was actually just a helper for me at the time to now have to do this when restarting. Will remove

src/components/Avatar/Avatar.tsx
```diff
@@ -20,7 +20,15 @@ export function Avatar({ src, name, size = "md", showName = false, preventToolti
   const styles = Css.br100.wPx(px).hPx(px).mwPx(px).oh.$;
 
   const img = showFallback ? (
-    <div css={{ ...styles, ...Css.typography(sizeToFallbackTypeScale[size]).bgGray400.gray100.df.aic.jcc.$ }} {...tid}>
+    <div
+      css={{
+        ...styles,
+        ...Css.typography(sizeToFallbackTypeScale[size])
+          .bgColor(Tokens.TextDisabled)
+          .color(Tokens.NeutralFillHoverSubtle).df.aic.jcc.$,
```
abadami: **nit**: It's a little odd to me that we're using the `TextDisabled` token for a bg color and the `NeutralFillHoverSubtle` for a non-hover text color. It's probably not a big deal if we do it sparingly but it is a little jarring to read.

**Edit (Thought)**: This could be something to talk about as we refine our Tokens, but I wonder if it'd be worth the redundancy to have another Token for these lighter disabled background type situations. Actually that might be a bigger question for the design of our Tokens as a whole, do we want multiple tokens that may hold the same value to differentiate their use cases?

src/components/Headers/BaseHeader.tsx
```diff
@@ -30,7 +30,7 @@ export function BaseHeader(props: BaseHeaderProps) {
       >
         <div css={Css.mw0.$}>
           {breadcrumbs && <Breadcrumbs {...breadcrumbs} />}
-          <h1 {...tid.title} css={Css.xl.$}>
+          <h1 {...tid.title} css={Css.xl.color(Tokens.OnSurface).$}>
```
abadami, line 33: **thought**: This was added when a color wasn't specified, but I do believe we need it. With that in mind, Breadcrumbs probably need to use Tokens as well, but that could be a follow-up.

src/components/Tabs.test.tsx
```diff
@@ -241,6 +241,12 @@ describe("TabsWithContent", () => {
   });
 });
 
+/** Active tabs use smSb (font-weight 600); tokenized border colors aren't asserted via toHaveStyle. */
+function expectActiveTab(el: HTMLElement) {
+  expect(el).toHaveAttribute("aria-selected", "true");
+  expect(el).toHaveStyle({ fontWeight: 600 });
+}
```
abadami, line 248: **praise/thought**: This little helper is nice. I wonder if our styling may get consistent enough to have a group of these helpers to help with testing this sort of thing.

scripts/generate-token-usages.ts
bdow: This builds out the "Used in" column for the Tokens.stories.tsx for the semantic tokens. I found it helpful to as a "let's make sure all these tokens are actually used" and to do my own analysis against. But maybe isn't needed forever. 
Happy to remove for now and see if we really find it helpful later to include usage documentation.

src/css/CssReset.css
```diff
@@ -107,6 +107,9 @@ Remove the margin in all browsers.
 
   body {
     margin: 0;
+    /* Theme-aware page chrome so unmarked text/bg follow Surface / OnSurface (incl. contrast scope). */
+    color: var(--b-on-surface);
+    background-color: var(--b-surface);
```
abadami: **thought**: I imagine not since the background is likely set in internal-frontend, but just to be sure, is there any risk to this messing with the background styling in Blueprint?

src/components/internal/DatePicker/DateRangePicker.tsx
```diff
@@ -27,7 +27,7 @@ export function DateRangePicker(props: DateRangePickerProps) {
   const tid = useTestIds(props, "datePicker");
 
   return (
-    <div css={Css.dib.bgWhite.xs.$} {...tid}>
+    <div css={Css.dib.bgColor(Tokens.SurfaceRaised).color(Tokens.OnSurface).xs.$} {...tid}>
```
abadami: **question**: Normally a `bgWhite` turns into a `Tokens.Surface`, but in this instance is `SurfaceRaised`. I did a check and noticed the overlay that is a parent of this has the same color setup as well, which is likely why you changed it, but do we even need these set here?

src/components/internal/DatePicker/DatePicker.tsx
```diff
@@ -22,7 +22,7 @@ export function DatePicker(props: DatePickerProps) {
   const tid = useTestIds(props, "datePicker");
 
   return (
-    <div css={Css.dib.bgWhite.xs.$} {...tid}>
+    <div css={Css.dib.bgColor(Tokens.SurfaceRaised).color(Tokens.OnSurface).xs.$} {...tid}>
```
abadami: **question**: question as DatePickerRange, do we need these if the overlay takes care of these colors?

src/inputs/DateFields/DateRangeField.stories.tsx
```diff
@@ -44,3 +45,19 @@ export function Example() {
...
+      value={range}
+      onChange={setRange}
+      placeholder="Date Range"
+      onBlur={action("Blur")}
+      onFocus={action("Focus")}
+      defaultOpen
+    />
+  );
+}
+DatePickerOpen.decorators = [withDimensions()];
```
abadami, line 63: **question (non-blocking)**: For my knowledge, what does the `withDimensions` decorator do?

src/components/Avatar/Avatar.tsx
bdow: >nit: It's a little odd to me that we're using the TextDisabled token for a bg color and the NeutralFillHoverSubtle for a non-hover text color. It's probably not a big deal if we do it sparingly but it is a little jarring to read.

Yea, great call out. This is incorrect usage. Will take a look.

src/components/Headers/BaseHeader.tsx
bdow: I looked at this one too and thought about removing it. It should be inherited from the body, which uses `OnSurface` as the default color. But I thought this extra guard against the color should be fine.

src/css/CssReset.css
bdow: oh yikes. I didn't see that `background-color` snuck in ehre. That'd deifitely blow up Blueprint. Will remove.

src/components/internal/DatePicker/DateRangePicker.tsx
bdow: Yes, this was intiention. SurfaceRaised is still `White` in light mode, but different in Dark mode. This is because the box-shadow applied in light mode gives it enough contrast between itself and the content under it. But in dark mode, that same box-shadow doesn't really show as its too dark. So instead the SurfaceRaised is a little lighter than Surface on dark mode.

src/components/internal/DatePicker/DatePicker.tsx
bdow: I think I removed this because the reason you describe.

src/inputs/DateFields/DateRangeField.stories.tsx
bdow: It gives the story specific height and width properties. Chromatic snapshots take the capture based on the height/width of the document. Overlays are absolutely position and do not contribute to the document's height/width, so they get cut off unless we set explicitly height/width of the chromatic snapshot.

src/inputs/Switch.tsx
```diff
@@ -136,16 +142,18 @@ const toggleWidth = (isCompact: boolean) => (isCompact ? 32 : 40);
 const circleDiameter = (isCompact: boolean) => (isCompact ? 12 : 20);
 
 // Switcher/Toggle element styles
-export const switchHoverStyles = Css.bgGray400.$;
+export const switchHoverStyles = Css.bgColor(Tokens.TextDisabled).$;
```
abadami: **suggestion**: I imagine this should probably not be using `TextDisabled`, similar to the first one I pointed out?

src/inputs/internal/radioStyles.ts
```diff
@@ -30,10 +30,11 @@ export const radioFocus = {
   ...Css.bshFocus.$,
 };
 
+// Blue900 hover has no semantic token — keep palette.
 export const radioHover = Css.blue900.bcBlue900.$;
 
 export const radioDisabled = {
-  ...Css.cursorNotAllowed.gray100.$,
+  ...Css.cursorNotAllowed.color(Tokens.NeutralFillHoverSubtle).$,
```
abadami: **suggestion**: I feel like this one should be TextDisabled, based on the first one I pointed out?

src/inputs/CheckboxBase.tsx
```diff
@@ -97,24 +97,26 @@ export function CheckboxBase(props: CheckboxBaseProps) {
...
+const baseStyles = Css.hPx(16).mw(px(16)).relative.ba.bc(Tokens.FieldBorderDefault).br4.bgColor(Tokens.Surface)
+  .transition.$;
+const filledBoxStyles = Css.bc(Tokens.ChoiceSelected).bgColor(Tokens.ChoiceSelected).$;
+// Blue900 hover has no semantic token — keep palette.
 const filledBoxHoverStyles = Css.bgBlue900.$;
-const disabledBoxStyles = Css.bgGray50.bcGray100.$;
-const disabledSelectedBoxStyles = Css.bgGray400.bcGray400.$;
-const disabledColor = Css.gray300.$;
+const disabledBoxStyles = Css.bgColor(Tokens.NeutralFillHoverSubtle).bc(Tokens.NeutralFillHoverSubtle).$;
+const disabledSelectedBoxStyles = Css.bgColor(Tokens.TextDisabled).bc(Tokens.TextDisabled).$;
```
abadami: **Thought (non-blocking)**: We're kinda using `TextDisabled` for a border disabled color quite a bit. I imagine the colors will be the same in contrast as well, but going back to my previous thought I wonder if we should be exhaustive in token names despite them potentially being similar in values, like introducing a `BorderDisabled` here.

OR, alternatively, we can have this value maybe be a universal `Disabled` instead? wdyt?

src/inputs/CheckboxBase.tsx
```diff
@@ -97,24 +97,26 @@ export function CheckboxBase(props: CheckboxBaseProps) {
...
-const filledBoxStyles = Css.bcBlue700.bgBlue700.$;
+const baseStyles = Css.hPx(16).mw(px(16)).relative.ba.bc(Tokens.FieldBorderDefault).br4.bgColor(Tokens.Surface)
+  .transition.$;
+const filledBoxStyles = Css.bc(Tokens.ChoiceSelected).bgColor(Tokens.ChoiceSelected).$;
+// Blue900 hover has no semantic token — keep palette.
 const filledBoxHoverStyles = Css.bgBlue900.$;
-const disabledBoxStyles = Css.bgGray50.bcGray100.$;
-const disabledSelectedBoxStyles = Css.bgGray400.bcGray400.$;
-const disabledColor = Css.gray300.$;
+const disabledBoxStyles = Css.bgColor(Tokens.NeutralFillHoverSubtle).bc(Tokens.NeutralFillHoverSubtle).$;
```
abadami: **suggestion**: I think this is suppose to use `SurfaceMuted` based on the other disabled bg colors?

src/inputs/internal/radioStyles.ts
bdow: Created a new `ChoiceDisabled`  token for this purpose. Nothing else really fit properly.

src/inputs/CheckboxBase.tsx
bdow: Used the same `ChoiceDisabled` here as well.

src/inputs/CheckboxBase.tsx
bdow: Went with `FieldBgDisabled` which is also an existing token.

abadami (approved):

PR #1347, opened 2026-07-28
feat: FormSection & FormSectionLayout

abadami:
### Form Section and Form Section Layout

Adds in the FormSection component and FormSectionLayout.

### DIFF
--- a/src/forms/FormSection.stories.tsx
+++ b/src/forms/FormSection.stories.tsx
@@ -1,0 +1,57 @@
+import { Meta } from "@storybook/react-vite";
+import { Css, Tokens } from "src/Css";
+import { FormSection } from "src/forms/FormSection";
+import { withBeamDecorator } from "src/utils/sb";
+
+export default {
+  component: FormSection,
+  decorators: [withBeamDecorator],
+} as Meta;
+
+export function Regular() {
+  return (
+    <FormSection
+      title="General Contractor"
+      description="The primary contractor responsible for this project."
+      fields={<PlaceholderFields count={2} />}
+    />
+  );
+}
+
+export function Child() {
+  return <FormSection title="Electrical" isChild fields={<PlaceholderFields count={2} />} />;
+}
+
+export function WithActions() {
+  return (
+    <FormSection
+      title="Sub-Contractors"
+      actions={[{ label: "Add", onClick: () => {}, variant: "tertiary" }]}
+      fields={<PlaceholderFields count={2} />}
+    />
+  );
+}
+
+export function WithChildSections() {
+  return (
+    <FormSection
+      title="Sub-Contractors"
+      actions={[{ label: "Add", onClick: () => {}, variant: "tertiary" }]}
+      childSections={[
+        { title: "Electrical", fields: <PlaceholderFields count={2} /> },
+        { title: "Plumbing", fields: <PlaceholderFields count={2} /> },
+        { title: "HVAC", fields: <PlaceholderFields count={2} /> },
+      ]}
+    />
+  );
+}
+
+function PlaceholderFields({ count }: { count: number }) {
+  return (
+    <div css={Css.df.fdc.gap1.$}>
+      {Array.from({ length: count }, (_, i) => (
+        <div key={i} css={Css.hPx(36).br4.bgColor(Tokens.SurfaceSeparator).$} />
+      ))}
+    </div>
+  );
+}
--- a/src/forms/FormSection.stories.tsx
+++ b/src/forms/FormSection.stories.tsx
@@ -8,25 +8,11 @@ export default {
   decorators: [withBeamDecorator],
 } as Meta;
 
-export function Regular() {
+export function WithoutChildren() {
   return (
     <FormSection
       title="General Contractor"
       description="The primary contractor responsible for this project."
-      fields={<PlaceholderFields count={2} />}
-    />
-  );
-}
-
-export function Child() {
-  return <FormSection title="Electrical" isChild fields={<PlaceholderFields count={2} />} />;
-}
-
-export function WithActions() {
-  return (
-    <FormSection
-      title="Sub-Contractors"
-      actions={[{ label: "Add", onClick: () => {}, variant: "tertiary" }]}
       fields={<PlaceholderFields count={2} />}
     />
   );
@@ -38,9 +24,17 @@ export function WithChildSections() {
       title="Sub-Contractors"
       actions={[{ label: "Add", onClick: () => {}, variant: "tertiary" }]}
       childSections={[
-        { title: "Electrical", fields: <PlaceholderFields count={2} /> },
+        {
+          title: "Electrical",
+          fields: <PlaceholderFields count={2} />,
+          actions: [{ label: "Add", onClick: () => {}, variant: "tertiary" }],
+        },
         { title: "Plumbing", fields: <PlaceholderFields count={2} /> },
-        { title: "HVAC", fields: <PlaceholderFields count={2} /> },
+        {
+          title: "HVAC",
+          fields: <PlaceholderFields count={2} />,
+          actions: [{ label: "Add", onClick: () => {}, variant: "tertiary" }],
+        },
       ]}
     />
   );
--- a/src/forms/FormSection.stories.tsx
+++ b/src/forms/FormSection.stories.tsx
@@ -27,6 +27,7 @@ export function WithChildSections() {
         {
           title: "Electrical",
           fields: <PlaceholderFields count={2} />,
+          description: "Electrical contracts are needed for the construction to continue",
           actions: [{ label: "Add", onClick: () => {}, variant: "tertiary" }],
         },
         { title: "Plumbing", fields: <PlaceholderFields count={2} /> },
--- a/src/forms/FormSection.test.tsx
+++ b/src/forms/FormSection.test.tsx
@@ -1,0 +1,85 @@
+import { FormSection } from "src/forms/FormSection";
+import { render } from "src/utils/rtl";
+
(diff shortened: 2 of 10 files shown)


src/layouts/FormSectionLayout/FormSectionLayout.tsx
```diff
@@ -0,0 +1,52 @@
+import { ReactNode } from "react";
+import { Button, ButtonProps } from "src/components/Button";
+import { Css, Tokens } from "src/Css";
+import { FormSection, FormSectionProps } from "src/forms/FormSection";
+import { useTestIds } from "src/utils";
+import { defaultTestId } from "src/utils/defaultTestId";
+
+export type FormSectionLayoutProps = {
+  /** `"full"` renders edge-to-edge; `"centered"` caps content at a 720px column and centers it. */
+  layout: "full" | "centered";
```
bdow: **suggestion:** Can we defualt to "centered" - I think that is what we're going to have most likely. 

Also, side note: I've been thinking about how we handle different layout. We really have 3 - "full width", "centered large (1440px)", and "centered small / 720px". These layouts will want to be used not just in something like FormSectionLayout, but potentially more generically like a `<ContentLayout size="lg | sm | full">` sort of thing. I haven't thought it completely through, so not sure what direction to go in. Another thought was to make a truss helper in the `truss-config` so users can do something like `Css.wLg` or `Css.wSm`, and of course, still use `Css.w100.$`.

abadami: Per discussion, I've removed the layout prop and we'll only have a "centered" layout available for `FormSection`.

src/forms/FormSection.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      title="General Contractor"
+      description="The primary contractor responsible for this project."
+      fields={<PlaceholderFields count={2} />}
+    />
+  );
+}
+
+export function Child() {
+  return <FormSection title="Electrical" isChild fields={<PlaceholderFields count={2} />} />;
+}
```
bdow: **thought/suggestion:** I think we might be able to reduce the need for this `Child` story and it could be validated via `WithChildSections`. At least, I don't think doing visual regressions on the `isChild` variant alone is necessary. I was originally thinking `Regular` wouldn't be necessary either, but perhaps its more of a "With Children" and "Without Children" visual distinction. 
I also wonder if maybe we should have a story here when the children also have actions, maybe?

src/forms/FormSection.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    [r.getByText("Electrical"), r.getByText("Plumbing")].forEach((title) =>
+      expect(title).toHaveStyle({ fontSize: "16px" }),
+    );
+  });
+
+  it("overrides a caller-provided isChild on recursive childSections", async () => {
+    // Given a childSections entry that (incorrectly) sets isChild itself
+    // When rendered as a child of a parent section
+    const r = await render(
+      <FormSection title="Trade Partners" childSections={[{ title: "Electrical", isChild: false }]} />,
```
bdow: **thought/question:** Should we even allow for setting `isChild`?

src/forms/FormSection.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  it("uses the smaller child title style when isChild is set", async () => {
+    // Given a FormSection explicitly marked isChild
+    // When rendered
+    const r = await render(<FormSection title="Sub-Contractors" isChild />);
+    // Then it renders at the child (md) heading size
+    expect(r.formSection_title).toHaveStyle({ fontSize: "16px" });
+  });
+
+  it("recursively renders childSections at the smaller child title size", async () => {
```
bdow: **nit:** Very nit-picky, but we're not validating that the rendering happens "recursively" here. It really doesn't matter how we rendered the child sections, this test could still pass.

src/forms/FormSection.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+      />,
+    );
+    // Then both child sections render, nested under the parent
+    expect(r.formSection_title).toHaveTextContent("Trade Partners");
+    expect(r.electricalFields).toBeInTheDocument();
+    expect(r.plumbingFields).toBeInTheDocument();
+    // And each child section's title renders at the child heading size
+    [r.getByText("Electrical"), r.getByText("Plumbing")].forEach((title) =>
+      expect(title).toHaveStyle({ fontSize: "16px" }),
+    );
```
bdow: **suggestion:** How about using the test ids here? `r.childSection_title_0` Would that work?

src/forms/FormSection.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+/** A titled section of form content — title/description/actions row, then `fields`, then any recursively-rendered `childSections`. */
+export function FormSection(props: FormSectionProps) {
+  const { title, description, actions, fields, childSections, isChild = false } = props;
+  const tid = useTestIds(props, "formSection");
+
+  return (
+    <div css={Css.df.fdc.gap2.$} {...tid}>
+      <div css={Css.df.jcsb.aifs.$}>
```
bdow: **nit (non-blocking):** `align-items: flex-start` is the default. I think we could just omit it here. fwiw, this was more of an issue, imo, when it was run time styling w/ custom classes to represent each css combination. Now that we do build time styles with atomic class names, this is not really a concern to me. More just being pedantic about only applying styles that need to be applied.

src/forms/FormSection.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <div css={Css.df.fdc.gapPx(12).$}>
+          <h2 css={isChild ? Css.mdSb.$ : Css.lg.$} {...tid.title}>
+            {title}
+          </h2>
```
bdow: **nit/suggestion:** Due to the nature of the leveling here, I think it'd make sense  to change from an `h2` to an `h3` when `isChild`. A small helper method for getting the proper title would probably be the cleanest.
```
function getTitle(title: string, isChild: boolean, tid: Record<string, string>) {
  if (isChild) { return <h3 css={Css.mdSb.$} {...tid.title}>{title}</h3> }
  return <h2 css={Css.lg.$} {...tid.title}>{title}</h2>
}

File: src/forms/FormSection.tsx [hunk 8]
```diff
@@ -0,0 +1,56 @@
...
+              <Button key={`${action.label}`} {...action} />
+            ))}
+          </div>
+        )}
+      </div>
+      {fields}
+      {childSections && (
+        <div css={Css.df.fdc.gap3.$}>
+          {childSections.map((child, i) => (
+            <div key={defaultTestId(child.title) || i} css={Css.bb.bc(Tokens.SurfaceSeparator).pb3.ifLastOfType.bn.$}>
```
bdow: **suggestion:** Keys can be any string, so we don't need to manipulate the `title` here via `defaultTestId`. We can keep it as is.

src/layouts/FormSectionLayout/FormSectionLayout.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+          { title: "General Contractor", fields: <div data-testid="gcFields" /> },
+          { title: "Sub-Contractors", fields: <div data-testid="subFields" /> },
+        ]}
+      />,
+    );
+    // Then the form-level title/description render
+    expect(r.formSectionLayout_title).toHaveTextContent("Trade Partners");
+    expect(r.formSectionLayout_description).toHaveTextContent("Set up trade partner assignments for this project");
+    // And each section is delegated to FormSection, rendering its own content
+    expect(r.getByText("General Contractor")).toBeInTheDocument();
```
bdow: **nit (non-blocking):** Would prefer to use `r.[test_id]` vs `r.getByText`. I also don't think that looking for these titles is completely necessary. as long as the field values are in there I think we can safely assume the section made it in. The responsibility for testing additional properties we can delegate to the FormSection itself, which I think it already does.

src/layouts/FormSectionLayout/FormSectionLayout.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+/** Centered (720px) width/column shell for a form built out of `FormSection`s — e.g. as a `WorkflowLayoutStep`'s `content`. */
+export function FormSectionLayout(props: FormSectionLayoutProps) {
+  const { title, description, actions, sections } = props;
+  const tid = useTestIds(props, "formSectionLayout");
+
+  return (
+    <div css={Css.df.fdc.gap3.w100.pt4.maxwPx(720).mxa.$} {...tid}>
+      <div css={Css.df.jcsb.aifs.$}>
```
bdow: **nit (non-blocking):** same thing about `aifs` here.

bdow (approved):
👏 Very clean, simple, and easy to follow. A couple minor notes/nits. Happy to have some of them addressed as part of this PR or follow ups.

src/forms/FormSection.test.tsx
abadami: Good point.

src/layouts/FormSectionLayout/FormSectionLayout.test.tsx
abadami: That's a good point on not needing to check the titles, I'll just omit them.

src/forms/FormSection.test.tsx
abadami: That's fair. When using the test ids it might technically be more clear, but that can also technically be anything that sets the data-testid to "formSection". I'll just omit recursively, no real reason to have it there anyway.

PR #1349, opened 2026-07-30
feat: SelectCard group improvements: layout, image, and link support

allangaldinosilva:
### [P-2608](https://linear.app/homebound-team/issue/P-2608)

This PR adds a few improvements to the `SelectCardGroup` / `MultiSelectCardGroup` grid cards, plus a `fullWidth` option on `Card`.

#### `layout` prop

Grid cards can now render horizontally, with the icon to the left of the text instead of above it. The default is still `vertical`.

```tsx
<SelectCardGroup
  label="Options"
  layout="horizontal"
  options={[{ label: "Warranty", icon: "shield", value: "warranty" }]}
  onChange={...}
/>
```

Storybook: https://60466f7d43c37c0021788867-cnzbqlqxmr.chromatic.com/?path=/story/inputs-selectcard-selectcardgroup--horizontal-layout

#### `image` support

Grid options can now show an image instead of an icon. The two are mutually exclusive — an option takes either `icon` or `image`.

```tsx
<SelectCardGroup
  label="Plans"
  options={[{ label: "The Aspen", image: "/plans/aspen.png", value: "aspen" }]}
  onChange={...}
/>
```

Storybook: https://60466f7d43c37c0021788867-cnzbqlqxmr.chromatic.com/?path=/story/inputs-selectcard-selectcardgroup--horizontal-layout-with-images

#### `Card` `fullWidth` prop

`Card` can now fill its container instead of using its fixed width (256px card / 520px list) — handy when cards live in a grid whose columns don't match those widths.

```tsx
<Card title="The Aspen" imgSrc="/plans/aspen.png" type="list" fullWidth />
```

Issue screenshot:
Screenshot 2026-07-30 at 11 33 15 AM

Storybook: https://60466f7d43c37c0021788867-cnzbqlqxmr.chromatic.com/?path=/story/components-card--full-width-list-cards


src/inputs/SelectCard/GridSelectCard.tsx
```diff
@@ -33,13 +42,25 @@ export function GridSelectCard(props: GridSelectCardProps) {
       disabled={isDisabled}
       tooltip={tooltip}
       view="grid"
+      layout={layout}
       __storyState={__storyState}
+      footer={
+        link && (
+          <span css={Css.df.jcc.aic.xsSb.bt.bc(Tokens.SurfaceSeparator).pyPx(8).$}>
```
stephenh: I think `pyPx(8)` could be just `py1`

src/inputs/SelectCard/SelectCardShell.tsx
```diff
@@ -57,7 +63,12 @@ export function SelectCardShell(props: SelectCardShellProps) {
     children: (
       <label
         css={{ ...styles, ...Css.cursorPointer.if(isDisabled).cursorNotAllowed.$ }}
-        {...mergeProps(hoverProps, focusProps, pressProps)}
+        // Keep focus where it is while clicking the card. Without this, mousedown moves focus to
+        // the nearest focusable ancestor (e.g. a modal, which has tabindex=-1) before the label
+        // click focuses our hidden input. react-aria sees that second focus event with no user
+        // event tied to it, assumes "virtual" (screen reader) focus, and shows the keyboard focus
+        // ring on a plain mouse click.
```
stephenh: I'm surprised we have to work around this -- per the "mousedown moves focuse to the nearest focusable ancestoer", does that mean we're using `useFocusRing` wrong?

I.e. I would have assumed `useFocusRing` would do exactly this, i.e. that _we're_ the focusable ancestor, and so "just work".

We're missing `within: true` to `useFocusRing` ... is that accidentally causing this? Or is there some other prop to say "we're the ancestor" or what not? 

"the modal has tabindex=-1" ... should one of our divs have `tabindex=-1` so it stops on/finds the card itself?

stephenh (approved):
Stories look great; just curious about the useFocusRing workaround we need.

src/inputs/SelectCard/SelectCardShell.tsx
allangaldinosilva: Great call @stephenh, I figured out react-aria has `preventFocusOnPress` that's exactly for this sort of use-case.

The solution required to bump `react-aria` though. Because of that, I've decided to make it in a different PR https://github.com/homebound-team/beam/pull/1350.

cc @bdow / @abadami

PR #1350, opened 2026-07-30
fix: SelectCard showing keyboard focus ring on mouse click inside modals

allangaldinosilva:
### Problem

Clicking a SelectCard inside a Modal with the mouse showed the keyboard focus ring (double border), as
if the card had been tabbed to.

https://github.com/user-attachments/assets/c59206f2-c0de-4058-812a-0305864d50d8

What happens: on mousedown, the browser moves focus to the nearest focusable ancestor, inside a Modal
that's the dialog element (tabindex=-1). Then the label click focuses the card's hidden input.
react-aria sees that second focus event with no user event tied to it, assumes "virtual" (screen
reader) focus, and shows the keyboard focus ring on a plain mouse click. Outside a modal there's no
focusable ancestor, so the bug only showed inside modals.

Solution

Use react-aria's API for exactly this: `usePress({ preventFocusOnPress: true })`, which keeps focus
where it is while clicking the card. Mouse clicks no longer show the ring; keyboard focus (Tab) still
does.

https://github.com/user-attachments/assets/c8192c14-112f-4564-ab5a-847c1ef3641f

Because we were loading two react-aria versions (react-aria@3.47.0 top-level + 3.50.0 nested
under @react-aria/interactions), the solution required to bump `react-aria` to work properly. So this PR also:

- Bumps react-aria → ^3.50.0 and react-stately → ^3.48.0 so yarn dedupes to a single copy.
- Fixes two type fallouts from the bump: useTextField<"textarea"> in TextAreaField, and ComboBoxInput
  accepting both single/multiple selection ComboBoxState.


src/inputs/SelectCard/SelectCardGroup.stories.tsx
```diff
@@ -48,6 +50,27 @@ export function OptionType() {
...
 }
 
+export function InModal() {
+  const { openModal } = useModal();
+  const open = () => openModal({ content: <InModalContent /> });
+  // Immediately open the modal for Chromatic snapshots
+  useEffect(open, [openModal]);
+  return <Button label="Open" onClick={open} />;
+}
+InModal.decorators = [withBeamDecorator];
```
stephenh: Can we use the `newStory` or what not to set decorators? Generally prefer that over "assigning fields on the function" which is storybook's kinda jank imo default API.

src/inputs/SelectCard/SelectCardShell.tsx
```diff
@@ -30,7 +30,12 @@ export function SelectCardShell(props: SelectCardShellProps) {
 
   const { hoverProps, isHovered: isHoveredFromEvents } = useHover({ isDisabled });
   const { isFocusVisible: isFocusVisibleFromEvents, focusProps } = useFocusRing({ within: true });
-  const { pressProps, isPressed: isPressedFromEvents } = usePress({ isDisabled });
+  // preventFocusOnPress keeps focus where it is while clicking the card. Without it, mousedown
+  // moves focus to the nearest focusable ancestor (e.g. a modal, which has tabindex=-1) before the
+  // label click focuses our hidden input. react-aria sees that second focus event with no user
+  // event tied to it, assumes "virtual" (screen reader) focus, and shows the keyboard focus ring
+  // on a plain mouse click. Keyboard focus is unaffected: it goes straight to the hidden input.
+  const { pressProps, isPressed: isPressedFromEvents } = usePress({ isDisabled, preventFocusOnPress: true });
```
stephenh, line 38: My only nit is that "moves to the nearest focusable ancestor", I'm still just curious why "the card itself" is not "the nearest focusable ancestor".

I.e. given the fact that "we're the focus ring" / "the card is focused", I'm not sure why react-aria thinks it needs to "keep walking up the tree".

Still great to bump react-aria, particularly since we had dup/shadow version, so +1 on the PR, just OCD curious about this.

stephenh (approved):

PR #1351, opened 2026-07-30
feat: Add fields prop to FormSectionLayout

abadami:
### Adding Fields prop to FormSectionLayout

Some form layouts have fields available right below the form title (including in the Plan-Based Bid workflow), so this PR has FormSectionLayout accommodate for that.


src/layouts/FormSectionLayout/FormSectionLayout.stories.tsx
```diff
@@ -14,6 +14,7 @@ export function Default() {
       title="Trade Partners"
       description="Assign and manage trade partners for this project."
       actions={[{ label: "Save draft", onClick: () => {}, variant: "tertiary" }]}
+      fields={<PlaceholderFields count={2} />}
```
stephenh: Maybe something like `initialFields` or `topFields`?

As-is it sounds a lot like "the fields that go inside of `sections`", so maybe we can differentiate.

Maybe also some in-source documentation / rationale / recommendations on when users (and UX designers 😅 ) should use this instead of the standard `sections` API.

abadami: Ah, that name change is a good idea. I think of the two I like `initialFields` more. I also added some documentation based on the current design use cases for it.

src/layouts/FormSectionLayout/FormSectionLayout.test.tsx
```diff
@@ -41,4 +41,18 @@ describe("FormSectionLayout", () => {
...
+    const r = await render(
+      <FormSectionLayout
+        title="Trade Partners"
+        initialFields={<div data-testid="topLevelFields" />}
+        sections={[{ title: "General Contractor", fields: <div data-testid="gcFields" /> }]}
+      />,
+    );
+    // Then the top-level fields render before the sections
+    expect(r.topLevelFields).toBeInTheDocument();
+    expect(r.topLevelFields.compareDocumentPosition(r.gcFields)).toBe(Node.DOCUMENT_POSITION_FOLLOWING);
```
bdow, line 56: Huh! Never seen these props before. 👍

bdow (approved):

bdow:
Looks simple enough. I am curious what the comment means by shaping the rest of the form beforehand.

src/layouts/FormSectionLayout/FormSectionLayout.test.tsx
abadami: Yeah! I'll admit the agent found this one and I thought it was a neat way to test this, so I kept it 😄

PR #1352, opened 2026-07-30
feat: Content Header

abadami:
### Content Header

Adds a Content Header component. Main usage is to provide a standard header specifically for content that we've been seeing pop up in design specs. This might be expanded on in the future as we better define our more content specific layouts.


src/components/Headers/ContentHeader.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+  component: ContentHeader,
+  decorators: [withBeamDecorator],
+} as Meta;
+
+export function Default() {
+  return (
+    <ContentHeader
+      title="Trade Partners"
+      description="Assign and manage trade partners for this project."
+      actions={<Button label="Add" onClick={() => {}} />}
```
henrysipp: So is this intended to be a react fragment if I have multiple actions? What happens if a naive user wraps in a div? 

I don't *love* that if we aren't careful, we could accidentally break the flex behavior + gap between actions there.

Pattern I've seen in the past is something like this, though I'm not 100% on this api either necessarily. At least would strongly enforce the slots!

```
<ContentHeader>
  <ContentHeader.TItle>title</ContentHeader.Title>
  <ContentHeader.Description>description</ContentHeader.Description>

  <ContentHeader.Actions> <-- mostly just this part
    <Button />
    <Button />
  </ContentHeader.Actions>
</ContentHeader>
```

src/components/Headers/ContentHeader.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Generic title/description/actions header for content sections; stays visible while the page scrolls
+ * horizontally, but scrolls away normally on the vertical axis.
+ *
+ * `position: sticky`'s horizontal offset is bounded by the element's containing block — unlike the
+ * vertical axis (where a block's height naturally grows to fit all scrolled content), a block's width
+ * does *not* grow to fit an overflowing sibling, so sticky has no "room" to operate unless something
+ * gives it that room. If this sits above/beside content that overflows horizontally (e.g. a wide
+ * table), wrap both in a shared container sized with `min-width: fit-content` (truss: `mw("fit-content")`)
```
bdow, line 20: The suggestion of using min-width: fitcontent feels like the opposite of why you'd first even use this component. I also can't think of a reason why we'd want to have this sort of escape hatch.

src/components/Headers/ContentHeader.tsx
bdow, line 1: Will this replace or be used for headers in a FormSection?

src/components/Headers/ContentHeader.stories.tsx
henrysipp: Alternatively... 

```
type Props = {
  actions: React.ComponentProps<typeof Button>[]
}
```

And loop over those to render buttons... Yuck. Though we appear to already to that in the FormSectionLayout!

src/components/Headers/ContentHeader.tsx
abadami: Likely not for the FormSection's header, but yes for the FormSectionLayout's title and description.

src/components/Headers/ContentHeader.tsx
abadami: Ah, I left this in as it was the solution to fix an issue with it sticky scrolling not sticking that I've found in the story I made for this. This may be a situation of nuance that I'm not aware of for sticky horizontal scrolling and there may be a better way / suggestion for this.

src/components/Headers/ContentHeader.tsx
abadami: Honestly, though, in some respects I do wonder if I'm jumping the gun here. I'm seeing a couple examples of this pattern being followed of a "Content" based header but I do wonder if we should discuss it a bit more with design. I can see a few routes we go about this with what I'm kinda deeming as "Content Layouts" (e.g., `FormSectionLayout`, `GridTableLayout`)

src/components/Headers/ContentHeader.stories.tsx
bdow: I get what you're going for here @henrysipp. Though, I'm not sure if we can be that strict about what is displayed in this area to only Buttons. We are still reserving a slot for these actions via the proposed API, but just allowing the user to put in whatever the design calls for. **Unless...** @abadami are you seeing examples where this needs to be more than just buttons? Or can we be more strict with the API here?

bdow: If I recall, the only other thing in here was the potential "Saved ✅" flag. I'm not sure yet if that is how we're doing the auto save is finalized. And if we decide to put it here, we could pass in a specific 'autosave' prop here so the saving indicator always is in the same place next to any actions. With that in mind, I think maybe we can make this API more strict for actions. Then if users need to add other things to this area, we can expand the API to support other properties. Then we keep a consistent layout.
Per that thought, I think it'd be good to start out more restrictive and open up other options/props as we go. That'd be easier than trying to make things more restrictive after the fact

abadami: No we haven't decided on that quiet yet. I think that has been the only non-button component that we've discussed putting here though. 

I think it may be worth putting that `FormSectionActions` type and maybe making that a little more generic for wider use, but I think that could be saved for a quick follow-up.

bdow: Cool. Then I'd suggest making this a more strict API and not allowing the actions area to be just any ReactNode

stephenh: Just wandering by but historically Beam has preferred "pass props on the component like `title={...}` and `description={...}`" instead of "children with `Foo.Title` / `Foo.Description`" -- would generally prefer continuing that.

src/components/Headers/ContentHeader.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const tid = useTestIds(props, "contentHeader");
+  const descriptionEl = description && (
+    <div css={Css.sm.color(Tokens.OnSurface).$} {...tid.description}>
+      {description}
+    </div>
+  );
+
+  return (
+    <div
+      css={Css.sticky.left(documentScrollChromeLeft()).w(documentScrollChromeWidth()).px3.df.fdc.gapPx(12).$}
```
bdow: **suggestion:** The `px3` should be replaced with the shared horizontal padding constant.

bdow: I'm wondering if we'd need some sort of background color set on this component. As it isn't sticky to the top, we might not need it as things may not render behind it.... But maybe good to still have a `bg(Tokens.Surface)` just in case for some unforeseen reason.

src/components/Headers/ContentHeader.tsx
```diff
@@ -0,0 +1,66 @@
...
+          {actions && (
+            <div css={Css.df.gap1.fs0.$} {...tid.actions}>
+              {actions.map((action) =>
+                action.kind === "icon" ? (
+                  <IconButton key={action.icon} {...action} />
+                ) : (
+                  <Button key={`${action.label}`} {...action} />
+                ),
+              )}
+            </div>
```
bdow: **thought:** This actions section is repeated in other places - Maybe we can create a component that is like a "HeaderActions" that can consistently render this. I'm guessing they all render the same way between `PageHeader`, `WorkflowHeader`, and now `ContentHeader`?

src/components/Headers/ContentHeader.tsx
```diff
@@ -0,0 +1,66 @@
...
+          )}
+        </div>
+      )}
+      {title && descriptionEl}
```
bdow, line 50: **question:** Trying to follow how this logic works, so just typing it out while processing:
1. If there is a title, then render description under actions.
2. If there isn't a title, render description in line with actions
Is that correct?

src/components/Headers/ContentHeader.tsx
bdow: I don't think anyone would use `ContentHeader` in this case. If they just used mark up directly, it'll automatically do a `width: auto`, which `auto` would translate to be the size of its container. Most likely, in the case with a very wide table, then it'd be as wide as the document, which is the same width as the table.
I don't think there is a need for this suggestion. I can't think of a reason where they'd want this. 
I might also be not following completely, so let's chat on a VC.

src/components/Headers/ContentHeader.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {description}
+    </div>
+  );
+
+  return (
+    <div
+      css={Css.sticky.left(documentScrollChromeLeft()).w(documentScrollChromeWidth()).px3.df.fdc.gapPx(12).$}
+      {...tid}
+    >
+      {(title || description) && (
```
bdow: **question:** Is it intentional to not allow a header when the user only provides `actions`?

src/components/Headers/ContentHeader.tsx
abadami: That is correct. I could potentially make this a named boolean just to make that a little more clear.

src/components/Headers/ContentHeader.tsx
abadami: Yeah they likely would. I actually think we could use this in `FormSection` and `FormSectionLayout` as well since they would all theoretically be the same.

src/components/Headers/ContentHeader.tsx
abadami: Admittedly no, but it does lead to the question of "should we have `actions` only headers", and I feel like the answer is no anyway, but wdyt?

src/components/Headers/ContentHeader.tsx
abadami: Ah yeah, good call on both.

bdow (commented):
WorkflowLayout to add `Css.wfc.$`
WorkflowHeader to remove wrapping element and use `position: sticky` instead of `fixed`.
ContentHeader no longer needs the `min-width: fit-content;` workaround, as everything within the WorkflowLayout will get `min-width: fit-content`.

src/components/Headers/ContentHeader.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  it("omits the title/description/actions slots when not provided", async () => {
+    // Given a ContentHeader with no props set
+    const r = await render(<ContentHeader />);
+    // Then none of the optional slots render
+    expect(r.query.contentHeader_title).not.toBeInTheDocument();
+    expect(r.query.contentHeader_description).not.toBeInTheDocument();
+    expect(r.query.contentHeader_actions).not.toBeInTheDocument();
+  });
```
bdow: **thought:** I wonder how beneficial this test really is? I guess we can tell it omits certain sections, but it's such a non-real-world case that I'd almost expect that the component shouldn't render at all and just `return null;`

src/components/Headers/ContentHeader.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  it("puts description in the title's slot when there is no title", async () => {
```
bdow: **suggestion:** I think renaming this test to "omits title when not included" might be more appropriate. This doesn't necessarily validate that the description is in a specific place in the DOM, and I would expect that sort of test to be a visual test via chromatic/storybook. Then again, the last test validates that we can omit elements by not passing their props - so is this really testing anything new?
I think the only variation of rendering that can happen is that we _can_ pass actions, but if there's no title or description, then they won't render. Not sure if that's a valid use case, but probably more likely than passing no props at all 🤷

src/components/Headers/ContentHeader.tsx
bdow: I don't see why we wouldn't allow for it. Probably file to support, especially since its a very easy thing to support.

src/components/Headers/ContentHeader.tsx
bdow: Yea, I think that might help.
Also curious, what example are you working off from where we have only a description & actions, no title?

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -129,7 +131,7 @@ export function WorkflowLayout(props: WorkflowLayoutProps) {
         <div ref={spacerRef} css={Css.fs0.w100.$} style={{ height: headerHeight }} {...tid.spacer}>
           <div
             ref={headerMetricsRef}
-            css={Css.fixed.w(headerWidth).z(zIndices.pageStickyHeader).top(outerTop).$}
+            css={Css.sticky.left(headerLeft).w(headerWidth).z(zIndices.pageStickyHeader).top(outerTop).$}
```
bdow: **suggestion (non-blocking):** I don't think we necessarily need `left(headerLeft)`. When we're in a workflow, it _should_ be a full screen page experience w/o a sidenav, similar to how we won't have the global nav as part of it either. So we could just do `left0` here, imo. Doesn't hurt having what you have, as it'll be `0`. So more of just a thought, I guess.

bdow (approved):
Giving an approval with some small nit-picky things around title/description/actions rendering that I think would be easy to take care of in this PR w/o needing a re-review.

src/components/Headers/ContentHeader.tsx
abadami: Yeah that's fair enough. I'll aim to support it.

src/components/Headers/ContentHeader.tsx
abadami: That would be [this one](https://www.figma.com/design/OdgcptKNTLp0L30TqNUD6w/Material-Catalog-Cleanup?node-id=247-37821&m=dev)

src/components/Headers/ContentHeader.test.tsx
abadami: Actually you know what? I think we should enforce that. I'll go ahead and have it return null if none of those 3 props are provided.

PR #1353, opened 2026-07-30
feat: Allow IconButtons to be passed into a FormSection's actions props

abadami:
### Allow `IconButtons` to be passed into a `FormSection`'s actions props

We previously restricted it to just `ButtonProps`, but it's probably best if we have it less restrictive, as we already have a use case for `IconButtons` as an example.


src/forms/FormSection.stories.tsx
```diff
@@ -22,19 +23,19 @@ export function WithChildSections() {
   return (
     <FormSection
       title="Sub-Contractors"
-      actions={[{ label: "Add", onClick: () => {}, variant: "tertiary" }]}
+      actions={<Button label="Add" onClick={() => {}} variant="tertiary" />}
```
henrysipp: Same conniption as here: https://github.com/homebound-team/beam/pull/1352/changes#r3687553521

Some react-y wrapper around actions here so that we don't let users break out of the internal styling.

src/forms/FormSection.tsx
```diff
@@ -1,13 +1,12 @@
 import { ReactNode } from "react";
-import { Button, ButtonProps } from "src/components/Button";
 import { Css, Tokens } from "src/Css";
 import { useTestIds } from "src/utils";
 
 export type FormSectionProps = {
   title: string;
   description?: ReactNode;
   /** Rendered top-right of the title row, e.g. an "Add" button. */
-  actions?: ButtonProps[];
```
henrysipp, line 10: Alternatively why could this not have just been like `(ButtonProps | IconButtonProps)[]`

abadami: This is what I wanted to initially (and for Content Header as well), but the issue here is that `ButtonProps` and `IconButtonProps`'s key props are shared between the two (namely, `label` and `icon`). There's no surefire way to denote between the two unless you include a prop that you may not need to actually set (like action's `inc`).

I am exploring another option right now of maybe creating a specialized type for this. We'll see...

src/forms/FormSection.tsx
```diff
@@ -1,13 +1,23 @@
...
 import { useTestIds } from "src/utils";
 
+/**
+ * A single action rendered in a `FormSection`/`FormSectionLayout` title row —
+ * either a full `Button` or an icon-only `IconButton`.
+ *
+ * Tagged with `kind` (not `type`) because `ButtonProps` already has its own
+ * unrelated `type?: "button" | "submit" | "reset"` HTML-attribute field.
+ */
+export type FormSectionAction = ({ kind: "button" } & ButtonProps) | ({ kind: "icon" } & IconButtonProps);
```
bdow: I like this approach. Wondering if we can use `kind?: "default"` instead of `kind: button`. Being a bit nit-picky there, I know. But maybe we can also allow for omitting `kind` and that'll be our normal Button if that's the case.

src/forms/FormSection.tsx
```diff
@@ -32,9 +42,13 @@ export function FormSection(props: FormSectionProps) {
...
         {actions && (
           <div css={Css.df.gap1.fs0.$} {...tid.actions}>
-            {actions.map((action) => (
-              <Button key={`${action.label}`} {...action} />
-            ))}
+            {actions.map((action) =>
+              action.kind === "button" ? (
+                <Button key={`button-${action.label}`} {...action} />
+              ) : (
+                <IconButton key={`icon-${action.icon}`} {...action} />
```
bdow: **suggestion:** Can the `key` just be the `label` or `icon`? I don't think we need to concatenate anything here.

src/layouts/FormSectionLayout/FormSectionLayout.tsx
```diff
@@ -1,16 +1,17 @@
...
 import { useTestIds } from "src/utils";
 import { defaultTestId } from "src/utils/defaultTestId";
 
 export type FormSectionLayoutProps = {
   /** The form's own title — one level up from any `FormSection`'s title. */
   title: string;
   description?: ReactNode;
-  /** Rendered top-right of the title row, e.g. a "Save draft" button. */
-  actions?: ButtonProps[];
+  /** Rendered top-right of the title row, e.g. a "Save draft" Button. */
```
bdow: **nit:** Didn't notice this before, but the example is a "Save Draft" button - which I don't think would ever show up here. This would be like a "Add items" or something, right?

src/layouts/FormSectionLayout/FormSectionLayout.tsx
```diff
@@ -34,9 +35,13 @@ export function FormSectionLayout(props: FormSectionLayoutProps) {
...
         {actions && (
           <div css={Css.df.gap1.fs0.$} {...tid.actions}>
-            {actions.map((action) => (
-              <Button key={`${action.label}`} {...action} />
-            ))}
+            {actions.map((action) =>
+              action.kind === "button" ? (
+                <Button key={`button-${action.label}`} {...action} />
+              ) : (
+                <IconButton key={`icon-${action.icon}`} {...action} />
```
bdow: **nit:** same thing here: Can the key just be the label or icon?

src/forms/FormSection.tsx
abadami: Not a bad idea.

src/layouts/FormSectionLayout/FormSectionLayout.tsx
abadami: Yes, that would make more sense. I don't think we have a "Save Draft" button in any form section...plus that'd be weird.

src/forms/FormSection.tsx
henrysipp: Funny, I think I'd lean further into it!


Just so theres no chance for overlap in the future:
```
{
  kind: "Button"
  props: ButtonProps
}
```

bdow (approved):

PR #1355, opened 2026-07-30
feat: Adjust Collapsed State

abadami:
### StepperTab Collapsed State Adjustments

Adjusts the Collapse State so that it is blue if it has been completed or active, rather than it having to be completed and not active. This better matches the current design specs.

Also includes some nice to have updates to the StepperTab storybook that design requested a bit ago.


src/components/StepperTabs/StepperTab.tsx
```diff
@@ -16,16 +16,23 @@ export type StepperTabProps = {
   disabled?: boolean;
   /** Collapses the tab down to its colored bottom border only, hiding the label — for the mobile view */
   collapsed?: boolean;
+  /** Storybook-only visual state overrides for snapshotting pseudo-interactions. */
+  __storyState?: {
```
henrysipp, line 20: Is this a common pattern that we use elsewhere? Just seems like I smell to have to mock this. If so, please ignore.

henrysipp (approved):

src/components/StepperTabs/StepperTab.tsx
abadami: It's a common pattern we're starting to use now, yep!

PR #1357, opened 2026-07-31
fix: Keep structure in TreeSelectField filtering.

stephenh:
Before any type we filtered a `TreeSelectField` it become flat:

image

Now we keep the structure:

image

ANd use the same logic as GridTable, i.e. roughly:

* If the search matches the parent, show the parent + all children
* If the search matches a child, show the child + its parent
* etc

I also included:

* Keep the ability to expand/collapse mid-search
* Don't reset the search on click

ALl of this comes from working with non-trivially large `TreeSelectField`s on this page, where currently it's not great:

https://blueprint.homebound.com/libraries/design-packages/rp:2076/rpav:51313/required-options

### DIFF
--- a/src/components/Filters/TreeFilter.test.tsx
+++ b/src/components/Filters/TreeFilter.test.tsx
@@ -4,9 +4,11 @@ import { treeFilter, TreeFilterProps } from "src/components/Filters/TreeFilter";
 import { FilterDefs } from "src/components/Filters/types";
 import { NestedOption } from "src/inputs";
 import { HasIdAndName } from "src/types";
-import { click, render } from "src/utils/rtl";
+import { render, select } from "src/utils/rtl";
 import { zeroTo } from "src/utils/sb";
 
+// Pick options with `select` (which matches on `data-label`), never `getByRole("option", { name })`:
+// the by-name lookup cost ~270ms a call here and timed this file out in CI. See AGENTS.md.
 describe("TreeFilter", () => {
   it("can select root values (default)", async () => {
     // Given the tree filter filtering by root option values (default)
@@ -15,10 +17,7 @@ describe("TreeFilter", () => {
     expect(r.filter_tree).toHaveValue("All");
     expect(r.value).toHaveTextContent("{}");
     // When selecting some options
-    click(r.filter_tree);
-    click(r.getByRole("option", { name: "Grandparent 0" }));
-    click(r.getByRole("option", { name: "Child 1-0-0" }));
-    click(r.getByRole("option", { name: "Parent 1-1" }));
+    select(r.filter_tree, ["Grandparent 0", "Child 1-0-0", "Parent 1-1"]);
     // Then the filter's value is empty now that we're making selections
     expect(r.filter_tree).toHaveValue("");
     // Then the filter is set to only return the "root" values that are selected.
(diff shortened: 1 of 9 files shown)


src/components/Filters/TreeFilter.test.tsx
```diff
@@ -4,9 +4,11 @@ import { treeFilter, TreeFilterProps } from "src/components/Filters/TreeFilter";
 import { FilterDefs } from "src/components/Filters/types";
 import { NestedOption } from "src/inputs";
 import { HasIdAndName } from "src/types";
-import { click, render } from "src/utils/rtl";
+import { render, select } from "src/utils/rtl";
 import { zeroTo } from "src/utils/sb";
 
+// Pick options with `select` (which matches on `data-label`), never `getByRole("option", { name })`:
+// the by-name lookup cost ~270ms a call here and timed this file out in CI. See AGENTS.md.
```
stephenh, line 11: Turns out my preference for `data-testids` grows even stronger, see the AGENTS.md, but basically something about the jsdom impl of `getByRole`-with-a-name is very shitty b/c it evals CSS conditions for the entire DOM or something.

I only caught this by getting a CI timeout, asking Claude "wtf was that", it profiled the js, and came up with this.

src/inputs/internal/ListBox.tsx
```diff
@@ -21,7 +21,6 @@ type ListBoxProps<O, V extends AriaKey> = {
   loading?: boolean | (() => JSX.Element);
   disabledOptionsWithReasons?: Record<string, string | undefined>;
   isTree?: boolean;
-  allowCollapsing?: boolean;
```
stephenh, line 24: No more collapsing, we always keep the structure -- structure is the point of using a TreeSelect.

src/inputs/TreeSelectField/TreeOption.tsx
```diff
@@ -46,7 +45,7 @@ export function TreeOption<O>(props: TreeOptionProps<O>) {
     ref,
   );
   const isGroup = groupKeys.includes(item.key);
-  const canCollapse = allowCollapsing && !!option.children?.length;
+  const canCollapse = !!option.children?.length;
```
stephenh, line 48: Suggest "hide whitespace", mostly indentation changes after this

src/inputs/TreeSelectField/TreeSelectField.test.tsx
```diff
@@ -905,6 +1028,20 @@ describe("TreeSelectField", () => {
...
+/** Returns the labels of the currently-shown options, in menu order, i.e. `["Basketball", "NBA"]`. */
+function getOptionLabels(r: RenderResult): string[] {
+  return r.queryAllByRole("option").map((o) => o.textContent ?? "");
+}
+
+/** Returns the nesting level of each currently-shown option, based on its indentation, i.e. `[0, 1]`. */
+function getOptionLevels(r: RenderResult): number[] {
+  return r.queryAllByRole("option").map((o) => {
+    // The level is applied as `plPx(16 + level * 8)` on the option's `li`, i.e. `--paddingLeft: 24px`
+    const paddingLeft = Number.parseInt(o.closest("li")!.style.getPropertyValue("--paddingLeft"), 10);
```
stephenh, line 1040: Claude wrote this ... too fancy for my personal taste but 🤷

src/inputs/internal/ListBox.tsx
bdow: Fwiw, if I recall correctly, we chose to flatten at the time because we chose to only display options that matched the search term and not worry about weird issues like, "do parents always show?" "What if my search term results in only showing 2 out of 5 child options, and the I choose the parent of those two. Because I selected the parent, does that mean we select all 5 children or only the 2 visible?"

stephenh: Right -- I think that was pragmatic at the time, but if you visit the `...required-options` link in the PR description, I think the usability of that TreeSelect is just terrible / borders on unusable, and this PR essentially fixes it.

So I'd like to still ship this / some version of this.

> select parent when only two options shown

You're right that I'm not directly addressing/handling that in this PR -- right now selecting the parent really would "select the parent" so all 5 children.

Personally I think that is fine, but do you consider that a blocking issue?

bdow: No not a blocking issue. Just was doing some passing comments while on vacation.

bdow (approved):

PR #1358, opened 2026-07-31
fix: Fix nav links won't open in a new tab.

stephenh:
I regularly get tripped up that Nav links won't open in a new tab.

### DIFF
--- a/src/components/internal/MenuItem.tsx
+++ b/src/components/internal/MenuItem.tsx
@@ -1,5 +1,5 @@
 import { Node } from "@react-types/shared";
-import { useRef } from "react";
+import { type KeyboardEvent, type MouseEvent, useRef } from "react";
 import { useHover, useMenuItem } from "react-aria";
 import { Link, useNavigate } from "react-router-dom";
 import { TreeState } from "react-stately";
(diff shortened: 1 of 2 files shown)


src/components/internal/MenuItem.tsx
```diff
@@ -24,6 +24,14 @@ export function MenuItemImpl(props: MenuItemProps) {
   const navigate = useNavigate();
   const { hoverProps, isHovered } = useHover({});
   const tid = useTestIds(props);
+  // react-aria's `onAction` callback isn't given the triggering event, so we can't inspect
+  // modifier keys there. Capture whether the user cmd/ctrl/shift-clicked (which should open a
+  // new tab instead of navigating the current page) in the capture phase, before react-aria's
+  // press handling fires `onAction`, and read it back below. I.e. cmd-click -> `true`.
```
stephenh, line 30: Ngl I don't like capturing modifiers out-of-band like this, but the LLM says react-aria is making us do it 🤷

src/components/internal/MenuItem.tsx
```diff
@@ -34,8 +42,8 @@ export function MenuItemImpl(props: MenuItemProps) {
         }
         const { onClick } = menuItem;
         if (typeof onClick === "string") {
```
bdow, line 44: This makes me think `useMenuItem` or whatever hook we're using is the wrong one here. Might be interesting to ask in the React Aria github to see what they'd recommend.

bdow: This might be something: https://react-aria.adobe.com/Menu#links

Haven't looked too closely yet. But will add a ticket to improve this.

stephenh: @bdow if that is something that would replace the ~hack in this PR, do you want to just take over this PR / create another? (I'm ooo for a few days and haven't pulled up that react-aria link yet to take a look.)

bdow: I created a ticket to follow up for improving. I think this is fine for now. I can take over your Beam PRs while you're out and see them through still, though. Enjoy the time off.

stephenh: Great, thanks!

bdow (approved):

PR #1359, opened 2026-08-04
feat: Adjust FormSection description and actions for improved mobile exp

abadami:


bdow (approved):
These are so similar to the ContentHeader that I wonder if we could utilize ContentHeader here instead. And maybe `ContentHeader` would take a `level` property to know what size it should render. But can be a follow up

PR #1361, opened 2026-08-04
feat: Support discard confirmation modal in Workflow layouts

bdow:


src/components/Modal/ConfirmCloseModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+    continueText = "Continue Editing",
+  } = props;
+  const { modalState } = useBeamContext();
+
+  // TODO: Change to closeModal from useModal when canCloseChecks are reset
+  function closeModal() {
+    // Not using closeModal from useModal since the canClose checks are not reset
+    // after a close and could/will cause other close attempts to fail.
+    modalState.current = undefined;
+  }
```
abadami, line 33: **question (non-blocking)**: Could you speak a little bit to why this TODO is here? This is more for me to get a little history here, but I'm also a little curious on why these `canClose` checks are preventing use from using `useModal`.

src/layouts/WorkflowLayout/useUnsavedChangesGuard.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+type NavigationBlockerActions = {
+  /** Allow the blocked in-app navigation to continue (leave the page). */
+  proceed: () => void;
+  /** Abort the pending navigation and stay on the current page. */
+  cancelNavigation: () => void;
+};
+
+/** Gates Cancel, `useBlocker`, and `beforeunload` when `isDirty` reports unsaved changes. */
+export function useUnsavedChangesGuard(options: UseUnsavedChangesGuardOptions): {
```
abadami, line 21: **praise**: I think this was a smart decision to split this out into its own hook, as I could see this also being used for the Focused Form layout.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -30,6 +31,8 @@ export type WorkflowLayoutProps = Pick<BaseHeaderProps, "title" | "documentTitle
     steps: WorkflowLayoutStep[];
     /** The step shown initially (matched against `defaultTestId(step.label)`); falls back to the first step if omitted or if it doesn't match any step. Uncontrolled — the layout owns step navigation from here. */
     defaultStep?: string;
+    /** When this returns true, Cancel / in-app route changes / tab close require confirmation. */
+    isDirty?: () => boolean;
```
abadami, line 35: **thought**: So something that I don't think this covers but currently exists in the current Plan-Based bid setup is support for the close confirmation in between switching tabs, since each of those tabs is persistent. 

I don't think it particularly matters since everything is saved as a `draft` state anyway, but I could potentially see it coming up again.

src/components/Modal/ConfirmCloseModal.tsx
bdow: This is an existing component that was moved from the `/SuperDrawer` folder up a level so it was clear its no longer _just_ used for the SuperDrawer. 
This is probably a pretty old TODO that I'd need to look into more. SuperDrawer has this concept of `canClose` checks that will run before the user closes the SuperDrawer to handle this same sort of "Are you sure?" behavior.
In the `useModal` there is also `// TODO: Should remove checks`:
``` 
      closeModal() {
        // TODO: Should remove checks
        for (const canCloseModal of modalCanCloseChecks.current) {
          if (!canCloseModal()) {
            return;
          }
        }
        maybeCall(modalState.current?.onClose);
        modalState.current = undefined;
      },
```

This seems to be a little bit bigger of a lift to address these TODO's, and would prefer to do that as a follow up. Can create a ticket to address this at a later date, maybe during some hackday code cleanup.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
bdow: I don't think this would be a problem given to the suggested set up when there is a `formState` per step. The docs suggest this: https://github.com/homebound-team/beam/blob/7fc2cc0e7f3c6b2a9ef4dfe447ca7cb5d59712e8/docs/layouts.md#multiple-form-states-one-per-step

TLDR - When there are multiple `formState`'s, then each are created at the layout level. They are passed into their respective Step's from there. Since the formState lives as part of the layout, not as part of each individual step, then we don't lose in progress work while switching between steps.

abadami: Yes, I see. I only mention it because I don't think Create Plan-Based bid packages are a multi-form setup, so it needed to have the discard modal in between tabs. If we intend to switch it to a multiform setup though then it's no problem, which if we're able to that'd be great I think. Makes the most sense based on what it's doing.

src/forms/WorkflowLayoutMultiFormApp.tsx
```diff
@@ -0,0 +1,194 @@
...
+type ProjectBasics = { name?: string | null; code?: string | null };
+type ProjectContact = { email?: string | null; phone?: string | null };
+type ProjectNotes = { summary?: string | null };
+
+/**
+ * Demos `WorkflowLayout` when each step owns its own form-state: keep forms alive at the page,
+ * load step data on demand via `init` (so hydration is not dirty), and OR their `.dirty` flags
+ * into `isDirty`.
+ */
+export function WorkflowLayoutMultiFormApp() {
```
abadami, line 17: **suggestion**: Perhaps adjust the setup here to make it a little more clear? It's a little hard to follow since it's doing things we wouldn't normally do.

abadami (approved):

PR #1362, opened 2026-08-04
feat: Allow Form Section Children to be draggable

abadami:
## feat: Allow Form Section Children to be draggable

Allows the option for Form Section Children to be draggable by introducing an `orderField` prop, which keys off form state. This PR introduces internally the splitting of `FormSection` and `FormSectionChild` rather than a single recursive `FormSection`. 

This also includes a couple fixes to how DnDGrid applied styles, mostly in introducing a `snapshot` and `restore` mechanism on styles, and helping it work with Truss.


src/components/DnDGrid/DnDGridItemHandle.tsx
```diff
@@ -1,7 +1,9 @@
 import { mergeProps, useFocusRing, useHover } from "react-aria";
-import { Css, Icon, IconKey, useTestIds } from "src";
 import { BeamColor } from "src/colors";
 import { DnDGridItemProps } from "src/components/DnDGrid/useDnDGridItem";
+import { Icon, IconKey } from "src/components/Icon";
+import { Css } from "src/Css";
+import { useTestIds } from "src/utils";
```
bdow, line 6: 👍 🙏

src/forms/FormSection.tsx
```diff
@@ -1,75 +1,194 @@
...
+  actions?: ButtonProps[];
   fields?: ReactNode;
-  childSections?: Omit<FormSectionProps, "isChild">[];
+  childSections?: Omit<FormSectionProps, "isChild" | "dragHandle">[];
   /** Nested-section styling (smaller title). Set automatically by `FormSection`'s own recursion over `childSections` — don't set this yourself. */
   isChild?: boolean;
+  /**
+   * Stable identifier for this section. Used as the React key for `childSections`, and — when the
+   * parent section has `draggableChildSections` set — as the id returned in `onReorderChildSections`'s
+   * new-order array. Falls back to `title`, then array index, when omitted; supply a stable `id`
```
bdow: **thought:** Feels like if we don't include an `id` then we're potentially creating bugs via these fallbacks. As mentioned, there could be more than one of the same `title`. And if the `key` falls back to the `index` value, then when we reorder via drag & drop, then indices all change again, right?
My thought for an approach here is maybe there's a separate type that represents a formsection that has re-orderable elements. There's a `type FormSectionBase` and `type ReorderableFormSection` (or something like that), which would extend from `FormSectionBase`. And that'll require the fields like `id` and whatever else, so we don't need to make things optional.

src/forms/FormSection.tsx
```diff
@@ -1,75 +1,194 @@
...
+   * `FormSection` renders a `<input type="hidden">` mirroring `orderField.value` (falling back to the
+   * section's position among its siblings when `value` is `null`/`undefined`), permutes it into sync via
+   * `orderField.set(...)` whenever a drag/keyboard reorder commits — reassigning the *existing* set of
+   * order values held by these `childSections` rather than writing fresh 0-based indices, so it stays
+   * correct even when those values aren't zero-based (e.g. they share a numbering space with sections
+   * outside this `childSections` list) — and sorts `childSections` by `orderField.value` on every render
+   * (only takes effect when *every* childSection has one; otherwise array order is preserved). Also
+   * unlocks dragging on its own, without requiring `onReorderChildSections`. Has no effect outside a
+   * draggable `childSections` list.
+   */
```
bdow: **suggestion:** Can we make these comments more concise? They're so long right now.

src/forms/FormSection.tsx
```diff
@@ -1,75 +1,194 @@
...
+  orderField?: FieldState<number | null | undefined>;
+  /** @internal Set automatically by `FormSection`'s own recursion when rendering a draggable `childSections` entry — don't set this yourself. */
+  dragHandle?: ReactNode;
```
bdow: **question:** Can we infer this? Especially if we have gate of something like 
```
if (isReorderableFormSection(props)) { 
  // then render drag handle here.
```

src/forms/FormSection.tsx
```diff
@@ -1,75 +1,194 @@
...
+   * drag handle rendered next to each child section's title. Has no effect without `childSections`,
+   * and is a no-op unless `onReorderChildSections` is also provided.
+   */
+  draggableChildSections?: boolean;
+  /**
+   * Called with the new order of `childSections`' `id`s (falling back to `title`, then index) once a
+   * drag-and-drop or keyboard reorder completes. Provide together with `draggableChildSections` to enable
+   * reordering; omitting either leaves rendering/behavior unchanged.
+   */
+  onReorderChildSections?: (newOrder: string[]) => void;
```
bdow: **question:** Similar to my last comment, can we infer any of these by the existence of the other one? Like if `onReorder` is true, then we know from them on out, we're rendering re-orderable child sections. With recursion, we'll probably want a simple boolean flag one we start rendering children though - might be weird to continue to pass that callback around. In that case, then I'd probably want to have more type enforcement, like create a 
```
// Add only props parents can have
type FormSectionParent = FormSectionBase & {
  onReorder: () => {};
  isChild?: false;
};

// Add only props  a child can have
type FormSectionChild = FormSectionBase & { 
  isChild: true; 
  draggable?: Boolean 
};
```

Then we can do a type union in the props and conditionally render specific elements and handle certain cases. If this explodes too much, then maybe we're looking at breaking out of recursion and having separate FormSectionParent and FormSectionChild components.

src/forms/FormSection.tsx
```diff
@@ -1,75 +1,194 @@
...
+function sortByOrderField(
+  childSections: NonNullable<FormSectionProps["childSections"]>,
+): NonNullable<FormSectionProps["childSections"]> {
+  if (!hasCompleteOrderFields(childSections)) return childSections;
+  return [...childSections].sort((a, b) => a.orderField!.value! - b.orderField!.value!);
+}
+
+function hasCompleteOrderFields(childSections: NonNullable<FormSectionProps["childSections"]>): boolean {
+  return childSections.every((c) => c.orderField && isDefined(c.orderField.value));
+}
```
bdow: **thought:** I'm not a huge fan of having to do this sort of check so that we can use things like `b.orderField!.value!` and not blow up. I think this also means that consumers will need to ensure that when adding a new item to the list, they initially populate the `orderField`. This could be a foot-gun a little bit in that sense. As right now when using the `formState.add` method, then it can create new empty objects and might not fill out that `orderField`. 
So it makes me wonder if we can, on our side, if we should help out with that at all 🤔.  It also looks like we're doing this just so we can get a value, which I don't think we necessarily need - instead we just need to query the list of items within this DnDGrid, `.map` over them, and register their `index` as the order.

src/forms/FormSection.tsx
```diff
@@ -1,75 +1,194 @@
...
+    isChild = false,
+    draggableChildSections = false,
+    onReorderChildSections,
+    dragHandle,
+  } = props;
   const tid = useTestIds(props, "formSection");
+  const hasOrderFields = !!childSections?.some((c) => !!c.orderField);
+  const isDraggable = draggableChildSections && (!!onReorderChildSections || hasOrderFields);
+
+  const handleReorder = (newOrder: string[]) => {
```
bdow: **suggestion:** This is where I was saying I think the `handlReorder` callback can be updated to include a `gridRef: HTMLElement`. Then on reorder, we just query for the `gridRef.current.querySelectorAll([input[name="orderField"]])`, and loop through those, as they'll be their order in the DOM. And on that loop we can do a `orderField.set(index+1)` sort of thing. Not exactly sure how you connect the childSection in the DOM to the childSection in the `handleReorder` - maybe that is via `id` still. And if `id` is set, then we can omit the whole idea of `gridRef` param, and just use the map we get back or the ids in the new order.

src/forms/FormSection.tsx
```diff
@@ -1,75 +1,194 @@
...
+      {childSections &&
+        (isDraggable ? (
+          <DnDGrid onReorder={handleReorder} gridStyles={Css.gtc("minmax(0, 1fr)").gap3.$}>
+            {/* Observer: `sortByOrderField` reads each child's `orderField.value` — without this, the
+                order here would only refresh on some unrelated re-render, not on the `.set()` calls
+                `handleReorder`/consumers make directly to those FieldStates. */}
+            <Observer>
+              {() => (
+                <>
+                  {sortByOrderField(childSections).map((child, i) => (
```
bdow: **question:** I wonder how necessary this sort is at render time like this. I think I'd rather do one of two things:
1. Expect the user will send these in the correct order
2. Create a pre-sorted shallow copy of the child sections before mapping over them. This should be a one-time thing no need to be observed via mobx

src/forms/FormSection/FormSection.test.tsx
```diff
@@ -0,0 +1,230 @@
...
+  });
+
+  it("does not wrap childSections in a DnDGrid when only some of them set orderField", async () => {
+    // Given a FormSection with a mixed (unsupported) childSections array -- only one sets orderField
+    const r = await render(
+      <FormSection
+        title="Trade Partners"
+        childSections={[
+          { id: "electrical", title: "Electrical", orderField: orderField(0) },
+          { id: "plumbing", title: "Plumbing" },
```
bdow: I haven't seen the types yet, but was assuming this would throw a type error because you can't have both a child with and without an orderField

src/forms/FormSection/FormSectionChild.tsx
```diff
@@ -0,0 +1,71 @@
...
+/** A single entry in a `FormSection`'s `childSections` — never itself nests further children. */
+export type FormSectionChildProps = Omit<FormSectionProps, "childSections"> & {
+  id: string;
+  /** Drives draggability: set on every child for a draggable, order-sorted list; omit on all for a plain list. Mixing is unsupported and not type-checked. */
+  orderField?: FieldState<number | null | undefined>;
+};
+
+type FormSectionChildComponentProps = FormSectionChildProps & {
+  /** @internal computed by the parent `FormSection`; not caller-supplied. */
+  isLast: boolean;
```
bdow: Question: did you try this with the css approach via last-of-type, or last-child? Could be nice to not need this.

src/forms/FormSection/FormSectionChild.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { title, description, actions, fields, orderField, isLast } = props;
+  const tid = useTestIds(props, "formSectionChild");
+  const itemRef = useRef(null);
+  const isDraggable = !!orderField;
+  const { dragItemProps, dragHandleProps } = useDnDGridItem({ id: props.id, itemRef });
+
+  return (
+    <div
+      ref={isDraggable ? itemRef : undefined}
+      {...(isDraggable ? dragItemProps : {})}
```
bdow: Might be able to combine these two lines together. 
```suggestion
      {...(isDraggable ? {...dragItemProps, ref: itemRef} : {})}
```

src/forms/FormSection/FormSection.test.tsx
bdow: I see why - `orderField` is just an optional field. I was more-so expecting a type union of:
```
type FormSectionChildBase = { ... };
type FormSectionChild = FormSectionChildBase & {
  orderField?: never;  // Ensures this cannot be set
};

type ReorderableFormSectionChild = FormSectionChildBase & {
  id: string;
  orderField: FieldState<...> // Required prop
};

type Props = {
  ...,
  childSections: FormSectionChild[] | ReorderableFormSectionChild[];
}
```
I think this would enforce you can't do some with and some without the `orderField`.

src/forms/FormSection/FormSection.test.tsx
```diff
@@ -0,0 +1,230 @@
...
+    );
+    // Then they render in orderField order (Electrical first), not array order
+    expect(r.formSection_childSection_title_0).toHaveTextContent("Electrical");
+    expect(r.formSection_childSection_title_1).toHaveTextContent("Plumbing");
+  });
+
+  it("permutes existing (non-zero-based) order values rather than resetting them to 0-based indices", async () => {
+    // Given two childSections whose orderFields start at non-zero values (2 and 3)
+    const electricalOrder = orderField(2);
+    const plumbingOrder = orderField(3);
```
bdow, line 194: Huh... feels like an edge case, but doesn't hurt to support 👍

src/forms/FormSection/FormSection.tsx
```diff
@@ -0,0 +1,105 @@
...
+export type FormSectionAction =
+  | ({ kind?: "default" } & ButtonProps)
+  | ({ kind: "icon" } & Omit<IconButtonProps, "variant">);
+
+export type FormSectionProps = {
+  title: string;
+  description?: ReactNode;
+  actions?: FormSectionAction[];
+  fields?: ReactNode;
+  childSections?: FormSectionChildProps[];
```
bdow: **suggestion:** So, I think here is where you'd want a type union. Here's an approach you could take 
```
childSections: FormSectionChild[] | ReorderableFormSectionChild[];

function isReorderableChildSections(
  childSections: FormSectionChild[] | ReorderableFormSectionChild[],
): childSections is ReorderableFormSectionChild[] {
  return childSections.length > 0 && childSections.every((c): c is ReorderableFormSectionChild => c.orderField != null);
}
```

Then I actually think you might be better off splitting to two different components based on if its draggable or not, so we can get rid of the `c.orderField!.value` below. 

Something like:
```
if (childSections && isReorderableChildSections(childSections)) {
  // orderField is required, no `!`
  // render small in-file component that encompasses handleReorder method & DnDGrid
  <DraggableChildren childSections={childSections} />
} else {
  // plain list or empty - render as normal  
}
```

src/forms/FormSection/FormSection.tsx
```diff
@@ -0,0 +1,105 @@
...
+                        {...child}
+                        isLast={i === sorted.length - 1}
+                        {...tid.childSection}
+                      />
+                    ))}
+                  </>
+                );
+              }}
+            </Observer>
+          </DnDGrid>
```
bdow: question: Why does this need to be observed?

src/forms/FormSection/FormSection.tsx
```diff
@@ -0,0 +1,105 @@
...
+          </div>
+        )}
+      </div>
+      {fields}
+      {childSections &&
+        (isDraggableParent ? (
+          <DnDGrid onReorder={handleReorder} gridStyles={Css.gtc("minmax(0, 1fr)").gap3.$}>
+            <Observer>
+              {() => {
+                const sorted = sortByOrderField(childSections);
```
bdow: **question:** If you go with my suggestion and make a new component specific for the DnD branch of this logic, could we start the component off with a `const sortedChildren = sortByOrderField(childSections);`
```
function DraggableChildren(childSections: ReorderableChildSections) {
  const sortedChildren = sortByOrderField(childSections);
  return (
    <DnDGrid ... >
      {sortedChildren.map(...)}
    </DnDGrid>
  );   
}
```

bdow (commented):
Looking good. I have a couple comments I'd like to see addressed first regarding:
1. Tighter typing to prevent orderable and non-orderable children in the same `childSections` list.
2. branching the FormSections component based on the childSections type to contain DnD specific logic and only create when needed. Also should remove the need to use `!` syntax.

bdow:
One other thing I noticed was that the placeholder element is having a box shadow applied, and I don't think we want that. Just a box-shadow on the element being moved around.
Screenshot 2026-08-06 at 10 28 49 PM

src/forms/FormSection/FormSectionChild.tsx
abadami: It's used for the draggable child sections. I think there is some styling weirdness that will go on if we use the css version. That being said I haven't actually verified it, so let me do so first.

src/forms/FormSection/FormSection.tsx
abadami: I originally thought it's because we're looking at `orderField`, but after some testing without it. Nah, it's not needed. I imagine since we're changing the `orderField` for each section via props it's able to detect updates alright.

src/forms/FormSection/FormSectionChild.tsx
abadami: Nope, we didn't need it.

src/forms/FormSection/FormSection.tsx
```diff
@@ -0,0 +1,102 @@
...
+function isReorderable(
+  childSections: PlainFormSectionChild[] | ReorderableFormSectionChild[],
+): childSections is ReorderableFormSectionChild[] {
+  return childSections.length > 0 && childSections.every((c) => !!c.orderField);
+}
+
+type DraggableChildrenProps = { childSections: ReorderableFormSectionChild[] };
+
+/** Renders reorderable childSections, sorted by `orderField.value`, in a `DnDGrid`. */
+const DraggableChildren = function DraggableChildren(props: DraggableChildrenProps) {
```
bdow: **request:** We don't use this syntax. can we just do `function DraggableChildren() {...}`?

abadami: Holy cow that slipped by me.  I think I missed this since this was originally wrapped in an observer before I got rid of it.

src/forms/FormSection/FormSection.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { childSections, ...tid } = props;
+  const sorted = sortByOrderField(childSections);
+
+  const handleReorder = (newOrder: string[]) => {
+    const existingValues = sorted.map((c) => c.orderField.value ?? 0);
+    const childById = new Map(sorted.map((c) => [c.id, c]));
+    newOrder.forEach((id, i) => childById.get(id)?.orderField.set(existingValues[i]!));
```
bdow: **question:** Why do we need this `!` here? makes me wonder if we can set this up differently so we don't need the `!`.
This reorder logic confuses me. It might be a lack of sleep thing, but I think some comments or something would selfishly help me follow better.
Will type it out as I see it:
1. Get the existing sort values as an array.
2. Make a map of the id -> child section.
3. for each id in `newOrder`, get that child and set the existing value based on the index...
That number three doesn't track with me quite well immediately. Why do we use the existing value to define the new sort value? 
OOOoooooOOO it just came to me. This is what that test "edge case" helps with. Because it just says "whatever the existing sort value was in the first slot, set that now to the current child that is in the first slot."
So the need for the ! now is understood. And potentially brittle, right? If the `newOrder` ever had a different size than the existing values, then this ! would throw a runtime error with an array out of bounds error - or maybe it'd return undefined and we'd set orderField to undefined. But either way, maybe this ability to handle the edge case is just creating a potential error case. Granted, they should always be the same length. But I'd rather have a better guard around this sort of thing and avoid the `!`.

src/components/DnDGrid/DnDGrid.test.tsx
bdow, line 1: **praise:** Thank you for creating this test file.

src/components/DnDGrid/DnDGrid.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  it("commits a keyboard-driven reorder and cleans up the active styles", async () => {
+    // Given a DnDGrid with two items and a reorder spy
+    const onReorder = vi.fn();
+    const r = await render(<Harness onReorder={onReorder} />);
+
+    // When the user grabs the first item via keyboard, moves it down one, and commits
+    const handle = r.dragHandle_0;
+    fireEvent.keyDown(handle, { key: " " });
+    fireEvent.keyDown(handle, { key: "ArrowDown" });
+    fireEvent.keyDown(handle, { key: "Enter" });
```
bdow, line 72: accessibility ftw! It'd be hard doing this in tests without having support keyboard interactivity.

src/components/DnDGrid/DnDGrid.tsx
```diff
@@ -163,6 +165,9 @@ export function DnDGrid(props: DnDGridProps) {
         // And finally place it in the DOM after the element being dragged. If there is no `nextSibling`, then it is appended to the grid element.
         gridEl.current.insertBefore(cloneEl.current, dragEl.current.nextSibling);
 
+        // Now that the clone has captured the item's normal appearance, apply the active/reorder styling to the real element.
+        initReorder();
+
```
bdow, line 170: **praise:** Appreciate the comment.

src/forms/FormSection/FormSection.tsx
abadami: Ah, yeah I can see how this can be confusing and a little odd. I'm about to push in a similar reorder that just doesn't use the `!`, it checks to see if the existing value is undefined instead and if it is, just don't do anything. I don't think the value should ever really be undefined so it shouldn't cause any issues.

src/forms/FormSection/FormSection.tsx
```diff
@@ -0,0 +1,102 @@
...
+): childSections is ReorderableFormSectionChild[] {
+  return childSections.length > 0 && childSections.every((c) => !!c.orderField);
+}
+
+type DraggableChildrenProps = { childSections: ReorderableFormSectionChild[] };
+
+/** Renders reorderable childSections, sorted by `orderField.value`, in a `DnDGrid`. */
+function DraggableChildren(props: DraggableChildrenProps) {
+  const { childSections, ...tid } = props;
+  const sorted = sortByOrderField(childSections);
```
bdow, line 83: **thought (non-blocking):** A potential perf improvement would be to see if this sort happens every time we reorder. I'm guessing it does because `handleReorder` uses this as the source of truth the current order.

abadami: I'm guessing it does too. I think it's something to keep our eye on but I don't think this will have any issues currently. At least with any of the pages that I know of that uses this.

bdow: Do we have tests that sort multiple times? to ensure that we're not always reading off the original sort order when reapplying the new sort values? And ensure we're reading off the latest sort order

abadami: Ooo, this is a good call. I think you called it exactly and it's reading off the wrong sort order in `handleReorder` after the first sort. We may have to do the sort in the reorder function here unfortunately unless you can see an alternative. FWIW, I don't think doing the sort again is that taxing on performance atm...

bdow (approved):

PR #1363, opened 2026-08-05
fix: Improve Stepper visuals & state determination

bdow:


src/components/StepperTabs/StepperTab.tsx
```diff
@@ -44,50 +44,70 @@ export function StepperTab(props: StepperTabProps) {
...
       {...tid[defaultTestId(value)]}
     >
-      <span css={Css.lineClamp1.$}>{label}</span>
-      {completed && (
-        <span css={Css.fs0.ml1.$}>
-          <Icon icon="check" inc={2.5} {...tid.check} />
-        </span>
+      {!collapsed && (
+        <>
+          <span css={Css.dg.jic.mw0.$}>
```
abadami, line 55: **question**: Just to check, we're using display grid here purely for making sure the text here overlaps, to better help with making sure the layout doesn't shift?

As a side note, if we use this method quite a bit, I wonder if we should wrap the behavior in a utility function / component somehow 🤔 . Could be useful.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -44,50 +44,70 @@ export function StepperTab(props: StepperTabProps) {
...
+                .if(completed)
+                .o100.add("transform", "scale(1) translateY(0%)").$
+            }
+          >
+            <Icon icon="check" inc={2.5} {...tid.check} />
+          </span>
+        </>
       )}
+      {/* Absolute so active vs inactive thickness never changes the tab's layout height. */}
+      <span aria-hidden css={getIndicatorStyles(active, completed, disabled)} {...tid.indicator} />
```
abadami: **suggestion**: I feel like this comment maybe belongs in the `getIndicatorStyles` function proper? I thought this was a leftover comment from a previous iteration for a little bit before I realized it was talking about the function's return value.

abadami: **question**: Going away from using a css border does make sense, but I'm curious as to what the obstacles were that caused you to use an element rather than a css border. Was it purely layout shifting?

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -19,10 +19,16 @@ import { useMeasuredHeight } from "../useMeasuredHeight";
...
+/**
+ * A `WorkflowLayout` step: a `StepperTabsStep` (minus `value`, which is derived from `label`) plus
+ * page content. Stepper chrome marks prior steps complete via current-step index; `isValid` only
+ * gates the Continue/Complete CTA for the active step.
+ */
 export type WorkflowLayoutStep = Omit<StepperTabsStep, "value"> & {
   /** Rendered as the page body while this is the active step. */
   content: ReactNode;
+  /** Gates Continue/Complete for the active step (form validity) */
+  isValid: boolean;
```
abadami, line 33: **thought (non-blocking)**: This makes sense matching form-state's conventions. I wonder if it'd be worth having the stepper tab match up as well rather than still use `completed`. Though I suppose that's still the name of the state, probably very nit-picky to switch it too.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -44,50 +44,70 @@ export function StepperTab(props: StepperTabProps) {
...
+              {label}
+            </span>
+            <span css={Css.add("gridArea", "1/1").lineClamp1.wordBreak("break-all").if(active).smSb.$}>{label}</span>
+          </span>
+          <span
+            css={
+              Css.fs0.ml1.transitionAll.o0
+                .add("transform", "scale(0.75) translateY(100%)")
+                .if(completed)
+                .o100.add("transform", "scale(1) translateY(0%)").$
```
abadami, line 70: **praise**: Transition animation for this looks nice and clean. I like it a lot!

src/components/StepperTabs/StepperTabs.test.tsx
```diff
@@ -3,33 +3,41 @@ import { click, render } from "src/utils/rtl";
...
     expect(r.stepperTabs_step_2).toHaveAttribute("aria-current", "false");
-    expect(r.stepperTabs_tab_send.querySelector("[data-icon='check']")).not.toBeInTheDocument();
+    const sendCheck = r.stepperTabs_tab_send.querySelector("[data-icon='check']");
+    expect(sendCheck).toBeInTheDocument();
+    expect(sendCheck?.parentElement).toHaveStyle({ opacity: "0" });
   });
 
-  it("shows the check icon for a step that is both active and completed", async () => {
-    // Given the first step is both completed and current
+  it("keeps the check hidden on the active step", async () => {
```
abadami: **suggestion**: I believe this test is a little redundant as this is being checked by the first test since that checks the state of all current steps, and there isn't really a difference between checking different current steps in the order really anymore. I think we can omit this one.

src/components/StepperTabs/StepperTab.tsx
```diff
@@ -44,50 +44,70 @@ export function StepperTab(props: StepperTabProps) {
...
   return {
-    ...Css.cursor("default").hPx(0).py0.$,
-    ...(active || completed ? Css.bcBlue600.$ : Css.bcGray300.$),
+    ...Css.absolute.left0.right0.bottom0
+      .hPx(active ? activeIndicatorHeightPx : inactiveIndicatorHeightPx)
+      .bgColor(Tokens.FieldBorderDefault)
+      .if(active || completed)
+      .bgColor(Tokens.Primary)
+      .if(disabled)
+      .bgColor(Tokens.FieldBorderDefault).$,
```
abadami: **suggestion**: Couldn't the `bgColor` portion of this be restructure to not include the disabled check to make it a little more concise? Something like:

```typescript
.bgColor(Tokens.FieldBorderDefault)
.if(active || completed)
.bgColor(Tokens.Primary).$
```

src/components/StepperTabs/StepperTab.test.tsx
```diff
@@ -44,21 +54,37 @@ describe("StepperTab", () => {
...
+  });
+
+  it("uses a 2px indicator when inactive", async () => {
+    // Given an inactive tab
+    const r = await render(
+      <StepperTab label="Step Label" value="step" active={false} completed={false} onClick={vi.fn()} />,
+    );
+    // Then its indicator is 2px tall
+    expect(r.stepperTab_indicator).toHaveStyle({ height: "2px" });
   });
```
abadami: **suggestion**: I wonder if we could just combine these with the other two indicator tests, since I believe they happen at the same time as their respective states (3px & primary color for active, 2px & default color when inactive)

src/layouts/WorkflowLayout/WorkflowLayout.tsx
bdow: Those feel different. completed means you're done with it. valid means you've filled out required fields. You might not be "done" yet, but atleast everything you've entered so far is valid.

src/components/StepperTabs/StepperTab.tsx
bdow: Yes, layout shifting. Changing the border height based on the active step moves the inner content.

src/components/StepperTabs/StepperTab.tsx
bdow: Yes, this is to put elements on top of each other by making them sit with the same grid cell. I don't see this as being something we would use too often, but good thought on whether or not it'd be worth it.

src/components/StepperTabs/StepperTabs.test.tsx
bdow: Will do

src/components/StepperTabs/StepperTab.tsx
bdow: Good call. Didn't see that. Will fix.

src/components/StepperTabs/StepperTab.test.tsx
bdow: Sure, makes sense. Thanks!

src/components/StepperTabs/StepperTab.tsx
bdow: I'm actually not sure how easy it would be to abstract. A lot of the styling and element choices are very specific to this component. Maybe if we end up needing to do this again it'll be more clear where abstraction makes sense.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
abadami: That's a good way to think about it. Aight, sounds good.

abadami (approved):

PR #1364, opened 2026-08-05
feat: StepperTabs Collapse on Scroll in WorkflowLayout

abadami:
## StepperTabs Collapse on Scroll in WorkflowLayout

Implement the collapse on scroll behavior specific to WorkflowLayout, rather than using the `useAutoHideonScroll` hook, as their needs are different.


src/layouts/WorkflowLayout/WorkflowLayout.stories.tsx
```diff
@@ -29,6 +29,23 @@ export function Default() {
...
+  return (
+    <WorkflowLayout
+      title="Workflow Layout"
+      onCancel={action("cancel clicked")}
+      completeLabel="Save"
+      onComplete={action("complete clicked")}
+      steps={makeSteps(50)}
+    />
+  );
+}
```
bdow: **suggestions:** If we want this story to validate it collapses on scroll, then I think we should add a `play` function that scrolls the page down far enough to trigger that behavior before the snapshot. I think we do this elsewhere, so there could be a good thing to base this on. Also, would prefer to do it via the play function in the `newStory` helper. Something like `newStory(<Component />, { play: () => {} }`

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -31,35 +31,99 @@ export type WorkflowLayoutProps = Pick<BaseHeaderProps, "title" | "documentTitle
...
+  useLayoutEffect(() => {
+    lastScrollHeightRef.current = document.documentElement.scrollHeight;
+  }, [headerHeight]);
 
-  // TODO: revisit scroll-driven collapsing (see plan); collapsed on mobile only for now.
-  const collapsed = isMobile;
+  useLayoutEffect(() => {
+    if (isMobile) return; // Mobile always collapses (see `collapsed` below) — no need to track scroll.
+
+    const updateScrollCollapsed = () => {
```
bdow: **thought:** This does feel like a good candidate for a separate hook. Mostly to reduce the lines of code of this file.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -31,35 +31,99 @@ export type WorkflowLayoutProps = Pick<BaseHeaderProps, "title" | "documentTitle
...
+      const doc = document.documentElement;
+      const currentY = window.scrollY;
+      const threshold = bannerAndNavbarHeightRef.current + headerHeightRef.current + SCROLL_COLLAPSE_THRESHOLD_PX;
```
bdow: **thought:** I feel like this `SCROLL_COLLAPSE_THRESHOLD_PX` might not be needed. I think the threshold of the banner height + the header's height would be sufficient. Would be interesting to try it out and see if the experience is wonky or just fine without the extra offset

src/layouts/WorkflowLayout/WorkflowLayout.tsx
abadami: Fair enough, I'll move it out.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
abadami: Yeah this might've been a remnant of the previous implementation that kinda stuck it out through iterations. I have removed it and I don't even notice a difference, so I think this'll do just fine without it.

src/layouts/WorkflowLayout/WorkflowLayout.test.tsx
```diff
@@ -200,6 +200,20 @@ describe("WorkflowLayout", () => {
...
+
+    // When the page scrolls down past the header's resting position
+    scrollWindow(0, { clientHeight: 800, scrollHeight: 1_000_800 });
+    scrollWindow(300, { clientHeight: 800, scrollHeight: 1_000_800 });
+
+    // Then the tabs collapse to a non-interactive indicator bar
+    // (the scroll-position state machine itself is covered by useScrollCollapse's own tests)
+    click(r.header_stepperTabs_tab_stepTwo);
+    expect(r.query.stepTwoBody).not.toBeInTheDocument();
+  });
```
bdow: **nit:** This seems to be testing that the element becomes disabled, not that it collapses. Maybe validating a `height: 0px` would be more of what you're looking for here.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -55,11 +58,18 @@ export function WorkflowLayout(props: WorkflowLayoutProps) {
...
 
+  const bannerAndNavbarHeight = useBannerAndNavbarHeight();
+
   const headerMetricsRef = useRef<HTMLDivElement>(null);
   const headerHeight = useMeasuredHeight(headerMetricsRef, true);
 
-  // TODO: revisit scroll-driven collapsing (see plan); collapsed on mobile only for now.
-  const collapsed = isMobile;
+  // No DOM anchor needed: WorkflowLayout only nests under `EnvironmentBannerLayout`, so the header
+  // always rests exactly `bannerAndNavbarHeight + headerHeight` px below the document top — plain
```
bdow: The AI likes to make comments based on its local history, which is what this feels like. I think we can omit, or make these comments more concise. If I read this without context I'd wonder "Why would a DOM anchor be needed in the first place?" So makes me wonder, why even call that out here? I think a good rule of thumb is that we comment about what we're doing vs what we're not doing/omitting.

bdow (approved):

bdow:
Looking at the Chromatic snapshots, doesn't appear the `play` function worked as expected. If you can get it to work in like 15-20min, then go for it. Otherwise, I think just remove the story.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
abadami: I would say omit in this case. I think the code here is clear enough, and yeah we absolutely don't need a comment on the history here.

PR #1365, opened 2026-08-06
feat: expose onContinue workflow step action

allangaldinosilva:
Allow clients to define custom step functions on click to `Continue` button.

https://github.com/homebound-team/internal-frontend/pull/7880


src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -86,7 +98,10 @@ export function WorkflowLayout(props: WorkflowLayoutProps) {
       completeLabel={completeLabel}
       onComplete={onComplete}
       primaryDisabled={!activeStep?.completed}
-      onContinue={() => setCurrentStep(tabSteps[currentIndex + 1].value)}
+      onContinue={async () => {
+        // A `false` result means the caller vetoed the advance (e.g. a save failed), so stay put.
+        if ((await onContinue?.(activeStep.value)) !== false) setCurrentStep(tabSteps[currentIndex + 1].value);
+      }}
```
bdow, line 96: **question:** How necessary is it to pass `activeStep.value` to the `onContinue`? Right now we don't require consumers to define a `value` property, it is automatically generated from the `id`. The `value` prop was not really meant to become consumer facing.
_If_ we really need the `value` here with `onContinue`, then I think we might want to change the API to enforce that the consumers pass in the step's `value`, which might be better named as `id`, as it'll act like a unique identifier per step.

bdow (commented):
Since `value` is an internal property only, I think leaking this into a `onComplete(activeStepValue)` might be quite odd. I also like the user didn't need to add a `value`/`id` to each step, which kept things simpler. 
I think instead, I would push to do a per-step `onContinue` handler:
```
steps={[
  {
    label: "Basics",
    completed: basicsForm.valid,
    onContinue: async () => {
      const result = await saveBasics();
      return result.ok; // false → stay put
    },
    content: <BasicsStep form={basicsForm} />,
  },
  // ...
]}
```
Would this work for what you want to do?

src/layouts/WorkflowLayout/WorkflowLayout.tsx
allangaldinosilva: The case I'm working on, each step will define its `onContinue` function, so because I don't have a way to retrieve what's the current step, I'm passing it in the args. 

https://github.com/homebound-team/internal-frontend/blob/p-2643-save-swap-product-ce-lines/src/routes/projects/change-events/guided/GuidedChangeEventPage.tsx#L70

Would be great if we have a `useWorkflow` hook to retrieve important info like `currentStep` ...

allangaldinosilva:
> Since `value` is an internal property only, I think leaking this into a `onComplete(activeStepValue)` might be quite odd. I also like the user didn't need to add a `value`/`id` to each step, which kept things simpler. I think instead, I would push to do a per-step `onContinue` handler:
> 
> ```
> steps={[
>   {
>     label: "Basics",
>     completed: basicsForm.valid,
>     onContinue: async () => {
>       const result = await saveBasics();
>       return result.ok; // false → stay put
>     },
>     content: <BasicsStep form={basicsForm} />,
>   },
>   // ...
> ]}
> ```
> 
> Would this work for what you want to do?

Yeah @bdow, it'd match what I need, I'm gonna update this PR to do so. Thanks 👏

src/components/StepperTabs/StepperTabs.tsx
```diff
@@ -3,7 +3,10 @@ import { Css } from "src/Css";
 import { useBreakpoint } from "src/hooks/useBreakpoint";
 import { useTestIds } from "src/utils";
 
-export type StepperTabsStep = Pick<StepperTabProps, "label" | "value" | "completed" | "disabled">;
+export type StepperTabsStep = Pick<StepperTabProps, "label" | "value" | "completed" | "disabled"> & {
+  /** Runs before leaving this step via Continue, e.g. to save it. Return `false` to stay put, like when the save failed. */
+  onContinue?: () => boolean | void | Promise<boolean | void>;
+};
```
bdow: **request:** `StepperTabProps` already has an `onContinue`. I'd prefer to pick from there as well and keep the signures for `onContinue` consistent.

src/layouts/WorkflowLayout/WorkflowLayout.test.tsx
```diff
@@ -200,6 +200,50 @@ describe("WorkflowLayout", () => {
...
+    const steps = makeSteps();
+    const r = await render(
+      <WorkflowLayout {...baseProps({ steps: [{ ...steps[0], onContinue: () => false }, steps[1]] })} />,
+      withRouter(),
+    );
+    // When Continue is clicked
+    await clickAndWait(r.continue);
+    // Then we stay on the first step
+    expect(r.body).toBeInTheDocument();
+    expect(r.query.stepTwoBody).not.toBeInTheDocument();
```
bdow, line 245: **nit:** Would prefer to validate the step 1 in still in the dom vs step 2 isn't.

bdow: My fault, this is doing that. Nevermind!

src/layouts/WorkflowLayout/WorkflowLayout.tsx
```diff
@@ -86,7 +86,10 @@ export function WorkflowLayout(props: WorkflowLayoutProps) {
       completeLabel={completeLabel}
       onComplete={onComplete}
       primaryDisabled={!activeStep?.completed}
-      onContinue={() => setCurrentStep(tabSteps[currentIndex + 1].value)}
+      onContinue={async () => {
```
bdow, line 89: **nit/suggestion:** I think this could be written a little bit clearer, maybe? The tests do validate that this works when `onComplete` is not defined, but I think we could make this easier to read, but might require a few more lines of code... Something like this:
```
const onContinue = activeStep?.onContinue;
if (onContinue) {
  const allowed = await onContinue();
  if (allowed === false) return;
}
setCurrentStep(tabSteps[currentIndex + 1].value);
```
It does the same thing, but I think its more easy to read.

src/components/StepperTabs/StepperTabs.tsx
allangaldinosilva: I didn't find it on `StepperTabProps` 🤔 I agree that's the right spot and then just "pick" it on `StepperTabsStep`.

src/layouts/WorkflowLayout/WorkflowLayout.tsx
allangaldinosilva: Looks great, much cleaner! Thanks 👏

src/components/StepperTabs/StepperTabs.tsx
bdow: That's my fault! I actually think that might be the wrong spot lookign again. Mostly, I want to share the type between definition, which here are currently on `type StepperTabsStep` and `type WorkflowActionsProps`.
So maybe we don't include it in `StepperTabProps` because, we don't use it in that component. So maybe update the shape in `WorkflowActionsProps.onContinue` to be `() => boolean | void | Promise<boolean | void>;` Then we can pick from there via:
`type StepperTabsStep = Pick<StepperTabProps, "..."> & Pick<WorkflowActions, "onContinue">;`
What do you think?

allangaldinosilva: yeah, makes sense, since these are ended up intercepting the continue button handler, having them sharing the same signature is good 👍

bdow (approved):

PR #1369, opened 2026-08-10
feat: Add optional tag to SelectCardGroup options 

allangaldinosilva:
### [PINT-68](https://linear.app/homebound-team/issue/PINT-68)

Adds an optional `tag` to `SelectCardGroup`, so consumers can surface a status (e.g. "PO Issued", "Configuration Required") on individual cards. Works in grid and list views for both single and multi select groups.

`view=list`
image

`view=grid`
image


src/inputs/SelectCard/ListSelectCard.tsx
```diff
@@ -32,6 +34,11 @@ export function ListSelectCard(props: SelectCardItemProps) {
       __storyState={__storyState}
       {...others}
     >
+      {tag && (
+        <div css={Css.ml3.if(isDisabled).o50.$}>
```
bdow: **question/suggestion:** Three thoughts/questions:
1. A `disabled` state of a tag is probably something we'd want to hoist up to the tag component, so we have a consistent way to considering them disabled. Though, `disabled` isn't really a concept for a tag, but probably fine to still add a `disabled` prop on it and just denote in a JS doc that `disabled` is set only when Tag is used within a disable-able component.
2. Has design given any feedback on what the Tag component would look like when the card is disabled? If not, we should ask them.
3. The `ml3` style is shared between this and the description. How can we ensure these stay in sync?

bdow (approved):
Approving to unblock, but I'd like to have the questions in the comment answered/addressed. Regarding working with Design on solidifying the styling for tags that should appear disabled, feel free to create a follow up ticket and assign to me and just point to the file as to where we need to update. Or better yet, add a `// TODO: DS-1234 (linear issue number)` as a comment where we need to fix.

src/inputs/SelectCard/types.ts
```diff
@@ -21,9 +24,9 @@ type SelectCardGroupItemOptionBase<V extends Value> = {
   selectionBehavior?: "exclusive";
 };
 
-/** Grid-view option; requires either `icon` or `image` (image url shown in place of the icon). */
+/** Grid-view option; requires either `icon` or `image`, and only image cards take a `tag`. */
 export type SelectCardGridGroupItemOption<V extends Value> = SelectCardGroupItemOptionBase<V> &
-  ({ icon: IconProps["icon"]; image?: never } | { image: string; icon?: never });
+  ({ icon: IconProps["icon"]; image?: never; tag?: never } | { image: string; icon?: never; tag?: CardTag });
```
allangaldinosilva, line 29: Per this [discussion](https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=320-332&m=dev#1882661104), we're not going to support tags on icon cards. This type enforcement will trigger a compile time issue when `icon+tag` usage is defined.

Image

src/inputs/SelectCard/ListSelectCard.tsx
allangaldinosilva: Thanks for the feedback @bdow, all of them were properly addressed ✅, here are my thoughts: 

1. Per this figma [discussion](https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=320-332&m=dev#1882661104), we're not going to have different tag layouts for `disabled` state.
2. Yes, the tag position is now in front of the label
3. Done as part of `2` changes.

bdow (commented):
Thanks for the adding in the Figma feedback to this PR!

PR #1370, opened 2026-08-10
fix: Fix WorkflowLayout's collapsed scroll transition

abadami:


bdow (commented):
I'm getting a very odd behavior when scrolling down far, and then scrolling up slightly. Things are jumping and re-rendering a bit crazy.  If you can't reproduce, let me know. I also see it on the existing workflow header tabs. Something with the scroll listener, I am guessing. i think it'd be good to fix as part of this PR.

PR #1371, opened 2026-08-10
chore: Figma as source of truth for color tokens

bdow:
New guidance around color primitives and tokens. The gist is that now Figma is the SoT for these colors. If any colors change in Figma, then use the MCP to get an export of the colors that saves to figma-colors.raw.json per tokens/README.md. Use that file to then run generation scripts as defined in the readme.


scripts/generate-design-tokens.ts
```diff
@@ -35,6 +36,170 @@ const themeScopesCssPath = join(rootDir, "src/css/generated/theme-scopes.css");
...
+      contrast: RawColorValue;
+    }
+  >;
+};
+
+function normalizeAlpha(alpha: number | undefined): number | undefined {
+  if (alpha === undefined) return undefined;
+  // Collapse float32 noise from Figma (e.g. 0.20000000298023224 → 0.2).
+  return Math.round(alpha * 1000) / 1000;
+}
```
abadami, line 58: **question (non-blocking)**: I think this is good to do for consistency regardless, but is the float32 noise really significant enough to affect the color?

tokens/figma-colors.raw.json
abadami, line 1: **thought**: Do we need to commit in the raw Figma values? As far as I can tell, it's not used for anything but translating into `color.json` and `motion.json` and we commit those.

Although, this does make a little more sense if we're not able to fetch the values programmatically and we have to manually export them. Is it not possible to fetch them?

tokens/README.md
```diff
@@ -1,14 +1,37 @@
...
 
 **Normative color format:** [Design Tokens Color Module 2025.10](https://www.designtokens.org/tr/2025.10/color/#color-tokens) — for `$type: "color"`, `$value` is `colorSpace: "srgb"`, `components: [r, g, b]` (each **0–1**), optional `alpha`, optional `hex`. Semantic baselines usually reference `{beam.color.primitive.*}`.
 
-## Root
+## Updating colors from Figma
+
+Do **not** use Figma’s per-collection JSON export — wrong shape, and you’d have to merge **Primitive Colors** + **Semantic Tokens** by hand.
+
+1. Edit Variables in [BEAM_27_LIBRARY](https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY) (`Primitive Colors`, `Semantic Tokens`). Set **Contrast** even when it matches Light.
+2. In Cursor: ask for an **MCP export** of those Variables into [`figma-colors.raw.json`](./figma-colors.raw.json) (one file, both collections). Needs an authenticated Figma MCP session.
```
abadami, line 22: **thought**: Ah I see, the values are expected to be gotten from a MCP session.

I know it advises not to use the per-collection JSON export, but is there no way to get the collection programmatically? I don't think merging the primitive colors and semantic tokens by hand would be a big deal and it would be nice to just hit "build-tokens" and have that handled all in one place.

Not necessary to handle now, but I think it may be a worthy QoL follow-up.

abadami (approved):
This all seems good. I don't think anything I commented on needs to block this PR, but I do think the one follow-up suggestion I have is a worthy follow-up.

PR #1372, opened 2026-08-10
feat: ai star loader;

apattersonATX-HB:
ds-305

https://github.com/user-attachments/assets/f2fffd59-5e7c-45e4-b33e-bdb18d83c1fb


src/components/AiLoader.tsx
```diff
@@ -0,0 +1,77 @@
...
+      .add("animationName", "aiStarLoader")
+      .add("animationDuration", `${cycleMs}ms`)
+      .add("animationIterationCount", "infinite")
+      .add("animationTimingFunction", "ease-in-out")
+      .add("animationDelay", `${(index * cycleMs) / 3}ms`).$
+  );
+}
+
+/** Eight-pointed sparkle, drawn to fill a 24x24 box. */
+const starPath =
```
apattersonATX-HB: Intentionally left this out of icons but maybe we can reuse it in `BPAiLogo` 🤔

apattersonATX-HB: Actually, looks like it does need to be moved into `icons`

I'm seeing it all over the place 🙃 🥊

src/components/Icon.tsx
```diff
@@ -200,6 +208,8 @@ export const Icons = {
       <path d="M11.0513 3.69095C11.3506 2.76968 12.654 2.76968 12.9534 3.69095L14.4722 8.36485C14.6061 8.77685 14.99 9.05579 15.4232 9.0558L20.3377 9.05599C21.3064 9.05603 21.7091 10.2956 20.9255 10.865L16.9497 13.7538C16.5992 14.0084 16.4526 14.4598 16.5864 14.8718L18.1049 19.5458C18.4042 20.4671 17.3498 21.2332 16.5661 20.6638L12.5901 17.7753C12.2396 17.5207 11.765 17.5207 11.4146 17.7753L7.43853 20.6638C6.65483 21.2332 5.6004 20.4671 5.89971 19.5458L7.4182 14.8718C7.55205 14.4598 7.4054 14.0084 7.05494 13.7538L3.07914 10.865C2.29548 10.2956 2.69823 9.05603 3.66692 9.05599L8.58142 9.0558C9.01461 9.05579 9.39854 8.77685 9.53242 8.36486L11.0513 3.69095Z" />
     </>
   ),
+  // Inset to ~18.8px so it reads at the same optical size as `star` and the rest of the set.
+  aiStar: <path d={aiStarPath} transform="translate(2.6 2.6) scale(0.78333)" />,
```
bdow: **question:** Could we not export this star in a 24x24 viewbox?

src/components/AiLoader.tsx
bdow: I wonder if this star is okay to keep as a regular icon. 🤔 It might be more-so considered an "illustration" as it includes multiple colors. I don't know if we'd want to ever show this in its non-gradient styles. Right now icon's would allow for setting the `fill`, but I would guess we always want this to be displayed only in one color.

src/components/AiLoader.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { aiStarPath } from "src/components/Icon";
+import { Css, increment, Margin, Palette, Xss } from "src/Css";
+import { useTestIds } from "src/utils/useTestIds";
+
+export type AiLoaderProps = {
+  inc?: number;
+  /** Accessible name for the indicator, defaults to "Loading". */
+  label?: string;
+  /** Styles overrides */
+  xss?: Xss<Margin>;
```
bdow: **nit:** I don't know if we'd need this margin. But we're also not actually restricting to it. We'd need to do the whole `<X extends Only<Xss<Margin>>>` as well.

src/components/AiLoader.tsx
```diff
@@ -0,0 +1,74 @@
...
+};
+
+/** An indeterminate loading indicator for AI-driven work. */
+export function AiLoader(props: AiLoaderProps) {
+  const { inc = 3, label = "Loading", xss } = props;
+  const tid = useTestIds(props, "aiLoader");
+  const size = increment(inc);
+  // Namespace the gradient so multiple loaders on a page don't collide; `useId` includes colons.
+  const gradientId = `aiLoaderGradient-${useId().replaceAll(":", "")}`;
```
bdow: **question:** Can `useId` take a dictionary so we can avoid characters we don't want to allow?

src/components/AiLoader.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    Css.add("transformBox", "fill-box")
+      .add("transformOrigin", "center")
+      // Also the keyframes' 0%, so stars waiting out their delay (and reduced-motion users) rest small.
+      .add("transform", "scale(0.5)")
+      .add("animationName", "aiStarLoader")
+      .add("animationDuration", `${cycleMs}ms`)
+      .add("animationIterationCount", "infinite")
+      .add("animationTimingFunction", "ease-in-out")
+      .add("animationDelay", `${(index * cycleMs) / 3}ms`).$
```
bdow: **thought/question (non-blocking):** I'm kind of surprised none of these are actual Truss props. I would guess that atleast `transform` is 🤔. Could be a nice truss addition to support these different properties.

bdow (approved):
Looks good. I'd like to have design AT this prior to merging, though. Let's see if we can get Andrea to take a look at the storybook file and see what she thinks. Thanks!
giving an approval to try and keep things moving, but would like to have the comments addressed prior to merging, please.

src/components/AiLoader.tsx
apattersonATX-HB: Well if we keep it as an `Icon` we can use it in all the icon friendly components easily. The way I wrote it `fill` gets ignored which might be a good thing since this aiStar is never meant to any other way. I haven't seen any `disabled` states or anything like that for this icon.

src/components/AiLoader.tsx
apattersonATX-HB: Took this out

src/components/AiLoader.tsx
apattersonATX-HB: Hmmmm, i'll add these in a follow up so we can update other callers.

PR #1373, opened 2026-08-10
feat: BP AI logo

apattersonATX-HB:
### ds-304

Screenshot 2026-08-10 at 5 19 58 PM


src/components/Logos/BlueprintAiLogo.tsx
```diff
@@ -0,0 +1,44 @@
+import { useId } from "@react-aria/utils";
+import { Css, Palette } from "src/Css";
+
+type BlueprintAiLogoProps = {
+  width?: number | "auto";
+  height?: number | "auto";
+};
```
bdow: **thought:** I wonder if we can prevent users from distorting the image by providing an incorrect aspect ratio.
Maybe we just provide guidance in the JS Doc that we only suggset providing one value, either height or width. If both are provided, only one (probably width?) will be respected.

src/components/Logos/Logos.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+  component: BlueprintAiLogo,
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=1198-190&m=dev",
+    },
+  },
+} as Meta;
+
+type LogoSize = "small" | "medium" | "large";
```
bdow: **suggestion/nit (non-blocking):** This LogoSize might be misleading in the stories. It displays as if it is a prop users can pass in, but really its just a faux sizing only for the story.
Maybe the story instead just defines what `width` or `height` values it provided to the component.

bdow (approved):
Small bits of feedback. Would like to see those addressed before merging. But happy to merge and follow up as well - as long as the follow up actually happens! 😄

src/components/Logos/BlueprintAiLogo.tsx
apattersonATX-HB: Went one better than JSDoc and made it a type error with a shared LogoSizeProps.


TIL: the artwork can't actually distort (SVG preserveAspectRatio letterboxes rather than stretches), but passing both did silently add dead space, so it's still worth blocking.

bdow: Do we have a standardized set of sizes?

bdow: Nevermind! Just saw the changes. Looks good! Thanks

PR #1377, opened 2026-08-12
feat: ai panel & import banner

apattersonATX-HB:
### [DS-308](https://linear.app/homebound-team/issue/DS-308/import-banner)

Adds `AiPanel`, the shared Blueprint AI surface, and `ImportBanner` as its loading preset. 

- **`AiPanel`** — the tinted wash, Blueprint AI logo, and a card that takes arbitrary `children`. The designs reuse this same chrome for a loader, a URL form, a review prompt, and an entire page's worth of form sections, so content is composed in rather than added as a variant per flow.
- **`ImportBanner`** — Import copy + `align="center"` + `loading` over `AiPanel`.
- **Two new truss properties**, `Css.aiWash` and `Css.aiGradientText`, since both were needed in more than one place and the gradient-text recipe (`background-clip` + `-webkit-text-fill-color`) is easy to get wrong.
- **`variant: "banner" | "page"`** — `banner` fills whatever column its container gives it; `page` caps content at 768 and centers it, for a panel dropped into a full-bleed page body. Not a size prop but logo, radius, padding, gap and column width all differ between the two.
- **Stories inside our 5 page layouts** — `WithAiPanel` on `EnvironmentBanner`/`Navbar`/`SideNav`/`PageHeader`/`Workflow`, plus `FullAiPanel` recreating the workflow import form.

**TBD**
- On a viewport taller than its content, the wash stops short of the bottom. That's `WorkflowLayout` having no height for its `fg1` body to grow into, not an `AiPanel` issue.


src/components/AiPanel.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { BlueprintAiLogo } from "src/components/Logos";
+import { Css, Properties, Tokens } from "src/Css";
+import { useTestIds } from "src/utils";
+
+/**
+ * `banner` sits above other page content and fills whatever column its container gives it; `page` is
+ * for when the AI surface *is* the page, and caps its content width. Styles vary slightly between the two
+ * within figma.
+ */
+export type AiPanelVariant = "banner" | "page";
```
apattersonATX-HB: Looking at this again I'm not sure we'd like to use the panel in this way to cover an entire page like in our `WorkflowLayout`. It might be better to just give those an `AiMode` prop

truss-config.ts
```diff
@@ -44,8 +44,41 @@ const transition: string = [
...
+    newMethod("aiWash", { backgroundImage: aiWash }),
+    newMethod("aiGradientText", {
+      // Fallback where a background can't be clipped to text; `transparent` would hide it outright.
+      color: palette.Purple700,
+      backgroundImage: aiGradient,
+      backgroundClip: "text",
+      WebkitBackgroundClip: "text",
+      // Clears the glyph fill so the background shows through.
+      WebkitTextFillColor: "transparent",
+    }),
```
bdow: **question:** How's this look cross browser?

truss-config.ts
```diff
@@ -44,8 +44,41 @@ const transition: string = [
...
+
+/** The Blueprint AI ramp — the same purple-to-blue as the `aiStar` sparkle, so AI marks read as one family. */
+const aiGradient = `linear-gradient(90deg, ${palette.Purple700}, ${palette.Blue700})`;
+
+/**
+ * A faint wash for AI surfaces. The designs draw it as a radial gradient, but it's stretched so far
+ * that within a panel's height it only varies left-to-right — so this is that one axis, with the stops
+ * snapped to palette entries rather than carrying one-off hexes.
+ */
+const aiWash = [
```
bdow: **thought/question:** `wash` is not a term I'm aware of. Design calls this style the `AI Background`. Could we just adopt that to keep consistency?
I'm also not sure what this comment above means. Does it mean "design used a radial gradient, but that won't work for us"? 
Design's gradient here is:
```
background: radial-gradient(12014% 100.51% at 35.83% 62.18%, var(--Blue50, rgba(239, 246, 255, 0.50)) 0%, var(--Purple100, rgba(237, 233, 254, 0.50)) 100%), #FFF;
```
Did you try this and it just didn't look good?

https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=1199-774&t=am7ZdF2MMqkl5wK1-4

truss-config.ts
```diff
@@ -44,8 +44,41 @@ const transition: string = [
   .map((property) => `${property} ${motion.duration.normal} ${motion.easing.standard}`)
   .join(", ");
 
+/** `palette` values are opaque `rgba(...)`, so re-alpha one to layer it as a tint. */
+function alpha(color: string, a: number): string {
+  return color.replace(/,\s*1\)$/, `, ${a})`);
+}
+
+/** The Blueprint AI ramp — the same purple-to-blue as the `aiStar` sparkle, so AI marks read as one family. */
+const aiGradient = `linear-gradient(90deg, ${palette.Purple700}, ${palette.Blue700})`;
```
bdow: **thought/suggestion:** The design team is denoting this as `AI Bold` in Figma. I'd push for us to mirror the same naming conventions they are using for better translation between figma & code.
Also, the gradient they define is:
```
background: linear-gradient(270deg, #1D4ED8 2.97%, #6D28D9 87.08%);
```
Nearly the same you have hear. The color order and `deg` are switched, but that's not a big deal. Though, would be good match their color stops.

src/layouts/WorkflowLayout/WorkflowLayout.stories.tsx
```diff
@@ -102,6 +107,175 @@ export function Composed() {
...
+          `FormSection`s directly rather than `FormSectionLayout`, which requires a `title` — here the
+          panel's own gradient title is the form-level heading (the design's "Form Title 20"), so a
+          layout title would duplicate it. This mirrors `FormSectionLayout`'s own column and spacing.
+        */}
+      <div css={Css.df.fdc.gap8.w100.maxwPx(720).mxa.pt4.$}>
+        {sections.map((section) => (
+          <FormSection key={section.title} {...section} />
+        ))}
+      </div>
+    </AiPanel>
```
bdow: **thought/suggestion:** My original thought here was that we'd be reusing the `FormSectionLayout`, which gives us that `maxwPx(720)` and the layout we'd want when in the forms, but just add a `aiEnabled` flag of sorts to give this extra styling. The `title` and `message` text sizes would match the existing sizes uses by `FormSectionLayout`'s `title` and `description`. At least to me, this AI styling seems just a decoration on top of the existing FormSectionLayout. Could we see what that'd look like?
And I'm not saying to just drop the AIPanel as part of this, but instead the composition of it would happen internally in Beam vs the engineer doing it. Maybe something close to:
```
// psuedo code
function FormSectionLayout(props) {
  const { aiEnabled, ... } = props;
  
  return aiEnabled ? <AiPanel><FormSectionLayoutContent {...props} /></AiPanel> : <FormSectionLayoutContent {...props} />;
```
Given that some pages Then from the implementation engineers point of view their page would look simpler:
```
function SomeWorkflow() {
  return <FormSectionLayout aiEnabled={boolean} />
```

I kind of see the AiPanel being more of just a wrapper - include the "Blueprint AI" logo, the background color, and container element w/padding, but everything inside is just children. And maybe the `container` element follows a `AiPanelProps.fullWidth` property to either do `Css.w100.if(!fullWidth).wfc.$`. This panel can be less aware of whether is a `page` or `banner` variant.

Let's chat through this one. I think this direction would simplify the AiPanel component and reduce what the caller needs to compose.

src/components/AiPanel.tsx
```diff
@@ -0,0 +1,98 @@
...
+          {title && (
+            <span css={Css.lg.aiGradientText.$} {...tid.title}>
+              {title}
+            </span>
+          )}
+          {message && (
+            <span css={{ ...Css.sm.color(Tokens.OnSurface).$, ...alignStyles[align].text }} {...tid.message}>
+              {message}
+            </span>
+          )}
```
bdow: **suggestion:** I think it'd be nice to have this component as dumb as possible. Just know whether it should be full width container or not. Then use other elements like a "AiImportBanner" that'd use this container and put its own contents in.

truss-config.ts
apattersonATX-HB: Looks fine to me
Screenshot 2026-08-14 at 1 57 19 PM

src/layouts/WorkflowLayout/WorkflowLayout.stories.tsx
apattersonATX-HB: Ya same here, its mostly a wrapper. I like exposing it as well for future work that we, inevitably, won't have full designs for. 

I can rip this out and then teach all of our layouts to handle an AiMode in this future [ticket](https://linear.app/homebound-team/issue/DS-302/support-ai-background-and-card-wrapper-for-ai-stepper-layouts)

apattersonATX-HB: Obviously happy to chat about it too

truss-config.ts
apattersonATX-HB: Screenshot 2026-08-14 at 1 59 47 PM
Screenshot 2026-08-14 at 1 59 42 PM

Wash is an art term, I picked it up painting minis actually 🙈. It's a semi transparent paint layer over a background, so it shifts with whatever it's over. But ya i named it wrong.

Nope, I never tried yours, I just missed the library component. Mine's also too faint as a result and isn't a perfect match.

The one thing that makes me hesitate is our slow push to theming. Ours is translucent over `Tokens.Surface` so it changes based on background (see screenshots, took those with the off values too). Do we want to future proof or just keep it simple?

src/components/AiPanel.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const Widths = () => (
+  <div css={Css.df.fdc.gap4.$}>
+    {[1440, 956, 640].map((w) => (
+      <div key={w} css={Css.wPx(w).$}>
+        <AiPanel>
+          <div css={Css.sm.color(Tokens.OnSurface).$}>{w}px container</div>
+        </AiPanel>
+      </div>
+    ))}
+  </div>
```
bdow: **thought/suggestion:** I don't think this is what we're going after in terms of `width`. We should always have the AiPanel's background go the full width of its container, that's expected and being done here. But the inner elevated card container should be able to choose whether its `fullWidth` or takes its content's width. I believe that can be done by doing a `width: fit-content`/`.wfc.$`. And, if its not full width, then it should center itself as well. So something like:
```
type AiPanelProps = {
  fullWidth?: boolean;
  children: ReactNode;
}

function AiPanel(props: AiPanelProps) {
  const { fullWidth = true } = props;

  return (
    <div css={Css.aiBackground.w100}.$}>
      {/** Wrapping div to keep logo & content aligned */}
      <div css={Css.w100.if(!fullWidth).wfc.mxa.$}>
        <BlueprintAiLogo />
        <div css={Css.br12.bshbasic.bgColor(Tokens.Surface).$}>{children}</div>
      </div>
   </div>
  );
}

File: src/components/AiPanel.tsx [hunk 8]
```diff
@@ -0,0 +1,35 @@
...
+      css={{
+        // `aiBackground` is opaque, so `Tokens.Surface` only shows if the gradient fails to paint.
+        ...Css.df.fdc.aifs.gapPx(4).w100.px3.py2.bgColor(Tokens.Surface).aiBackground.$,
+        ...(rounded ? Css.br12.$ : {}),
+      }}
+      {...others}
+      {...tid}
+    >
+      <BlueprintAiLogo height={2} />
+      <div css={Css.df.fdc.w100.br12.ptPx(12).px2.pb2.bgColor(Tokens.Surface).bshBasic.$} {...tid.card}>
```
bdow: **suggestion:** I think we need to put some spacing between the logo and the inner panel here. Looks like the design calls for 8px of spacing between them.

src/components/AiPanel.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <div
+      css={{
+        // `aiBackground` is opaque, so `Tokens.Surface` only shows if the gradient fails to paint.
+        ...Css.df.fdc.aifs.gapPx(4).w100.px3.py2.bgColor(Tokens.Surface).aiBackground.$,
+        ...(rounded ? Css.br12.$ : {}),
+      }}
+      {...others}
+      {...tid}
+    >
+      <BlueprintAiLogo height={2} />
```
bdow: **question:** Where do you see a height of 16px for this? Most places I've seen are 24px when its outside the card, and it was 20px height within the card. FWIW, I dont' think we're placing this within the card ever, and moving towards just outside of the card as you hav here. 
So TLDR: I think we want this to be `height={3}`

src/components/AiPanel.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * The Blueprint AI surface: the wordmark over the AI background, wrapping a card of whatever the flow
+ * needs.
+ */
+export function AiPanel(props: AiPanelProps) {
+  const { rounded = false, children, ...others } = props;
+  const tid = useTestIds(others, "aiPanel");
+  return (
+    <div
+      css={{
+        // `aiBackground` is opaque, so `Tokens.Surface` only shows if the gradient fails to paint.
```
bdow: **questions:** Is this true that the "`aiBackground` is opaque"? It doesn't appear that way to me, but could be missing something.
And why would the gradient fail to paint?

truss-config.ts
```diff
@@ -44,8 +44,34 @@ const transition: string = [
   .map((property) => `${property} ${motion.duration.normal} ${motion.easing.standard}`)
   .join(", ");
 
+/** Design's "AI Bold" ramp */
+const aiBold = `linear-gradient(270deg, ${palette.Blue700} 2.97%, ${palette.Purple700} 87.08%)`;
+
+/**
+ * Design's "AI Background". Drawn as a radial gradient, but its ellipse is elongated far enough that
+ * the render only varies left-to-right, so this is that one axis.
+ */
```
bdow, line 52: **question:** I'm not sure I follow this comment completely, but my assumption is that the "ellipse" it references is the size/spread of the radial gradient. So if I understand it right, I don't think its true. The center point does start from the more left side, but not enough where it gives more of a linear gradient look. If it were going to be a linear look, then it would be more like Purple50 -> Blue50 -> Purple100. 
Is there a reason why we can't use design's proposed radial gradient and instead we tried to massage it into a linear gradient?

PR #1378, opened 2026-08-13
feat: ds-306 - banner in superdrawer ai btn variant

apattersonATX-HB:
### SuperDrawer

- `SuperDrawerContent` takes a `banner?: ReactNode` slot that renders above the children and cancels the content padding, so the panel spans the drawer's full width while the content stays inset

### AiPanel actions

- `primaryAction` / `secondaryAction` named slots typed as `ActionButtonProps`, matching our  `fooBarLayout` components. Caller supplies label/onClick and the component owns variants (`ai` for primary, `quaternary` for secondary)

### `ai` Button variant

- New variant on the same purple→blue gradient as designed

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -2,7 +2,7 @@ import { AriaButtonProps } from "@react-types/button";
 import { ButtonHTMLAttributes, ReactNode, RefObject, useMemo, useState } from "react";
 import { useButton, useFocusRing, useHover } from "react-aria";
 import { Icon, IconProps, Loader, maybeTooltip, navLink, resolveTooltip } from "src/components";
-import { Css, Properties, Tokens } from "src/Css";
+import { Css, Palette, Properties, Tokens } from "src/Css";
 import { useGetRef } from "src/hooks/useGetRef";
 import { BeamButtonProps, BeamFocusableProps } from "src/interfaces";
 import { isAbsoluteUrl, isPromise, noop } from "src/utils";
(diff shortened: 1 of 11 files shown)


src/components/Button.tsx
```diff
@@ -142,6 +142,17 @@ const variantStyles: Record<
...
   }
 > = {
+  ai: {
+    baseStyles: Css.aiGradientBg.color(Tokens.OnPrimary).$,
+    hoverStyles: Css.add("backgroundImage", `linear-gradient(90deg, ${Palette.Purple800}, ${Palette.Blue800})`).$,
+    pressedStyles: Css.add("backgroundImage", `linear-gradient(90deg, ${Palette.Purple900}, ${Palette.Blue900})`).$,
+    disabledStyles: Css.add(
+      "backgroundImage",
+      `linear-gradient(90deg, ${Palette.Purple200}, ${Palette.Blue200})`,
+    ).color(Tokens.ButtonPrimaryDisabledFg).$,
```
bdow, line 152: **question:** Can these gradients be the `aiBold` background we're adding to truss-config? I think that's what we're doing here. In the other PR, `aiBold` is a text color. Maybe that becomes `aiBoldText` and then there's also a `aiBoldBg` - both using the same underlying `aiBold` defined in the truss-config.

bdow (approved):
Understood that we're merging here, not into `main`, but just to the other PR, which is where we'll review everything together, and then merge to main.

bdow:
I wonder if you'd be better off just merging these two changes yourself locally. Then killing these two PRs and starting fresh. Wdyt?

apattersonATX-HB:
> I wonder if you'd be better off just merging these two changes yourself locally. Then killing these two PRs and starting fresh. Wdyt?

Ya thatd be great at this point. Will address the feedback from the other one n start fresh

PR #1382, opened 2026-08-17
feat: ai panels

apattersonATX-HB:
### Components

- **`AiPanel`** — container only: BPAiLogo over the AI background, wraps a white card. `rounded` for a panel inside page content
- **`AiImporting`** — in-progress loading state. 
- **`AiReview`** — the finished state. Copy sits beside its actions rather than above them, keeping it one row tall over the content being reviewed. `primaryAction` takes the AI gradient, `secondaryAction` is quaternary.

### SuperDrawer

- `SuperDrawerContent` gains `banner?: ReactNode` — rendered above the children in the same scroll container, negative-margined out to the drawer's full width while the body stays inset.


src/components/AiReview.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Css, Tokens } from "src/Css";
+import { useTestIds } from "src/utils";
+
+export type AiReviewProps = {
+  title: string;
+  message?: ReactNode;
+  /** Accepts the AI's work, e.g. "Accept Import". Drawn with the AI gradient. */
+  primaryAction?: ActionButtonProps;
+  /** Discards it, e.g. "Clear Import". */
+  secondaryAction?: ActionButtonProps;
```
bdow: **thought (non-blocking):** Wondering if we would be better off with an `actions: ActionButtonProps[]` array here? We lose the forcing of using the AI variant for the primary, and the quaternary for secondary.

**_tangent:_**
One other thought I had was to create a new "scoped theme". Where allc hildren inside of a sort of: `<div data-theme="ai">{children}</div>` would automatically get the AI treatment. The `primary` button's background color would change appropriately. Not something to do now, but could be interesting to see what other styling can just switched based on css variables.

src/components/AiPanel.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    expect(r.aiPanel_column).toHaveStyle({ width: "fit-content", marginLeft: "auto", marginRight: "auto" });
+  });
+
+  it("says nothing of its own", async () => {
+    // Given a bare panel, i.e. no caller-supplied ARIA
+    const r = await render(<AiPanel />);
+    // Then it stays inert — announcements belong to whoever fills it
+    expect(r.aiPanel).not.toHaveAttribute("role");
+    expect(r.aiPanel).not.toHaveAttribute("aria-busy");
+  });
```
bdow: **suggestion:** can we change this test name, maybe to:
`does not set assistive attributes by default"`

```suggestion
  it("does not set assistive attributes by default", async () => {
    // Given a bare panel, i.e. no caller-supplied ARIA
    const r = await render(<AiPanel />);
    // Then it stays inert — announcements belong to whoever fills it
    expect(r.aiPanel).not.toHaveAttribute("role");
    expect(r.aiPanel).not.toHaveAttribute("aria-busy");
  });
```

src/components/AiPanel.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * The Blueprint AI surface: the wordmark over the AI background, wrapping a card of whatever the flow
+ * needs.
+ *
+ * The background always spans its container; `fullWidth` only governs the card.
+ */
+export function AiPanel(props: AiPanelProps) {
+  const { rounded = false, fullWidth = true, children, ...others } = props;
+  const tid = useTestIds(others, "aiPanel");
+  return (
+    <div css={{ ...Css.df.fdc.w100.px3.py2.aiBackground.$, ...(rounded ? Css.br12.$ : {}) }} {...others} {...tid}>
```
bdow: **nit:** Probably don't need `df.fdc` any longer as there is only one child in this element. Could also shorten the `css` prop via using the Truss `.if` statement:

```suggestion
    <div css={Css.w100.px3.py2.aiBackground.if(rounded).br12.$} {...others} {...tid}>
```

bdow: I wouldn't apply this directly, probably needs some formatting to run afterwards.

src/components/AiReview.tsx
bdow, line 1: **suggestion/thought:** So, I'm wondering if this AI Review component is still a thing? Given how the new material import / material clean up Figma doesn't have a "post loading state", it just goes straight to the [form](https://www.figma.com/design/OdgcptKNTLp0L30TqNUD6w/Material-Catalog-Cleanup?node-id=1300-13733&m=dev), with a much smaller AI banner ("slim").
I don't think this one, composed in this way, is being used anywhere in upcoming designs.

src/components/AiLoadingPanel.tsx
bdow, line 1: **question/thought:** Bear with, I know this is maybe too deep over a name, but is `AiImporting` the right term here? It really names itself based on a single use case of Material Importing. There's also the document upload, which uses the headline: "We’re processing you’re document...", which doesn't use "importing" as term anywhere.
I'd suggest something like `AiLoadingPanel`, maybe?

src/components/AiReview.tsx
apattersonATX-HB: 🤔 We'd end up needing to expand `ActionButtonProps` to allow for `variant` which i don't think we want to do since our layouts use this same prop. That also opens the door for someone to slide in 3 buttons into this panel which i dont think we'd like.


Who would control that scoped theme? Our layout component?

bdow: re: scoped theme: Layout component most likely. But that is not something to do here in this change.

exposing `variant` is maybe fine, but I 100% get the hesitation. I also don't think ActionButtonProps is something we'd end up using - we might want to also support IconButtons here too. I think its fine to be more of a "slot" for these buttons rather than dictating only 2 buttons and only these styles. 
Granted, I think this component is going away for now, so let's not spend too much time figuring this out here. We can keep the discussion going for the next component that requires a set of actions to be tied to it.

bdow (commented):
The simplicity of the AI panel looks great. It's a nicely abstracted building block. The use of it in the loading panel component looks good. I just have a small nit on the name of that component. Also, I don't think the AiReview component is something we want to ship at the moment. If you want to save it in a separate branch, that's fine, but I don't know if that is a pattern we'll be using yet.

src/components/AiLoadingPanel.tsx
apattersonATX-HB: 😆 Valid

src/components/AiReview.tsx
apattersonATX-HB: In the new designs no 🙁 .

What I think will work is turning it into `AiBanner` and use the proposed `slim` prop to turn it into just a title with text button variants. That is getting ahead of ourselves though.

I'll pop it out now and do that work in another PR

src/components/AiPanel.tsx
apattersonATX-HB: Thank you 🫡

src/components/AiLoadingPanel.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const r = await render(<AiLoadingPanel />);
+    // Then assistive tech waits for a pause rather than interrupting, and knows content is settling
+    expect(r.aiLoadingPanel).toHaveAttribute("role", "status");
+    expect(r.aiLoadingPanel).toHaveAttribute("aria-busy", "true");
+  });
+
+  it("centers its copy", async () => {
+    const r = await render(<AiLoadingPanel />);
+    expect(r.aiLoadingPanel_message).toHaveStyle({ textAlign: "center" });
+  });
```
bdow: **nit (non-blocking):** This sort of styling test can be validated via storybook. Whenever we/claude write tests, we should try to reduce redundancy between jest & storybook.

bdow: fwiw, I think copy tests are good. Minor styling tests like this one, probably not necessary.

src/components/AiLoadingPanel.tsx
```diff
@@ -0,0 +1,38 @@
...
+    // `aria-busy` so it knows the surrounding content is still settling.
+    <AiPanel role="status" aria-busy={true} {...tid}>
+      <div css={Css.df.fdc.aic.gap1.w100.$}>
+        <AiLoader />
+        <span css={Css.lg.aiBoldText.$} {...tid.title}>
+          {title}
+        </span>
+        <span css={Css.sm.color(Tokens.OnSurface).tac.$} {...tid.message}>
+          {message}
+        </span>
```
bdow, line 31: **suggestion(non-blocking):** based on the Figma's I saw, it appears the description/message area is actually more narrow that the container. 

Image

Notice the message wraps earlier. Not a blocker, but might end up having design ask about that in AT. It shows its another **71px** of horizontal padding, but I think we just do something like `Css.px8.$` here, which would give us 64px on either side. Then, on mobile, we probably want to remove that. So maybe the style is more like:
```
Css.existing.styles.ifMdAndUp.px8.$
```

bdow (approved):

src/components/AiLoadingPanel.test.tsx
apattersonATX-HB: 😅

PR #1383, opened 2026-08-18
feat: SideNav is hoisted into global nav menu on mobile

bdow:
Using a context, we pass the SideNav props needed for rendering the Side Nav in the mobile menu. Mobile Menu now is smart enough to start at the side nav, supporting going "back" to the Main Menu. Once a user does this, the only way back to the SideNav/Sub menu is by closing the menu and re-opening it. This is a corner cut for now until we can revisit for IA work down the road

### DIFF
--- a/src/components/Navbar/NavbarMobileMenu.tsx
+++ b/src/components/Navbar/NavbarMobileMenu.tsx
@@ -5,12 +5,16 @@ import { createPortal } from "react-dom";
 import { useLocation } from "react-router-dom";
 import { AppNavItems } from "src/components/AppNav/AppNavItems";
 import type { AppNavItem } from "src/components/AppNav/appNavTypes";
+import { Button } from "src/components/Button";
 import { IconButton } from "src/components/IconButton";
 import { Css, Tokens } from "src/Css";
 import { useEnvironmentBannerLayoutHeight } from "src/layouts/EnvironmentBannerLayout/EnvironmentBannerLayoutHeightContext";
 import { pageContentPaddingX } from "src/layouts/layoutSpacing";
+import { type MobileSubNavContent, useMobileSubNav } from "src/layouts/NavbarLayout/MobileSubNavContext";
 import { useTestIds } from "src/utils";
 import { zIndices } from "src/utils/zIndices";
+
+type MobileNavLevel = "sub" | "global";
 
 type NavbarMobileMenuProps = {
   items: AppNavItem[];
@@ -22,17 +26,19 @@ export function NavbarMobileMenu(props: NavbarMobileMenuProps) {
   // its own `navbar` prefix (which wins), so ids are the same either way.
   const tid = useTestIds(props, "navbar");
   const [isOpen, setIsOpen] = useState(false);
+  const [level, setLevel] = useState<MobileNavLevel>("sub");
   const { pathname, search } = useLocation();
+  const mobileSubNav = useMobileSubNav();
 
   usePreventScroll({ isDisabled: !isOpen });
+
+  const close = () => setIsOpen(false);
 
   // Close when navigation changes the route — covers programmatic `navigate()` and any item whose
   // handler pushes a new location. Same-route taps are handled by the drawer's anchor-click capture.
   useEffect(() => {
-    setIsOpen(false);
+    close();
   }, [pathname, search]);
-
-  const close = () => setIsOpen(false);
 
   return (
     <>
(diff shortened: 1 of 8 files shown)


src/layouts/NavbarLayout/MobileSubNavContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+ */
+export function useRegisterMobileSubNav(content: MobileSubNavContent, enabled: boolean): boolean {
+  const ctx = useContext(MobileSubNavContext);
+  const setMobileSubNav = ctx?.setMobileSubNav;
+  // Depend on fields, not the `content` object — callers pass `sideNav={{ items }}` inline.
+  const { top, items, footer } = content;
+
+  useLayoutEffect(() => {
+    if (!setMobileSubNav || !enabled) return;
+    setMobileSubNav({ top, items, footer });
```
stephenh, line 40: Calling a useState setter from a `useLayoutEffect` seems kind of odd, but I suppose that's basically what we have to do, to allow a child render function to pass props up the tree?

src/layouts/SideNavLayout/SideNavLayout.tsx
```diff
@@ -46,9 +47,12 @@ function SideNavLayoutContent(props: SideNavLayoutProps) {
   const bp = useBreakpoint();
   const tid = useTestIds(props, "sideNavLayout");
   const railCollapsedWidthPx = 56;
+  // Under NavbarLayout on `sm`, the rail is hidden and items show as the mobile menu's nested pane.
+  const hasMobileSubNavProvider = useRegisterMobileSubNav(sideNav, !bp.mdAndUp);
```
stephenh: Should `!bp.mdAndUp` be pushed into `useRegisterMobileSubNav`? Seems like maybe something it could/should own.

Also curious just b/c the line below this is having to "re-check `&& !mdAndUp`" which seems redundant -- like maybe `useRegisterMobileSubNav` should just return `showMobileSubNav` directly?

src/layouts/NavbarLayout/MobileSubNavContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/** Registered mobile sub-nav; `null` when the drawer has only the main pane. */
+export function useMobileSubNav(): MobileSubNavContent | null {
+  return useContext(MobileSubNavContext)?.mobileSubNav ?? null;
+}
+
+/**
+ * Publish `content` as the mobile menu's nested pane while `enabled`.
+ * Returns whether a {@link MobileSubNavProvider} exists (i.e. under `NavbarLayout`).
```
stephenh: Maybe document when/why there would/wouldn't be a `MobileSubVarProvider` under `NavbarLayout`? I kinda thought that was under our control, so would always be the case / not be the case.

src/components/Navbar/NavbarMobileMenu.tsx
```diff
@@ -22,30 +26,51 @@ export function NavbarMobileMenu(props: NavbarMobileMenuProps) {
   // its own `navbar` prefix (which wins), so ids are the same either way.
   const tid = useTestIds(props, "navbar");
   const [isOpen, setIsOpen] = useState(false);
+  const [level, setLevel] = useState<MobileNavLevel>("sub");
   const { pathname, search } = useLocation();
+  const mobileSubNav = useMobileSubNav();
```
stephenh, line 31: Kinda wonder if we should keep "side nav" instead of switching to a "sub nav" term -- I get that, to this `NavbarMobleMenu` this is now "a sub menu", but semantically to the rest of the app "this is the logical side data, just moved around".

I.e. reading the term "mobile sub nav" by itself might make me think there is some net-new / mobile-only "sub nav" that isn't "oh right just the moved around side nav".

stephenh (approved):

src/components/Navbar/NavbarMobileMenu.tsx
bdow: I'd personally rather keep it as "sub nav", mostly because we don't know where this nav may live in the future once we do some IA work, or it also **doesn't** need to come from the side nav. Any other sort of sub navigation could get hoisted up to here.

src/layouts/NavbarLayout/MobileSubNavContext.tsx
bdow: Yes — SideNavLayout is a descendant, so it can’t pass props up to NavbarLayout. `useLayoutEffect` publishes before paint so we don’t flash the global-only drawer.

src/layouts/SideNavLayout/SideNavLayout.tsx
bdow: Good call. The hook now owns `!mdAndUp` and returns whether the mobile pane is active, so SideNavLayout can use that directly as `showMobileSubNav`.

src/layouts/NavbarLayout/MobileSubNavContext.tsx
bdow: NavbarLayout always mounts the provider. The `false` case is standalone `SideNavLayout` (no NavbarLayout ancestor) or `mdAndUp` (rail stays). Updated the JSDoc to say that.

PR #1384, opened 2026-08-18
feat: Support 'CompanionRow' in GridTable

bdow:
A CompanionRow is a "companion"/related row to one of the "data" rows in the table. This can be used for different circumstances:
1. AI row actions - per new designs
2. Row details/actions like in the existing Product Option variants table


henrysipp (approved):
I was going to complain about the `make*()` methods but then saw the `useComputed` hook. Very cool!

PR #1386, opened 2026-08-18
feat: CenteredLayout component for supporting standardized non-full screen layout widths

bdow:


src/layouts/FormSectionLayout/FormSectionLayout.tsx
```diff
@@ -20,13 +21,13 @@ export type FormSectionLayoutProps = {
...
 
-/** Centered (720px) width/column shell for a form built out of `FormSection`s — e.g. as a `WorkflowLayoutStep`'s `content`. */
+/** Centered `sm` width shell for a form of `FormSection`s — e.g. a `WorkflowLayoutStep`'s `content`. */
 export function FormSectionLayout(props: FormSectionLayoutProps) {
   const { title, description, actions, withAutoSave, initialFields, sections } = props;
   const tid = useTestIds(props, "formSectionLayout");
 
   return (
-    <div css={Css.df.fdc.gap8.w100.pt4.maxwPx(720).mxa.ifSm.px2.$} {...tid}>
+    <div css={{ ...centeredWidthCss("sm"), ...Css.df.fdc.gap8.pt4.$ }} {...tid}>
```
henrysipp: Why does this spread the `centeredWidthCss` rather than just using the new component? 
Anything preventing us from simply adding a `css` field to the `CenteredLayout` and passing through the `df.fdc.gap8.pt4`?

henrysipp: Or embedding a div as a child similar to the WorkflowLayout stories below.

bdow: Mostly precedent. We typically like to keep CSS contracts very tight on what sort of styles the component will support. We could pass CSS to that element, and avoid this, or nesting an extra div as well.

src/layouts/layoutSpacing.ts
```diff
@@ -1,6 +1,15 @@
 import { Css } from "src/Css";
 import { pageContentPaddingXValue } from "./layoutVars";
 
+/** Outer max-width for {@link CenteredLayout}. `sm` is 768 so md+ 24px pad leaves 720 of content. */
+export const centeredMaxWidthPx = { sm: 768, lg: 1440 } as const;
+export type CenteredLayoutSize = keyof typeof centeredMaxWidthPx;
+
+/** Centered shell: `w100` + outer max-width + 12px pad, 24px from `md` up. */
+export function centeredWidthCss(size: CenteredLayoutSize) {
```
henrysipp: For reasons specified in the other comment I feel like exporting this could be "dangerous". A user (or agent) could inadvertently grab this rather than the preferred component, reducing strictness.

Why not local to the layout file?

henrysipp (approved):
Nits but I might just not fully understand some of our conventions.

src/layouts/layoutSpacing.ts
bdow: Yea, fair. It could be local to the layout file.

PR #1387, opened 2026-08-18
feat: ai banners

apattersonATX-HB:
**`AiBanner`** — composes `AiPanel`, so it carries the wordmark and card. Title over an optional message, with `secondaryAction` (quaternary) and `primaryAction` (AI gradient). [Figma](https://www.figma.com/design/DchiwVkssXeYi2Er8sMU2k/H2-2026-Plans---Automated-Construction-Doc-Capture?node-id=1608-25908&m=dev)

**`AiSlimBanner`** — deliberately *not* an `AiPanel`: no card, no logo. Just the gradient strip, an `aiStar`, the title, and one text-variant action. [Figma](https://www.figma.com/design/DchiwVkssXeYi2Er8sMU2k/H2-2026-Plans---Automated-Construction-Doc-Capture?node-id=1518-51815&m=dev)


src/components/AiBanner.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function Sample({ title, width = 1440, children }: { title: string; width?: number; children: ReactNode }) {
+  return (
+    <div>
+      <h2 css={Css.lg.mb1.$}>{title}</h2>
+      <div css={Css.wPx(width).$}>{children}</div>
+    </div>
+  );
+}
```
bdow: **nit:** Regarding these widths, we should probably just use the full screen, and when we want to show what it looks like on a mobile view, use the chromatic snapshots to capture on the mobile viewport as well. This is done via the `export default` at the top like:
```
export default {
  component: StepperTabs,
  parameters: {
    chromatic: { modes: viewportModes("desktop", "mobile1") },
  },
} as Meta;
```

I noticed we're doing this same thing in some other stories, but would prefer to just have one story that takes up the full width of the page, and not define widths within stories.

apattersonATX-HB: NP 👍🏽

src/components/AiSlimBanner.tsx
```diff
@@ -0,0 +1,34 @@
+import { Button } from "src/components/Button";
+import { Icon } from "src/components/Icon";
+import type { ActionButtonProps } from "src/components/Layout/layoutTypes";
+import { Css } from "src/Css";
+import { useTestIds } from "src/utils";
+
+export type AiSlimBannerProps = {
+  title: string;
+  action?: ActionButtonProps;
```
bdow, line 9: **thought(non-blocking):** I'll be curious to see what they decide here, if this should be a slot for multiple actions potentially - or if its really just one. I'll follow up with design on this. I have a working session tomorrow with them around some of this stuff.

src/components/AiSlimBanner.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <Sample title="Without an action">
+        <AiSlimBanner title={title} />
+      </Sample>
+
+      <Sample title="Long copy wraps">
+        <AiSlimBanner
+          title={`${title} found across every plan in this project, including the ones you archived last month`}
+          action={action}
+        />
+      </Sample>
```
bdow: **question:** Did design give you an example of long text to see how it should wrap? Like would the icon be centered vertically or aligned to the top... that sort of thing. Probably can be validated in AT with them.

bdow (approved):
Looks great. Small nits, but happy to have any changes needed to be follow ups.

src/components/AiSlimBanner.tsx
apattersonATX-HB: This banner is so thin i can't imagine them wanting multiple actions.

src/components/AiSlimBanner.stories.tsx
apattersonATX-HB: Nope. I will call that out. I'm pushing an update for mobile views as well where the buttons stack since this looks pretty awful. 
 
image-1787152865021

Screenshot 2026-08-19 at 10 23 37 AM

PR #1388, opened 2026-08-18
feat: added handling for material card designs

Jacob-Diebold:
## Summary

Adds GridTable / TableCard support for the Material Catalog card layout

- Split the eyebrow into `cardLeftEyebrowSlot` / `cardRightEyebrowSlot. `cardEyebrowSlot` still works as a deprecated alias for the left slot.
- Add `cardCarouselSlot` for a footer thumbnail strip (title, swatch links, overflow chevrons). Carousel cards stretch the row `rowLink` / `onClick` across the card instead of wrapping it, so thumbnail links stay clickable and keyboard-reachable.
- Add `cardHeight` and `cardImageFit` on `GridTable` (`cover` by default, `contain` for product photos). Interactive cards get a hover wash.
- `GridTableLayout` can lock to card view with `defaultView="card"` alone — `withCardView` is only needed to show the list/card toggle.

Screenshot 2026-08-18 at 3 46 12 PM
Screenshot 2026-08-18 at 3 46 27 PM


Jacob-Diebold:
@bdow Let me know what you think! I tried to avoid nested buttons; but let me know if this is the wrong approach.

src/components/Table/components/TableCard.tsx
```diff
@@ -102,60 +134,134 @@ export function TableCard<R extends Kinded>(props: TableCardProps<R>) {
...
+
+  return (
+    <div
+      css={{
+        ...Css.w100
+          .hPx(height)
+          .ba.br12.bc(Tokens.FieldBorderDefault)
+          .bgColor(Tokens.SurfaceRaised)
+          .df.fdc.gap2.relative.if(isInteractive)
+          .onHover.bc(Tokens.FieldBorderHover).onHover.cursorPointer.$,
```
bdow: **nit/question:** I haven't seen two `onHover`'s before. Is that needed? Secondly, `cursorPointer` doesn't need to just be on hover, its a state that'll only show on hover anyways. So we could move that up.

src/components/Table/components/TableCard.tsx
```diff
@@ -102,60 +134,134 @@ export function TableCard<R extends Kinded>(props: TableCardProps<R>) {
...
+      {!to && onClick && (
+        <button
+          type="button"
+          aria-label={title}
+          onClick={onClick}
+          css={stretchedAction}
+          {...focusProps}
+          {...tid.action}
+        />
+      )}
```
bdow: **question:** So when we have a carousel, it appears we jump ship early above, and then we handle adding links down here, correct? If that's right, I think I'd prefer to just do this one way.
Here's some psuedo code of what I'd be thinking:
```
function TableCard() {
  const cardContent = (
    {/** Use focus-within psuedo selector to add focus styles on parent */}
    <div css={Css.all.container.styles.NO.PADDING.$}>
      {/** 
         maybe the helper `getButtonOrLink`.
         Add padding and other styles needed here
      */}
      <Link or Button>
        {/** Current card content */}
      </Link or Button>
      {hasCarousel && <CarouselSection />}
    </div>
  );
}
```
The idea is that we can change our nesting around so that the card out element is not a link or button. Instead its just a div, but the interactive element directly inside takes up the full width/height of that container. Doesn't really change anything except the need of using `focus-within` to properly style the parent and not the element in focus.
Then the carousel section can be appended separately, within the card, but as a sibling to the interactive element. 
I think that approach would simplify the render logic. Then you can get rid of that "hoverWash" element too.

src/components/Table/components/TableCard.tsx
```diff
@@ -176,24 +282,186 @@ export function TableCardView(props: TableCardViewProps) {
...
+    </div>
+  );
+}
+
+type CarouselProps = {
+  thumbnails: CardCarouselThumbnail[];
+};
+
+/** Horizontal thumbnail strip with overflow chevrons — private to TableCard. */
+function Carousel(props: CarouselProps) {
```
bdow: **suggestion** I'd love to see this carousel component become abstracted so it can be used in other cases. I don't know what those other cases are yet, but its a good component that we might want to use later. I could see it being used on any other horizontally scrolling element.

src/components/Table/components/TableCard.tsx
```diff
@@ -176,24 +282,186 @@ export function TableCardView(props: TableCardViewProps) {
...
+        {canScrollLeft && (
+          <div
+            css={Css.absolute.left0.top0.bottom0.wPx(16).z1.add("pointerEvents", "none").$}
+            style={{ background: leftFade }}
+          />
+        )}
+        {/* py/my: oxa also clips overflow-y, so pad for the 4px bshFocus ring without growing layout */}
+        <div
+          ref={stripRef}
+          css={Css.df.gap1.oxa.fs0.sbwn.pyPx(4).myPx(-4).element("::-webkit-scrollbar").dn.$}
```
bdow: **suggestion:** there are some other properties I think that'd be useful here, like [`scroll-snap-align`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scroll-snap-align) & [`scroll-snap-type`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scroll-snap-type). These kind of keep carousel items positioned well when scrolling side to side

src/components/Table/components/TableCard.tsx
```diff
@@ -176,24 +282,186 @@ export function TableCardView(props: TableCardViewProps) {
...
+  useEffect(() => {
+    updateOverflow();
+    const el = stripRef.current;
+    if (!el) return;
+    el.addEventListener("scroll", updateOverflow);
+    return () => el.removeEventListener("scroll", updateOverflow);
+  }, [updateOverflow, thumbnails.length]);
+
+  function scrollByDir(dir: -1 | 1) {
+    stripRef.current?.scrollBy({ left: dir * 96, behavior: "smooth" });
```
bdow: **question/nit(non-blocking):** why `96`? Could we identify the width of the carousel and scroll by that amount?

src/components/Table/components/TableCard.tsx
bdow: Looking through this more, I definitely think carousel deserves its own component.

src/components/Table/components/TableCard.tsx
```diff
@@ -176,24 +282,186 @@ export function TableCardView(props: TableCardViewProps) {
...
+            <Icon icon="chevronLeft" inc={3} color={Tokens.OnSurface} />
+          </button>
+        )}
+      </div>
+      <div css={Css.relative.fg1.mw0.$}>
+        {canScrollLeft && (
+          <div
+            css={Css.absolute.left0.top0.bottom0.wPx(16).z1.add("pointerEvents", "none").$}
+            style={{ background: leftFade }}
+          />
```
bdow: **thought/suggestoin:** I think we get rid of this separate `fade` element, and just work that into the button element.

src/components/Table/components/TableCard.tsx
bdow: **thought/question:** Do you need `.element("::-webkit-scrollbar").dn` if we are already defining `scrollbar-width: none`?

src/components/Table/components/TableCard.tsx
Jacob-Diebold: Ahh, you're right. CC snuck that in and I didn't catch it. Thanks!

src/components/Table/components/TableCard.tsx
Jacob-Diebold: Thats a way better approach than the convoluted way I was doing it, thanks!

One thing that I differed slightly on, when I tried the focus-within approach and focused one of the thumbnails I would get a focus ring around the card and the thumbnail; this made it hard to differentiate between what I was actually focusing.

I ended up doing this to get around that, but let me know if theres a better solution!
```
.when(":has(> a:focus-visible, > button:focus-visible)")
.bshFocus.end.if(!!action)
.cursorPointer.onHover.bc(Tokens.FieldBorderHover)
.bgColor(Tokens.SurfaceRaisedHover).$
```

I was able to get rid of the hoverWash element, but I had to add a marker for when the card was hovered and something like this to the hero image:
```
css={Css.h100.w100.objectFit(imageFit).if(!!action).when(cardHovered, "ancestor", ":hover").o(0.5).$}
```

There isn't anything else in beam really using markers like this, but this feels kind of unique. Whenever any part of the card is hovered we need to add some opacity to the hero image so it gives the same kind of effect as the bg color hover change. I feel like theres probably a more simple way to do this; but I couldn't find it if there is.

src/components/Table/components/TableCard.tsx
Jacob-Diebold: Good call out, I haven't needed to use those before. It is really nice here.

src/components/Table/components/TableCard.tsx
Jacob-Diebold: Yep, that is a way better approach.

src/components/Table/components/TableCard.tsx
Jacob-Diebold: I removed the fade element and ended up using a mask-image so the items themselves fade out underneath the chevron. The gradient needs to match whatever is behind the carousel so any sort of background color wont break it i.e. when hovered, dark mode, etc.
```
  // Fade the strip's own content out under whichever chevron is showing. Masking the items means the
  // fade works on any background without being told what it is.
  const maskImage = [
    "linear-gradient(90deg,",
    canScrollBack ? `transparent 0, transparent ${clearWidth}px, black ${fadeWidth}px,` : "black 0,",
    canScrollForward
      ? `black calc(100% - ${fadeWidth}px), transparent calc(100% - ${clearWidth}px), transparent 100%)`
      : "black 100%)",
  ].join(" ");
Css.df.fg1.mw0.oxa.sbwn
.sst("x mandatory")
.gapPx(gap)
// Padding so the focus ring isn't clipped
.pPx(4)
.mPx(-4)
.add("scrollPaddingInline", `${4}px`)
.add("maskImage", maskImage).$
  ```
  
  I added some different stories for the carousel so you can see how it handles with different backgrounds, chevron sizes, etc.

File: src/components/Table/components/TableCard.tsx (hunk 4 above)
Comment (Jacob-Diebold) [reply src/components/Table/components/TableCard.tsx]: Nope, thanks for catching that.

File: src/components/Table/components/TableCard.tsx [hunk 7]
```diff
@@ -176,24 +238,101 @@ export function TableCardView(props: TableCardViewProps) {
...
         </div>
-      </div>
+      )}
+    </>
+  );
+
+  return (
+    <div
+      css={{
+        ...Css.markerOf(cardHovered)
```
bdow: **question:** Do we still need this marker

Jacob-Diebold: Not if design approves the change to the card box shadow on hover

Just waiting to hear back before I change that part and then I'll remove this too

src/components/Carousel.tsx
```diff
@@ -0,0 +1,162 @@
...
+            .pPx(4)
+            .mPx(-4)
+            .add("scrollPaddingInline", `${4}px`)
+            .add("maskImage", maskImage).$
+        }
+        {...tid.items}
+      >
+        {Children.map(children, (child) => (
+          <div css={Css.fs0.ssa("start").$}>{child}</div>
+        ))}
```
bdow, line 99: **thought:** I've never used `Children.map` before. I'm curious to see how this works out.

src/components/Carousel.tsx
```diff
@@ -0,0 +1,162 @@
...
+}
+
+type ScrollButtonProps = {
+  direction: "back" | "forward";
+  label: string;
+  /** The chevron's `Icon` size; its box is derived from this. */
+  inc: number;
+  visible: boolean;
+  onClick: () => void;
+  "data-testid"?: string;
```
bdow: **nit:** We usually (at least I think) do not define `data-testid` as a prop. When you spread it into a component via `<ScrollButton {...tid.next} />` then it shouldn't need an explicit prop for it.

bdow (approved):
Great start to the carousel component. I could see us updating in the future, but a great starting point. 
Tablecard looks good too. Only suggestion is to remove the current hover behvvior with the marker and just do that box-shadow while waiting for Andrea's response.

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
```diff
@@ -220,6 +220,36 @@ describe("GridTableLayout", () => {
       expect(r.query.viewToggleButton).not.toBeInTheDocument();
     });
 
+    it("renders cards with defaultView=card without withCardView and without a toggle", async () => {
+      setRunningInJest();
```
stephenh, line 224: Isn't this done automatically for all of our tests?

src/components/Table/cardSlots.ts
```diff
@@ -49,3 +79,8 @@ export function cardDataBlockSlot(props: { label: string; value: ReactNode | str
 export function cardProgressSlot(value: number): CardProgressSlot {
   return { kind: "progress", value };
 }
+
+/** Footer carousel (title + thumbnail links). */
+export function cardCarouselSlot(props: { title: string; thumbnails: CardCarouselThumbnail[] }): CardCarouselSlot {
```
stephenh: @bdow curios if "carousel slot" is too specific vs. just "there is a slot at the bottom of the card and this page just happens to put a carousel component in it".

src/components/Carousel.tsx
```diff
@@ -0,0 +1,161 @@
...
+  chevronInc?: number;
+};
+
+/**
+ * A horizontally scrolling strip of items, with chevrons that appear only when the strip overflows.
+ *
+ * The chevrons scroll a "page" (the strip's visible width) at a time, and items snap into place so
+ * scrolling never leaves an item half-shown.
+ */
+export function Carousel(props: CarouselProps) {
```
stephenh, line 26: cc @bdow kinda surprised we didn't have a Carousel already, makes me wonder if we have some 3rd party carousels floating around in `internal-frontend` that we could eventually remove / move over to this component.

src/components/Table/cardSlots.ts
bdow: I could this this being more of a `cardActions` section. Something to denote that this is a footer with separate interactions

src/components/Layout/GridTableLayout/GridTableLayout.test.tsx
Jacob-Diebold: Doesn't look like it, this test fails without it and other tests in this file that have to do with cards are calling it as well. 

`FAIL  src/components/Layout/GridTableLayout/GridTableLayout.test.tsx > GridTableLayout > view toggle > renders cards with defaultView=card without withCardView and without a toggle
Error: Unable to find an element by: [data-testid="tableCard_title"]`

src/components/Table/cardSlots.ts
Jacob-Diebold: Would we want to bake the footer slot options in? i.e. {type: carousel |(other card action types we declare here eventually)}, but the tableCard still owns this thumbnail component?

Or let the caller pass whatever they want (i.e. internal frontend passes in the thumbnail component into a cardActions slot?)

This should be the last thing to resolve before I can merge this.

bdow: Maybe a better name would be `interactiveFooter`. I could see it being a union type of supported interactive footers.`interactiveFooter: CarouselProps | ActionBarProps` or something... I'd prefer to keep it strict so we have more control over what is put in that area and that it is consistent.

Jacob-Diebold: I like it!

Jacob-Diebold: @bdow I have the changes merged to support this in the latest commit if you want to look over it at all.

src/components/Table/cardSlots.ts
```diff
@@ -21,19 +21,27 @@ export type CardStatusSlot = CardSlotBase<"status"> & { tag: CardTag };
...
   swatchUrl: string;
   label: string;
   to: string;
 };
 
-export type CardCarouselSlot = CardSlotBase<"carousel"> & {
+export type CardCarouselFooter = {
+  kind: "carousel";
   title: string;
   thumbnails: CardCarouselThumbnail[];
```
bdow, line 35: **thought (non-blocking):** Happy to have this done as a follow up, but I think at this point we're halfway or more through abstraction. The one that sticks out to me is this `thumbnails: CardCarouselThumbnail[];`. I almost think that we'd just want to pass in whatever the Carousel props would be. I think Carousel really just accepts children right now, right? But again, I haven't thought enough through - but just leaning towards this idea that if we say "this footer is the carousel", then it should probably just be passed the carousel props.

src/components/Table/cardSlots.ts
```diff
@@ -21,19 +21,27 @@ export type CardStatusSlot = CardSlotBase<"status"> & { tag: CardTag };
 export type CardDataBlockSlot = CardSlotBase<"dataBlock"> & { label: string; value: ReactNode | string | number };
 export type CardProgressSlot = CardSlotBase<"progress"> & { value: number };
 
-/** One thumbnail link in a thumbnail carousel footer. */
+/** One thumbnail link in a carousel interactive footer. */
 export type CardCarouselThumbnail = {
   id: string;
   swatchUrl: string;
   label: string;
   to: string;
```
bdow, line 29: **thought:** swatchUrl might be a bit too specific. Maybe just `imgUrl`

bdow (approved):

src/components/Table/cardSlots.ts
Jacob-Diebold: I might make a follow-up ticket to look into this a bit more to see how we want to tackle it. Since this carousel footer is a title + the carousel component, it might not be quite as straightforward.

PR #1389, opened 2026-08-19
fix: Responsive support for CenteredLayout

bdow:


truss-config.ts
```diff
@@ -90,9 +90,9 @@ const sections: Sections = {
   ],
   layoutContainer: () => [
     newMethod("layoutContainer", {
-      width: `min(100%, ${documentScrollChromeWidth()})`,
-      left: documentScrollChromeLeft(),
-      position: "sticky",
+      width: `min(100%, ${documentScrollContentWidth()})`,
+      left: documentScrollContentLeft(),
+      position: "sticky" as const,
```
allangaldinosilva, line 95: nit: seems like we don't need this `as const` 🤔

allangaldinosilva (approved):
Looks great! thanks 👏

PR #1390, opened 2026-08-20
feat: text fields support ai proposals

apattersonATX-HB:
## [DS-248](https://linear.app/homebound-team/issue/DS-248/text-fields-to-support-strike-through-original-value-and-ai-background): AI-proposed values on text fields

- `useAiProposal(value, proposedValue)` returns `{ isAiMode, effectiveValue, onUserEdit }`. The proposal is the field's *effective* value — displayed and handed to the field so editing starts from it and commits through the normal `onChange`/`onSelect`.
- AI mode ends on the user's **first edit** and never returns to that state. This is driven by the edit event, not by watching `value`: a user rejecting a proposal by re-entering the on-record value (5 → 7 proposed → types 5) produces no change to `value`, so there is nothing to observe.
- `TextFieldBase.onDomChange` is the single call site for every typed edit across all nine fields. Only `ComboBoxBase` (option select), `DateFieldBase` (calendar) and `Autocomplete` (option select) hook in separately, because those commits update a controlled value and fire no DOM event.


bdow:
Without going through the code, there are a couple UX changes that we'd like to update:
1. When a user clicks into a field, the old value should still be visible as a strike through. If the user makes an update to the field, then not until the user `blur`'s the field would the strikethrough value disappear. This allows it to stay as a reference while the user is updating the AI proposal.
2. Disabled state shouldn't allow a user to click into it (non-blocking, we're focused really on just getting the Numberfield out as a prototype)
3. We want to keep the AI bg, text color, and font weight while the user is focused on the input. And only change when once the user types. (this might be tricky to do a re-render while typing, we don't want to lose their cursor spot. So if its overly complicated, let's punt on it)

src/inputs/TextFieldBase.tsx
bdow, line 1: **question/thought:** Could we have done your `useAiProposal` hook just here in TextFieldBase?

apattersonATX-HB: I wish. `TextFieldBase` wrappers need the proposal *before* `TextFieldBase` exists
 - `ComboBoxBase` uses it to mark the proposed option as selected in the dropdown
 - `NumberField` feeds it through `useNumberFieldState`
 - `DateField` opens the picker on the effectiveValue

src/inputs/DateFields/DateFieldBase.tsx
```diff
@@ -52,15 +53,20 @@ type DateFieldCommonProps = Pick<
   useYearPicker?: boolean;
 };
 
-export type DateFieldProps = DateFieldCommonProps & {
-  value: PlainDate | undefined;
-  onChange: (value: PlainDate | undefined) => void;
-};
+/** Value proposed by an AI model; puts the field in AI mode. */
+type ProposedValue<V> = { proposedValue?: V };
```
bdow, line 57: **thought:** ProposedValue is nearly always going to be a `string`. at least that's the plan from the backend. I wonder where we do the conversion... I'm not sure where it makes the most sense to do the conversion. Should Beam just always expect the type it would for both `value` and `proposedValue`? 🤷  Or would it be better for dev ergonomics if the `proposedValue` can be whatever, and Beam tries to massage it into what it would expect. FWIW, we kind of already do this. TextFieldBase always has a string `value`, from `inputProps.value`. So maybe it'd be better here. One less step for most engineers to deal with.

src/inputs/DateFields/DateFieldBase.tsx
```diff
@@ -137,8 +153,8 @@ export function DateFieldBase(props: DateRangeFieldBaseProps | DateSingleFieldBa
           // When focused, change to use the "short" date format, as it is simpler to update by hand and parse.
           setInputValue(
             (isRangeMode
-              ? formatDateRange(props.value, dateFormats.short)
-              : formatDate(props.value, dateFormats.short)) ?? "",
+              ? formatDateRange(effectiveValue as DateRange | undefined, dateFormats.short)
+              : formatDate(effectiveValue as PlainDate | undefined, dateFormats.short)) ?? "",
```
bdow, line 157: **nit:** Would like to avoid this `as` casting. But seems safe enough at the moment.

src/inputs/DateFields/DateFieldBase.tsx
bdow, line 1: **thought/suggestion:** honestly would've preferred to not have done these more nuanced fields yet. Doesn't hurt to have done it, but does feel like we're front loading more than needed. especially for a POC where we are not sure if this will be the final direction.

src/inputs/hooks/useAiProposal.test.ts
bdow, line 1: **suggestion:** Can this follow the `Given/When/Then` comment flow, please?

src/inputs/hooks/useAiProposal.ts
bdow, line 1: **praise:** I like the centralized logic handling this hook does for us.

bdow (approved):

apattersonATX-HB:
We merged this in to get ready for demo. Cleanup to follow. 

One additional thing we noticed is that if the suggested value is the same as the proposed value it shows both in the aiMode state. It shouldn't do that

PR #1391, opened 2026-08-20
fix: fixed icon menu link styling

Jacob-Diebold:
## Description

- Menu items that are both an icon item and an in-app `Link` were wrapping the label under the icon. `Icon` renders a block-level SVG, and the relative-URL `Link` had no flex layout, unlike the external `<a>` wrapper.
- Apply `display: flex` / `align-items: center` / `width: 100%` to in-app `Link` menu items so the icon and label stay on one row.
- Cover the icon + relative URL case in `ButtonMenu` tests and the `IconMenuItems` story.

### Old
Screenshot 2026-08-20 at 8 32 20 AM

### New
Screenshot 2026-08-20 at 8 32 43 AM

### DIFF
--- a/src/components/ButtonMenu.test.tsx
+++ b/src/components/ButtonMenu.test.tsx
@@ -207,6 +207,22 @@ describe("ButtonMenu", () => {
     openSpy.mockRestore();
   });
 
+  it("keeps an icon and label on one row when the item is an in-app link", async () => {
+    // Given a menu item that is both an icon item and a relative URL
+    const r = await render(
+      <ButtonMenu
+        trigger={{ label: "Trigger" }}
+        items={[{ label: "Install Tasks", onClick: "/install-tasks", icon: "linkExternal" }]}
+      />,
+      withRouter(),
+    );
+    // When opening the menu
+    click(r.trigger);
+    // Then the in-app link is a flex row so the block-level SVG does not wrap the label
+    expect(r.trigger_installTasks.querySelector("a")).toHaveStyle({ display: "flex" });
+    expect(r.trigger_installTasks).toHaveTextContent("Install Tasks");
+  });
+
   it("closes the menu when clicking the trigger again", async () => {
     // Given a ButtonMenu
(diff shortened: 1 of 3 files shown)


src/components/ButtonMenu.test.tsx
```diff
@@ -207,6 +207,22 @@ describe("ButtonMenu", () => {
...
+        items={[{ label: "Install Tasks", onClick: "/install-tasks", icon: "linkExternal" }]}
+      />,
+      withRouter(),
+    );
+    // When opening the menu
+    click(r.trigger);
+    // Then the in-app link is a flex row so the block-level SVG does not wrap the label
+    expect(r.trigger_installTasks.querySelector("a")).toHaveStyle({ display: "flex" });
+    expect(r.trigger_installTasks).toHaveTextContent("Install Tasks");
+  });
```
bdow: **suggestion:** I think the visual storybook test would be sufficient. I'm not sure this really would ensure properly layout. This is an extreme edgecase,  but if someone for some reason set all flex containers to be flex-wrap: wrap, then this test wouldn't catch it.

Jacob-Diebold: Sounds good, I'll remove it!

src/components/internal/MenuItem.tsx
```diff
@@ -170,7 +170,7 @@ function maybeWrapInLink(
       </span>
     </a>
   ) : (
-    <Link className="navLink" to={onClick}>
+    <Link className="navLink" css={Css.df.aic.w100.$} to={onClick}>
```
bdow: **nit:** I don't think the w100 is doing anything here. its basically the `df` that is helping keeping things on the one line.

bdow (approved):
Thanks for fixing!

src/components/internal/MenuItem.tsx
Jacob-Diebold: ahh, you're right. Thanks. Fixed!

PR #1393, opened 2026-08-21
feat: Consolidate headers into single component

bdow:
FormSectionLayout, FormSection, and FormSectionChild did their own header styles. ContentHeader followed the same makeup, but was used for outside of forms. Due to their increasing similarities, we are consolidating all of them.

### DIFF
--- a/src/components/Headers/ContentHeader.tsx
+++ b/src/components/Headers/ContentHeader.tsx
@@ -1,14 +1,23 @@
 import { ReactNode } from "react";
+import { AutoSaveIndicator } from "src/components/AutoSaveIndicator";
 import { HeaderAction, HeaderActions } from "src/components/Headers/HeaderActions";
 import { Css, Only, Padding, Tokens, Xss } from "src/Css";
 import { useTestIds } from "src/utils";
 
 type ContentHeaderXss = Xss<Padding>;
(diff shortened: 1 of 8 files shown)


src/components/Headers/ContentHeader.tsx
```diff
@@ -1,14 +1,23 @@
...
 type ContentHeaderXss = Xss<Padding>;
 
+export type ContentHeaderLevel = 2 | 3 | 4;
+
 export type ContentHeaderProps<X = ContentHeaderXss> = {
   title?: string;
   description?: ReactNode;
   actions?: HeaderAction[];
+  /** When true, prepends `AutoSaveIndicator` in the actions area. */
+  withAutoSave?: boolean;
```
stephenh, line 16: From a DX perspective, ideally pages wouldn't have to tell the content header "btw we are/are-not going to auto-save" -- it would "just know" or "just work".

Could we always reserve the whitespace for the auto-save indicator, even if ~50% of the time it's never enabled (b/c the form isn't autowave) but then for pages that are auto-save, which we won't technically know until the auto-save state changes, we could show the indicator, and not cause any content reflows?

stephenh (approved):

PR #1394, opened 2026-08-21
fix: panel spacing; slim banner icon alignment

apattersonATX-HB:
Design feedback


src/components/AiLoadingPanel.tsx
```diff
@@ -21,7 +21,7 @@ export function AiLoadingPanel(props: AiLoadingPanelProps) {
     // `status` rather than `alert` so assistive tech waits for a pause instead of interrupting, and
     // `aria-busy` so it knows the surrounding content is still settling.
     <AiPanel role="status" aria-busy={true} {...tid}>
-      <div css={Css.df.fdc.aic.gap1.w100.$}>
+      <div css={Css.df.fdc.aic.gap1.w100.px1.$}>
```
bdow, line 24: Just an fyi. I'm hoping to define these paddings based on direction from design. Sounds like there might be two different variants of this. Non-blocking.

bdow (approved):

PR #1395, opened 2026-08-23
feat: add AI workflow chrome and split AiPanel from AiCard

bdow:
Workflow and form layouts can opt into aiMode for the page background, logo card, gradient title, and AI CTAs. AiPanel is only the background; AiCard is the wordmark and raised surface. Prefer AiBanner, AiLoadingPanel, AiSlimBanner, or aiMode — treat the primitives as an escape hatch.

The AiPanel no longer composes the AiCard as the number of differences in when to use which sizes and paddings made that composition have too many knobs. By breaking them out we simplify the prop usage.


bdow:
Moved this forward to unblock other work. Will address feedback still.

PR #1398, opened 2026-08-25
feat: gridtable row banner

apattersonATX-HB:
### [DS-320](https://linear.app/homebound-team/issue/DS-320/ai-table-row-inline-banner) 

Inline error/warning notice for a GridTable row's companion slot


src/components/RowBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const tid = useTestIds(props, "rowBanner");
+  return (
+    <div css={Css.df.ais.gap(0.5).w100.p1.br4.xs.bgColor(Tokens.Surface).ba.bc(color).bshBasic.$} {...tid}>
+      {/* Matches the copy's 16px line box, so the taller icon centers on the first line rather than
+          drifting to the middle of wrapped copy — and overflows the line instead of growing it. */}
+      <span css={Css.df.aic.fs0.h2.$}>
+        {/* Padding and icon spacing is slightly different than our default icon atm */}
+        <span css={Css.df.aic.jcc.fs0.hPx(18).wPx(18).br4.bgColor(bgColor).$}>
+          <Icon icon={icon} inc={1.75} color={color} {...tid.type} />
+        </span>
```
apattersonATX-HB: Need to check with design on the icon padding here. There is more spacing in the designs than our usual icon with bg gives

src/components/RowBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+  /** Rendered in order, always as text buttons. */
+  actions?: ActionButtonProps[];
+};
+
+/**
+ * An error/warning notice for a `GridTable` row's `companion` slot.
+ *
+ * Any bgColor behind the banner belongs to the row, not to the banner.
+ */
+export function RowBanner(props: RowBannerProps) {
```
apattersonATX-HB: The more I see these components in isolation the less i want to call them out as `Ai` specific components

bdow: +1. Won't always be used for Ai purposes, but that is where it originated from.

bdow: Let's rename this to `InlineFeedbackBanner`. This matches the Figma component name of `Inline Feedback`: https://www.figma.com/design/DchiwVkssXeYi2Er8sMU2k/H2-2026-Plans---Automated-Construction-Doc-Capture?node-id=922-67294&t=l3Txa70w1QaYvzA4-4

src/components/RowBanner.tsx
bdow: FWIW, I think this is going to be the `Tag` component. It could be an `iconOnly` version of the Tag, but sometimes might include text as well per this Figma: https://www.figma.com/design/qDCLrDQECBRgPh5Rx7UFvy/H2-2026-Cost-Intell?node-id=160-11435&t=az0pkVxIgwy60So7-4

src/components/RowBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <span css={Css.df.aic.jcc.fs0.hPx(18).wPx(18).br4.bgColor(bgColor).$}>
+          <Icon icon={icon} inc={1.75} color={color} {...tid.type} />
+        </span>
+      </span>
+      <span css={Css.fg1.mw0.color(Tokens.OnSurface).$} {...tid.description}>
+        {description}
+      </span>
+      {actions.length > 0 && (
+        // Pinned to the same line box so a text button's taller one can't grow the banner either.
+        <div css={Css.df.aic.gap(1.5).fs0.h2.$}>
```
bdow: **question:** I don't really follow the comment. Is it referencing the `h2` here? I'm wondering if we really need that height here?

src/components/RowBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+  description: ReactNode;
+  /** Rendered in order, always as text buttons. */
+  actions?: ActionButtonProps[];
+};
+
+/**
+ * An error/warning notice for a `GridTable` row's `companion` slot.
+ *
+ * Any bgColor behind the banner belongs to the row, not to the banner.
+ */
```
bdow: This may or may not be used in GridTable row's companion slot. Comment might confuse the AI if we want to use this outside of the table

src/components/RowBanner.tsx
apattersonATX-HB: Updated

src/components/RowBanner.tsx
apattersonATX-HB: Ya i just saw that thread from Figma. Updated comments

src/components/RowBanner.tsx
apattersonATX-HB: Was supposed to be a `Tag` 🤦🏽‍♂️

src/components/InlineFeedbackBanner.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <span css={Css.fg1.mw0.color(Tokens.OnSurface).$} {...tid.description}>
+        {description}
+      </span>
+      {actions.length > 0 && (
+        // Held to the tag's height so a text button's taller line box can't grow the banner.
+        <div css={Css.df.aic.gap(1.5).fs0.hPx(18).$}>
+          {actions.map((action) => (
+            <Button key={`${action.label}`} {...action} variant="text" />
+          ))}
+        </div>
```
bdow: **thought:** I still don't get the need for this `hPx`, I'm guessing the Button is taller than 18px? Maybe it's just the line-height do that? Only concern I'd have with setting a fixed height is that we won't allow this to wrap on mobile screens. Fine for now until we run into the issue, I think. We might want a different sort of mobile behavior anyways.

bdow (approved):

src/components/InlineFeedbackBanner.tsx
apattersonATX-HB: 😂 I just forgot to commit this change. My bad

PR #1399, opened 2026-08-25
feat: aiMode Tabs; AiTagIcon;

apattersonATX-HB:


apattersonATX-HB:
https://linear.app/homebound-team/issue/DS-323/tab-decoration

src/components/AiTagIcon.tsx
bdow, line 1: **nit:** Design created an "ai" variant of our Tags: https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=181-1669&t=zWGguKpZ9qn28xUn-4
Naming this "AiTagIcon" might conflict with the `Tag type="ai"` in human brains. 
Maybe "AiBadge"?

src/components/AiTagIcon.tsx
```diff
@@ -0,0 +1,32 @@
+import { Icon } from "src/components/Icon";
+import { Css, increment, Margin, Only, Tokens, Xss } from "src/Css";
+import { useTestIds } from "src/utils";
+
+export type AiTagIconXss = Margin | "flexShrink";
```
bdow: **thought/question:** I would think we'd always want a `flex-shrink: 0` here. Why would we ever change it?

bdow: Might not even need the `Margin` or the XSS property at all yet.

src/components/Tabs.tsx
```diff
@@ -18,6 +19,8 @@ export type Tab<V extends string = string> = {
   icon?: IconKey;
   // Suffixes label with specified node. Expected to be used for cases where the decoration is not just an icon.
   endAdornment?: ReactNode;
+  /** Whether the tab's content has AI proposals waiting on it. Takes precedence over `icon` and `endAdornment`. */
+  hasAiProposals?: boolean;
```
bdow: **nit:** Not sure if all AI stuff will always be proposals... Might want to just call this `aiMode` as well - wdyt? Keep consistency.

bdow (approved):

src/components/AiTagIcon.tsx
apattersonATX-HB: Ya thats way better

src/components/AiTagIcon.tsx
apattersonATX-HB: removed 🫡

src/components/Tabs.tsx
apattersonATX-HB: Ya good point. I was thinking too deep about wanting `aiMode` to mean more than a coat of paint but we've already been doing it that way 🙃

PR #1400, opened 2026-08-25
fix: make RightSidebar fill the remaining viewport height

Jacob-Diebold:
# fix: make RightSidebar fill the remaining viewport height

## Description

- An open `RightSidebar` now stretches to the full remaining viewport (`100vh` minus the page header) instead of shrinking to its content.
- Short panes keep a full-height column so the close control, icon strip, and divider line stay aligned with the page instead of collapsing.

### Previous
Screenshot 2026-08-25 at 9 37 55 AM
Screenshot 2026-08-25 at 9 38 04 AM

### Updated
Screenshot 2026-08-25 at 9 38 04 AM
Screenshot 2026-08-25 at 9 38 09 AM

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

Mental Checklist

- [ ] Add your squad to the reviewers list.
- [ ] Updated tests for this change.
- [x] Tested changes locally.
- [ ] Updated Documentation (if necessary).

### DIFF
--- a/src/components/RightSidebar.tsx
+++ b/src/components/RightSidebar.tsx
@@ -49,7 +49,12 @@ export function RightSidebar({ content, headerHeightPx }: RightSidebarProps) {
             animate={{ x: 0, opacity: 1 }}
             transition={{ delay: 0.2, ease: [0.51, 0.92, 0.24, 1], duration: 0.3 }}
             exit={{ transition: { ease: "linear", duration: 0.2 }, x: "100%" }}
-            css={Css.w100.mw(RIGHT_SIDEBAR_MIN_WIDTH).z0.maxh(`calc(100vh - ${headerHeightPx}px)`).oya.pl4.pr3.$}
+            css={
+              Css.w100
+                .mw(RIGHT_SIDEBAR_MIN_WIDTH)
+                .z0.mh(`calc(100vh - ${headerHeightPx}px)`)
+                .maxh(`calc(100vh - ${headerHeightPx}px)`).oya.pl4.pr3.$
+            }
           >
             <>
               {/* Sticky header section */}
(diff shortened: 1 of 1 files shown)


src/components/RightSidebar.tsx
```diff
@@ -49,7 +49,12 @@ export function RightSidebar({ content, headerHeightPx }: RightSidebarProps) {
             animate={{ x: 0, opacity: 1 }}
             transition={{ delay: 0.2, ease: [0.51, 0.92, 0.24, 1], duration: 0.3 }}
             exit={{ transition: { ease: "linear", duration: 0.2 }, x: "100%" }}
-            css={Css.w100.mw(RIGHT_SIDEBAR_MIN_WIDTH).z0.maxh(`calc(100vh - ${headerHeightPx}px)`).oya.pl4.pr3.$}
+            css={
+              Css.w100
+                .mw(RIGHT_SIDEBAR_MIN_WIDTH)
+                .z0.mh(`calc(100vh - ${headerHeightPx}px)`)
+                .maxh(`calc(100vh - ${headerHeightPx}px)`).oya.pl4.pr3.$
```
bdow: appears `min-height` and `max-height` are the same thing. Can we just do `height` instead?

Jacob-Diebold: Yes, I was literally in the process of changing that when you commented this lol

bdow:
I'm curious on how we plan to use this. I think this is part of some old layouts, like `FormLayout`, which isn't used the same way as our new component of `FocusedFormLayout`. There might be more work to integrate a right pane to the FocusedFormLayout. 
Also happy to _not_ use FocusedFormLayout yet and instead just make a "// TODO: Update to FocusedFormLayout"

bdow (approved):

Jacob-Diebold:
> I'm curious on how we plan to use this. I think this is part of some old layouts, like `FormLayout`, which isn't used the same way as our new component of `FocusedFormLayout`. There might be more work to integrate a right pane to the FocusedFormLayout. Also happy to _not_ use FocusedFormLayout yet and instead just make a "// TODO: Update to FocusedFormLayout"

I'm using it in a FormPageLayout for a history pane of changes to mvs. I noticed the divider was only as high as the content inside of it; but had a lot of scrollable distance (you can also see that in the first storybook screenshot) and it hadn't looked intended to be that way/ wasn't great UX. This should look and work better for all the implementations we have for it.

bdow:
> I'm using it in a FormPageLayout for a history pane of changes to mvs

is this an existing page? Let's chat on Slack. Could potentially be more to do here if we want to use the new layouts. Happy to push this PR through, though.

PR #1405, opened 2026-08-25
feat: LinkCard; AiLinkCardGroup

apattersonATX-HB:
LinkCard — a raised card with a title, optional message, and a trailing arrow.

AiLinkCardGroup — cards: LinkCardProps[] rendered into its own AiPanel + BPAiLogo


src/components/AiLinkCardGroup.tsx
```diff
@@ -0,0 +1,30 @@
...
+export type AiLinkCardGroupProps = {
+  cards: LinkCardProps[];
+};
+
+/**
+ * A stack of {@link LinkCard}s on the AI surface, each linking to the finding it describes.
+ *
+ * Renders nothing when `cards` is empty.
+ */
+export function AiLinkCardGroup(props: AiLinkCardGroupProps) {
```
stephenh, line 16: @apattersonATX-HB / @bdow normally I'm "pages should have zero layout!!!" but I almost wonder if this `AiLinkCardGroup` really needs to be a component? Np, just a musing.

stephenh (approved):

src/components/LinkCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <span css={Css.smSb.color(Tokens.OnSurface).$} {...tid.title}>
+          {title}
+        </span>
+        {message && (
+          <span css={Css.xs.color(Tokens.OnSurface).$} {...tid.message}>
+            {message}
+          </span>
+        )}
+      </div>
+      <div css={Css.df.aic.fs0.hPx(40).$}>
```
bdow: **question:** Why a height of 40px?

src/components/AiLinkCardGroup.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { AiLinkCardGroup } from "src/components/AiLinkCardGroup";
+import { LinkCardProps } from "src/components/LinkCard";
+import { Css } from "src/Css";
+import { viewportModes, withRouter } from "src/utils/sb";
+
+export default {
+  component: AiLinkCardGroup,
+  decorators: [withRouter()],
+  parameters: {
+    chromatic: { modes: viewportModes("desktop", "mobile1") },
```
bdow: **question:** Do you think this one is worth doing a mobile snapshot of? I didn't notice any mobile specific styling.

src/components/AiLinkCardGroup.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    action: { onClick: "/plans/1/elevations" },
+  },
+  {
+    title: "Review location changes: 1 added, 1 renumbered, 1 removed.",
+    message: summary,
+    action: { onClick: "/plans/1/locations" },
+  },
+  { title: "Review 2 new options, 1 changed, 1 removed.", message: summary, action: { onClick: "/plans/1/options" } },
+  { title: "Review 1 changed program value.", message: summary, action: { onClick: "/plans/1/program" } },
+];
```
bdow: **nit/ask:** can we move these into the Default story? Is there a reason for them to be module level. 

In terms of set up, would prefer to have helper constants and methods at the bottom of the file, if reused in many stories. So when coming in here, a user finds the stories they're looking for at the top of the file. This is fairly minor here, but would like to have the same structure applied everywhere.

src/components/AiLinkCardGroup.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const cards = [
+  {
+    title: "Review 1 changed elevation.",
+    message: "Describe change",
+    action: { onClick: "/plans/1/elevations" },
+  },
+  { title: "Review 1 changed program value.", action: { onClick: "/plans/1/program" } },
+];
+
```
bdow: **nit/ask:** Same structure thing here - could we move the `cards` to the bottom of the file?
fwiw, when jest or storybook load the module, they'll eagerly evaluate these module level bits of code. Not a huge concern here, but if we used factories like in `internal-frontend`, which can build a lot, then its better to wrap those sort of fixtures within a function like `function getCards() {...}`.

src/components/AiLinkCardGroup.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const { cards } = props;
+  const tid = useTestIds(props, "aiLinkCardGroup");
+  if (cards.length === 0) return null;
+  return (
+    <AiPanel padding="lg" {...tid}>
+      <div css={Css.df.fdc.aifs.gap2.w100.$}>
+        <BlueprintAiLogo height={3} />
+        {cards.map((card) => (
+          <LinkCard key={`cardGroup_${card.title}_card`} {...card} />
```
bdow: **nit:** probably could've just used the `card.title` here.

src/components/LinkCard.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+const longMessage = (
+  <>
+    According to the latest document upload <a href="/documents/1">Document Name Here</a> uploaded by Drew Whiting on
+    July 15th 2026, here is a summary of changes to option configuration. Locations that moved between elevations keep
+    their original numbering, so the take-off quantities you already approved are unaffected, but any option attached to
+    a removed location needs to be reassigned before this plan can be released to the field. Three of the renumbered
+    locations also appear in the structural package, which has not yet been re-uploaded, so those may change again.
+  </>
+);
+
```
bdow: **nit/ask:** Same around structure again

bdow (approved):
a few small requests, otherwise good to go.

src/components/LinkCard.tsx
apattersonATX-HB: Keeps it centered while it wraps

src/components/AiLinkCardGroup.stories.tsx
apattersonATX-HB: I guess not, this was mostly habit

src/components/AiLinkCardGroup.stories.tsx
apattersonATX-HB: 🫡 fixed

src/components/AiLinkCardGroup.tsx
apattersonATX-HB: We like how plug n play it is as it's own component. The inner LinkCardGroup might get used outside of an Ai flow on its own but this gives it a singular purpose that's easy for us to use.

stephenh: I'm generally supportive, but I'm also just slightly incredulous about the amount of net-new comments that designers are putting through Figmas + pulling in Brandon to help implement.

Like I get we're revamping BP, I technically like everything going by, I'm just hoping we eventually turn the corner on "another figma, another 2-3 new components b/c :shrug: why not".

PR #1411, opened 2026-08-27
fix: use `Tag` in Tabs aiMode instead of aiBadge;

apattersonATX-HB:
Design requested that we use an ai mode Tag instead of the `aiBadge`. I haven't removed that component yet but it is going unused atm.
Screenshot 2026-08-27 at 12 57 38 PM

### DIFF
--- a/src/components/Tabs.tsx
+++ b/src/components/Tabs.tsx
@@ -253,7 +253,15 @@ function TabImpl<V extends string>(props: TabImplProps<V>) {
   });
 
   const decoration = aiMode ? (
-    <Tag text="" preventTooltip icon="aiStar" type="ai" variant="primary" iconOnly {...tid.aiTag} />
+    <Tag
+      text="Tab has AI proposals ready for review"
+      preventTooltip
+      icon="aiStar"
+      type="ai"
+      variant="primary"
+      iconOnly
+      {...tid.aiTag}
+    />
   ) : icon ? (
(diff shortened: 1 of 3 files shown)


stephenh (approved):

src/components/Tabs.tsx
```diff
@@ -252,7 +252,13 @@ function TabImpl<V extends string>(props: TabImplProps<V>) {
     ...(isRouteTab(tab) ? {} : { onClick: () => onClick(tab.value) }),
   });
 
-  const decoration = aiMode ? <AiBadge /> : icon ? <Icon icon={icon} /> : endAdornment;
+  const decoration = aiMode ? (
+    <Tag text="" preventTooltip icon="aiStar" type="ai" variant="primary" iconOnly {...tid.aiTag} />
```
stephenh: I already approved this, but seems kinda odd to use a text-less Tag -- I would not have anticipated a "text-less Tag" was a kosher usage of `Tag` -- @bdow wdyt?

bdow: Should still have text/label defined. IconOnly is what hides it.  Something for accessibility, like "AI proposals ready for review"

apattersonATX-HB: Thought it was weird label was still required with `iconOnly`. 

I added a label and added a jsdoc comment to `Tag.text` on why its required 🫡

PR #1412, opened 2026-08-27
fix: fix inlineBanner text/tag alignment

apattersonATX-HB:
Screenshot 2026-08-27 at 4 19 32 PM


.claude/launch.json
bdow, line 1: does this allow claude to test in its own browser?

apattersonATX-HB: Basically yeah. Anytime it thinks "i should check this visually" it'll come to this to launch storybook.

I can remove it if we want, i didn't mean to commit this 🙏🏽

src/components/InlineFeedbackBanner.tsx
```diff
@@ -34,11 +34,12 @@ export function InlineFeedbackBanner(props: InlineFeedbackBannerProps) {
...
     <div css={Css.df.ais.gap1.w100.p1.br4.xs.bgColor(Tokens.Surface).ba.bc(borderColor).bshBasic.$} {...tid}>
-      {/* `df` so Tag's inline-flex box doesn't pick up a line box's descender space. */}
-      <span css={Css.df.fs0.$}>
+      {/* The 2px nudge lets a single line read centered against the tag while wrapped copy stays top-aligned;
+          the tag's box is that nudge plus the description's 16px line height, so it centers on the first line. */}
+      <span css={Css.df.aic.fs0.ptPx(2).hPx(18).$}>
         <Tag {...tagProps} {...tid.tag} />
       </span>
-      <span css={Css.fg1.mw0.color(Tokens.OnSurface).$} {...tid.description}>
+      <span css={Css.fg1.mw0.ptPx(2).color(Tokens.OnSurface).$} {...tid.description}>
```
bdow: did we mean to also add this padding to the description? Seems like from the comment that the padding was meant to just be on the Tag.

src/components/InlineFeedbackBanner.tsx
```diff
@@ -34,11 +34,12 @@ export function InlineFeedbackBanner(props: InlineFeedbackBannerProps) {
 
   return (
     <div css={Css.df.ais.gap1.w100.p1.br4.xs.bgColor(Tokens.Surface).ba.bc(borderColor).bshBasic.$} {...tid}>
-      {/* `df` so Tag's inline-flex box doesn't pick up a line box's descender space. */}
-      <span css={Css.df.fs0.$}>
+      {/* The 2px nudge lets a single line read centered against the tag while wrapped copy stays top-aligned;
+          the tag's box is that nudge plus the description's 16px line height, so it centers on the first line. */}
+      <span css={Css.df.aic.fs0.ptPx(2).hPx(18).$}>
```
bdow: I think there's a bit of extra CSS in here that isn't needed. 
Should only need what was there before, or similar. 

✅  The `display: flex` makes it block level, so the span doesn't end up being the height of whatever `line-height` its inheriting as it would if it were an `inline` element. 

✅  The `fs0` ensures it doesn't shrink and get cut off/show an ellipsis.

❌ The `aic` is unnecessary, as there only one item within this flex container.
❌ The `hPx(18)` is unnecessary as the flex container will just inherit the height of the element within.
❌ The `ptPx(2)` seems unnecessary too, as we really are just trying to line up the `description` with the Tag's text.

src/components/InlineFeedbackBanner.tsx
bdow: **thought:** Another way to have achieved this not using padding would've been to have the line-height of this element be the same height of the Tag, `line-height: 18px;`. That would align the text with the tag. But the 2px of padding top works too.

bdow (approved):
Fine moving this forward, but I do think we can reduce the amount of CSS needed.

PR #1414, opened 2026-08-28
feat: Support turning on on table rows to turn purple

bdow:


src/components/Table/components/Row.tsx
```diff
@@ -565,6 +571,8 @@ export type GridDataRow<R extends Kinded> = {
   draggable?: boolean;
   /** Image src for the row, to be used for card view */
   imgSrc?: string;
+  /** Paints the row with `Tokens.AiFieldBg`. Re-read on every `rows` update (unlike `initSelected`). */
+  aiMode?: boolean;
```
stephenh: @bdow maybe this could be a `MaybeFn<boolean>` so that if a caller did:

```
  aiMode: () => ...access some mobx things...,
```

Then `aiMode` turns into its own observable basically.

src/components/Table/utils/RowState.ts
```diff
@@ -103,6 +105,9 @@ export class RowState<R extends Kinded> {
     // then anyone watching `.row` will see the new row instance + new data.
     this._row = row;
     this._data = row.data;
+    // Primitive so unchanged `aiMode` across 500-row `setRows` does not notify.
```
stephenh: I don't really follow this comment, can you reword?

stephenh (approved):

src/components/Table/components/Row.tsx
bdow: Makes sense, will do

src/components/Table/utils/RowState.ts
bdow: Yup, updated

PR #1415, opened 2026-08-28
feat: table cards to support ai styling

apattersonATX-HB:
### [DS-253](https://linear.app/homebound-team/issue/DS-253/table-cards-to-support-ai-styling) — [Figma](https://www.figma.com/design/DchiwVkssXeYi2Er8sMU2k/H2-2026-Plans---Automated-Construction-Doc-Capture?node-id=1602-57854&m=dev)

Adds AI styling to `GridTable as="card"`

- `aiMode: "updated"` — purple border + background, default text
- `aiMode: "new"` — same, plus purple text throughout

`aiMode` sits on `GridDataRow` next to `imgSrc` , so `GridTable.tsx` is untouched.

`ProposedValue` moves out of `src/inputs/internal/` and becomes public: cards need it for showing `~original~ proposed` data and our row tables will probably want it too.


bdow:
>aiMode: "updated" — purple border + background, default text
aiMode: "new" — same, plus purple text throughout

When we we have "updated", then some of the text still needs to be purple depending on what was updated. At that point we can probably keep these two things separated. The card can just know "did I have anything proposed" and the content of the card will handle all the changing of purple colors. So I'm thinking we just need this to be a boolean flag still to provide the generic card styles, and let the content dictate what is purple or not.

bdow:
Also, I just pushed up a change to support `aiMode` on Rows, which used just a boolean flag. So I think maybe we can reuse that for properly styling these cards as well.

src/components/Table/components/TableCard.tsx
```diff
@@ -43,8 +44,9 @@ function TableCardImpl<R extends Kinded>(props: TableCardProps<R>) {
   const tid = useTestIds(props, "tableCard");
 
   let title: string | undefined;
-  let leftEyebrow: string | undefined;
-  let rightEyebrow: string | undefined;
+  let titleProposed: { original?: string } | undefined;
+  let leftEyebrow: ReactNode;
+  let rightEyebrow: ReactNode;
```
apattersonATX-HB: @bdow I was having a hard time coming up with how to handle proposed values for these slots

bdow: Let's make a simple ProposedValue component that handles the strike through of an optional "originalValue" prop, purple text color, and font weight. Then people can just pass that in as the ReactNode

bdow: Another option would be to not do a ReactNode, and do a sort of: `leftEyebrow: string | AiFieldProposal`. Could match/`satisfies` the gql object, which would be a nice DX.

apattersonATX-HB: went with `string | ProposedValueProps`

bdow:
Screenshot 2026-09-01 at 12 18 09 PM

I believe the AI styling when not changing an original value - only has a proposed value - shouldn't be the bold font weight. Can we fix that?
I also think they'd want the `label` to show purple per the designs (e.g. "Sqft", "Height", etc)

src/components/Table/components/TableCard.test.tsx
```diff
@@ -28,6 +28,49 @@ describe("TableCardView", () => {
     expect(r.tableCardView).toHaveStyle({ backgroundColor: maybeCssVar(Tokens.AiFieldBg) });
   });
 
+  it("draws the AI border and tints the progress track when aiMode is true", async () => {
```
bdow, line 31: **thought (non-blocking):** This isn't really worth a jest test imo, could be / is being validated just via chromatic.

src/components/Table/components/TableCard.tsx
```diff
@@ -183,10 +187,15 @@ export function TableCardView(props: TableCardViewProps) {
   const content = (
     <>
       <div
-        css={Css.relative.hPx(184).w100.bb.bc(Tokens.FieldBorderDefault).oh.borderRadius("12px 12px 0 0").$}
+        css={
+          Css.relative
+            .hPx(184)
```
bdow, line 192: **thought (non-blocking):** For some reason, I thought we made this height dynamic, so a caller could change it if needed. Maybe that never made it through.

src/components/Table/components/TableCard.tsx
```diff
@@ -225,28 +234,36 @@ export function TableCardView(props: TableCardViewProps) {
...
-              <div css={Css.df.fdc.fg1.$}>
+              <div css={Css.df.fdc.fg1.fb2.mw0.$}>
                 {col2.map((d) => (
                   <div key={d.label} css={Css.df.gapPx(4).$} {...tid[defaultTestId(d.label)]}>
-                    <dt>{d.label}:</dt>
-                    <dd>{d.value}</dd>
+                    <dt css={Css.fs0.$}>{d.label}:</dt>
+                    {/* `mw0` lets a long value — e.g. a `ProposedValue` carrying both halves — wrap
+                        instead of widening the column. */}
+                    <dd css={Css.mw0.$}>{d.value}</dd>
```
bdow, line 253: **question:** Do we really need `mw0` on both this wrapping `div` and on the `dd` element? Thought maybe we'd only need it in one place, but fine if needed in both.

bdow (approved):
All looks good from a tech side to me. Would like to just get some styles updated, which would probably get called out as part of AT as well. Approving for now, and happy to ship and follow up, or do as part of this PR.

PR #1418, opened 2026-09-01
feat: Abstract JumpLink for usage on any workflows

bdow:
JumpLinks were only part of the FocusedFormLayout. New design direction has JumpLinks as part of StepperLayout body as well. JumpLinks now can be integrated as part of FormSectionLayout. This also updates the FocusedFormLayout to not be tied to only the FormSectionLayout, and instead just expects a user to use the FormSectionLayout as the children element for FocusedFormLayout.


src/layouts/FormSectionLayout/FormSectionLayout.tsx
```diff
@@ -63,4 +98,16 @@ export function FormSectionLayout(props: FormSectionLayoutProps) {
...
     </CenteredLayout>
   );
+
+  if (!showRail) return form;
+
+  return (
+    <div css={Css.df.w100.$}>
+      <JumpLinksRail links={jumpLinks} activeId={activeId} {...tid.jumpLinks} />
+      {/* Mirror the rail's width so the form stays centered on the page, as it is without the rail. */}
+      <div css={Css.fg1.mw0.mb4.mr(jumpLinksRailReservation).$} {...tid.column}>
```
apattersonATX-HB: question:

`mb4` only gets added in rail mode, so bottom spacing now depends on whether the rail renders or not.

Before, `FocusedFormLayout` wrapped the body in `Css.fg1.mw0.mb4`, and `WorkflowPageLayout`'s body is `pt4 / ifMdAndUp pt6` with no bottom padding — so a FocusedForm without `withJumpLinks` (and every FocusedForm on `sm`, where the rail never shows) loses mb4 and ends flush at the page bottom. It also makes a Stepper step's bottom gutter differ based on the rail.

Should we move that mb4 up? I didn't see any big diffs in the UI diffs so maybe I'm not following this correctly

apattersonATX-HB (approved):

src/layouts/FormSectionLayout/FormSectionLayout.tsx
bdow: Good call, will update.

PR #1419, opened 2026-09-01
feat: rejected Tag type

apattersonATX-HB:
Add a new `rejected` tag type that will dim the tag and strikethrough the text.

Works with both `TagVariant`s

### DIFF
--- a/src/components/Tag.stories.tsx
+++ b/src/components/Tag.stories.tsx
@@ -83,14 +83,14 @@ export function Secondary() {
   );
 }
 
-export function Rejected() {
+export function Strikethrough() {
   return (
     <div css={Css.df.fdc.aifs.gap1.$}>
-      <h3 css={Css.mb1.mdSb.$}>Rejected</h3>
+      <h3 css={Css.mb1.mdSb.$}>Strikethrough</h3>
       <p>Indicates a value that has been declined, i.e. a rejected AI suggestion.</p>
-      <Tag text="TRMCLG0004" type="rejected" variant="secondary" />
-      <Tag text="TRMCLG0004" type="rejected" variant="secondary" icon="xCircle" />
-      <Tag text="Rejected Label" type="rejected" />
+      <Tag text="TRMCLG0004" type="strikethrough" variant="secondary" />
+      <Tag text="TRMCLG0004" type="strikethrough" variant="secondary" icon="xCircle" />
+      <Tag text="Rejected Label" type="strikethrough" />
     </div>
   );
 }
--- a/src/components/Tag.stories.tsx
+++ b/src/components/Tag.stories.tsx
@@ -82,3 +82,15 @@ export function Secondary() {
     </div>
   );
 }
+
+export function Rejected() {
+  return (
+    <div css={Css.df.fdc.aifs.gap1.$}>
+      <h3 css={Css.mb1.mdSb.$}>Rejected</h3>
+      <p>Indicates a value that has been declined, i.e. a rejected AI suggestion.</p>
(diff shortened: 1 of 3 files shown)


src/components/Tag.stories.tsx
```diff
@@ -82,3 +82,15 @@ export function Secondary() {
     </div>
   );
 }
+
+export function Rejected() {
+  return (
+    <div css={Css.df.fdc.aifs.gap1.$}>
+      <h3 css={Css.mb1.mdSb.$}>Rejected</h3>
+      <p>Indicates a value that has been declined, i.e. a rejected AI suggestion.</p>
```
stephenh, line 90: @apattersonATX-HB just curious, this is "rejecting an AI-suggested removal" 

Image

Do I read that as:

* the grey text -> means the AI wanted to remove this
* the strikethrough -> means I said no, keep it

And in general does this pattern of "the color of the text denotes 'what the AI wanted to do'" and the strikethrough is "the user saying accept/reject"?

Kinda wonder @bdow if we could have that articulated / written down somewhere -- maybe in an AI documentation storybook here in beam?

src/components/Tag.tsx
```diff
@@ -5,13 +5,13 @@ import { Css, Margin, Only, Palette, Properties, Xss } from "src/Css";
 import { useTestIds } from "src/utils";
 
 export type TagXss = Margin | "backgroundColor" | "color";
-export type TagType = "info" | "update" | "warning" | "error" | "success" | "neutral" | "ai";
+export type TagType = "info" | "update" | "warning" | "error" | "success" | "neutral" | "ai" | "rejected";
```
stephenh: ...we're adding `rejected`, should there be an `accepted`?

What's the difference between `type=ai` and `type=rejected`? Could we document that different with an in-source comment?

apattersonATX-HB: Nope, that'd just be a normal tag.

ai is just purple and rejected is strikethrough with a fade. I added a comment

src/components/Tag.stories.tsx
apattersonATX-HB: No its really just styling. "rejected" could also be called "strike" or whatever you prefer (names are hard!).

The reason I'm heading this direction is because it's going to be on the caller/design to style use these two TagTypes correctly.

In our table use case ([conflicts](<https://www.figma.com/design/DchiwVkssXeYi2Er8sMU2k/H2-2026-Plans---Automated-Construction-Doc-Capture?node-id=1499-48636&m=dev>)) we have a possible instance of Ai "rejecting" multiple values and then proposing one new value. Or maybe rejecting one value, proposing a new one, then leaving one alone. That means we cant do anything cute like turning text to `ReactNode | ProposedValueProps` and rendering these values as a pair.

So to answer your question, no. Gray AND strikethrough mean Ai rejected but Beam.Tag doesn't know anything about that.

stephenh: I guess if "Tag doesn't know about (AI rejected)" is the current assertion, which I'm fine with, I think the `type=strikethrough` is a better name then.

I think in general I'm skeptical about using tags as a way to "diff a list", but I can't quite articulate why atm. 🤔 🤷

stephenh (approved):

src/components/Tag.stories.tsx
apattersonATX-HB: I'll rename 👍🏽 

Im guessing theres a design somewhere that's converted these from chips to tags but I havent seen it.

Screenshot 2026-09-01 at 4 45 10 PM

PR #1421, opened 2026-09-02
feat: base card component

apattersonATX-HB:
## Summary
Extracts a shared `BaseCard` foundation — border/radius/background, a 184px hero image, hover/focus states, and optional status-tag / hero-action / footer slots — and refactors `TableCardView` to consume it.

One card to rule them all 🧙🏽‍♂️


src/components/BaseCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { useFocusRing } from "react-aria";
+import { ButtonMenu, ButtonMenuProps } from "src/components/ButtonMenu";
+import { CardTag, ImageFitType } from "src/components/Card";
+import { IconButton, IconButtonProps } from "src/components/IconButton";
+import { Tag } from "src/components/Tag";
+import { Css, Tokens } from "src/Css";
+import { useTestIds } from "src/utils";
+import { getButtonOrLink } from "src/utils/getInteractiveElement";
+
+export type BaseCardProps = {
```
apattersonATX-HB, line 13: @bdow I intentionally left off `title | eyebrows` but lmk if you think we should add them here

src/components/Table/components/TableCard.tsx
```diff
@@ -163,141 +162,104 @@ export function TableCardView(props: TableCardViewProps) {
...
-            <Tag {...status} />
+  return (
+    <BaseCard
+      {...tid}
+      imgSrc={imgSrc}
+      imgAlt={titleText}
+      imageFit={imageFit}
+      tag={status}
+      to={to}
+      onClick={onClick}
```
bdow: **thought:** I'm kind of surprised that `to` and `onClick` are separate props. I would expect that a user cannot do both - you're either going to have it perform an inpage action or navigate elsewhere. Typically these are a single `onClick` prop. Maybe we can keep the `action = to || onClick` that we had above. I'm guessing we have both because of some old GridTable reason... but not sure.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { render, withRouter } from "src/utils/rtl";
+
+describe("BaseCard", () => {
+  it("treats the hero as decorative by default", async () => {
+    // Given a card with no imgAlt
+    // When rendered
+    const r = await render(<BaseCard imgSrc="home.jpg">body</BaseCard>);
+    // Then the hero is decorative, since the body names the card
+    expect(r.baseCard_image).toHaveAttribute("src", "home.jpg");
+    expect(r.baseCard_image).toHaveAttribute("alt", "");
```
bdow: **suggestion:** If we're considering the image to just be decorative, then I'd suggest using the `aria-hidden="true"` attribute on it.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+  });
+
+  it("uses imgAlt when the hero carries meaning", async () => {
+    // Given a card with an explicit alt
+    // When rendered
+    const r = await render(<BaseCard imgSrc="home.jpg" imgAlt="123 Main St" children="body" />);
+    // Then it wins over the decorative default
+    expect(r.baseCard_image).toHaveAttribute("alt", "123 Main St");
+  });
```
bdow: **question:** Are there existing cases where the we would and wouldn't provide `alt` text? Seems like we probably should always have alt text. If you said the image is just decorative, that's fine. But it caught me off guard to see it framed as "sometimes we care about it, and sometimes we don't." I'm looking to understand when those cases happen.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+    expect(r.baseCard_image).toHaveStyle({ objectFit: "cover" });
+    // When contain is requested
+    r.rerender(
+      <BaseCard imgSrc="home.jpg" imageFit="contain">
+        body
+      </BaseCard>,
+    );
+    // Then the full image is shown
+    expect(r.baseCard_image).toHaveStyle({ objectFit: "contain" });
+  });
```
bdow, line 21: **question:** I wonder if we could get away with just a storybook example here? And maybe that's a better test for what we're doing anyways for this.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+    expect(r.baseCard).not.toHaveStyle({ cursor: "pointer" });
+  });
+
+  it("only invites a click when the card is actually interactive", async () => {
+    // Given a card that navigates
+    // When rendered
+    const r = await render(<BaseCard imgSrc="home.jpg" to="/homes/1" children="body" />, withRouter());
+    // Then it shows a pointer cursor
+    expect(r.baseCard).toHaveStyle({ cursor: "pointer" });
+  });
```
bdow: **suggestion:** I probably wouldn't have written this test. our CSS Reset sets this `cursor: pointer` on buttons, and by default it happens on `a` tags. This component isn't really in charge of that style, and so testing it here seems unnecessary.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+    expect(r.baseCard_action).toHaveAttribute("type", "button");
+  });
+
+  it("labels the action from imgAlt, without a separate label prop", async () => {
+    // Given an interactive card with a meaningful imgAlt
+    // When rendered
+    const r = await render(<BaseCard imgSrc="home.jpg" onClick={() => {}} imgAlt="123 Main St" children="body" />);
+    // Then that text becomes the action's accessible name
+    expect(r.baseCard_action).toHaveAttribute("aria-label", "123 Main St");
+  });
```
bdow: **thought:** I don't think that we need to describe a aria-label on these buttons. The button wraps the whole card. So anything in the card is representative of that button. It doesn't hurt to distill it down. But since imgAlt is optional, it feels like we're trying to do too much here, imo.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  it("leaves the action unlabeled when the hero is decorative", async () => {
+    // Given an interactive card with no imgAlt, i.e. a self-describing body
+    // When rendered
+    const r = await render(<BaseCard imgSrc="home.jpg" onClick={() => {}} children="123 Main St" />);
+    // Then no aria-label is set — the browser derives the accessible name from the button's own text
+    expect(r.baseCard_action).not.toHaveAttribute("aria-label");
```
bdow: **suggestion:** I do think the `imgAlt` should be required. Feels a little odd to not include it, imo.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+  });
+
+  it("renders the tag inside the hero so it rides along with the link", async () => {
+    // Given a linked card with a status tag
+    // When rendered
+    const r = await render(
+      <BaseCard imgSrc="home.jpg" to="/homes/1" tag={{ text: "Active", type: "success" }} children="body" />,
+      withRouter(),
+    );
+    // Then the tag sits in the hero, inside the anchor — a Tag isn't focusable, so that nesting is legal
```
bdow: **nit:** this comment about what nesting is legal reads very much like the AI providing itself some comments.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+  it("renders a link for to", async () => {
+    // Given a card that navigates
+    // When rendered
+    const r = await render(<BaseCard imgSrc="home.jpg" to="/homes/1" children="body" />, withRouter());
+    // Then its body is an anchor
+    expect(r.baseCard_action.tagName).toBe("A");
+    expect(r.baseCard_action).toHaveAttribute("href", "/homes/1");
+  });
+
+  it("renders a non-submitting button for onClick", async () => {
```
bdow: **suggestion:** I also am not sure if this test is worth it as is. Probably should do a `click()` on the element and validate the onClick was called. I guess validating that its a button and of type button is fine, but what we really want to ensure we got right is also that `onClick` works. maybe that's somewhere else down in these tests?

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+    expect(r.baseCard_action).not.toHaveAttribute("aria-label");
+  });
+
+  it("keeps action and footer outside the card's link", async () => {
+    // Given a linked card with its own hero button and an interactive footer
+    // When rendered
+    const r = await render(
+      <BaseCard
+        imgSrc="home.jpg"
+        to="/homes/1"
```
bdow: **nit:** I think BaseCard should just accept a `onClick` the same that Button does, and can render the correct element that way.

src/components/BaseCard.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  it("paints the AI background and border when aiMode is true", async () => {
+    // Given a card in aiMode
+    // When rendered
+    const r = await render(<BaseCard imgSrc="home.jpg" aiMode children="body" />);
+    // Then it takes the AI fill and the heavier AI border. Tokenized *border colors* can't be asserted
+    // via toHaveStyle — jsdom resolves an unresolved var() to black — so the width stands in.
+    expect(r.baseCard).toHaveStyle({ backgroundColor: maybeCssVar(Tokens.AiFieldBg) });
+    expect(r.baseCard).toHaveStyle({ borderWidth: "2px" });
+  });
```
bdow: **nit:** This could just be a chromatic test

src/components/BaseCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+  action?: IconButtonProps | ButtonMenuProps;
+  /** Card body. Unpadded to allow callers to full-bleed sections; the standard inset is usually `p3`. */
+  children: ReactNode;
+  /** Content with its own interactive controls, rendered outside the card's link/button area. */
+  footer?: ReactNode;
+  /** Makes the body a link. Takes precedence over `onClick`. */
+  to?: string;
+  /** Makes the body a button, for cards that act on click instead of navigating. */
+  onClick?: () => void;
+  /** Fixed card height in px. Defaults to sizing to content. */
```
bdow: **suggestion:** Mentioned this in another comment, but I would move towards making these a single `onClick` property - same that we do for `Button`. Then determine the correct element using the `getButtonOrLink` helper. I think that should work for us here.

src/components/BaseCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { useTestIds } from "src/utils";
+import { getButtonOrLink } from "src/utils/getInteractiveElement";
+
+export type BaseCardProps = {
+  /** Hero image url.  */
+  imgSrc: string;
+  /** Doubles as the link/button's accessible name when set.*/
+  imgAlt?: string;
+  /** Defaults to `"cover"`. */
+  imageFit?: ImageFitType;
```
bdow, line 17: **thought:** Right now, all of our cards have a hero. Thinking for that maybe some point in the future we might not have a hero image 🤷. Not something to worry about yet, but maybe at that point we'd make the props more like:
```
heroImg?: { src: string, alt: string; fit?: ImageFitType };
```

src/components/BaseCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+  } = props;
+  const tid = useTestIds(props, "baseCard");
+  const { isFocusVisible, focusProps } = useFocusRing();
+  const cardAction = to || onClick;
+
+  const actionAttrs = {
+    ...tid.action,
+    // Without this, an interactive card with a decorative hero (`imgAlt` unset) still gets a correct
+    // accessible name — the browser computes it from the link/button's own text content.
+    ...(imgAlt ? { "aria-label": imgAlt } : {}),
```
bdow: **thought:** This comment seems to say we don't really need a `aria-label` here as the browser computes it from the text content. So maybe we can drop it?

src/components/BaseCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+        ...(aiMode ? Css.bw2.bc(Tokens.AiFieldFg).$ : {}),
+      }}
+      {...tid}
+    >
+      {cardAction ? (
+        getButtonOrLink(content, cardAction, actionAttrs)
+      ) : (
+        // Keep the same box as the action, so the card's layout doesn't depend on it being interactive.
+        <div css={contentStyles}>{content}</div>
+      )}
```
bdow: Per:
> // Keep the same box as the action, so the card's layout doesn't depend on it being interactive.

At first I didn't see what was the same about it. But looking back up I see that `contentStyles` is what we also apply to the `actionAttrs`.
Makes me think the comment probably isn't even necessary.

src/components/Card.tsx
```diff
@@ -31,6 +31,7 @@ export type CardProps = {
   tag?: CardTag;
 };
 
+/** @deprecated Prefer {@link BaseCard} if a card is needed. */
```
bdow, line 34: **praise:** Thank you

src/components/BaseCard.tsx
bdow: Yea, its interesting... Like what do we consider the "customizable body". Is it only what is below the title?
Looking at the [cards](https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=1345-2592&t=EePVwSHPYjj8Dfvt-4) design has so done far, that seems to be the case. 
The current usages both go through TableCard. But the new one for the "Create Change Event" flow isn't in a table. So it would have to rebuild the same eye brow & title styles. That makes me think we either need to build it in, or (maybe better) we create a "CardBody" component that can take a title, eyebrows, badge, etc and lay those elements out consistently. It could do something like:
```
type CardBodyProps = {
  title: String;
  leftEyebrow?: String;
  rightEyebrow?: String;
  dataBlock?: CardData[];
  ...
};
```
Basically everything the body can potentially do now, but also give it `children` so someone can place whatever they want underneath our predefined slots. 
What do you think?

apattersonATX-HB: I think that's great

src/components/Table/components/TableCard.tsx
```diff
@@ -161,143 +152,43 @@ export function TableCardView(props: TableCardViewProps) {
...
+        {...tid}
+        title={title}
+        leftEyebrow={leftEyebrow}
+        rightEyebrow={rightEyebrow}
+        badge={badge}
+        badgeTags={badgeTags}
+        data={data}
+        progress={progress}
+        aiMode={aiMode}
+        hasFooter={!!shownFooter}
```
bdow: **thought:** I haven't read down yet, but curious as to why the body cares whether or not a footer is going to be shown...

src/components/CardBody.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const {
+    title,
+    leftEyebrow,
+    rightEyebrow,
+    badge,
+    badgeTags,
+    data = [],
+    progress,
+    aiMode = false,
+    hasFooter = false,
```
bdow: **question:** I see this drives the `isLastSection` which just conditionally adds `pb3`. I wonder if we can do that further up. Like, maybe the `CardBody` doesn't need to care about how its positioned in the card. That feels like its more of the card's job to me. I'd like to maybe also get rid of `topGap` if we can. But if we can't, maybe `topGap` becomes an xss property for padding?

bdow (approved):

bdow:
@allangaldinosilva - FYI. This is coming through soon. You should be able to update to use `BaseCard` and `CardBody` for you needs on that new workflow.

allangaldinosilva:
> @allangaldinosilva - FYI. This is coming through soon. You should be able to update to use `BaseCard` and `CardBody` for you needs on that new workflow.

Thanks a lot @apattersonATX-HB & @bdow 👏

PR #1423, opened 2026-09-03
feat: Support tooltips on primary workflow actions

bdow:
Changes 'isValid' prop to 'primaryDisabled' to better line up with Button.disabled prop, so it can easily pass through a "disabled reason" for the button. 'isValid' was thought to basically be 'formState.valid', but all it did was gate the Continue/Create buttons. The new name is more clear as to what the prop actually does vs just some information.


allangaldinosilva (approved):
looks great!

PR #1424, opened 2026-09-03
feat: ai menu item

apattersonATX-HB:


src/components/internal/MenuItem.tsx
```diff
@@ -141,6 +143,19 @@ function ImageMenuItem(item: ImageMenuItemType) {
...
+      {/* The aiStar artwork fills its full viewBox, but Design draws it inset within a 24px slot, so
+          scale it down and center it in a same-size box to match other menu items' icon alignment. */}
+      <span css={Css.df.aic.jcc.fs0.sqPx(24).mr2.$}>
+        <Icon icon="aiStar" inc={2.125} />
+      </span>
+      <span css={Css.if(!isDisabled).color(Palette.Purple600).$}>{label}</span>
+    </>
+  );
+}
+
```
apattersonATX-HB, line 158: I have a follow up ticket to handle this

https://linear.app/homebound-team/issue/DS-350/update-aistar-to-new-proper-size

Since it'll affect all of our other Ai components and i want a clean diff

src/components/ButtonMenu.tsx
bdow: **question:** We did Image and icon separate types so a user can't mix and `icon` and `src` prop. Would it make sense for `ai` to do the same thing? Otherwise, someone can try to do both an `aiMode: true, icon: "warning"`, but that icon wouldn't show through. Maybe fine?

bdow (approved):

src/components/ButtonMenu.tsx
apattersonATX-HB: Updated comment. We're gonna expand on this anyway later as discussed

PR #1425, opened 2026-09-03
feat: Refactor document-scroll right pane overlay for tables

bdow:
## Summary
- Refactors table right pane to `DocumentScrollOverlayRightPaneLayout` with stable main subtree (no table re-render on open/close)
- Moves open state to external stores on `RightPaneProvider`; row handlers should use `useRightPaneActions()`
- Adds viewport geometry hook for fixed overlay positioning below sticky table chrome

**Stack:** Part 1 of 2 — merge before [#TBD forms PR].

## Test plan
- [x] `yarn tsc --noEmit`
- [x] Right pane layout tests
- [x] GridTableLayout tests


bdow:
Split into stacked PRs: #1428 (table overlay + perf) and #1429 (forms inline). Full branch preserved at `ds-339-right-pane-support-full`.

PR #1426, opened 2026-09-04
feat: Support axis lock for DnDGrid

bdow:
This is very helpful for reordering a linear list. Helps keep the UI look clean by keeping the item being dragged on its track.

Also fixes an issue where we'd lose track of the mouse movements due to scoping the mouse move events too local to the component.

### DIFF
--- a/src/components/DnDGrid/DnDGrid.tsx
+++ b/src/components/DnDGrid/DnDGrid.tsx
@@ -1,5 +1,5 @@
 import equal from "fast-deep-equal";
-import React, { KeyboardEvent, ReactNode, useCallback, useRef } from "react";
+import React, { KeyboardEvent, ReactNode, useCallback, useEffect, useRef } from "react";
 import { Css, Palette, Properties, useTestIds } from "src";
 import { ElementStyleSnapshot, isDefined, restoreElementStyle, setInlineStyles, snapshotElementStyle } from "src/utils";
 import { zIndices } from "src/utils/zIndices";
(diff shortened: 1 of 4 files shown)


src/forms/FormSection/FormSection.tsx
```diff
@@ -77,7 +77,7 @@ function DraggableChildren(props: DraggableChildrenProps) {
   };
 
   return (
-    <DnDGrid onReorder={handleReorder} gridStyles={Css.gtc("minmax(0, 1fr)").gap3.$}>
+    <DnDGrid onReorder={handleReorder} lockAxis="y" gridStyles={Css.gtc("minmax(0, 1fr)").gap3.$}>
```
bdow, line 80: @RoseAndresHB - this should make the drag & drop on that workflow page more smooth by locking in the y axis. I also fixed some bugs regarding when the mouse moves out of the draggable area. Should be a good improvement.

src/components/DnDGrid/DnDGrid.tsx
```diff
@@ -11,12 +11,14 @@ export type DnDGridProps = {
...
   activeItemStyles?: Properties;
+  /** Restrict drag movement to one axis. Omit to allow free 2D movement. */
+  lockAxis?: "x" | "y";
   /** Returns the new order of the GridItems. */
   onReorder: (items: string[]) => void;
 };
 
 export function DnDGrid(props: DnDGridProps) {
-  const { children, gridStyles, onReorder, activeItemStyles } = props;
+  const { children, gridStyles, onReorder, activeItemStyles, lockAxis } = props;
```
stephenh, line 21: Given how many of our DNDs are for reordering a vertical list (i.e. probably all of them except the Todo kanban board 😅 ) almost wonder if we should just default to lockAxis=y as a default. 🤔

stephenh (approved):

PR #1428, opened 2026-09-04
feat: Refactor document-scroll right pane overlay for tables

bdow:
## Summary
- Refactors table right pane into `DocumentScrollOverlayRightPaneLayout`: fixed overlay pane, spacer for horizontal scroll, stable main subtree (table does not re-render on open/close)
- Replaces Context + nested providers with a module-scoped `rightPaneStore` (`useSyncExternalStore`); removed `RightPaneProvider` from `BeamContext`
- Adds `useRightPaneActions()` for row handlers (open/close only, no open-state subscription); use `useRightPane()` when you need `isRightPaneOpen`
- Adds viewport geometry hook so the fixed pane pins below sticky table chrome

**Stack:** Part 1 of 2 — merge before #1429.


src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+/** Full-width main + fixed overlay pane on desktop; full-bleed pane on `sm`. */
+export function DocumentScrollOverlayRightPaneLayout({
+  children,
+  paneWidth = defaultDocumentScrollRightPaneWidth,
+}: DocumentScrollOverlayRightPaneLayoutProps) {
+  const { sm } = useBreakpoint();
+  const tid = useTestIds({}, "documentScrollRightPaneLayout");
+  const mainTid = useTestIds({}, "rightPaneMain");
+  const spacerTid = useTestIds({}, "rightPaneSpacer");
```
bdow: **suggestion:** We don't need so many testids. Just use one

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  return <DocumentScrollRightPane paneWidth={paneWidth} mobile={false} behavior="overlay" anchorRef={anchorRef} />;
+}
+
+/** Overlay main column — full chrome width; spacer adds horizontal scroll when open. */
+function DocumentScrollOverlayRightPaneMain({ mainPart, children }: { mainPart: object; children: ReactNode }) {
+  return (
+    <div
+      css={Css.fs0.mw("fit-content").w(`min(100%, ${documentScrollChromeWidth()})`).$}
+      data-right-pane-main
```
bdow: **suggestion:** Remove - not being used. Look for other instances like this.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const waitForPaneExit = () => {
+      if (document.querySelector("[data-right-pane-content]")) {
+        frame = requestAnimationFrame(waitForPaneExit);
+        return;
+      }
+      setReserveOverlayChrome(false);
+    };
+    frame = requestAnimationFrame(waitForPaneExit);
+    return () => cancelAnimationFrame(frame);
+  }, [isRightPaneOpen]);
```
bdow: **ask:** Can we do this another way? Isn't there a way for the AnimatePresence to handle this?

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  useLayoutEffect(() => {
+    const layoutRoot = anchorRef.current;
+    const spacer = spacerRef.current;
+    if (!layoutRoot) return;
```
bdow: **question:** What is this layout effect doing? Really hard to understand this one looking at it without any comments.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+    <DocumentScrollRightPaneLayoutRoot anchorRef={anchorRef} tid={tid}>
+      {sm ? (
+        <>
+          <DocumentScrollRightPaneMainChildren>{children}</DocumentScrollRightPaneMainChildren>
+          <DocumentScrollRightPane paneWidth={paneWidth} mobile={true} />
+        </>
+      ) : (
+        <>
+          <div css={Css.df.aifs.mw100.w("fit-content").$}>
+            <DocumentScrollOverlayRightPaneMain mainPart={mainTid.overlay}>
```
bdow: **request:** `mainPart=`mainTid.overlay` is pretty gross.
Suggestion:
```
<DocumentScrollOverlayRightPaneMain {...tid.overlay}>
```
```
function DocumentScrollOverlayRightPaneMain(props: { children: ReactNode }) { 
  const { children, ...others } = props
  const tid = useTestIds(others, "overlay");
}
```

Or, don't even pass a testid through... This extra prop is not how we do passing of test ids.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+    </div>
+  );
+}
+
+/** Zero-width slot; width toggled imperatively when the overlay pane opens. */
+const DocumentScrollOverlayRightPaneSpacer = forwardRef<HTMLDivElement, Record<string, unknown>>(
+  function DocumentScrollOverlayRightPaneSpacer(tid, ref) {
+    return <div ref={ref} aria-hidden css={Css.fs0.fg0.h1.$} style={{ width: 0 }} {...tid} />;
+  },
+);
```
bdow: **question:** Does this really need to be a separate component. It'll always render with the side panel no? Just not on mobile, right? Can we consolidate. Not need for something like this to be its own component.

src/components/Layout/RightPaneLayout/DocumentScrollRightPane.tsx
```diff
@@ -12,55 +11,130 @@ import {
...
-  useEffect(() => closePane, [closePane]);
+    let frame = 0;
+    const waitForPaneExit = () => {
+      if (document.querySelector("[data-right-pane-content]")) {
+        frame = requestAnimationFrame(waitForPaneExit);
+        return;
+      }
+      releaseAfterExit();
+    };
+    frame = requestAnimationFrame(waitForPaneExit);
```
bdow: This again? Is this another right pane - why do we handle this one separately like this. Makes me think we're doing things wrong here.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+  anchorRef,
+  spacerRef,
+  paneWidth,
+}: {
+  anchorRef: RefObject<HTMLDivElement | null>;
+  spacerRef: RefObject<HTMLDivElement | null>;
+  paneWidth: number;
+}) {
+  const { isRightPaneOpen } = useRightPaneOpenState();
+  const effectivePaneWidth = documentScrollRightPaneWidth(paneWidth);
```
bdow: This returns a css property. It doesn't look like it. can we make this more clear.

src/components/Layout/RightPaneLayout/DocumentScrollRightPaneLayout.tsx
bdow: **question:** Whats the difference between DocumentScrollOverlayRightPaneLayout and this file? Do we really need both?

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
bdow, line 1: FWIW, I'm planning to remove the "**DocumentScroll**" prefix from these files as a follow up PR. This will require putting a `Legacy` prefix on others if there are conflicts. Not doing as part of this PR as its already fairly large.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
```diff
@@ -6,54 +6,64 @@ import { EnvironmentBannerLayoutHeightProvider } from "src/layouts/EnvironmentBa
...
+
+describe("DocumentScrollOverlayRightPaneLayout", () => {
+  beforeEach(() => {
+    document.documentElement.style.setProperty(beamFloatingRightOffsetVar, "0px");
+  });
 
-describe("DocumentScrollRightPaneLayout", () => {
   afterEach(() => {
-    document.documentElement.style.removeProperty(beamFloatingRightOffsetVar);
+    document.documentElement.style.setProperty(beamFloatingRightOffsetVar, "0px");
```
bdow: **question:** Do we really need to do this in a beforeEach & afterEach?

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
```diff
@@ -6,54 +6,64 @@ import { EnvironmentBannerLayoutHeightProvider } from "src/layouts/EnvironmentBa
...
     // When the pane is opened
     await clickAndWait(r.openPaneBtn);
 
-    // Then the pane/spacer render; scoped width and root floating offset match the effective pane width
+    // Then the fixed overlay pane renders; scoped width and root floating offset match it
     expect(r.rightPaneContent).toBeInTheDocument();
-    expect(r.rightPaneSpacer).toBeInTheDocument();
-    expect(r.rightPaneSpacer).toHaveStyle({ width: expectedWidth });
+    expect(r.documentScrollRightPaneLayout_spacer).toHaveStyle({ width: expectedWidth });
+    expect(r.rightPaneContent.style.marginLeft).toBe("");
```
bdow, line 48: **question:** Why do we test for this `marginLeft`? Can that have a comment if its a valid assertion.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
```diff
@@ -6,54 +6,64 @@ import { EnvironmentBannerLayoutHeightProvider } from "src/layouts/EnvironmentBa
...
 
-    // Then the spacer and pane clear; width vars return to 0
+    // Then the pane clears; spacer width and width vars return to 0 after the exit animation
     await waitFor(() => {
       expect(r.query.rightPaneContent).toBeNull();
     });
-    expect(r.query.rightPaneSpacer).toBeNull();
+    await waitFor(() => {
+      expect(r.documentScrollRightPaneLayout_spacer).toHaveStyle({ width: "0px" });
+    });
```
bdow: **question:** Can we use a single `await waitFor` or do we need one per assertion?

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
```diff
@@ -6,54 +6,64 @@ import { EnvironmentBannerLayoutHeightProvider } from "src/layouts/EnvironmentBa
...
+    expect(r.documentScrollRightPaneLayout_spacer).toBeInTheDocument();
     expect(r.documentScrollRightPaneLayout.style.getPropertyValue(beamRightPaneWidthVar)).toBe("0px");
     expect(document.documentElement.style.getPropertyValue(beamFloatingRightOffsetVar)).toBe("0px");
```
bdow, line 61: **suggestion:** comments would be help here to understand why these are set.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
```diff
@@ -106,18 +116,49 @@ describe("DocumentScrollRightPaneLayout", () => {
...
+
+    // When the pane is opened
+    await clickAndWait(r.openPaneBtn);
+
+    // Then a single overlay pane uses the outer width
+    expect(r.queryAllByTestId("rightPaneContent")).toHaveLength(1);
+    expect(r.documentScrollRightPaneLayout_spacer).toBeInTheDocument();
+    expect(r.documentScrollRightPaneLayout.style.getPropertyValue(beamRightPaneWidthVar)).toBe(
+      documentScrollRightPaneWidthCss(320),
+    );
```
bdow: **question:** Is this necessary to validate? Seems like we already ensured that `documentScrollRightPaneLayout` is in the dom. Is that enough? Do we really need to open things again and validate. Seems superfluous.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const spacerRef = useRef<HTMLDivElement | null>(null);
+
+  if (nestedInLayout && process.env.NODE_ENV !== "production" && !nestedWarnedRef.current) {
+    nestedWarnedRef.current = true;
+    console.warn(
+      "DocumentScrollOverlayRightPaneLayout is nested inside another document-scroll right-pane layout. Use a single layout (compose at the body; do not set withRightPane on both a parent and a nested layout).",
+    );
+  }
+
+  if (nestedInLayout) return <>{children}</>;
```
bdow: **question:** Is this warning needed? Seems like we handle things properly by just returning `children`. Maybe this guard is unnecessary?

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
bdow: Dropped `beforeEach`; `afterEach` resets the root var so other tests do not inherit it. Removed the pre-open `documentElement` assertion — covered after open/close.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
bdow: Added a comment — asserts the fixed overlay does not pick up inline clear mode's negative `marginLeft`.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
bdow: Would prefer one `waitFor`, but `testing-library/no-wait-for-multiple-assertions` requires separate callbacks — added a comment.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
bdow: Added — desktop overlay keeps the spacer node mounted; only its width resets while closed.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
bdow: Trimmed the open assertions — single layout root in the tree is enough here.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
bdow: Keeping it — behavior is correct without the warn, but double `withRightPane` is easy to misconfigure silently. Dev-only `console.warn` nudges toward a single layout at the body.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
```diff
@@ -6,54 +6,63 @@ import { EnvironmentBannerLayoutHeightProvider } from "src/layouts/EnvironmentBa
...
     expect(r.documentScrollRightPaneLayout.style.getPropertyValue(beamRightPaneWidthVar)).toBe(expectedWidth);
     expect(document.documentElement.style.getPropertyValue(beamFloatingRightOffsetVar)).toBe(expectedWidth);
 
     // When the pane is closed
     click(r.closePaneBtn);
 
-    // Then the spacer and pane clear; width vars return to 0
+    // Then the pane clears; spacer width and width vars return to 0 after the exit animation
+    // Separate waitFor calls — eslint disallows multiple assertions per waitFor callback.
     await waitFor(() => {
```
stephenh: I bet these `await waitFor`s could go await if line 54 was a `await clickAndWait` 

...or maybe this is an animation thing?

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { waitForRightPaneExit } from "./waitForRightPaneExit";
+
+export type DocumentScrollOverlayRightPaneLayoutProps = {
+  children: ReactNode;
+  /** Width (px) of the detail pane opened via `useRightPane`. */
+  paneWidth?: number;
+};
+
+/** Full-width main + fixed overlay pane on desktop; full-bleed pane on `sm`. */
+export function DocumentScrollOverlayRightPaneLayout({
```
stephenh, line 25: Just curious, do we have a _non_ overlay version of the right pane?

This is just such a wordy name, wondering if `DocumentScrollWithRightPaneLayout` (or similar) would be simpler

bdow: Yes, we will have one. That's a follow up PR. It'll be "Inline" vs "Overlay"

bdow: fwiw, I'm also planning to remove the "DocumentScroll" prefix in the future.

src/components/Layout/RightPaneLayout/rightPaneStore.ts
```diff
@@ -0,0 +1,73 @@
+import type { ReactNode } from "react";
+import type { OpenRightPaneOpts } from "./types";
+
+export type RightPaneOpenActions = {
```
stephenh, line 8: I won't complain too much, since this is not 3 contexts, but seems fairly "when in rome" to just use a mobx store with useComputeds or something, vs. a hand-rolled-but-admittedly-only-50 LOC state library 😅

src/setupTests.tsx
```diff
@@ -17,6 +18,7 @@ beforeEach(() => {
   // Reset to a desktop viewport so a viewport set in one test can't leak into the next.
   resetViewport();
   resetWindowScroll();
+  resetRightPaneStore();
```
stephenh, line 21: Oh, this is like a global / module-level store -- seems fine I guess

src/layouts/useScrollPinnedViewportBounds.ts
```diff
@@ -0,0 +1,49 @@
+import { type RefObject, useCallback, useLayoutEffect, useRef, useState } from "react";
+
+export type ScrollPinnedViewportBounds = {
+  topPx: number;
+  heightPx: number;
+};
+
+/** Viewport top/height for fixed overlays that follow an in-flow anchor, then pin below sticky chrome on scroll. */
```
stephenh: What's an in-flow anchor? Maybe ask the AI to give some `i.e.` examples.

An ASCII diagram would be cute too, if it can make one easily that makes sense.

src/components/Layout/RightPaneLayout/waitForRightPaneExit.ts
```diff
@@ -0,0 +1,15 @@
+import { rightPaneContentDataAttribute } from "./types";
+
+/** Poll until {@link rightPaneContentDataAttribute} leaves the DOM (exit animation done). */
```
stephenh: Is this a test-only thing? Maybe explain why non-test callers would want to do this, i.e. `i.e. we wait for this to do ...behavior...`

src/components/Layout/RightPaneLayout/types.ts
```diff
@@ -0,0 +1,30 @@
+import type { ReactNode } from "react";
+
+/** Desktop right-pane strategy. See `docs/layouts.md` for overlay / push / auto outcomes. */
+export type DocumentScrollRightPaneMode = "auto" | "overlay" | "push";
```
stephenh: Generally like to avoid `types.ts` files and put these closest to whatever method/function is accepting them.

src/components/Layout/RightPaneLayout/withRightPane.ts
```diff
@@ -0,0 +1,27 @@
...
+/** Desktop split-pane outcome (`md+` only). */
+export type ResolvedDocumentScrollRightPaneBehavior = "overlay" | "push" | "clear";
+
+export type ResolvedWithRightPane = {
+  width: number;
+  mode: DocumentScrollRightPaneMode;
+};
+
+/** Normalize `withRightPane` prop into width + mode, or `undefined` when opted out. */
+export function resolveWithRightPaneOptions(
```
stephenh, line 31: The file name / function name seem fairly mismatched; kinda wonder if this file could just be inlined into its main caller

src/components/Layout/RightPaneLayout/documentScrollRightPaneLayoutShared.tsx
```diff
@@ -0,0 +1,44 @@
...
+/** Nested guard, anchor ref, test id, and pane width var for document-scroll right-pane hosts. */
+export function DocumentScrollRightPaneLayoutRoot({
+  anchorRef,
+  tid,
+  children,
+}: {
+  anchorRef: ReturnType<typeof useDocumentScrollRightPaneAnchorRef>;
+  tid: object;
+  children: ReactNode;
+}) {
```
stephenh: A dedicated `...RootProps` type and/or removing the inline destructure would make this 8 LOC param declaration a lot cleaner.

src/components/Layout/RightPaneLayout/documentScrollRightPaneLayoutShared.tsx
```diff
@@ -0,0 +1,44 @@
...
+      ref.current = node;
+    },
+    [closePane],
+  );
+
+  return { ref, setRef };
+}
+
+/** Nested guard, anchor ref, test id, and pane width var for document-scroll right-pane hosts. */
+export function DocumentScrollRightPaneLayoutRoot({
```
stephenh: Same comment on another file, but the file name seems somewhat random compared to the file contents.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
stephenh: Gotcha!

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }, [nestedInLayout]);
+
+  if (nestedInLayout) return <>{children}</>;
+
+  return (
+    <DocumentScrollRightPaneLayoutRoot anchorRef={anchorRef} tid={tid}>
+      {sm ? (
+        <>
+          {children}
```
stephenh, line 49: I see `children` in this `sm ?` ... where is `children` in the else `:` block?

stephenh (approved):

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
bdow: line 58

stephenh: 👓 ah right!

stephenh: I think I was thrown off b/c I expected `...PaneHost` to be "doing the hosting" and have the children in it.

It looks like `...PaneHost` is basically orchestrating `DocumentScrollRightPane` when not in mobile mode ... kinda wonder if some of that should be `DocumentScrollRightPane`s job, but 🤷 lgtm.

bdow: Yea, its a bit confusing for sure. But basically the host orchestrates the desktop behavior for including that "spacer" to ensure we get the extra scroll area so everything below the overlay is still accessible. 
On mobile we don't need that because mobile is a full screen take over.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.test.tsx
bdow: Yep — `await clickAndWait` on close is enough; dropped the extra `waitFor`s.

src/layouts/useScrollPinnedViewportBounds.ts
bdow: Sounds good. Will add.

src/components/Layout/RightPaneLayout/waitForRightPaneExit.ts
bdow: Adding comments to provide clarity.

src/components/Layout/RightPaneLayout/types.ts
bdow: Moved these next to the functions that accept them (`withRightPane.ts` / `rightPaneStore`).

src/components/Layout/RightPaneLayout/withRightPane.ts
bdow: Renamed to `withRightPane.ts` so it matches `resolveWithRightPaneOptions`.

src/components/Layout/RightPaneLayout/documentScrollRightPaneLayoutShared.tsx
bdow: Extracted `DocumentScrollRightPaneLayoutRootProps`.

src/components/Layout/RightPaneLayout/documentScrollRightPaneLayoutShared.tsx
bdow: Split — hook is `useDocumentScrollRightPaneAnchorRef`; root/context is `DocumentScrollRightPaneLayoutRoot`.

PR #1429, opened 2026-09-04
feat: Add overlay right pane for forms and centered layouts

bdow:
Wire FormSectionLayout and CenteredLayout via withRightPane onto the
overlay host. `reserveScroll: "auto"` inserts a pane-width spacer when
leftover chrome stays usable (≥480px); otherwise the pane overlays in
place. Pin workflow headers and add RightPanePanel chrome.


src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -13,24 +13,36 @@ import { DocumentScrollRightPaneLayoutRoot, NestedRightPaneLayoutContext } from
...
+  type ReserveScroll,
+} from "./withRightPane";
 
 export type DocumentScrollOverlayRightPaneLayoutProps = {
   children: ReactNode;
   /** Width (px) of the detail pane opened via `useRightPane`. */
   paneWidth?: number;
+  /**
+   * Whether to insert a horizontal-scroll spacer while the pane is open.
+   * `true` for tables; `"auto"` for forms (pane-width spacer when leftover chrome is usable).
```
stephenh, line 29: Could we say why it's true for tables, and when users should/should-not set this?

I.e. wondering if this is an internal prop that we could always set somehow, i.e. detect that we're in a table or what not.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
```diff
@@ -43,19 +55,24 @@ export function DocumentScrollOverlayRightPaneLayout({
...
         <>
           {children}
           <DocumentScrollRightPane paneWidth={paneWidth} mobile={true} />
         </>
       ) : (
         <>
-          <div css={Css.df.aifs.mw100.wfc.$}>
-            <div css={Css.fs0.mwfc.w(`min(100%, ${documentScrollChromeWidth()})`).$}>{children}</div>
+          <div css={growWithContent ? Css.df.aifs.wfc.mw100.$ : Css.df.aifs.w100.$}>
+            <div css={growWithContent ? Css.w100.mwfc.$ : Css.fg1.mw0.$}>{children}</div>
```
stephenh, line 67: Wouldn't mind an in-source comment of "when growWithContent=true, we want ...behavior x... because of ...rationale y..."

src/components/Layout/RightPaneLayout/DocumentScrollRightPane.tsx
```diff
@@ -4,31 +4,23 @@ import { usePreventScroll } from "react-aria";
...
-  /** Desktop only: fixed overlay, or in-flow push/clear. Ignored when `mobile` is true. */
-  behavior?: ResolvedDocumentScrollRightPaneBehavior;
-  /** Desktop overlay only: layout root for fixed pane top/height. */
+  /** Desktop only: layout root for fixed pane top/height. */
   anchorRef?: RefObject<HTMLElement | null>;
 };
 
-/** Detail pane UI: mobile takeover, desktop fixed overlay, or desktop sticky in-flow (push/clear). */
-export function DocumentScrollRightPane({ paneWidth, mobile, behavior, anchorRef }: DocumentScrollRightPaneProps) {
+/** Detail pane UI: mobile takeover or desktop fixed overlay. */
```
stephenh, line 22: What does `desktop fixed overlay` mean? Like a right pane that doesn't scroll i.e. it's statically overlaid?

src/components/Layout/RightPaneLayout/DocumentScrollRightPaneLayoutRoot.tsx
```diff
@@ -10,14 +10,21 @@ export type DocumentScrollRightPaneLayoutRootProps = {
   anchorRef: ReturnType<typeof useDocumentScrollRightPaneAnchorRef>;
   tid: object;
   children: ReactNode;
+  /** Tables grow the document with wide content; forms stay leftover-sized. */
+  expandToMinContent?: boolean;
```
stephenh, line 14: ...so tables are the one that expand to min content?

src/components/Layout/RightPaneLayout/RightPanePanel.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    // Then the pane content clears after exit animation
+    await waitFor(() => {
+      expect(r.query.rightPaneContent).toBeNull();
+    });
+  });
+
+  it("puts the close control in the header on sm", async () => {
+    // Given a mobile viewport
+    setViewport("sm");
```
stephenh, line 39: I'd really like these `setViewport`s to become `opts` to `render` at some point

src/components/Layout/RightPaneLayout/withRightPane.test.ts
```diff
@@ -0,0 +1,147 @@
...
+import {
+  defaultDocumentScrollRightPaneWidth,
+  minDocumentScrollMainWidthPx,
+  readDocumentScrollChromeWidthPx,
+  resolveOverlaySpacerWidthPx,
+  resolveReserveScroll,
+  resolveWithRightPaneOptions,
+} from "./withRightPane";
+
+describe("readDocumentScrollChromeWidthPx", () => {
```
stephenh, line 12: Would like our tests to have only a single top-level `describe` that is the file name -- need to get this into our AGENTS.md b/c I've noticed my agent doing the same.

src/components/Layout/RightPaneLayout/DocumentScrollRightPane.tsx
bdow: Sorry, I thought this was still in draft. Going over it myself still as I find some of this also confusing. Trying to come up with better names in places. Thanks for the feedback though.

src/components/Layout/RightPaneLayout/withRightPane.ts
```diff
@@ -1,46 +1,90 @@
-/** Desktop right-pane strategy. See `docs/layouts.md` for overlay / push / auto outcomes. */
-export type DocumentScrollRightPaneMode = "auto" | "overlay" | "push";
+import { beamLayoutViewportWidthVar, beamSideNavLayoutWidthVar } from "src/layouts/layoutVars";
 
-/** Desktop strategies for the upcoming inline (push/clear) pane only. */
-export type DocumentScrollInlineRightPaneMode = "auto" | "push";
+/** Whether the overlay host inserts a horizontal-scroll spacer while the pane is open. */
```
stephenh, line 3: Do users have to know when to set this, or it's just handled for them?

src/layouts/CenteredLayout/CenteredLayout.stories.tsx
```diff
@@ -34,6 +37,33 @@ export function Small() {
...
+      <NavbarLayout navbar={createNavbar()}>
+        {withSideNav ? <SideNavLayout sideNav={{ items: sideNavItems() }}>{body}</SideNavLayout> : body}
+      </NavbarLayout>
+    );
+  },
+  {
+    args: { size: "sm", withSideNav: true },
+    argTypes: {
+      size: { control: { type: "select" }, options: ["sm", "lg"] },
+      withSideNav: { control: "boolean" },
```
stephenh, line 61: Does chromatic snapshot each combination of these argTypes?

src/utils/sbComponents.tsx
```diff
@@ -359,3 +361,59 @@ function GridTableLayoutRightPaneDetail({ name }: { name: string }) {
     </div>
   );
 }
+
+/** Placeholder fields plus a control that opens a document-scroll right pane (`RightPanePanel`). */
+export function RightPaneStoryOpenFields({ count = 2 }: { count?: number }) {
```
stephenh, line 366: Kinda feels like an odd place for these.

stephenh (approved):
I don't feel like I'm close enough to judge whether "is this really too complicated?" -- it kinda feels like it, but ngl the storybook examples are slick/just work so 🤷 lgtm! 😅

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
bdow: Dropped `reserveScroll` / `"auto"` — users only set `withRightPane` (`true` / px / `{ width? }`). Tables and forms share one grow + spacer path.

src/components/Layout/RightPaneLayout/DocumentScrollRightPane.tsx
bdow: Yes — desktop is `position: fixed` over the page, pinned below sticky chrome. The page scrolls underneath; a pane-width spacer keeps covered content reachable. Updated the JSDoc.

src/components/Layout/RightPaneLayout/DocumentScrollOverlayRightPaneLayout.tsx
bdow: That flag is gone. On `md+` the document always grows to fit content + the pane spacer; `sm` stays 100% wide.

src/components/Layout/RightPaneLayout/DocumentScrollRightPaneLayoutRoot.tsx
bdow: Not tables-only. `md+` grows for forms and tables. The `expandToMinContent` prop is gone; the host root applies that itself.

src/components/Layout/RightPaneLayout/RightPanePanel.test.tsx
bdow: Agreed — would rather pass viewport on `render`. Out of scope here; happy to do it as a follow-up.

src/components/Layout/RightPaneLayout/withRightPane.test.ts
bdow: Renamed the describe to `withRightPane` and added the “one top-level describe = filename” rule to AGENTS.md.

src/components/Layout/RightPaneLayout/withRightPane.ts
bdow: Dropped `reserveScroll` / `"auto"` — users only set `withRightPane` (`true` / px / `{ width? }`). Tables and forms share one grow + spacer path.

src/layouts/CenteredLayout/CenteredLayout.stories.tsx
bdow: No, chromatic won't do snapshots of each. This was just to add controls while testing. This story has been removed and consolidated all examples under a Layouts/With Right Pane set of stories.

src/utils/sbComponents.tsx
bdow: Moved them to local helpers in the With Right Pane stories.

PR #1448, opened 2026-09-09
feat: update aiStar svg; ProposoedValue orginalValue color;

apattersonATX-HB:


src/components/AiBadge.tsx
```diff
@@ -15,8 +15,8 @@ export function AiBadge(props: AiBadgeProps) {
   const tid = useTestIds(props, "aiBadge");
   return (
     <span {...tid} css={Css.dif.aic.jcc.fs0.br100.sqPx(increment(inc)).bgColor(Tokens.AiFieldBg).$}>
-      {/* Design insets the sparkle by a quarter of the badge on each side, so it always draws at half that size. */}
-      <Icon icon="aiStar" inc={inc / 2} />
+      {/* Design insets the sparkle a quarter of the badge on each side; adjusted for aiStar's own inset. */}
+      <Icon icon="aiStar" inc={inc * 0.706} />
```
bdow: **thought:** This multiplier feels like something I thought we'd be able to get away from. And `.706` feels sooo precise. What does this size result in?

apattersonATX-HB: 12 for a 3 inc.

The old figma's used the wonky star so to keep it pixel perfect when we swap came up with this.

bdow: Just looking at where `AiBadge` is used - appears to be nowhere 😄. Do we still need this component. I think it was meant for Tabs, but we moved away from it.

apattersonATX-HB: No its not. Forgot that i already hooked these up to tabs and had them reviewed. Old [figmas never got updated ](https://www.figma.com/design/DchiwVkssXeYi2Er8sMU2k/H2-2026-Plans---Automated-Construction-Doc-Capture?node-id=3971-2092&m=dev), unless im looking at the wrong spot 🤷🏽‍♂️

bdow (approved):

PR #1450, opened 2026-09-09
feat: add ButtonMenu support for InlineFeedbackBanner

roberthdg:
image


src/components/InlineFeedbackBanner.tsx
```diff
@@ -9,12 +10,16 @@ import { useTestIds } from "src/utils";
 
 export type InlineFeedbackBannerType = "error" | "warning";
 
+/** A banner action: a text `Button`, or a `menu` that opens a `ButtonMenu` from a text trigger. */
+export type InlineFeedbackBannerAction =
+  ActionButtonProps | ({ kind: "menu"; label: string } & Pick<ButtonMenuProps, "items" | "disabled" | "tooltip">);
```
bdow, line 15: **thought (non-blocking):** This `kind` is very similar to what we're doing on `type HeaderAction`. I wonder if we could combine these `HeaderAction` and `ActionButton` types... Will create a follow up ticket to see about consolidating these.

bdow (approved):

PR #1451, opened 2026-09-09
feat: Add SelectedOptionPill and MultiSelectField withPillList.

bdow:
Show selections as removable pills below the field instead of chips, so the combobox stays a picker and the list can carry helper text and AI styling.
## Within MultiSelectField via (with `withPillList: true`)
Screenshot 2026-09-10 at 10 33 29 AM

## AI styles applied
Screenshot 2026-09-10 at 10 33 53 AM

### DIFF
--- a/src/forms/SelectedOptionPill.tsx
+++ b/src/forms/SelectedOptionPill.tsx
@@ -1,0 +1,38 @@
+import type { ReactNode } from "react";
+import { IconButton } from "src/components/IconButton";
+import { Css, Palette, Tokens } from "src/Css";
+import { useTestIds } from "src/utils/useTestIds";
+
+export type SelectedOptionPillProps = {
+  value: ReactNode;
+  helperText?: ReactNode;
+  onRemove: () => void;
+  aiMode?: boolean;
+};
+
+/** A selected-option capsule with a remove control. `aiMode` paints the capsule purple. */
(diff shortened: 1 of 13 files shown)


src/forms/SelectedOptionPill.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { IconButton } from "src/components/IconButton";
+import { Css, Palette, Tokens } from "src/Css";
+import { useTestIds } from "src/utils/useTestIds";
+
+export type SelectedOptionPillProps = {
+  value: ReactNode;
+  helperText?: ReactNode;
+  onRemove: () => void;
+  aiMode?: boolean;
+};
```
apattersonATX-HB, line 13: blocking: The pills need a disabled state. As it is now, if we disable a multiselect thats using pills options can still get removed

apattersonATX-HB (changes requested):

src/forms/SelectedOptionPill.tsx
bdow: Fair point. Good catch.

bdow: Fixed this. Ready to re-review. Thanks!

apattersonATX-HB (approved):

PR #1452, opened 2026-09-10
chore: Remove nested barrels and run tests without module isolation.

stephenh:
vitest's module runner is not real ESM: `export * from "./x"` copies the set
of keys present on x at that moment. When a module imports a barrel that
re-exports it, the barrel evaluates while that module is still in progress
and permanently lacks its exports. With per-file isolation the missing key
was always the module the file entered through, which it already held by
direct import, so this never surfaced. Sharing one module graph per worker
exposed it as `undefined` components in whichever file ran second.

Every folder `index.ts` is gone. src/index.ts remains as the only barrel and
public API, expanded to explicit re-exports so its exported names are exactly
the 688 it exported before. Product code, tests, and stories import the module
that declares a symbol; the helpers that lived in src/utils/index.ts moved to
src/utils/helpers.ts. oxlint forbids importing `src` from product code.

With that, vitest runs with `isolate: false`, which also needed: Testing
Library cleanup registered per file in setupTests, since its own hook only
registers on first import; framer-motion swapped for a stub through
`resolve.alias` instead of a setup-file `vi.mock`, which does not apply
consistently across files sharing a graph; the one test that mocks
react-virtuoso kept in an isolated project; and two tests asserting absolute
`objectId` counter values now assert identity stability instead. Node's own
`localStorage` global is turned off in workers so jsdom's is used.

The suite goes from 65s to about 15s, and passes under four shuffled file
orders. Storybook and the built bundle are unchanged.

---
Part of the dependency-update stack (16 PRs, land in order).
- ⬇️ Based on #1447


bdow (approved):
Quick browse looks good. Thanks

PR #1455, opened 2026-09-10
feat: Add TableSummary and StackBarGraph

RoseAndresHB:
## Summary
- add an accessible, responsive TableSummary component for displaying issues on an associated table, with buttons to activate filters on said table as well
- include desktop, mobile, no-issues, and four-status Storybook states plus focused interaction tests
- configure the project Node runtime and Yarn to consume the existing npm token

## Screenshots
verifier-ds-336-after-default
verifier-ds-336-after-four-statuses
verifier-ds-336-after-mobile
verifier-ds-336-after-no-issues

## Test plan
- [x] `yarn test src/components/TableSummary/TableSummary.test.tsx`
- [x] Live Storybook verification of filtering and scrolling


.yarnrc.yml
```diff
@@ -2,6 +2,8 @@ compressionLevel: mixed
 
 enableGlobalCache: false
 
+npmAuthToken: "${NPM_TOKEN-}"
```
RoseAndresHB, line 5: Note: needed to add this to access one of our internal npm packages

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 26.7.0
```
RoseAndresHB, line 1: Note: for `asdf` users

src/components/TableSummaryReport/types.ts
```diff
@@ -0,0 +1,32 @@
+import type { ReactNode } from "react";
+
+export type TableSummaryReportStatus = "success" | "neutral" | "warning" | "error";
```
bdow, line 3: **question/thought:** I'd be curious to see if these statuses map the same colors that our tags do.

src/components/TableSummaryReport/types.ts
bdow, line 1: **suggestion:** Would prefer to have these types inline with the component(s)

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,143 @@
...
+}
+
+function percentage(count: number, total: number): number {
+  return total === 0 ? 0 : Math.round((Math.max(0, count) / total) * 100);
+}
+
+const statusColors: Record<TableSummaryReportStatus, Palette> = {
+  success: Palette.Green500,
+  neutral: Palette.Gray500,
+  warning: Palette.Orange500,
```
bdow, line 136: **thought:** I believe there is also a `yellow600` used by the "4 Warnings" metric. 
Also, I noticed that the color of the icon for "Incomplete" is actually Orange600, while the bar graph shows Orange500. Not sure if that was intention on the design side, but am noting that. Asking clarifying questions here: https://www.figma.com/design/62R8KiDklvgBBSH0mQGWHo/BEAM_27_LIBRARY?node-id=2065-1093#1921843307

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,143 @@
...
+            />
+          ))}
+        </div>
+      )}
+      <footer
+        css={Css.df.fdc.gap2.p2.if(visibleMetrics.length > 0).bt.bc(Tokens.FieldBorderDefault).$}
+        {...tid.coverage}
+      >
+        <div css={Css.df.aic.jcsb.gap2.$}>
+          <span css={Css.xs2Sb.ttu.add("letterSpacing", "0.5px").$}>Coverage by status</span>
```
bdow, line 56: **suggestion:** I think this "graph title" should be  passed in by the user.

src/components/TableSummaryReport/TableSummaryReport.tsx
bdow, line 1: fwiw, I'm seeing the graph section and the "X Missing|Incomplete|etc" as separate components. I _think_ they should be built upon the same data, but it does look like they might not map 1:1 based on the designs. Getting clarity around that in Figma. 
Either way, I do think it'd be nice to build them as separate components. Whether the implementer passes in the graph via a prop (e.g. `footer: ReactNode`), or the graph is generated via the data passed to the bid package coverage. 
~Part~ Most of me wants to keep them decoupled. If this is a reusable component, than we might not want a horizontal stacked bar graph in every instance. So letting the implementer compose these together could be just as easy.

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,143 @@
...
+    [segments],
+  );
+
+  return (
+    <section css={Css.df.fdc.bgColor(Tokens.Surface).bc(Tokens.FieldBorderDefault).ba.br12.oh.bshBasic.$} {...tid}>
+      <header css={Css.df.aic.jcsb.gap2.p2.bb.bc(Tokens.FieldBorderDefault).$}>
+        <div css={Css.mdSb.mw0.$}>{title}</div>
+        {visibleMetrics.length > 0 && issueAction && (
+          <Button
+            label={isMobile ? "Issues" : issueAction.label}
```
bdow, line 26: **thought:** This might not always be "Issues". I think we should provide the user a way to specify this mobile label as well. not sure the best way to handle that... 🤔

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const { sm: isMobile } = useBreakpoint();
+  const tid = useTestIds(props, "tableSummaryReport");
+  const visibleMetrics = metrics.slice(0, 4);
+  const totalCount = useMemo(
+    () => segments.reduce((total, segment) => total + Math.max(0, segment.count), 0),
+    [segments],
+  );
+
+  return (
+    <section css={Css.df.fdc.bgColor(Tokens.Surface).bc(Tokens.FieldBorderDefault).ba.br12.oh.bshBasic.$} {...tid}>
```
bdow, line 21: **thought:** I don't think there's a border on this outer component. Appears to just be the box-shadow.

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,143 @@
...
+      <header css={Css.df.aic.jcsb.gap2.p2.bb.bc(Tokens.FieldBorderDefault).$}>
```
bdow, line 22: **suggestion:** So the header has a padding-x of 16px on both side. Padding-y on this should be 12px. Then there's another padding-y on the `<div>{title}</div>` that would be of 8px. That's result in the 20px of vertical padding we want to display.

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: yep agreed on that, will refactor on second pass.

src/components/TableSummaryReport/types.ts
RoseAndresHB: roger that. I remembered seeing that elsewhere but wasn't sure if it was convention or not. will refactor :+1:

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: :+1:

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: Roger. I agree with your thoughts and points. Going to go with the approach of passing in as a ReactNode prop for now and can pivot if we don't like that 👍

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: Could provide an optional override label prop and use "Issues" as the default?

src/components/TableSummaryReport/types.ts
RoseAndresHB: implemented

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: implemented

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: implemented

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: implemented

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: implemented

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: implemented

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: Watching the comment thread in figma to see what the desired solution is here.

src/components/TableSummaryReport/types.ts
RoseAndresHB: Same four names (`success` / `neutral` / `warning` / `error`), but not the same tokens. Report/graph fills are Palette `*500` (Green/Gray/Orange/Red). Tags use `*100`/`*200` backgrounds and `*600`/`*700` icons, and Tag `error` icons are Orange700 rather than red.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,143 @@
...
+}
+
+function percentage(count: number, total: number): number {
+  return total === 0 ? 0 : Math.round((Math.max(0, count) / total) * 100);
+}
+
+const statusColors: Record<TableSummaryReportStatus, Palette> = {
+  success: Palette.Green500,
+  neutral: Palette.Gray500,
+  warning: Palette.Orange500,
```
RoseAndresHB: sounds like we want to leave colors as is for now

src/components/TableSummaryReport/index.ts
bdow, line 1: Per new guidance, we're removing these barrel exports. See https://github.com/homebound-team/beam/pull/1452

src/components/StackBarGraph.tsx
```diff
@@ -0,0 +1,70 @@
+import { useMemo, type ReactNode } from "react";
+import { Css, Palette } from "src/Css";
+import { useTestIds } from "src/utils/useTestIds";
+
+export type StackBarGraphProps = {
+  title: ReactNode;
+  totalLabel: ReactNode;
+  segments: readonly StackBarGraphSegment[];
+  "data-testid"?: string;
```
bdow: **nit/suggestion:** We never (or at least I don't think we do) have an explicit "data-testid" prop. Usually that's just caught as an extra prop in a spread like `const { title, ...others } = props;` Or we don't catch it in the spread and just do something like `const tid = useTestIds(props, "default");` Would prefer to keep going down that path.

src/components/StackBarGraph.tsx
```diff
@@ -0,0 +1,70 @@
...
+};
+
+/** Horizontal stacked bar with a legend. Compose into `TableSummaryReport` via `footer`, or use standalone. */
+export function StackBarGraph(props: StackBarGraphProps) {
+  const { title, totalLabel, segments } = props;
+  const tid = useTestIds(props, "stackBarGraph");
```
bdow, line 14: Looks like you're doing it this way anyways. Maybe we can remove the prop? If the problem is on the other end, then we just need to spread the test id into this component via `<StackBarGraph title="Foo" {...tid} />`

src/components/StackBarGraph.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <span css={Css.xs2Sb.ttu.add("letterSpacing", "0.5px").$}>{title}</span>
+        <span css={Css.xs.wsnw.$}>{totalLabel}</span>
+      </div>
+      <div css={Css.df.hPx(20).borderRadius("6px").oh.$} {...tid.bar}>
+        {segments.map((segment) => (
+          <div
+            key={segment.label}
+            css={
+              Css.flexGrow(Math.max(0, segment.count))
+                .add("minWidth", segment.count > 0 ? "1px" : 0)
```
bdow: **question:** Any reason to have chosen a flexGrow over doing a hard coded width percentage? If we continue with flex-grow, then maybe we should flex-shrink to 0. The default is `1` so elements would shrink here too. I think the percentage would be a better explicit sizing. Then we don't have to worry about the "flex" parth of flexbox.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+  );
+}
+
+type MetricButtonProps<V extends string | number> = {
+  metric: TableSummaryReportMetric<V>;
+  active: boolean;
+  onClick: ((value: V) => void) | undefined;
+  divider: boolean;
+  mobile: boolean;
+  "data-testid"?: string;
```
bdow: **nit/suggestion:** Same feedback regarding the "data-testid" prop here as before. And if this exists anywhere else in the PR.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+      </span>
+    </button>
+  );
+}
+
+const statusColors: Record<TableSummaryReportStatus, Palette> = {
+  success: Palette.Green500,
+  neutral: Palette.Gray500,
+  warning: Palette.Orange500,
+  error: Palette.Red500,
```
bdow, line 127: **thought:** I'm hesitant that these colors will be all we use, and specifically in this sort of view.
For instance, on our Bills & Credits page, we show circular graph where each segment represents a Market. So its possible that the colors we use in this Report may not be "status" colors, but could represent _anything_. 
With that in mind, I think having the user pass in an `IconKey` & `BeamColor` pair for each metric would keep this open for future cases. 
We still control the composition of the elements, but gives more flexibility for what this summary can be used for.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+  );
+  const { hoverProps, isHovered } = useHover({ isDisabled: metric.disabled });
+  const { focusProps, isFocusVisible } = useFocusRing();
+  const tid = useTestIds(props, "metric");
+
+  return (
+    <button
+      ref={ref}
+      css={{
+        ...Css.bn.outline0.df.flexGrow(1).aic.jcc.gap1.p2.mw0.bgColor(Tokens.Surface).color(Tokens.OnSurface).$,
```
bdow: **question:** If you don't put on `bn` and `outline0`, what happens? I'm guessing these were done specifically by AI trying to do a sort of "button style reset". I believe our CSS Reset file already does these. If they aren't adding anything, can we remove them?

bdow: **nit/suggestion:** You can replace `flexGrow(1)` with `.fg1`.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+        ...Css.if(divider && mobile).bb.bc(Tokens.FieldBorderDefault).$,
+        ...Css.if(divider && !mobile).br.bc(Tokens.FieldBorderDefault).$,
+        ...(isHovered && !metric.disabled ? Css.bgColor(Tokens.NeutralFillHoverSubtle).$ : {}),
+        ...(isPressed || active ? Css.bgColor(Tokens.NeutralSurfacePressed).$ : {}),
+        ...(isFocusVisible ? Css.bshFocus.$ : {}),
```
bdow, line 109: **question:** Were these hover, pressed, and focus states coming from Design. I don't recall seeing them. If not, we should ask design for anything missing.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+        ...(metric.disabled ? Css.cursorNotAllowed.o50.$ : Css.cursorPointer.$),
```
bdow, line 110: Same question for `disabled` regarding design input. 
I can't imagine a case where we'd disable these, but I still think its good to handle/define this state as it is an interactive element.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+    onStatusClick,
+    statusDisabled,
+    footer,
+  } = props;
+  const { sm: isMobile } = useBreakpoint();
+  const tid = useTestIds(props, "tableSummaryReport");
+  const visibleMetrics = metrics.slice(0, 4);
+
+  return (
+    <section css={Css.df.fdc.bgColor(Tokens.Surface).br12.oh.bshBasic.$} {...tid}>
```
bdow: **question:** Any reason for the `oh`? I'd initially advise against it, unless we have content that is bleeding out and we _really_ don't want it to. Sometimes keeping the overflow visible will help identify problems, and will ensure text doesn't get cut off for our users. If we had to choose between text being hidden vs text overflowing the element, I'd chose overflowing the element. That at least makes it accessible to the user. Either way content flowing out of the component would be a problem that should be fixed, but at least its not hiding potentially meaningful information from the user.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <header css={Css.df.aic.jcsb.gap2.px2.pyPx(12).bb.bc(Tokens.FieldBorderDefault).$}>
+        <div css={Css.mdSb.mw0.py1.$}>{title}</div>
+        {visibleMetrics.length > 0 && statusLabel != null && onStatusClick && (
+          <Button
+            label={isMobile ? statusMobileLabel : statusLabel}
+            variant="tertiary"
+            icon={null}
```
bdow: **request:** `icon={null}` is not necessary, I don't think. If it is.. let me know. Seems like something we can denote as a bug and fix outside of this PR.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+            endAdornment={<Icon icon="arrowRight" />}
+            onClick={onStatusClick}
+            disabled={statusDisabled}
+            {...tid.statusAction}
+          />
+        )}
```
bdow, line 59: **thought (non-blocking):** This is potentially a place where we could also repurpose the "HeaderActions" component. What'd probably happen is that this would become something more like:
```
export type TableSummaryReportProps = {
  ..., // existing props
  headerActions: HeaderActions[]
}
```
And the consumer would just pass in
```
<TableSummaryReport headerActions={[{ label: "View 15 issues", endAdornment: <Icon icon="arrowRight" />`, ... }]} ... />
```
Also opens that section up for any future actions we might need to support.

bdow: FWIW, I don't like over future-proofing, but I also try not to pigeon hole us into a situation where we need to update Beam too often.
Again, this is a **non-blocking** comment. I think its nice we handle the label change on smaller screens isolated here for the consumer.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+export type TableSummaryReportProps<V extends string | number> = {
+  title: ReactNode;
+  metrics?: readonly TableSummaryReportMetric<V>[];
+  activeMetricValues?: readonly V[];
+  onMetricClick?: (value: V) => void;
+  statusLabel?: ReactNode;
+  statusMobileLabel?: ReactNode;
+  onStatusClick?: VoidFunction;
+  statusDisabled?: boolean;
```
bdow, line 27: **thought:** What do you think of these props becoming somehting like 
```
statusButton: { 
  label: ReactNode;
  mobileLabel: ReactNode;
  onClick: VoidFunction;
  disabled: boolean; 
}
```
Might make these props a little cleaner looking, and keep alike props together.

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,139 @@
...
+    onMetricClick,
+    statusLabel,
+    statusMobileLabel = "Items",
+    onStatusClick,
+    statusDisabled,
+    footer,
+  } = props;
+  const { sm: isMobile } = useBreakpoint();
+  const tid = useTestIds(props, "tableSummaryReport");
+  const visibleMetrics = metrics.slice(0, 4);
```
bdow: **suggestion:** let's remove this slice. There's a case coming up soon where we'll want more than 4 potentially. We're looking to use this in Takeoffs soon.

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,139 @@
...
+  label: string;
+  count: number;
+  status: Exclude<TableSummaryReportStatus, "neutral" | "success">;
+  disabled?: boolean;
+};
+
+export type TableSummaryReportProps<V extends string | number> = {
+  title: ReactNode;
+  metrics?: readonly TableSummaryReportMetric<V>[];
+  activeMetricValues?: readonly V[];
```
bdow: **thought:** Was there an "active" state that design asked for? If not, can we drop this?

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+            icon={null}
+            endAdornment={<Icon icon="arrowRight" />}
+            onClick={onStatusClick}
+            disabled={statusDisabled}
+            {...tid.statusAction}
+          />
+        )}
+      </header>
+      {visibleMetrics.length > 0 && (
+        <div css={Css.df.fdr.if(isMobile).fdc.$} {...tid.metrics}>
```
bdow, line 62: **nit:** Would prefer to use CSS media queries for this than relying on Beam's JS to do the re-render for `isMobile`. It'd be done via `Css.fdr.ifSm.fdc.$`.

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { Button } from "src/components/Button";
+import { Icon, type IconKey } from "src/components/Icon";
+import { Css, Palette, Tokens } from "src/Css";
+import { useBreakpoint } from "src/hooks/useBreakpoint";
+import { useTestIds } from "src/utils/useTestIds";
+
+export type TableSummaryReportStatus = "success" | "neutral" | "warning" | "error";
+
+export type TableSummaryReportMetric<V extends string | number> = {
+  value: V;
```
bdow: **thought/suggestion** Looking at `value`, it appears its mostly used for the callback function, right? Also used as a key. I wonder if each of these TableSummaryReportMetrics could just have their own `onClick` and `active` (if active is an ask from design - fwiw I wouldn't do it 😄), and we can drop the generics.

src/components/TableSummaryReport/TableSummaryReport.tsx
```diff
@@ -0,0 +1,139 @@
...
+              active={activeMetricValues.includes(metric.value)}
+              onClick={onMetricClick}
+              divider={index < visibleMetrics.length - 1}
+              mobile={isMobile}
+              {...tid[`metric_${String(metric.value)}`]}
+            />
+          ))}
+        </div>
+      )}
+      {footer !== undefined && (
```
bdow: **nit:** Could just do `footer && (...)`

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,139 @@
...
+  mobile: boolean;
+  "data-testid"?: string;
+};
+
+function MetricButton<V extends string | number>(props: MetricButtonProps<V>) {
+  const { metric, active, onClick, divider, mobile } = props;
+  const ref = useRef<HTMLButtonElement>(null);
+  const { buttonProps, isPressed } = useButton(
+    { onPress: () => onClick?.(metric.value), isDisabled: metric.disabled, "aria-pressed": active },
+    ref as RefObject<HTMLButtonElement>,
```
bdow, line 94: **question:** Do we need this casting of `ref as RefObject<HTMLButtonElement>,`

src/components/StackBarGraph.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const tid = useTestIds(props, "stackBarGraph");
+  const totalCount = useMemo(
+    () => segments.reduce((total, segment) => total + Math.max(0, segment.count), 0),
+    [segments],
+  );
+
+  return (
+    <div css={Css.df.fdc.gap2.p2.$} {...tid}>
+      <div css={Css.df.aic.jcsb.gap2.$}>
+        <span css={Css.xs2Sb.ttu.add("letterSpacing", "0.5px").$}>{title}</span>
```
bdow, line 23: **thought:** Stephen's latest truss changes may have added letter-spacing as a truss helper.. not sure 100% , but worth taking a look. He tried to get rid of a bunch of these `.add()` methods.

src/components/StackBarGraph.tsx
RoseAndresHB: good idea; implemented 👍

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: my initial implementation had those props nested and i ended up flattening them because i thought that was our convention in Beam haha. If the nested props is an acceptable pattern, i can pivot back to it 🙂

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: implemented (dropped active state)

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: dropped cast

src/components/StackBarGraph.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

RoseAndresHB: followed up w/ Jenn and got hover + focus state, active not needed 👍

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: open question out to Jenn on `disabled` styling 👍

src/components/StackBarGraph.tsx
RoseAndresHB: looks like that hasn't been added just yet.

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

RoseAndresHB: implemented

src/components/TableSummaryReport/TableSummaryReport.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: got input from Design that we should never expect to see `disabled` for any of these buttons; the button should simply not be rendered if it isn't usable. Thus we do not need a disabled state, so will drop that.

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented IconKey + BeamColor props for each metric 🙂

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: dropped disabled state

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: I think gonna leave as-is for now, and can revisit if it becomes a need later 🙂

src/components/TableSummary/TableSummary.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+  component: TableSummary,
+} as Meta;
+
+export function Default() {
+  return <TableSummary {...createProps()} />;
+}
+
+export const Mobile = newStory(() => <TableSummary {...createProps()} />, {
+  parameters: { chromatic: { modes: viewportModes("iphone12") } },
+});
```
bdow: nit (non-blocking): Meant to point this out before, but we could use one story and put in two modes for chromatic here to capture both and wouldn't need a story specifically for mobile and desktop. Also would prefer to use the `mobile1` mode we've been using elsewhere. 
```
chromatic: { modes: viewportModes("desktop", "mobile1") },
```

src/components/TableSummary/TableSummary.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+          { label: "Incomplete", count: 9, icon: "errorCircle", color: Palette.Orange500 },
+          { label: "Warnings", count: 3, icon: "errorCircle", color: Palette.Orange500 },
+        ],
+      })}
+    />
+  );
+}
+
+/** Demonstrates applying a report status filter and scrolling to the filtered table. */
+export function FiltersTableAndScrolls() {
```
bdow, line 40: **praise:** Appreciate the functional story here for testing 👏

src/components/StackBarGraph.tsx
```diff
@@ -0,0 +1,73 @@
...
+  );
+}
+
+export type StackBarGraphStatus = "success" | "neutral" | "warning" | "error";
+
+export type StackBarGraphSegment = {
+  label: string;
+  count: number;
+  status: StackBarGraphStatus;
+};
```
bdow: **suggestion:** I'm thinking `StackBarGraphStatus` might be too limiting. I appreciate the semantic naming for these colors/statuses, but these might not be enough or not hold true for every stack bar graph we want to use. I think we'll probably want to allow the user to choose from any of our colors. I think the status just does map to a color, so maybe the prop is just `color: BeamColor` instead? Then we can remove the `statusColors` map and the `StackBarGraphStatus` type.

src/components/StackBarGraph.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <span css={Css.xs2Sb.ttu.add("letterSpacing", "0.5px").$}>{title}</span>
+        <span css={Css.xs.wsnw.$}>{totalLabel}</span>
+      </div>
+      <div css={Css.df.hPx(20).borderRadius("6px").oh.$} {...tid.bar}>
+        {segments.map((segment) => (
+          <div
+            key={segment.label}
+            css={
+              Css.w(`${segmentWidth(segment.count, totalCount)}%`)
+                .fs0.add("minWidth", segment.count > 0 ? "1px" : 0)
```
bdow: **thought/question:** I like the idea of ensuring the segment is shown by at least 1px, but I wonder if we'll even get to this point. And if we did, would we wind up with an overflow in this graph, then?

Quick math for an edge case; say our totalCount is 200. We're on a 100px wide stack bar graph. So every 1% maps to `0.5px`. And given our segments are:
Segment | count | % | px value
-- | -- | -- | --
1 | 100 | 50% | 50px
2 | 50 | 25% | 25px
3 | 40 | 20% | 20px
4 | 2 | 1% | 1px
5 | 2 | 1% | 1px
6 | 2 | 1% | 1px
7 | 1 | 0.5% | (min) 1px
8 | 1 | 0.5% | (min) 1px
9 | 1 | 0.5% | (min) 1px
10 | 1 | 0.5% | (min) 1px
-- | -- | -- | --
**Total** | **200** | **100%** | **102px**

And now we've overflown the container by 2px. Again, this is probably quite an edge case, but maybe its not? 🤷  We're just introducing this component, so we don't know what the edge cases really are yet. But anyways. I'd push for removing the minWidth to avoid the accidental overflow. Though, I see you have set `overflow: hidden;` should really shouldn't be a problem, unless those `1px`'s are at the end of the bar and just don't get shown... so seems like a pick your poison.

src/components/TableSummary/TableSummary.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+            { label: "Four", count: 4, icon: "errorCircle", color: Palette.Orange500 },
+            { label: "Five", count: 5, icon: "xCircle", color: Palette.Red500 },
+          ],
+        })}
+      />,
+    );
+    // Then all five are rendered
+    expect(r.tableSummary_metric_one).toBeInTheDocument();
+    expect(r.tableSummary_metric_four).toBeInTheDocument();
+    expect(r.tableSummary_metric_five).toBeInTheDocument();
```
bdow, line 38: **question:** These seems to only test 3 out of the 5. Should it have validated all of them?

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+export function TableSummary(props: TableSummaryProps) {
+  const { title, metrics = [], action, footer } = props;
+  const { sm: isMobile } = useBreakpoint();
+  const tid = useTestIds(props, "tableSummary");
+  const actionMobileLabel = action?.mobileLabel ?? "Items";
+
+  return (
+    <section css={Css.df.fdc.bgColor(Tokens.Surface).br12.bshBasic.$} {...tid}>
+      <header css={Css.df.aic.jcsb.gap2.px2.pyPx(12).bb.bc(Tokens.FieldBorderDefault).$}>
```
bdow: **praise:** Appreciate the semantic HTML

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+};
+
+export function TableSummary(props: TableSummaryProps) {
+  const { title, metrics = [], action, footer } = props;
+  const { sm: isMobile } = useBreakpoint();
+  const tid = useTestIds(props, "tableSummary");
+  const actionMobileLabel = action?.mobileLabel ?? "Items";
+
+  return (
+    <section css={Css.df.fdc.bgColor(Tokens.Surface).br12.bshBasic.$} {...tid}>
```
bdow: **nit:** We might actually want to use the `bgColor(Tokens.SurfaceRaised)` here. Anything that is "card-like" or renders above the background should be considered elevated/raised above its current surface. For the light mode, its the same background and we get the elevation contrast thanks to the box-shadow. For dark mode, it's a different background color as our dark box-shadow doesn't really display to provide that contrast between the background and this card on top of it. 
for reference, the SurfaceRaised token is used on Menus and the TableCard component at the moment.

src/components/TableSummary/TableSummary.tsx
bdow: **suggestion/request:** `FieldBorderDefault` turns out to be the wrong color here, and semantically incorrect as well. In Figma this border is Gray200. `FieldBorderDefault` is Gray300. We do have another semantic token of `SurfaceSeparator`, which is Gray200 and should be used here instead.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <header css={Css.df.aic.jcsb.gap2.px2.pyPx(12).bb.bc(Tokens.FieldBorderDefault).$}>
+        <div css={Css.mdSb.mw0.py1.$}>{title}</div>
```
bdow: **suggestion:** This could probably be an `h4`/`h3`? We're within a `<header>` but we don't signify where the `h[N]` title is. I'd also suggest adding the expected font size here, so if a user passes just a string vs a ReactNode, we'll render the proper size and not accidentally inherit something we shouldn't.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+            const metricId = defaultTestId(metric.label);
+            return (
+              <MetricButton
+                key={metricId}
+                metric={metric}
+                divider={index < metrics.length - 1}
+                {...tid[`metric_${metricId}`]}
+              />
+            );
+          })}
```
bdow: **nit (non-blocking):** super minor, but could inline the metricId and reduce lines of code by a couple here.

```
          {metrics.map((metric, index) => (
              <MetricButton
                key={metricId}
                metric={metric}
                divider={index < metrics.length - 1}
                {...tid[`metric_${defaultTestId(metric.label)}`]}
              />
           ))}
```

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+        // outline0 drops the UA focus outline (avoids a second ring beside our box-shadow).
+        // relative + z1 paints the focused cell above neighbors so the ring isn't tucked under.
+        ...Css.outline0.relative.df.fg1.aic.jcc.gap1.p2.mw0.bgColor(Tokens.Surface).color(Tokens.OnSurface).$,
+        // Desktop: right divider; small screens: bottom divider (and clear the right edge).
+        ...Css.if(divider)
+          .br.bc(Tokens.FieldBorderDefault)
+          .ifSm.bb.bc(Tokens.FieldBorderDefault)
+          .add("borderRight", "none").$,
+        ...(isHovered ? Css.bgColor(Tokens.NeutralFillHoverSubtle).$ : {}),
+        // Single blue ring (bshFocus's outer color). FocusRingMuted is near-black — not for these cells.
```
bdow, line 98: **praise;** Thanks for the comment here

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+      ref={ref}
+      css={{
+        // outline0 drops the UA focus outline (avoids a second ring beside our box-shadow).
+        // relative + z1 paints the focused cell above neighbors so the ring isn't tucked under.
+        ...Css.outline0.relative.df.fg1.aic.jcc.gap1.p2.mw0.bgColor(Tokens.Surface).color(Tokens.OnSurface).$,
+        // Desktop: right divider; small screens: bottom divider (and clear the right edge).
+        ...Css.if(divider)
+          .br.bc(Tokens.FieldBorderDefault)
+          .ifSm.bb.bc(Tokens.FieldBorderDefault)
+          .add("borderRight", "none").$,
```
bdow: **question:** Is `br0` a thing that could replace this `.add`?

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+        ...(isHovered ? Css.bgColor(Tokens.NeutralFillHoverSubtle).$ : {}),
+        // Single blue ring (bshFocus's outer color). FocusRingMuted is near-black — not for these cells.
+        ...(isFocusVisible ? Css.boxShadow(`0px 0px 0px 2px ${Palette.Blue700}`).z1.$ : {}),
+        ...Css.cursorPointer.$,
+      }}
+      {...mergeProps(buttonProps, hoverProps, focusProps)}
+      {...tid}
+    >
+      <Icon icon={metric.icon} color={metric.color} inc={4} />
```
bdow: **suggestion:** I think we should wrap this in a `<span css={Css.fs0.$}>` to ensure the icon doesn't shrink. Looks a little odd when it does.

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <span css={Css.lg.wsnw.$}>
```
bdow: **thought:** I wonder if we should allow this to wrap 🤔. Probably would be fine. If it didn't wrap, we're not doing an `overflow: hidden` anywhere so, text would just push out of its container. Otherwise, we could do a `Css.truncate.$` to prevent this from going out of its container if we really want to keep it on one line.

src/components/StackBarGraph.tsx
```diff
@@ -0,0 +1,73 @@
...
+};
+
+/** Horizontal stacked bar with a legend. Compose into `TableSummary` via `footer`, or use standalone. */
+export function StackBarGraph(props: StackBarGraphProps) {
+  const { title, totalLabel, segments } = props;
+  const tid = useTestIds(props, "stackBarGraph");
+  const totalCount = useMemo(
+    () => segments.reduce((total, segment) => total + Math.max(0, segment.count), 0),
+    [segments],
+  );
```
bdow: **thought (non-blocking):** I wonder if this is worth memo-izing 🤔. Seems very quick to compute and I don't really see what this memo is actually helping. We're not passing it down to any other components from here.

src/components/StackBarGraph.tsx
RoseAndresHB: Ah yeah i missed that. I agree with this; similar to the metric buttons

src/components/StackBarGraph.tsx
RoseAndresHB: Yeah i hear that. I think I'd lean more toward removing the minWidth.

src/components/StackBarGraph.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.stories.tsx
RoseAndresHB: implemented

src/components/StackBarGraph.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.test.tsx
RoseAndresHB: whoops! fixed that, thanks!

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: `br0` is border-radius: 0 in Truss, not border-right

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

src/components/StackBarGraph.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented — dropped `wsnw` so the label can wrap

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

src/components/TableSummary/TableSummary.tsx
bdow: Ahh. that's right.  I think I've done that mistake before 🤦

RoseAndresHB: Found an alternative but it seemed more unwieldy, so gonna leave as-is 🙂

src/components/TableSummary/TableSummary.tsx
```diff
@@ -0,0 +1,112 @@
...
+export function TableSummary(props: TableSummaryProps) {
+  const { title, metrics = [], action, footer } = props;
+  const { sm: isMobile } = useBreakpoint();
+  const tid = useTestIds(props, "tableSummary");
+  const actionMobileLabel = action?.mobileLabel ?? "Items";
+
+  return (
+    <section css={Css.df.fdc.bgColor(Tokens.SurfaceRaised).br12.bshBasic.$} {...tid}>
+      <header css={Css.df.aic.jcsb.gap2.px2.pyPx(12).bb.bc(Tokens.SurfaceSeparator).$}>
+        <h3 css={Css.mdSb.mw0.py1.m0.$}>{title}</h3>
```
bdow, line 43: **nit(non-blocking):** I don't think `m0` is necessary. our css reset should've removed these default margins on everything.

bdow (approved):
This is great. Thank you for the effort and patience going through all my comments. I really appreciate you jumping into Beam and helping out here. (Also a great note on the end-of-year review to note contributions to the whole eng org, not just your team via working in the design system 😄 🚀)

src/components/TableSummary/TableSummary.tsx
RoseAndresHB: implemented

PR #1456, opened 2026-09-10
feat: hide struck-through original when the AI proposal already matches

Jacob-Diebold:
## Summary
- [PLAN-1795](https://linear.app/homebound-team/issue/PLAN-1795/fe-drop-the-struck-through-original-value-for-fields-on-optimistically): when an AI proposal already matches the on-record value (e.g. an optimistic create), Beam no longer strikes through a duplicate of the same text.
- Comparison is on formatted display text, so dates, option lists, and cents that read identically are treated as unchanged even when the underlying values are not `===`.
- Fields stay in AI mode so the proposal treatment remains; only the redundant original is omitted, including while the user is mid-edit.
- Storybook `AiProposals` now has an "already matches the proposal" section so this case is visible next to a real change and a blank fill-in.

## Screenshots
Screenshot 2026-09-10 at 2 36 16 PM

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

Mental Checklist
- [ ] Add your squad to the reviewers list.
- [x] Updated tests for this change.
- [ ] Tested changes locally.
- [ ] Updated Documentation (if necessary).


src/inputs/hooks/useAiProposal.ts
```diff
@@ -43,12 +46,17 @@ export function useAiProposal<V>(
   const hasProposal = proposedValue !== undefined;
   const isAiMode = hasProposal && !hasTyped;
 
+  // Compared as display text, so each field's own formatting decides what counts as "the same",
+  // i.e. two `PlainDate`s or two option lists that read identically.
+  const proposedText = hasProposal ? format(proposedValue as V) : undefined;
+  const originalText = originalValue.current !== undefined ? format(originalValue.current) : undefined;
+  const isUnchanged = originalText === proposedText;
```
bdow, line 53: **praise:** Very nice. This is a much cleaner implementation to handle your use case. Thanks for championing this through!

bdow (approved):

apattersonATX-HB (approved):

src/inputs/hooks/useAiProposal.ts
stephenh: Agreed, looks great, only nit is maybe having an in-source comment that explicitly calls that this is our our creation flows will work -- right now it's a little implicit I think. But still looks great.

Jacob-Diebold: Will do! I never know how much homebound/ blueprint related work to include context for since this is open source.

PR #1459, opened 2026-09-12
fix: Move keyframes into truss-config.

stephenh:
Truss 2.32.0 lets us declare `@keyframes` in `truss-config.ts` and gives each one a `Css` abbreviation, so our four timelines no longer have to be hand-written CSS that truss cannot check.

### What moved

`loadingDots`, `spin`, `pulse`, and `aiStarLoader` are now a `keyframes` block in `truss-config.ts`, carrying their existing comments over from `CssReset.css`.

Truss writes each block only when some rule uses it, and validates every animation value against the declared names, so a typo is a build error instead of a silently dead animation.

### Callsites

Each keyframe gets a `Css.<name>(value)` method that sets `animation: <name> value`:

- `LoadingSkeleton` — `.pulse("2s cubic-bezier(0.4, 0, 0.6, 1) infinite")`
- `Loader` — `.spin("800ms linear infinite")`, replacing four longhand `.add(…)` calls
- `AiLoader` — `.aiStarLoader(…)`, replacing five
- `LoadingDots` — one `.loadingDots(…)` per dot

The `animation` shorthand resets `animation-delay` to `0s`, so the per-element delays in `LoadingDots` and `AiLoader` now ride in the shorthand as its second time value. Both have a comment saying so.

`LoadingDots` keeps three literal chains rather than a `dotCss(delayMs)` helper: a helper's value cannot resolve at build time, and truss falls back to a runtime `var(--animation)` instead of static atomic classes.

### `CssReset.css`

All four `@keyframes` blocks are gone from the reset — truss owns them outright now.

Apps pick them up from `dist/truss.css` through the truss plugin's `libraries`. That needs the app on truss 2.30 or later: the merge takes every `@keyframes` block from a library stylesheet unconditionally, but a pre-2.30 `parseTrussCss` has no branch for the `/* @truss @keyframes */` annotation and drops them. internal-frontend is on 2.28.0 and is being upgraded alongside this PR.

### The `virtual:truss.css` change in 2.32.0

Truss no longer writes a `<link>` into the generated `index.html`; entry modules import the stylesheet instead. Two code paths here, and only one needed a change:

- **The library build** needs nothing. `tsup.config.ts` uses `trussEsbuildPlugin`, and lib builds still emit a standalone stylesheet for consuming apps to merge.
- **Storybook** does use `trussPlugin` (via `vite.config.mts`) in app mode, so it was relying on the old `<link>` injection. Without a change, `yarn build-storybook` warns `No module imported "virtual:truss.css", so this build ships no Truss stylesheet` and produces an unstyled Storybook — confirmed by building it. `.storybook/preview.tsx` now imports `virtual:truss.css`.

We have no `index.html`, so there is no stale `<link>` tag for the plugin's new build-time error to catch.

### The `GridTable.borderStyle` test

CI's failure was real and belongs to this branch, not to `main`. The test passed `Css.bc(Tokens.TextDisabled).pl2.$` as a custom `borderStyle` and asserted `paddingLeft: "calc(var(--t-spacing) * 2)"`.

The cell really does get the padding — it carries the `pl2` class, and `.pl2 { padding-left: calc(var(--t-spacing) * 2); }` is the only `padding-left` rule matching it. But only the *first* `getComputedStyle(el)` call for an element returns a declaration containing it:

```
call 1: paddingLeft "calc(var(--t-spacing) * 2)"  paddingRight "12px"  borderLeftWidth "1px"   (43 props)
call 2: paddingLeft ""                            paddingRight ""      borderLeftWidth "1px"   (39 props)
```

Every later call drops all four padding longhands — `padding-right: 12px` goes too, despite being a plain px value — while the border properties are unaffected. One padding longhand that cssstyle cannot round-trip as a length takes the whole group down with it. `toHaveStyle` calls `getComputedStyle` once per asserted property, so `paddingLeft`, third in the object, always read an already-degraded declaration and saw `<empty>`, while the two border assertions passed.

So the component behavior is fine and the assertion was unobservable. Truss started compiling increments to `calc(var(--t-spacing) * n)`, which is why this surfaced with the upgrade. The test now asks for `plPx(8)`: with no unparseable longhand in the group the padding survives every call, and the test still proves that `GridStyle.borderStyle` reaches the cell. The padding assertions are also split out from the border-geometry ones, with a comment recording the limitation.

### The `var(var(--b-primary))` regression

Chromatic caught a missing blue bar next to the active section in `FormPageLayout`'s side nav. The cause is a second behavior change in 2.32.0, unrelated to keyframes.

Truss now hoists a *literal* `variable`-kind value into an atomic class instead of leaving it on the runtime path. On the way it resolves a `Tokens` member and wraps it in `var()` — `tryResolveTemplateLiteral` applies `maybeCssVar` to every interpolation. We were already writing the `var()` ourselves:

```ts
Css.smSb.boxShadow(`inset 3px 0px 0 0px var(${Tokens.Primary})`).$
```

which compiled to:

```css
.bxs_inset_3px_0px_0_0px_var_var_b_primary { box-shadow: inset 3px 0px 0 0px var(var(--b-primary)); }
```

`var(var(…))` is invalid at computed-value time, so the browser discards the whole `box-shadow` declaration. There is no console warning and no build error — the only signal is the visual diff.

Under 2.29.6 the same chain stayed on the runtime path (`box-shadow: var(--boxShadow)` plus an inline custom property), so the double wrap was harmless and invisible.

Six rules were affected, all box-shadows:

- `FormPageLayout` — the active/hover section nav bar
- `Button` — the default and danger focus rings
- `Accordion`, `MenuItem`, `Option`, `TreeOption` — focus rings
- `AvatarButton` — the hover glow

The fix is to interpolate the token bare and let truss wrap it:

```ts
Css.smSb.boxShadow(`inset 3px 0px 0 0px ${Tokens.Primary}`).$
```

The `var()` stays where the plugin does not reach: the SVG `fill` attributes in `Icon.tsx` and `CheckboxBase.tsx`, and the inline-style string in `RichTextField.tsx`. Those are plain JSX and strings, not `Css` chains.

Worth raising upstream, since the failure is silent. Truss could skip `maybeCssVar` when the preceding quasi already ends in `var(`, so both spellings work; and it could fail the build on a resolved value containing `var(var(`, next to the existing `validateAnimationValue` check.

### Testing

- `yarn tsc --noEmit` is clean.
- `yarn truss` output is unchanged from 2.31.0.
- `yarn build` emits all four `@keyframes` blocks and the expected `animation:` classes into `dist/truss.css`.
- `yarn build-storybook` no longer warns, and the bundled `iframe-*.css` contains the truss rules and all four keyframes.
- `yarn build-storybook` before the `var()` fix: 6 `var(var(` rules in `iframe-*.css`. After: 0, and the nav bar rule reads `box-shadow: inset 3px 0px 0 0px var(--b-primary)`.
- Test suite: 1487 pass, 1 skipped, 0 fail.


src/components/Table/GridTable.test.tsx
```diff
@@ -369,20 +369,18 @@ describe("GridTable", () => {
       <GridTable
         columns={[nameColumn, borderedColumn]}
         rows={rows}
-        style={{ ...defaultStyle, borderStyle: Css.bc(Tokens.TextDisabled).pl2.$ }}
+        // And the custom border style adds a padding in px, because jsdom drops the
+        // `calc(var(--t-spacing) * n)` that an increment like `pl2` compiles to.
+        style={{ ...defaultStyle, borderStyle: Css.bc(Tokens.TextDisabled).plPx(8).$ }}
```
stephenh, line 377: AI comment:

Why this is `plPx(8)` and not `pl2`, which is what CI was failing on:

`pl2` compiles to `calc(var(--t-spacing) * n)` now, and jsdom returns that only on the 1st `getComputedStyle` call for an element. Later calls drop all four padding longhands — the static ones too. `toHaveStyle` calls `getComputedStyle` once per asserted property, so `paddingLeft` read empty while the two border properties, read before it, passed.

The cell was styled correctly the whole time; the assertion just couldn't see it. A static px value survives every call.

src/components/Avatar/AvatarButton.tsx
```diff
@@ -95,7 +95,7 @@ export function AvatarButton(props: AvatarButtonProps) {
 }
 
 const resetStyles = Css.br100.cursorPointer.outline0.relative.$;
-export const hoverStyles = Css.boxShadow(`0 0 4px var(${Tokens.OnSurface})`).$;
+export const hoverStyles = Css.boxShadow(`0 0 4px ${Tokens.OnSurface}`).$;
```
stephenh, line 98: Truss auto-adds the `var(...)` -- technically it did that before but only in a codepath that, previously, the `Css.boxShadow(value)` calls did not go through.

The truss change cleaned up the `boxShadow(value)` pass to get more optimizations, which as a side-effect made it go through the existing "auto-add var(...)" code.

src/components/Table/GridTable.test.tsx
bdow: `pl2` computes to `16px`, not `8px` as the spacing increment is 8px. So this may be a regression.

bdow: Actually, just noticing its just a test. So still fine.

bdow (approved):

src/components/Table/GridTable.test.tsx
stephenh: Right -- in general this was very confusing, like something about jsdom internally calculating / but then invalidating styles that use `var`s ... honestly seems kinda suss, but right, it's just a test :thinking:

PR #1460, opened 2026-09-12
chore: Use sqPx for equal width and height.

stephenh:
Stacked on #1459.

`Css.sqPx(n)` already sets both `height` and `width`, so the 25 places that spelled out `hPx(n).wPx(n)` (or `wPx(n).hPx(n)`) say the same thing in one call. 16 files, mostly components plus a handful of stories.

Came out of review on #1459, where `LoadingDots` had the same pair.

### Testing

- `yarn tsc --noEmit` is clean.
- `dist/truss.css` is byte-for-byte identical before and after, so no visual change.
- Test suite: 1487 pass, 0 fail.


stephenh:
Merging b/c it's trivial.

PR #1462, opened 2026-09-15
feat: Support AutoSaveIndicator on PageHeader

bdow:


src/components/Headers/PageHeader.tsx
```diff
@@ -1,29 +1,36 @@
...
-import type { Only } from "src/Css";
+import { Css, type Only } from "src/Css";
 import { pageContentPaddingX } from "src/layouts/layoutSpacing";
 import { useTestIds } from "src/utils/useTestIds";
 
 export type PageHeaderProps<V extends string, X> = Omit<BaseHeaderProps, "bottomSlot"> & {
   /** Rendered as buttons on desktop; two or more collapse into a `ButtonMenu` at `sm`. */
   actions?: HeaderAction[];
+  /** When true, prepends `AutoSaveIndicator` in the actions area. */
+  withAutoSave?: boolean;
```
stephenh, line 13: Could we just always include `AutoSaveIndicator`, and if there are no auto saves, then it just does nothing?

I.e. would prefer the DX of "it just works" vs. pages having to remember "oh right you need to call withAutoSave".

If there is a strong rationale for "yes we need the explicit hint", might be worth articulating that as an in-source coment.

stephenh (approved):

src/components/Headers/PageHeader.tsx
bdow: Sounds good, will do

PR #1463, opened 2026-09-15
feat: ButtonMenu to support new Cycle Tracking UX

bdow:
Screenshot 2026-09-15 at 12 07 48 PM


src/components/internal/Menu.tsx
```diff
@@ -16,10 +16,12 @@ type MenuProps<T> = {
   persistentItems?: MenuItem[];
   selectedItem: string | undefined;
   onChange: ((key: string) => void) | undefined;
+  /** Non-interactive content above the items. The slot is min 240px so short menus don't hug. */
+  header?: ReactNode;
```
stephenh, line 20: This looks just fine, curious what the header is from the screenshots on the PR description?

bdow: The `header` is the
> **Plan Cycle** Running 9d
Started Sept 5, 2026

stephenh: Oh I see ... this is like a fake static/first item? If so, maybe add some variation of "this is a fake first item" to the jsdocs 😅

stephenh: Maybe we could call it `headerItem` to highlight it's "item-ness"?

src/components/Button.tsx
```diff
@@ -32,6 +32,8 @@ export type ButtonProps = {
   labelInFlight?: string;
   /** Shows pressed/active styles (useful when a menu is open) */
   active?: boolean;
+  /** Tint fill + border. Only applied when `variant` is `secondary`; ignored otherwise. */
+  colorScheme?: ButtonColorScheme;
```
stephenh, line 36: Nice, I like a new prop vs "these all become their own variant" 😅

stephenh (approved):
Looks great!

src/components/internal/Menu.tsx
bdow: I'd prefer not to name it as an `item` as those are the elements a user can select. I could see that being just as confusing thinking its like a fixed top item or something...  Happy to expand the JS doc

PR #1464, opened 2026-09-15
feat: Allow for actions to stay visible in header on sm viewports

bdow:
Screenshot 2026-09-15 at 2 55 18 PM


src/components/Headers/PageHeader.tsx
```diff
@@ -17,23 +18,33 @@ export type PageHeaderProps<V extends string, X> = Omit<BaseHeaderProps, "bottom
...
-        tabs && (
-          <div css={pageContentPaddingX}>
-            <Tabs {...tabs} {...tid.tabs} />
-          </div>
+        (bottomSlotActions.length > 0 || tabs) && (
+          <>
+            {bottomSlotActions.length > 0 && (
+              <div css={pageContentPaddingX} {...tid.bottomSlot}>
+                <HeaderActions actions={bottomSlotActions} />
+              </div>
```
apattersonATX-HB, line 40: nit: BaseHeader has `Css.if(!bottomSlot).mb2.` Until now bottomSlot was only true with tabs, and `Tabs` brings its own bottom spacing — so an action with no tabs drops mb2 and the button sits flush on the bb border. The screenshot in the description has tabs, which is why it looks fine there.

We should bring the extra spacing here `...pageContentPadding, ...Css.if(!tabs).mb2`

src/components/Headers/HeaderActions.tsx
```diff
@@ -16,14 +16,19 @@ import { useTestIds } from "src/utils/useTestIds";
...
+  keepVisible?: boolean;
+};
+
 export type HeaderAction =
-  | ({ kind?: "default" } & Omit<ButtonProps, "label"> & { label: string })
-  | ({ kind: "icon" } & Omit<IconButtonProps, "variant" | "label"> & { label: string })
-  | ({ kind: "menu" } & ButtonMenuProps);
+  | ({ kind?: "default" } & HeaderActionKeepVisible & Omit<ButtonProps, "label"> & { label: string })
+  | ({ kind: "icon" } & HeaderActionKeepVisible & Omit<IconButtonProps, "variant" | "label"> & { label: string })
+  | ({ kind: "menu" } & HeaderActionKeepVisible & ButtonMenuProps);
```
apattersonATX-HB, line 27: nit: `HeaderAction` also gets consumed by

- GridTableLayoutActions.tsx
- FormSection.tsx
- FormSectionLayout.ts

But none of them use `keepVisible`. Least not yet 🤔

src/components/Headers/HeaderActions.tsx
```diff
@@ -36,30 +41,47 @@ export function HeaderActions(props: HeaderActionsProps) {
...
+  sm: boolean,
+): { bottomSlotActions: HeaderAction[]; rightSlotActions: HeaderAction[] } {
+  if (!sm) return { bottomSlotActions: [], rightSlotActions: actions ?? [] };
+  return {
+    bottomSlotActions: actions?.filter((action) => action.keepVisible) ?? [],
+    rightSlotActions: actions?.filter((action) => !action.keepVisible) ?? [],
+  };
+}
+
+const actionRowCss = Css.df.aic.fww.gap2.fs0.$;
```
apattersonATX-HB, line 69: question: was the gap bump intentional?

apattersonATX-HB (approved):

src/components/Headers/PageHeader.tsx
bdow: Thanks — follow-up in https://github.com/homebound-team/beam/pull/1465. Added `mb2` on the keepVisible row when there are no tabs.

src/components/Headers/HeaderActions.tsx
bdow: Good catch. FormSection / FormSectionLayout already go through ContentHeader, so `keepVisible` works there. GridTableLayout uses HeaderActions directly and does not collapse, so the flag is a no-op for now — we can wire a bottom slot later if a consumer needs it.

src/components/Headers/HeaderActions.tsx
bdow: Yes, intentional.

PR #1466, opened 2026-09-16
fix: Switch order of right slot and actions on PageHeader

bdow:
Introducing the new Cycle Tracking action into the Plan Package page resulted in the wrong order as that page currently puts all of its actions into the `rightSlot` vs using the `actions` prop. Switching the order of render here, rather than trying to do too much on `internal-frontend`. We don't want the Cycle Tracking action to be rendered directly into the `rightSlot` because it needs to render in a different slot whether its on mobile or desktop, so we expect it to be defined as a HeaderAction passed into PageHeader.actions


bdow:
Super minor change - has no impact downstream. No pages were using both. FWIW, Ideally everything migrates towards using the `actions` prop so we can smartly collapse them down into a single button menu on mobile.

PR #1467, opened 2026-09-16
feat: update `useUnsavedChangesGuard` with `allowNavigation`

apattersonATX-HB:
[DS-327](https://linear.app/homebound-team/issue/DS-327/allow-focusedformlayout-to-permit-selected-in-app-navigations-while)

`FocusedFormLayout` blocks *every* in-app navigation while the form is dirty, which is too blunt for pages that keep their selected item in the URL. 

Adds an optional `allowNavigation` callback to the workflow layouts:

- Only consulted when `isDirty()` is true. Return `true` and the route change goes through with no prompt; everything else still prompts.
- Receives React Router's blocker args (`currentLocation`, `nextLocation`, `historyAction`), so apps can scope the allowlist to the current listing, and can treat browser Back differently if they want.
- Cancel and tab close/refresh are deliberately untouched — both lose the edits regardless of where the user was headed.
- Still one `useBlocker` per page, so the "don't register a parallel app-level check" rule in `docs/layouts.md` stands. This is now the supported way to carve out same-form routes.

`StepperLayout` gets the prop too, since both layouts share the guard. No known use for it there yet — it just seemed worse to have the two APIs differ for no reason.


bdow (approved):
awesome 👏

PR #1469, opened 2026-09-17
feat: Add built-in Right Pane Trigger elements for Workflows

bdow:


src/layouts/Workflow/RightPaneTriggers.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    await waitFor(() => {
+      expect(r.query.rightPaneContent).toBeNull();
+    });
+    expect(r.rightPaneTriggers).toHaveAttribute("aria-hidden", "false");
+    expect(r.rightPaneTriggers).not.toHaveAttribute("inert");
+    expect(r.rightPaneTriggers).toHaveStyle({ transform: "translateX(0)" });
+    expect(r.focusedFormLayout_header.contains(r.rightPaneTriggers)).toBe(false);
+  });
+
+  it("keeps the open pane when changing stepper steps", async () => {
```
bdow: Hmm.. I don't think we'd actually want this. I think the history/comments might be dependent on what is being displayed in the step - not for everything across the stepper workflow. We might want to close/reset on each step change by default. We can introduce persistent drawer per step later if we need it.

src/layouts/Workflow/RightPaneTriggers.tsx
```diff
@@ -0,0 +1,69 @@
...
+/** Page-level trigger chrome: floating stack on desktop (slides off-screen with the pane), header slot on `sm`. */
+export function WorkflowPageRightPaneTriggers(props: { triggers: RightPaneTrigger[] }) {
+  const { triggers } = props;
+  const { sm: isMobile } = useBreakpoint();
+  const { isRightPaneOpen } = useRightPane();
+  const { openRightPane } = useRightPaneActions();
+  const tid = useTestIds(props, "rightPaneTriggers");
+  const pageTriggerInsetPx = 24;
+
+  const onSelect = useCallback(
```
bdow: use `onClick` - keeps consistent with everywhere else.

src/layouts/Workflow/WorkflowPageLayout.tsx
bdow, line 1: If this is a Stepper example and each step has its own context for comments/history, then we can't add the right triggers here, right? And if we did put them _in_ the step, how would we get the triggers into the header?

src/layouts/Workflow/RightPaneTriggers.test.tsx
bdow: Agreed — comments/history should follow the active step. Triggers now live on each `StepperLayoutStep`, and tab / Continue / Back all go through `goToStep`, which closes the pane before changing steps. Persistent-across-steps can wait until we need it.

src/layouts/Workflow/RightPaneTriggers.tsx
bdow: Renamed to `onClick`.

src/layouts/Workflow/WorkflowPageLayout.tsx
bdow: Right — they should not be defined on the workflow page. Each step now owns `rightPaneTriggers`. `StepperLayout` passes `activeStep.rightPaneTriggers` into `WorkflowPageLayout`, which still renders the chrome (floating on desktop, header slot on `sm`). `FocusedFormLayout` stays layout-level since it has no steps.

docs/layouts.md
```diff
@@ -211,8 +209,8 @@ scrollbars while that chrome stays in place:
...
...g. `DocumentScrollToTopButton` via `getFloatingRightOffset`). A pane-width spacer (always mounted on `md+`, `0px` while closed) CSS-transitions to the pane width with the slide so centered content eases instead of jumping; it grows the document so content under the pane stays reachable. While open on `md+`, publishes `--beam-right-pane-content-min: 480px` (`0px` when closed) for the form/shell column to read — not overlay `main` — so JumpLinks sit outside the floor (wide leftover = side-by-side; tight leftover = horizontal scroll). Page-header chrome is `position: fixed` so horizontal scroll cannot move it. The overlay host shell does not subscribe to open state — chrome updates run in a sibling subtree. **`sm`:** full-bleed portal overlay below the env banner (same pattern as `NavbarMobileMenu`); scroll locked via `usePreventScroll`; no spacer / min-width / width vars stay `0px`. Optional pane chrome: **`RightPanePanel`** (title + close). Hooks: **`useRightPane()`** (open, close, `isRightPaneOpen`); **`useRightPaneActions()`** (open/close only — use in table row handlers). Used by `GridTableLayout` around the table body and by `CenteredLayout` / `FormSectionLayout` for form panes.
```
stephenh, line 212: I continue to not read anything in this file 😅

src/components/IconButton.stories.tsx
```diff
@@ -86,6 +86,10 @@ export const Circle = Template.bind({});
 // @ts-ignore
 Circle.args = { variant: "circle" };
 
+export const CircleCompact = Template.bind({});
+// @ts-ignore
+CircleCompact.args = { variant: "circle", compact: true };
```
stephenh: Would prefer `newStory` / no `Template.bind`s

src/components/IconButton.tsx
```diff
@@ -24,7 +24,7 @@ export type IconButtonProps = {
   /** HTML attributes to apply to the button element when it is being used to trigger a menu. */
   menuTriggerProps?: AriaButtonProps;
   buttonRef?: RefObject<HTMLButtonElement | null>;
-  /** Whether to show a 16x16px version of the IconButton */
+  /** Shrinks the control: 18px for default, 32px for `circle`. */
```
stephenh: ...huh, maybe include in the jsdoc why circles are special and get a larger compact?

stephenh (approved):

src/components/IconButton.stories.tsx
bdow: Updated

src/components/IconButton.tsx
bdow: Updated

PR #1470, opened 2026-09-17
fix: modal banner padding so spacing doesn't cutoff in scroll

apattersonATX-HB:
Fixes this

Screenshot 2026-09-17 at 4 28 09 PM


src/components/Modal/Modal.tsx
```diff
@@ -188,7 +196,9 @@ export function Modal(props: ModalProps) {
                   Use `fdrr` so that the close icon won't sit between "modal header search field"
                   and the modal body results in the DOM focus order, i.e. in our global search modal.
                 */}
-                <header css={Css.df.fdrr.p3.fs0.if(drawHeaderBorder).bb.bc(Tokens.SurfaceSeparator).$}>
+                <header
+                  css={Css.df.fdrr.p3.fs0.if(hasBanner).pb2.if(drawHeaderBorder).bb.bc(Tokens.SurfaceSeparator).$}
```
bdow: What if we just always did `pb2`? and add a `pt2` to the top of `<main>` and get rid of the `mb3` in the banner's portal element?
If we adjust some styles here I don't think it'd be too bad. Basically what you're doing here, but not conditionally - just always have this padding shared between body and main.
I guess if we want to keep it consistent as possible, then we currently have a total of 24px of padding at the bottom of the header, and none on `main`. So we could do a `ptPx(12)` on main and `pbPx(12)` on the header. And maybe the banner's element just adds a margin-top of 4px if we really need that extra separation.

I guess what I'm saying is let's see if we can do this without a context.

bdow (approved):
This looks good. Only thing I can think of is that we went from 24px of space between header & body, to now being 32px. If we want to limit churn on storybook we could change the `pb2` to `pbPx(12)` and the `pt2` to `ptPx(12)`. Either way works for me.

PR #1472, opened 2026-09-18
feat: textField presents `originalValue` beneath input;

apattersonATX-HB:
## [DS-359](https://linear.app/homebound-team/issue/DS-359/implement-new-original-value-pattern)

Moves the AI "original value" out of the input and into a slot below the field.

Previously the struck-through on-record value rendered *inside* the input wrapper, left of the proposal. Fine for `20 → 25`, bad for anything long — a SKU or a note would push the proposal out of view entirely.

Two gating decisions worth a look:

- **Disabled fields still show the original.** It's the value being replaced, not supporting copy, so it sits outside the `showErrorAndHelper` gate that suppresses helper text.
- **Read-only and compound fields opt out.** Read-only renders no input for the original to sit under (it keeps the inline path). Compound halves live inside their own bordered boxes, so a block below would break the seam — nothing passes `originalValue` to one today, but the guard makes the intent explicit.

Screenshot 2026-09-18 at 12 29 03 PM


src/inputs/TextFieldBase.tsx
```diff
@@ -401,38 +397,60 @@ export function TextFieldBase<X extends Only<TextFieldXss, X>>(props: TextFieldB
...
+          </>
+        )}
       </div>
-      {/* Error message and helper text for "left" labelStyle */}
+      {/* Original value, error message, and helper text for "left" labelStyle, whose container is a row.
+       * TODO: check with design on this look. This block has always hung at the container's left edge,
+       * i.e. under the label rather than under the field, which reads fine for helper prose but leaves
+       * the struck original stranded away from the value it's replacing. Aligning it would mean
+       * offsetting the whole block by `labelLeftFieldWidth`, which moves every left-label field in Beam. */}
       {labelStyle === "left" &&
```
apattersonATX-HB, line 420: We dont have designs for AiInputs with a left label but it looks pretty bad IMO

bdow (approved):
Nice. Probably worth finding time with design to walk through examples with the different label styles to make sure all looks good.

